PR #2, opened 2017-04-01
Add .gitignore to ignore compilation products

zmwangx:
I'm allergic to dirty worktrees...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*.o
+config.h
+nnn


zmwangx:
By the way, mind reminding me why create an `config.h` when it's just a copy of `config.def.h`? 😅

jarun:
Thank you!

jarun:
> By the way, mind reminding me why create an config.h when it's just a copy of config.def.h? ￼

Legacy legacy legacy :+1: 
But I think it would make sense if we remove the config.h removal stuff from the Makefile. Then it would have someone's temp preferences.

zmwangx:
> But I think it would make sense if we remove the config.h removal stuff from the Makefile. 

Makes sense.

PR #3, opened 2017-04-03
Test on Travis

zmwangx:
This PR sets up Travis testing on Ubuntu Trusty and macOS, with gcc and clang on both platforms. Warnings are turned into errors. See https://travis-ci.org/zmwangx/nnn/builds/218196438 for instance.

Notes:
1. I'm using the Trusty image because the Precise image comes with GCC 4.6.3, and the error messages are just abysmal.
2. Travis's macOS infrastructure is not big enough to handle its demand in real time, so there are crazy backlogs on workdays that IIRC are usually only cleared past midnight (in the U.S.). See https://www.traviscistatus.com/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: c
+matrix:
+  include:
+    # Access more recent gcc and clang via a Trusty image
+    - os: linux
+      dist: trusty
+      compiler: gcc
+    - os: linux
+      dist: trusty
+      compiler: clang
+    - os: osx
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

zmwangx:
> ... so there are crazy backlogs on workdays...

Well, immediately after saying that my build https://travis-ci.org/zmwangx/nnn/builds/218196438 completed. Rather surprising.

PR #12, opened 2017-04-16
Makefile: use libncurses instead of libncursesw on Darwin

zmwangx:
macOS 10.12.4 and many earlier versions have libncurses 5.4 with wide character support, but not libncursesw.


jarun:
Thank you!

PR #14, opened 2017-04-24
Add definition of PATH_MAX on GNU/Hurd

szlin:
MAX_PATH is not defined in some platforms, most notably GNU/Hurd. Therefore, I define it here to some constant.


jarun:
Thank you!

PR #15, opened 2017-04-25
Exit if parent process exited

extrasharp:
nnn was getting stuck in a loop if I spawned a shell, then closed the terminal. This should fix that!

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Noice is Not Noice, a noicer fork...
   - [Boost chdir prompt](#boost-chdir-prompt)
   - [Change file associations](#change-file-associations)
 - [Why fork?](#why-fork)
+- [Mentions](#mentions)
 - [Developers](#developers)
 
 ### Introduction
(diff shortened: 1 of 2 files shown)


jarun:
- I tried it but couldn't reproduce it. Can you please provide the step-by-step instructions to elaborate the scenario?
- Can you verify if the PPID check goes under the `nochange:` label if works or not? That way it would cover all cases. As you might have noticed we also call system() to spawn `nlay`.
- Also, kindly rebase your master. I did a force commit without noticing your PR. Sincere apologies.

extrasharp:
To reproduce 100% on my machine (using zsh as the shell and urxvt as the terminal):
1. run nnn
2. press ! to spawn a new shell out of nnn
3. ```kill -9``` the zsh process that is the parent of nnn

then nnn becomes a zombie with 80-100% cpu usage.
Earlier today, all I had to do was close the terminal window normally, or end the X session and this would happen. But just now I could only reproduce it by killing the parent process directly. So the issue doesnt seem to be deterministic.

Yes, it works under the ```nochange:``` label

No worries!

jarun:
Thank you! Eagerly waiting for the updated PR. :+1:

extrasharp:
Okay I am not very good at git haha I may have messed something up here

jarun:
Would you mind closing this PR and raising a new one using latest master? I think that's the best option here.

jarun:
Please:

1. Update your fork (or delete the old one, fork again)
2. Make the changes
3. Raise the PR

I'll not push new changes before you raise the PR. Sounds fair?

extrasharp:
Sounds good, thanks!

PR #28, opened 2017-05-13
Add -Werror to makefile

fennm:


jarun:
Actually we wanted to avoid `-Werror` in the Makefile earlier. So closing this. I'll take care of this one. No problem! :+1:

PR #32, opened 2017-05-29
fix compile warning on gcc7

shaggytwodope:
While this removes any warnings, it may be worth
exploring in more detail in the future.


jarun:
Thank you!

Issue #36, opened 2017-08-28
uneven splitting of codepoints


markroxor:
![correct](https://user-images.githubusercontent.com/11366315/29776177-75642354-8c25-11e7-8ede-a41c4091ecc8.png)
![jumbled](https://user-images.githubusercontent.com/11366315/29776176-756252b8-8c25-11e7-94ba-93f9b3c1e77d.png)

As per the above screenshots the codepoints on the edges of the terminal session are messed up because the number of columns "breaks-through" the code points.

jarun:
Copying from our earlier email conversation:

I think we need to check if the codepoint at COL is valid or not. If not, try the prev one and so on...
The function mblen() might be of help (no need to do a wchar conversion).

markroxor:
`mblen` returns 1 for the string 
> Malgudi Days - मालगुडी डेज - E05. Swami and Friends - स्वामी और उसके दोस्त (Part 1)

Is this justified?

markroxor:
I figured out the actual length of the file name (including the MB length).  How can we decide that while resizing the terminal the string should break at a particular code point? any ideas/?

jarun:
I think you need to keep checking for valid codepoints (using mblen) till it exceeds the total length of the columns.

PR #37, opened 2017-08-28
uneven splitting of codepoints

markroxor:
refer #36

### DIFF
--- a/nnn.c
+++ b/nnn.c
@@ -1189,7 +1189,6 @@ unescape(const char *str)
 static void
 printent(struct entry *ent, int sel)
 {
-    return;
 	static int ncols;
 
 	if (PATH_MAX + 16 < COLS)
@@ -1278,13 +1277,14 @@ printent_long(struct entry *ent, int sel)
         {
             // experimenting in this section
             const char *c = unescape(ent->name);
-            unsigned int i;
-            int count=0;
-            for(i=0;i<strlen(c);i++)
-                count+=mblen(c+i, MB_CUR_MAX);
 
-			snprintf(g_buf, ncols, "%s%-16.16s %8.8s  %s--%d", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name), count);
+            int count = 30;
+            int goback=0;
+            if(ncols+goback>count)
+                while((ncols+goback)>count && mblen(c+(ncols+goback)-count, MB_CUR_MAX)==-1)
+                    goback--;
 
+            snprintf(g_buf, ncols+goback, "%s%-16.16s %8.8s %s", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
         }
 	} else {
 		if (S_ISDIR(ent->mode))
--- a/nnn.c
+++ b/nnn.c
@@ -1186,6 +1186,21 @@ unescape(const char *str)
 	return buffer;
 }
 
+int
+getgoback(struct entry *ent,int sel, int ncols)
+{
+    const char *c = unescape(ent->name);
+
+    // counts the number of columns before it enounters the filename
+    int count = snprintf(g_buf, ncols, "%s ", CURSYM(sel));
+
+    int goback = 0;
+    while((ncols + goback) > count && mblen(c + (ncols + goback) - count, MB_CUR_MAX) == -1)
+        --goback;
+
+    return goback;
+}
+
 static void
 printent(struct entry *ent, int sel)
 {
@@ -1197,7 +1212,7 @@ printent(struct entry *ent, int sel)
 		ncols = COLS;
 
 	if (S_ISDIR(ent->mode))
-        snprintf(g_buf, ncols, "%s%s/", CURSYM(sel), unescape(ent->name));
+		snprintf(g_buf, ncols, "%s%s/", CURSYM(sel), unescape(ent->name));
 	else if (S_ISLNK(ent->mode))
 		snprintf(g_buf, ncols, "%s%s@", CURSYM(sel), unescape(ent->name));
 	else if (S_ISSOCK(ent->mode))
@@ -1206,9 +1221,10 @@ printent(struct entry *ent, int sel)
 		snprintf(g_buf, ncols, "%s%s|", CURSYM(sel), unescape(ent->name));
 	else if (ent->mode & 0100)
 		snprintf(g_buf, ncols, "%s%s*", CURSYM(sel), unescape(ent->name));
-	else
-		snprintf(g_buf, ncols, "%s%s", CURSYM(sel), unescape(ent->name));
-
+	else {
+		int goback = getgoback(ent, sel, ncols);
+		snprintf(g_buf, ncols + goback, "%s%s", CURSYM(sel), unescape(ent->name));
+    }
 	/* Dirs are always shown on top */
 	if (cfg.dircolor && !S_ISDIR(ent->mode)) {
 		attroff(COLOR_PAIR(1) | A_BOLD);
@@ -1242,6 +1258,21 @@ coolsize(off_t size)
 	return size_buf;
 }
 
+int
+getgoback_long(struct entry *ent, int sel, int ncols, char buf[18])
+{
+    const char *c = unescape(ent->name);
+
+    // counts the number of columns before it enounters the filename
+    int count = snprintf(g_buf, ncols, "%s%-16.16s %8.8s  ", CURSYM(sel), buf, coolsize(ent->size));
+
+    int goback = 0;
+    while((ncols + goback) > count && mblen(c + (ncols + goback) - count, MB_CUR_MAX) == -1)
+        --goback;
+
+    return goback;
+}
+
 static void
 printent_long(struct entry *ent, int sel)
 {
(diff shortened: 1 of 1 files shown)


markroxor:
Not the final patch, I'm just experimenting. I'll ping you once it is done.

markroxor:
@jarun just gave it a quick fix, can you confirm on your machine?

nnn.c
```diff
@@ -1274,7 +1274,18 @@ printent_long(struct entry *ent, int sel)
 		else if (ent->mode & 0100)
 			snprintf(g_buf, ncols, "%s%-16.16s %8.8s* %s*", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
 		else
-			snprintf(g_buf, ncols, "%s%-16.16s %8.8s  %s", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
+        {
+            // experimenting in this section
+            const char *c = unescape(ent->name);
+
+            int count = 30;
```
jarun: What's this count? Why 30?

nnn.c
```diff
@@ -1274,7 +1274,18 @@ printent_long(struct entry *ent, int sel)
 		else if (ent->mode & 0100)
 			snprintf(g_buf, ncols, "%s%-16.16s %8.8s* %s*", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
 		else
-			snprintf(g_buf, ncols, "%s%-16.16s %8.8s  %s", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
+        {
+            // experimenting in this section
```
jarun: Is this the final code or a test code?

nnn.c
```diff
@@ -1274,7 +1274,18 @@ printent_long(struct entry *ent, int sel)
...
+            const char *c = unescape(ent->name);
+
+            int count = 30;
+            int goback=0;
```
jarun: Please maintain spacing (e.g. `goback = 0`) everywhere. I tend to follow the linux kernel coding style in my projects (other that the 80 char limit per line):

https://01.org/linuxgraphics/gfx-docs/drm/process/coding-style.html

jarun: This applies to the rest of the changes too.

nnn.c
```diff
@@ -1274,7 +1274,18 @@ printent_long(struct entry *ent, int sel)
...
+            if(ncols+goback>count)
+                while((ncols+goback)>count && mblen(c+(ncols+goback)-count, MB_CUR_MAX)==-1)
+                    goback--;
+
+            snprintf(g_buf, ncols+goback, "%s%-16.16s %8.8s %s", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
```
jarun: This case takes care of only regular files. Our problem is with names notwithstanding the type of file.

jarun:
@markroxor in addition to my review comments, a general comment (as you have removed the `work in progress` label):

We tend to push only **_production_ready_** code. Your code, unfortunately, is far from that. Many of the utilities are available in multiple Linux distros and OSes. So when we make a release lots of people use them. Which is why we do _NOT_ release hacks or non-optimal patches without a valid reason.

nnn.c
jarun: OK, I got it from your PR comment. Please leave these code related comments in the diff as general comments. Makes it easier to find.

We can't limit to 30. In this case we can fall back to the non-detailed view mode.

nnn.c
```diff
@@ -1274,7 +1274,18 @@ printent_long(struct entry *ent, int sel)
...
 			snprintf(g_buf, ncols, "%s%-16.16s %8.8s* %s*", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
 		else
-			snprintf(g_buf, ncols, "%s%-16.16s %8.8s  %s", CURSYM(sel), buf, coolsize(ent->size), unescape(ent->name));
+        {
+            // experimenting in this section
+            const char *c = unescape(ent->name);
+
+            int count = 30;
+            int goback=0;
+            if(ncols+goback>count)
```
jarun: Similar changes should go into the function `printent()` (non-detail mode) too.

jarun:
@markroxor in addition to my review comments, a general comment (as you have removed the `work in progress` label):

We tend to push only **_production_ready_** code. Your code, unfortunately, is far from that. Many of the utilities are available in multiple Linux distros and OSes. So when we make a release lots of people use them. Which is why we do _NOT_ release hacks or non-optimal patches without a valid reason.

nnn.c
jarun: I think the best option would be to check the number of cols in the beginning. If it is < 30 call `printent()` with the same args.

markroxor: Instead I am using what `snprintf` is returning. Refer code. What do you think?

nnn.c
markroxor: final code, waiting for your suggestions.

nnn.c
markroxor: you mean I should extend this to other "else" clause as well?

nnn.c
```diff
@@ -1186,6 +1186,21 @@ unescape(const char *str)
 	return buffer;
 }
 
+int
+getgoback(struct entry *ent,int sel, int ncols)
+{
+    const char *c = unescape(ent->name);
+
+    // counts the number of columns before it enounters the filename
+	int count = snprintf(g_buf, ncols, "%s ", CURSYM(sel));
```
jarun: This is fixed to 30. Using 30 was OK. What I intended to say is why not fallback to non-detail mode when we have less than 30 columns?

nnn.c
```diff
@@ -1197,7 +1212,7 @@ printent(struct entry *ent, int sel)
 		ncols = COLS;
 
 	if (S_ISDIR(ent->mode))
-		snprintf(g_buf, ncols, "%s%s/", CURSYM(sel), unescape(ent->name));
+        snprintf(g_buf, ncols, "%s%s/", CURSYM(sel), unescape(ent->name));
```
jarun: Indentation issue. Please use tabs.

nnn.c
markroxor: > This is fixed to 30. Using 30 was OK.

Isn't this better?
> What I intended to say is why not fallback to non-detail mode when we have less than 30 columns?

On it.

nnn.c
```diff
@@ -1186,6 +1186,21 @@ unescape(const char *str)
...
+int
+getgoback(struct entry *ent,int sel, int ncols)
+{
+    const char *c = unescape(ent->name);
+
+    // counts the number of columns before it enounters the filename
+	int count = snprintf(g_buf, ncols, "%s ", CURSYM(sel));
+
+	int goback = 0;
+    while((ncols + goback) > count && mblen(c + (ncols + goback) - count, MB_CUR_MAX) == -1)
```
jarun: Please explain what you are trying to do here. I am unable to understand.

jarun: Please start with what `goback` is. :)

markroxor: Kind of hard to explain but -
`mblen` returns -1 for multi-byte characters which have not ended. Suppose the character is "ं", it is of "length" `3` so mblen will return a non ''-1'' value only when this char is completely printed. So I decremented the variable `goback` until mblen doesn't return a non "-1" value, this is the point from where breaking the code-point is safe.

Did you try executing this locally? Is this what you ask?

nnn.c
```diff
@@ -1186,6 +1186,21 @@ unescape(const char *str)
...
+        goback--;
```
jarun: it's always better to use pre-incr and pre-decr. Ref: https://embeddedgurus.com/stack-overflow/2008/08/efficient-c-tips-3-avoiding-post-increment-decrement/

nnn.c
markroxor: `goback` is the difference between the current column length and the length where breaking the code-point is safe.

jarun: It's makes sense to me now. `goback` can only be negative. That's what I missed earlier.

jarun:
I think the changes look good now. I will run a some tests and pull it in.

jarun:
I did some testing. This isn't working. It seems to work because in the line:

    while((ncols + goback) > count && mblen(c + (ncols + goback) - count, MB_CUR_MAX) == -1)

`c + (ncols + goback) - count` makes an illegal memory access if `ncols + goback - count` is > length of the string. So it _cuts_ the string beyond its end.

Thanks for your efforts. I will have to handle this myself because it needs deeper understanding of how the printing works today and curses printing.

jarun:
Please take a look at https://github.com/jarun/nnn/commit/1e5a0b8c5bc78462956d01c647293a7da594c36f.

My understanding is - these kind of issues involve wide char conversion one way or another. It would have been great to be able to avoid it but we are doing it already at `unescape()` so I mixed both and could throw in some optimization. The conversion is always safe as it guarantees split at a codepoint.

markroxor:
I know this is resolved but -
> `c + (ncols + goback) - count` makes an illegal memory access if `ncols + goback - count` is > length of the string. So it _cuts_ the string beyond its end.

This can be easily fixed by limiting it to the total length of printable string? Isn't it?

jarun:
I tried that already. Didn't work as expected. Try it out.

jarun:
I mean test by resizing the terminal live.

jarun:
Also, it's more accurate but slower to calculate the fixed length for each entry. So I did an approximation there including the newline char at the end.

markroxor:
Okay. Anyways the issue is resolved so no point in further tweaking. Deleting the branch. Thanks.

PR #39, opened 2017-09-15
Makefile: LICENSE file will be handled by install/uninstall targets

Arlon1:


jarun:
Why is this required?

Arlon1:
To be accurate there is no technical or legal reason to include the license text (as the license doesn't require it).

Although in my opinion it is always a good idea to share the license with each installation, especially since we're having here a custom license.
Also the license text is only 1399 bytes large.

jarun:
The [license](https://github.com/jarun/nnn/blob/master/LICENSE) is not a custom one. It's standard [BSD-2-Clause](https://opensource.org/licenses/BSD-2-Clause). Also the license is mentioned in program help (`nnn -h`).

I am closing this PR with the above argument.

Issue #1, opened 2017-03-29
ToDo list


jarun:
Feel free to suggest new features to add to the list below. Some of the features may still be master-only.

- [x] list and extract archives
- [x] show current selection number along with total entries in dir
- [x] support `NNN_NOWAIT` to unblock nnn (DE specific)
- [x] file and directory creation
- [x] change to light mode if number of cols < 35
- [x] in-place file rename
- [x] watch out for current dir changes (if not in filter or du mode)
- [x] pin and jump to a directory
- [x] show vol capacity and vol free in help
- [x] option to use exiftool instead of mediainfo
- [x] option to show dirs in custom color (default: enabled in blue)
- [x] show dir content size and number of files in du mode
- [x] support bookmarks string
- [x] add option to support opening text files in EDITOR
- [x] use `xdg-open` on Linux and `open(1)` on OS X as desktop opener
- [x] support multiple apps in `nlay`
- [x] use `vlock` as default terminal screensaver
- [x] *navigate-as-you-type* mode
- [x] subtree search (using gnome-search-tool or catfish)
- [x] support custom nlay
- [x] *search-as-you-type* at filter prompt
- [x] RPM packages for RHEL/CentOS (thanks @fennm)
- [x] support cd ..... (with . as PWD, stops at /)
- [x] jump to initial directory (where nnn was started)
- [x] `nlay` - a shell script for actions like open desktop search utility, run screensaver
- [x] show file details, mediainfo output, help in `less`
- [x] show media information (needs mediainfo)
- [x] support readline at chdir prompt
- [x] use `Q` to quit to current dir (needs external shell integration)
- [x] use `^Q` to quit (even from filter mode)
- [x] UTF-8 support
- [x] add option to start in disk usage analyzer mode
- [x] open current dir in desktop file manager
- [x] disk usage analyzer more
- [x] show unprivileged free blocks in du mode
- [x] shortcut `-` to jump to last directory
- [x] resolve `~` (HOME) at chdir prompt
- [x] option to show complete stats of just the current entry (name, file type, human readable permission string, byte size... all `lstat()` details)
- [x] update noice man page
- [x] sort entries by size
- [x] support file path copier script (`NNN_COPIER`)
- [x] support detail view (default)
  - [x] file type
  - [x] modification time
  - [x] human-readable file size
  - [x] current item in reverse video
  - [x] full name of currently selected file in bottom "bar"
- [x] roll navigation with `Up` and `Down` at the first and last entries
[...]

PR #40, opened 2017-09-18
Prevent nnn from waiting on open file process

Rahi374:
#1 Tested and necessary on Debian.

### DIFF
--- a/nnn.c
+++ b/nnn.c
@@ -2212,7 +2212,7 @@ nochange:
 				}
 
 				/* Invoke desktop opener as last resort */
-				spawn(utils[2], newpath, NULL, NULL, F_NOTRACE|nowait);
+				spawn(utils[2], newpath, NULL, NULL, nowait);
 				continue;
 			}
 			default:
@@ -2865,7 +2865,7 @@ main(int argc, char *argv[])
 	copier = getenv("NNN_COPIER");
 
 	/* Get nowait flag */
-	nowait = getenv("NNN_NOWAIT") ? F_NOWAIT : 0;
+	nowait = F_NOTRACE | (getenv("NNN_NOWAIT") ? F_NOWAIT : 0);
 
 	signal(SIGINT, SIG_IGN);
 
--- a/nnn.c
+++ b/nnn.c
@@ -222,6 +222,7 @@ static char *player;
 static char *copier;
 static char *editor;
 static char *desktop_manager;
+static char nowait;
 static blkcnt_t ent_blocks;
 static blkcnt_t dir_blocks;
 static ulong num_files;
@@ -2211,7 +2212,7 @@ nochange:
 				}
 
 				/* Invoke desktop opener as last resort */
-				spawn(utils[2], newpath, NULL, NULL, F_NOTRACE|F_NOWAIT);
+				spawn(utils[2], newpath, NULL, NULL, F_NOTRACE|nowait);
 				continue;
 			}
 			default:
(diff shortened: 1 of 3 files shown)


jarun:
So what happens to users on Ubuntu and OS X where the current code works?

Rahi374:
I imagined that this version would still work, but you do have a point.
Would a build flag do the trick?

jarun:
Yes, please add this change under a Debian-specific flag.

jarun:
This is what I see on Ubuntu:

```
arun@CBQM882 /m/c/U/a/V/aj887757_ENHREQ1369409> cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.3 LTS"
arun@CBQM882 /m/c/U/a/V/aj887757_ENHREQ1369409> cat /etc/debian_version
stretch/sid
arun@CBQM882 /m/c/U/a/V/aj887757_ENHREQ1369409>
```

Check what's different for you on Debian. Then pass some Debian-only macro from the Makefile as explained here:

https://stackoverflow.com/a/32274389

zmwangx:
I would guess this isn't a Debian/Ubuntu distinction, but rather one with the desktop, say Unity vs Gnome vs KDE.

jarun:
We can keep Unity out of the equation I guess. What's the default on Debian? XFCE or Gnome?

jarun:
OK, Gnome.

jarun:
With the number of DEs we have and may have on Linux, this would be quite difficult to maintain and track.

I think the best way forward would be to support an env variable, which, when set, adds `F_NOWAIT`. Then we document the symptoms and ask users to set this variable if they notice this problem.

What do you guys think?

Rahi374:
Yeah I didn't think it would be a distro problem.
I'm actually on i3 so that might affect how desktop processes are handled. I'll try it on gnome later.

I personally thought that it would be better to be decided at build time, but if it is indeed dependent on the DE then an environment variable might be better.

jarun:
> I personally thought that it would be better to be decided at build time

As you can see, the behaviour is not defined. DEs may choose to implement it their own way and it may change even for the same distro. We can't test all of those. So the env variable will take care of that easily.

jarun:
We also need to document this. Both under the howto section in the README and in the manpage.

What's the point of this change if users don't know it exists and how to use it? All of the effort amounts to 0 (incidentally, the default value of NNN_NOWAIT) in that case, isn't it? Or, are you thinking you are the only one affected by the issue?

nnn.c
```diff
@@ -2863,6 +2864,9 @@ main(int argc, char *argv[])
 	/* Get the default copier, if set */
 	copier = getenv("NNN_COPIER");
 
+	/* Get nowait flag */
+	nowait = getenv("NNN_NOWAIT") ? F_NOWAIT : 0;
```
jarun: Why not something like:

    nowait = F_NOTRACE | (getenv("NNN_NOWAIT") ? F_NOWAIT : 0);
Didn't compile it... but something in that line. This way we can skip the extra bit-OR every time.

jarun:
Thanks @Rahi374! I'll merge this soon.

jarun:
@Rahi374 can you please confirm if https://github.com/jarun/nnn/commit/55643a5fcca90fe3998bee3901019fc703962083 works fine for you?

Rahi374:
@jarun Thank you!
Your language composition is a lot better than mine.
I will leave code comments directly in the commit.

Rahi374:
Also, do we need to increment the version number?

jarun:
Generally we change the version number when we do a release. I'll make a release probably next month.

If you want any other features in nnn, or are interested in working as a collaborator, please let me know on gitter.

Rahi374:
I see.

I don't usually have good ideas, plus for me this program is already perfect so I'm not sure what there is to collaborate with. Though if I do find anything I want to add I'll send you another PR

PR #47, opened 2017-10-05
Use temporary file name to avoid race condition

bkhl:
In quitcd.bash, use mktemp to make a filename, rather than hardcoding '/tmp/nnn'.

### DIFF
--- a/scripts/quitcd/quitcd.bash
+++ b/scripts/quitcd/quitcd.bash
@@ -1,4 +1,4 @@
-export NNN_TMPFILE="/tmp/nnn"
+export NNN_TMPFILE="$(mktemp -u nnn.XXXXXXXX)"
 
 n()
 {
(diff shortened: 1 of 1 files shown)


jarun:
I guess it's quite difficult for a user to do it that fast, even on terminator with multiple panes. Could you reproduce it manually?

I am open to any other argument(s) too.

If the answer is yes, we will need a few more changes:

- remove the default in https://github.com/jarun/nnn/blob/master/nnn.c#L2143 and write to the tmp file only if NNN_TMPFILE is defined. You can completely remove `tmpfile`.
- I believe the same change should go in the scripts for fish and zsh.
- Also, please remove the sentence "You can change it." from the https://github.com/jarun/nnn/blob/master/README.md#cd-on-quit section.

bkhl:
Ah, didn't notice the file is only created at the time nnn closes. In that case you are right it would be a bit overzealous to use mktemp.

jarun:
No problem! Thanks for using nnn!

PR #48, opened 2017-10-07
OpenBSD compatibility support

vaygr:
This one fixes 2 things:

1. Compilation on OpenBSD (checked **6.1**) as there's no `O_EVTONLY` there
2. Indefinite blocking using `kevent()`, which results in nnn's freeze


jarun:
Thank you very much!

Issue #50, opened 2017-10-07
Create command doesn't work on OpenBSD


vaygr:
Recently added create command (`n`) results in the following error message:

> No traversal delimiter allowed

jarun:
You cna't use delims like `/` or `\` while creating a new file/dir (as designed).
- Are you able to create files/dirs without these chars?
- Or nothing works?

jarun:
Resolved at https://github.com/jarun/nnn/commit/8959a5b57dd75a0a0402fedd7a3705146631d419.

vaygr:
Yeah, on Linux it worked perfectly fine, on OpenBSD it didn't at all. Turns out `basename()` implementation was slightly different there.

jarun:
Got it! We have a native implementation on (x)dirname() sans this handicap but I never cared to extend it to basename().

PR #53, opened 2017-10-08
Correct name checking for NEW and RENAME

vaygr:
According to the [man page](https://man.openbsd.org/man3/basename.3):

```
CAVEATS
     basename() returns a pointer to internal static storage space that will
     be overwritten by subsequent calls.

     Other vendor implementations of basename() may modify the contents of the
     string passed to basename(); this should be taken into account when
     writing code which calls this function if portability is desired.
```

So pointers will point to different memory areas, thus we should rely on string comparison.

This is a proposed fix for #50

### DIFF
--- a/nnn.c
+++ b/nnn.c
@@ -2673,7 +2673,7 @@ nochange:
 				break;
 
 			/* Allow only relative, same dir paths */
-			if (tmp[0] == '/' || basename(tmp) != tmp) {
+			if (tmp[0] == '/' || xstrcmp(basename(tmp), tmp) != 0) {
 				printmsg(STR_INPUT);
 				goto nochange;
 			}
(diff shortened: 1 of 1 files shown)


jarun:
Awesome! Thank you!

jarun:
hey Vlad, I received a mail from OpenHub (to claim my contribution to nnn) but the link couldn't be resolved. Did you add and remove nnn?

vaygr:
No, absolutely not, but maybe it's something weird with the links they generate in notifications.

The project [stands still](https://www.openhub.net/p/nnn). Just create your account there, so you'll be able to claim your contributions and I'll be able to add you to project managers.

jarun:
OK. You can add arunj.

vaygr:
Ah, I think you should [claim you contributions](https://www.openhub.net/p/733199/contributors/3149067874808395) first, and then [apply as a manager](https://www.openhub.net/p/nnn/managers/new). Their system is not straightforward :)

jarun:
The process is unbelievably time twisted. I would have added 6/7 of my other projects. But now I think 1 is enough. ;)

vaygr:
hehe @jarun once you're done with the application (I haven't seen it yet) there, do you think we should clean up this thread and leave only PR-related comments?

jarun:
Sure! I am deleting these now.

PR #54, opened 2017-10-08
Fix a couple of warnings on BSDs

vaygr:
This fixes a couple of warnings on OpenBSD, but I suspect the same will be valid for the rest:

```text
nnn.c:1510: warning: format '%lx' expects type 'long unsigned int', but argument 3 has type 'dev_t'
nnn.c:1510: warning: format '%lu' expects type 'long unsigned int', but argument 4 has type 'dev_t'
```

`dev_t` is actually plain `int` there, cause: `dev_t` -> `__dev_t` -> `__int32_t` -> `int`

(the last `typedef` is from `/usr/include/<amd64|i386>/_types.h`)


jarun:
Thank you!

PR #68, opened 2017-12-26
CMake based build 

dilawar:
- Contains CMake (>=2.6) based build.
- Builds on OpenSUSE.
- Packages are build here https://build.opensuse.org/package/show/home:dilawar/nnn


jarun:
You retired Makefile!!! Did you notice that we use Makefile to package for other distros in using packagecore?

If you are trying to handle the first item in the ToDo list, we want to have all the packages built in Travis CI. We don't want any distro specific stuff. The make-based build system has done an excellent job so far and I don't see a reason to retire it.

vaygr:
I also do not see any reason for adding another overloaded dependency to the build.

You shouldn't need a C++ compiler and a massive build system to compile a lightweight C program. It's like building hello world program with Meson.

dilawar:
@vaygr `CMake` is available on almost all distributions these days; and it is replacing 'pkg-config'. However, it is only useful  when path of dependencies varies from distribution to distribution. I've found cmake to be much easier to use that `autotools`. If the project never felt the need of `./configure` then there is nothing to gain by adding `cmake`, period.

But I am going to maintain my fork (just in case) for building packages on open-build-services in my personal repository (linked above). I keep other hard to packages there as well.

I am currently on `openSUSE_Leap_42.3` and package seems to be working (in case someone is reading it and interested).

jarun:
> But I am going to maintain my fork

That would be great!

> in case someone is reading it and interested

Of course! I was looking for it yesterday and noticed your package. Thank you so much!

jarun:
Here you go: https://github.com/jarun/nnn/commit/4778445da79a11f7625ac4adc865af340c5659a9

Merry Christmas and a Happy New Year!

I couldn't help notice you are in Bangalore. Beautiful sunny days with awesome breeze here. ;)

dilawar:
Cool. Thanks.

PR #70, opened 2017-12-26
Lookup xdg-open in PATH on non-Apple platforms

dmgk:
On FreeBSD (and other *BSD), xdg-open is installed in /usr/local/bin/, not /usr/bin/. Remove hardcoded path prefix and rely on PATH lookup by execlp() instead.


jarun:
Saves a few bytes. Awesome!

PR #74, opened 2018-01-05
Use standard date formats.

codeliveroil:
- For standard view, use ISO 8601 date format.
- For detailed view (i.e. stat), use default Linux date
  format. Also, remove GMT offset to avoid confusion in
  determining source of offset when reading alongside
  the local time zone.

### DIFF
--- a/nnn.c
+++ b/nnn.c
@@ -1319,7 +1319,7 @@ printent_long(struct entry *ent, int sel, uint namecols)
 {
 	static char buf[18], *pname;
 
-	strftime(buf, 18, "%d-%m-%Y %H:%M", localtime(&ent->t));
+	strftime(buf, 18, "%Y-%m-%d %H:%M", localtime(&ent->t));
 	pname = unescape(ent->name, namecols);
 
 	/* Directories are always shown on top */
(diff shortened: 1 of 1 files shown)


jarun:
Currently it's somewhat similar to the stat output:

```
stat:
Access: 2018-01-06 01:28:21.083618686 +0530
Modify: 2018-01-05 22:41:57.526575120 +0530
Change: 2018-01-05 22:41:57.526575120 +0530

current:
Access: Sat 06-Jan-2018 01:28:21 +0530,IST
Modify: Fri 05-Jan-2018 22:41:57 +0530,IST
Change: Fri 05-Jan-2018 22:41:57 +0530,IST

PR:
Access: Sat Jan 06 01:28:21 IST 2018
Modify: Fri Jan 05 22:41:57 IST 2018
Change: Fri Jan 05 22:41:57 IST 2018
```

I really don't see a reason to change it as long as someone is not complaining.

If you are interested in contributing some features please let me know.

jarun:
However, I removed the redundant hyphens at ab7c6f2af8071deb43d128f972b9c9951af987aa.

jarun:
OK. I see your point now. I will revert my commit. Can you make the following changes?

- Replace IST with +0530 as in `stat` output.
- Bring year after the month and weekday

I am looking for:

    Fri Jan 05 2018 22:41:57 +0530

jarun:
I am pulling this. I will make the minor changes. Thank you!

codeliveroil:
Thanks for making the change. Also, very practical and fast file browser! Thanks for maintaining it.

To clarify the changes:
standard view: dd-mm-yyyy is not a widely accepted format since it is often confused with mm-dd-yyyy when dd<=12; yyyy-mm-dd is the international date format that was created to remedy such issues (i.e. if the date begins with yyyy, it is interpreted as being in ISO8601 format and the next four digits are parsed as mmdd and not ddmm).
detailed view: either a GMT offset or a timezone is provided but rarely both.

jarun:
The main reason I pulled it is - the format is automatically sorted by date. So the time sorted view becomes very easy to understand.

Also, I retained the offset because a user of `nnn` would very likely be aware of his local standard time.

PR #76, opened 2018-01-07
make Makefile more packaging friendly 

rindeal:


jarun:
Can't review without any description.

jarun:
Can't review without any description.

rindeal:
I'd like to package this software, which is currently impossible without patching the Makefile. With these changes merged I won't have to patch anything, while everything will be working at least the same as before.

If you don't understand certain changes then comment directly on those and I'll try my best to explain them.

jarun:
> If you don't understand certain changes

Please set up a context when you raise a PR. For example, PR #74 tells me what the PR is about. Going through all the changes to get an idea of what's going on is difficult for me given I am not an expert in packaging myself.

I have the following questions:

- which distro are you packaging for?
- why decouple the `strip` option?
- why override `-O3` when the authors are certifying the binary works well that level of optimization and are ready to work on it if there are issues?

rindeal:
>    which distro are you packaging for?

https://dev.gentoo.org/~ulm/pms/head/pms.html, which can then be used by various distros like Gentoo, Exherbo, ...

>    why decouple the strip option?

because PMS-compatible package managers run `strip` themselves

>    why override -O3 when the authors are certifying the binary works well that level of optimization and are ready to work on it if there are issues?

because users of PMS-compatible package managers use their own optimization flags and don't want any default ones

jarun:
@shaggytwodope @szlin @zmwangx please have a look at this change. We would love to avoid surprises on our standard channels at next release.

Makefile
```diff
@@ -23,30 +25,31 @@ all: $(BIN) $(PLAYER)
...
+	$(CC) -DDEBUGMODE -g $(CFLAGS) $(LDFLAGS) -o $(BIN) $^ $(LDLIBS)
 
 install: all
-	mkdir -p $(DESTDIR)$(PREFIX)/bin
-	cp -f $(BIN) $(DESTDIR)$(PREFIX)/bin
-	cp -f $(PLAYER) $(DESTDIR)$(PREFIX)/bin
-	mkdir -p $(DESTDIR)$(MANPREFIX)/man1
-	cp -f $(BIN).1 $(DESTDIR)$(MANPREFIX)/man1
+	install -m 0755 -D -t $(DESTDIR)$(PREFIX)/bin $(BIN) $(PLAYER)
+	install -m 0644 -D -t $(DESTDIR)$(MANPREFIX)/man1 $(BIN).1
```
zmwangx: macOS's `install(1)` does not support `-D`.[1] I hate it, but that's what we've got.

[1] [Full HTML man page](https://rawgit.com/zmwangx/37e2232ed4fa7a4c9199266a9f62b70c/raw/a5daeda133c685cdaa865d1a256f6d577df0e804/install.1.html) of `install(1)` on macOS 10.13.2 (latest).

Makefile
```diff
@@ -23,30 +25,31 @@ all: $(BIN) $(PLAYER)
...
-	cp $(DISTFILES) nnn-$(VERSION)
+	$(CP) $(DISTFILES) nnn-$(VERSION)
 	tar -cf nnn-$(VERSION).tar nnn-$(VERSION)
 	gzip nnn-$(VERSION).tar
-	rm -rf nnn-$(VERSION)
+	$(RM) -r nnn-$(VERSION)
 
 clean:
-	rm -f $(BIN) nnn-$(VERSION).tar.gz
+	$(RM) -f $(BIN) nnn-$(VERSION).tar.gz
```
zmwangx: `$(RM)` already includes `-f`.

Makefile
```diff
@@ -23,30 +25,31 @@ all: $(BIN) $(PLAYER)
...
+
+.PHONY: install uninstall clean debug strip dist
```
zmwangx: `all` is also a phony target.

Makefile
rindeal: I know, I left it there in case someone would overwrite the variable, because in `clean` target all `rm`'s should have `-f` flag IMO.

jarun:
@szlin has updated offline he is fine with the changes.

I am guessing @shaggytwodope might be occupied with something pressing. He is due for another activity as well. Pulling in the changes.

I went through the latest iteration and things look good.

jarun:
Thanks @rindeal for your contribution!

PR #78, opened 2018-01-21
Makefile: use PKG_CONFIG instead of hardcoding.

maxice8:
This allows Exherbo that uses an arch prefix ( i686-linux-pc-gnu.pkg-config) before pkg-config to build cleanly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,15 @@ VERSION = 1.6
 PREFIX ?= /usr/local
 MANPREFIX ?= $(PREFIX)/share/man
 STRIP ?= strip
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

maxice8:
I have changed from Void Linux to Exherbo and maintainership of nnn and other packages i maintain from you might change. in the foreseeable future i will continue to maintain and update nnn, imgp, bcal, etc.

on the good side i now take requests to package stuff for Exherbo

jarun:
Thanks for the update! :+1:

PR #82, opened 2018-02-28
openSUSE: Add zypper instructions

avindra:
adds the command to install for openSUSE


jarun:
- openSUSE is a very popular distro and we probably do not need the instruction for its users.
- the openSUSE build system generates packages for other distros also.

I am closing this PR without merging.

PR #86, opened 2018-03-01
Minor tweaks on quitcd.bash

Jackymancs4:
Hello,
on MacOS 10.13.3 (and I think any BSD system) rm ask for confirmation before deleting ```/tmp/nnn```, and when it's done print out the name of the deleted file.
It's pretty annoying.

This PR makes no difference on Debian/Ubuntu.


jarun:
Thank you!

PR #88, opened 2018-03-02
Fix zsh path autocompletion

Nexion:
I'm not sure what that _guard function was supposed to do, but autocompletion of the path argument wasn't working with that. With this change the autocompletion is working both for options and a path.


jarun:
copy pasta errorestaaaa! Thanks!

Issue #90, opened 2018-03-04
Size is not showing correctly


jarun:
Steps to reproduce:

1. Create the file:

       fallocate -l 154140260 out.file
3. Check the file size in `nnn`:

       2018-03-04 22:02 146.100M  out.file
3. Check the file size in v1.7:

       2018-03-04 22:02  147.00M  out.file

154140260 / (1024*1024) = 146.999607

So `147.00M` is the correct approximate size.

The regression was introduced by 7be0726164442a83f47e5a9a0cdf2db343832d23.

jarun:
@arsv please take a look.

arsv:
Yeah, my bad. Carry is not handled properly when rounding.

https://github.com/jarun/nnn/pull/91

PR #91, opened 2018-03-04
fix rounding carry in coolsize()

arsv:
Fix for https://github.com/jarun/nnn/issues/90

Updated tests: https://gist.github.com/arsv/425db2e53f23209d3587ed1c061fa9de

### DIFF
--- a/nnn.c
+++ b/nnn.c
@@ -1442,6 +1442,7 @@ coolsize(off_t size)
 	static char size_buf[12]; /* Buffer to hold human readable size */
 	static int rem, i;
 	static int fdig; /* number of fractional digits to show */
+	static int frac;
 
 	i = 0;
 	rem = 0;
@@ -1452,12 +1453,16 @@ coolsize(off_t size)
 		++i;
 	}
 
-	rem = (1000 * rem) >> 10; /* convert 1024th fractions to 1000th */
 	fdig = 3;
+	frac = 1000;
+	rem = (10000 * rem) >> 10; /* convert 1024th fractions to 10000th */
+	rem = (rem / 10) + (rem % 10 >= 5 ? 1 : 0); /* round to 1000th */
 
 	/* Show 1 decimal for KB sizes and 2 decimals for MBs. */
-	if (i < 3) { --fdig; rem = (rem + 5) / 10; }
-	if (i < 2) { --fdig; rem = (rem + 5) / 10; }
+	if (i < 3) { --fdig; rem = (rem + 5) / 10; frac /= 10; }
+	if (i < 2) { --fdig; rem = (rem + 5) / 10; frac /= 10; }
+	/* carry, in case rounding above overflows the fractional part */
+	if (i < 3 && rem >= frac) { size++; rem = 0; }
 
 	if (i > 0)
 		snprintf(size_buf, 12, "%" PRId64 ".%0*i%c", size, fdig, rem, U[i]);


nnn.c
```diff
@@ -1454,10 +1455,13 @@ coolsize(off_t size)
 
 	rem = (1000 * rem) >> 10; /* convert 1024th fractions to 1000th */
 	fdig = 3;
```
jarun: We take care of kB and MB, but what about GB, TB and so on?

For example, in case of GB we show up to 3 decimal places. 

I see these diffs:

v1.7:

    2018-03-04 21:57   9.925G/ new_vids/
    2018-03-02 08:13   3.460G/ The.Big.Blue.1988.720p.BluRay.x264-x0r/
    2018-02-27 22:58   3.048G/ Cold Mountain 2003 (1080p Bluray x265 HEVC 10bit AAC 5.1 Tigole)/

Int calc:

    2018-03-04 21:57   9.924G/ new_vids/
    2018-03-02 08:13   3.459G/ The.Big.Blue.1988.720p.BluRay.x264-x0r/
    2018-02-27 22:58   3.047G/ Cold Mountain 2003 (1080p Bluray x265 HEVC 10bit AAC 5.1 Tigole)/

I think we are limiting both `fdig` and `frac` till MBs.

nnn.c
```diff
@@ -1454,10 +1455,13 @@ coolsize(off_t size)
...
+	frac = 1000;
 
 	/* Show 1 decimal for KB sizes and 2 decimals for MBs. */
-	if (i < 3) { --fdig; rem = (rem + 5) / 10; }
-	if (i < 2) { --fdig; rem = (rem + 5) / 10; }
+	if (i < 3) { --fdig; rem = (rem + 5) / 10; frac /= 10; }
```
jarun: We have to show accurate results till [yobibyte](https://en.wikipedia.org/wiki/Yobibyte). So I believe these if conditions should have a greater range and fit in a loop.

jarun: I missed the fact that we can show only upto 3 decimal places for GB and above.

jarun:
Please update the gist as well so we can run some profiling at the end.

arsv:
Hm, yeah, there's one more rounding problem there. Patch updated, tests updated.
This should match the old code, but it's not exactly correct. And same for sizes >= 1TB.

The problem: rem = size & 0x3FF. This only keeps the last 3 digits, 1/1024th of size. The old code tries to round the 4th digit, and tries to format 4+ decimals for 1TB and above, but there's no 4th digit there. This can be fixed, by keeping more digits in rem, but it won't match the numbers the old code produced. And I don't think there's any point in showing more than 3 decimals anyway, even for TB sizes and above.

arsv:
Gist updated. The last test there should be 9.977G, with the exact number that triggers 6/7 rounding in the last digit.

jarun:
Can we have a few test cases with TB? Specially corner cases like the one we found.

jarun:
Compilation errors with gcc 7.1.0:

```
~/G/nnn$ make
cc -O3 -Wall -Wextra -Wno-unused-parameter -D_GNU_SOURCE -I/usr/include/ncursesw  -o nnn nnn.c -lreadline -lncursesw -ltinfo
nnn.c: In function ‘coolsize’:
nnn.c:1468:26: warning: ‘%ld’ directive output may be truncated writing between 1 and 20 bytes into a region of size 12 [-Wformat-truncation=]
   snprintf(size_buf, 12, "%" PRId64 ".%0*i%c", size, fdig, rem, U[i]);
                          ^~~
nnn.c:1468:27: note: format string is defined here
   snprintf(size_buf, 12, "%" PRId64 ".%0*i%c", size, fdig, rem, U[i]);
nnn.c:1468:26: note: directive argument in the range [-9223372036854775808, 1025]
   snprintf(size_buf, 12, "%" PRId64 ".%0*i%c", size, fdig, rem, U[i]);
                          ^~~
In file included from /usr/include/stdio.h:936:0,
                 from /usr/include/ncursesw/curses.h:171,
                 from nnn.c:56:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:10: note: ‘__builtin___snprintf_chk’ output between 5 and 2147483671 bytes into a destination of size 12
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        __bos (__s), __fmt, __va_arg_pack ());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
nnn.c:1470:26: warning: ‘%ld’ directive output may be truncated writing between 1 and 20 bytes into a region of size 12 [-Wformat-truncation=]
   snprintf(size_buf, 12, "%" PRId64 "%c", size, U[i]);
                          ^~~
nnn.c:1470:27: note: format string is defined here
   snprintf(size_buf, 12, "%" PRId64 "%c", size, U[i]);
nnn.c:1470:26: note: directive argument in the range [-9223372036854775808, 1025]
   snprintf(size_buf, 12, "%" PRId64 "%c", size, U[i]);
                          ^~~
In file included from /usr/include/stdio.h:936:0,
                 from /usr/include/ncursesw/curses.h:171,
                 from nnn.c:56:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:10: note: ‘__builtin___snprintf_chk’ output between 3 and 22 bytes into a destination of size 12
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        __bos (__s), __fmt, __va_arg_pack ());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~/G/nnn$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc.real
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 7.1.0-10ubuntu1~16.04.york0' --with-bugurl=file:///usr/share/doc/gcc-7/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-7 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 7.1.0 (Ubuntu 7.1.0-10ubuntu1~16.04.york0) 
~/G/nnn$
```

jarun:
The following patch works:

```
~/G/nnn$ gd
diff --git a/nnn.c b/nnn.c
index 9364e16..14ef1de 100644
--- a/nnn.c
+++ b/nnn.c
@@ -1465,9 +1465,9 @@ coolsize(off_t size)
        if (i < 3 && rem >= frac) { size++; rem = 0; }
 
        if (i > 0)
-               snprintf(size_buf, 12, "%" PRId64 ".%0*i%c", size, fdig, rem, U[i]);
+               snprintf(size_buf, 12, "%lu.%0*i%c", size, fdig, rem, U[i]);
        else
-               snprintf(size_buf, 12, "%" PRId64 "%c", size, U[i]);
+               snprintf(size_buf, 12, "%lu%c", size, U[i]);
 
        return size_buf;
 }
~/G/nnn$
```

I believe it's reasonable given we are never going to print U64.

jarun:
There's a new difference now:

    fallocate -l 200064791 out.file

jarun:
New size - 190.81M
Old size - 190.80M

$ wcalc "200064791/(1024*1024)"
 = 190.796653

190.80 is the correct approximation.

jarun:
Considering our adventure so far, 2 things come to my mind

- it's tricky to ensure everything works fine without spending a considerable time framing accurate test cases
- the code has become _hacky_, and is not easy to understand for a new dev anymore (so maintenance becomes costly)

I am reverting to our old method of floating point calculation. It was definitely a very good approach but I think on modern processors the difference is negligible.

jarun:
gist from @arsv for tests:

```
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <inttypes.h>
#include <sys/time.h>

static char *
coolsize_f(off_t size)
{
	static const char * const U = "BKMGTPEZY";
	static char size_buf[12]; /* Buffer to hold human readable size */
	static int i;

	static long double rem;
	static const double div_2_pow_10 = 1.0 / 1024.0;

	i = 0;
	rem = 0;

	while (size > 1024) {
		if (rem >= 512) ++size;
		rem = size & (0x3FF); /* 1024 - 1 = 0x3FF */
		size >>= 10;
		++i;
	}

	snprintf(size_buf, 12, "%.*Lf%c", i, size + rem * div_2_pow_10, U[i]);
	return size_buf;
}

static char *
coolsize_i(off_t size)
{
	static const char * const U = "BKMGTPEZY";
	static char size_buf[12]; /* Buffer to hold human readable size */
	static int i;

	static off_t rem;
	static const double div_2_pow_10 = 1.0 / 1024.0;

	i = 0;
	rem = 0;

	while (size > 1024) {
		if (rem >= 512) ++size;
		rem = size & (0x3FF); /* 1024 - 1 = 0x3FF */
		size >>= 10;
		++i;
	}

	snprintf(size_buf, 12, "%.*f%c", i, size + (float)rem * div_2_pow_10, U[i]);
	return size_buf;
}

static const off_t sizes[] = {
	1234567890,
	154140260,
	200064791,
	1000,
	1023,
	1024,
	0,
	1,
	15,
	123,
	1234,
	12345,
	123456,
	1234567,
	12345678,
	123456789,
	1234567890,
	1000000,
	1100000,
	1200000,
	1500000,
	1600000,
	1800000,
	1990000,
	1990234,
	1024*1024,
	1024*1024*1024,
	(9LL*1024*1024*1024LL + 1000LL*1024*1024)
	//(9LL*1024*1024*1024*1024LL + 1023LL*1024*1024*1024),
};

static void test_static_cases(void)
{
	int i, n = sizeof(sizes) / sizeof(*sizes);

	for(i = 0; i < n; i++) {
		off_t size = sizes[i];

		char* str_f = coolsize_f(size);
		char* str_i = coolsize_i(size);

		if(!strcmp(str_f, str_i)) {
			printf("OK %20" PRId64 " %12s %12s\n", size, str_f, str_i);
		} else {
			printf("FAIL %" PRId64 "\n> %s\n> %s\n", size, str_f, str_i);
		}
	}
}

static void timediff(struct timeval* tv, struct timeval* t1, struct timeval* t0)
{
	time_t ds = t1->tv_sec - t0->tv_sec;
	long dus = t1->tv_usec - t0->tv_usec;

	if(dus < 0) {
		dus += 1000000;
		ds -= 1;
	}

	tv->tv_sec = ds;
	tv->tv_usec = dus;
}

static void dumptime(const char* tag, struct timeval* tv)
{
	printf("%s %li.%06lu us\n", tag, tv->tv_sec, tv->tv_usec);
}

static void test_relative_speed(void)
{
	struct timeval t0, t1;
	struct timeval ti, tf;
	unsigned i, n = 500*1000;

	srand48(17);
	gettimeofday(&t0, NULL);

	for(i = 0; i < n; i++)
		coolsize_f(lrand48());

	gettimeofday(&t1, NULL);
	timediff(&tf, &t1, &t0);

	srand48(17);
	gettimeofday(&t0, NULL);

	for(i = 0; i < n; i++)
		coolsize_i(lrand48());

	gettimeofday(&t1, NULL);
	timediff(&ti, &t1, &t0);

	printf("Timing %u iterations\n", n);
	dumptime("F time", &tf);
	dumptime("I time", &ti);
}

int main(void)
{
	test_static_cases();
	printf("\n");
	test_relative_speed();

	return 0;
}
```

jarun:
@arsv can you please review the following patch? It's written in the same line as yours and seems to be working.

```
diff --git a/nnn.c b/nnn.c
index dc2b8cf..02d4f84 100644
--- a/nnn.c
+++ b/nnn.c
@@ -1439,10 +1439,8 @@ coolsize(off_t size)
 {
 	static const char * const U = "BKMGTPEZY";
 	static char size_buf[12]; /* Buffer to hold human readable size */
-	static int i;
-
 	static off_t rem;
-	static const double div_2_pow_10 = 1.0 / 1024.0;
+	static int i;
 
 	i = 0;
 	rem = 0;
@@ -1453,7 +1451,47 @@ coolsize(off_t size)
 		++i;
 	}
 
-	snprintf(size_buf, 12, "%.*f%c", i, size + (float)rem * div_2_pow_10, U[i]);
+	if (i == 1) {
+		rem = (rem * 1000) >> 10;
+
+		rem /= 10;
+		if (rem % 10 >= 5) {
+			rem = (rem / 10) + 1;
+			if (rem == 10) {
+				++size;
+				rem = 0;
+			}
+		} else
+			rem /= 10;
+	} else if (i == 2) {
+		rem = (rem * 1000) >> 10;
+
+		if (rem % 10 >= 5) {
+			rem = (rem / 10) + 1;
+			if (rem == 100) {
+				++size;
+				rem = 0;
+			}
+		} else
+			rem /= 10;
+	} else if (i > 0) {
+		rem = (rem * 10000) >> 10;
+
+		if (rem % 10 >= 5) {
+			rem = (rem / 10) + 1;
+			if (rem == 1000) {
+				++size;
+				rem = 0;
+			}
+		} else
+			rem /= 10;
+	}
+
+	if (i > 0)
+		snprintf(size_buf, 12, "%lu.%0*lu%c", size, i, rem, U[i]);
+	else
+		snprintf(size_buf, 12, "%lu%c", size, U[i]);
+
 	return size_buf;
 }
 
```

arsv:
Seems fine to me. It could be possible to move common code out of if()s, but keeping each case completely separate has its benefits as well.

jarun:
> It could be possible to move common code out of if()s

True, but we have to use extra tmp vars. So I thought of keeping the units distinct. The results are correct in my tests so far. So I pushed it.

Thanks for the review! :+1:

PR #96, opened 2018-03-14
Make install(1) commands override-able

Dridi:
Hello,

I'm reviewing a package submission for `nnn` on Fedora and one thing that goes against our packaging policy is the install step. This is a trivial patch to solve that and maintain the same behavior by default.

https://bugzilla.redhat.com/show_bug.cgi?id=1548761

Dridi


jarun:
Thank you!

PR #98, opened 2018-03-14
nlay: use more portable shebang for bash.

maxice8:
Should make it able to run on more distros like `NixOS`, `OpenBSD` and whoever doesn't have /bin/bash really.


jarun:
Thank you! :-1:

PR #101, opened 2018-04-09
Source code release tarball build on Travis-CI

jolpaz:
This PR enables building the source code release tarball on Travis-CI along
with packages for linux distributions that were built thus far. Formal release
of source code tarball produced this way is important for OpenBSD (possibly
for other BSDs as well) port creation. There is no need to change the workflow
around release process, everything will work as it was with the addition of
said tarball. Example of released files can be found at https://github.com/jolpaz/nnn/releases


.travis.yml
```diff
@@ -22,11 +26,16 @@ script:
   - make clean;
 
 before_deploy:
+  - cd ..
```
jarun: Can we move this to _after_deploy_ so it's not even generated in case of a package build failure?

jarun:
Thank you!

jolpaz:
Cheers

PR #109, opened 2018-05-18
Clarify utils[7] to utils[UNKNOWN]

refacto:
We have a `#define UNKNOWN 7`, so let's use it.


jarun:
I saw the PR after I pushed in the change. Closing it.

PR #115, opened 2018-08-21
Fix backspace rendering and error handling.

codeliveroil:
@jarun - can you please review this change:

```
Fixes the following issues:
- When in filter mode, hitting backspace on the first character renders
  the backspace character on the screen along with the first character
  instead of just the search prompt (i.e. '/').
  For example, hitting backspace 3 times on the search string 'abc'
  produces this sequence:
     /abc
     /ab
     /a
     /a^?
- Go to the 'end' label if matches() fails (via regcomp()) when hitting
  backspace.
```

### DIFF
--- a/nnn.c
+++ b/nnn.c
@@ -1077,8 +1077,13 @@ filterentries(char *path)
 
 			wcstombs(ln, wln, REGEX_MAX);
 			ndents = total;
-			if (matches(pln) == -1)
-				continue;
+			char *tmppln = pln;
+			if (strcmp(pln, "") == 0) {
+				tmppln=".*"; //otherwise setfilter will fail for the empty string.
+			}
+			if (matches(tmppln) == -1) {
+				goto end;
+			}
 			redraw(path);
 			printprompt(ln);
 			continue;


jarun:
I can't reproduce this on Ubuntu. Which platform is it failing on?

jarun:
Also, did you notice this on master?

codeliveroil:
On macOS. Yes, the issue is in master since about 6 or so months.

codeliveroil:
On Ubuntu, I assume you still see /a even though pln is empty?

jarun:
> On Ubuntu, I assume you still see /a even though pln is empty?

No I don't.

Which release?

codeliveroil:
Yeah, just checked on Ubuntu and MacOS using the same terminal emulator on the same hardware and I only see the problem on macOS. Looks like "^?" is echo'ed each time backspace is hit.

jarun:
I believe you need to use the `Del` key? @zmwangx can you please help us here?

codeliveroil:
Correct, `ASCII 127` but that's also what I pass in Ubuntu and that seems to work fine.

jarun:
@zmwangx uses nnn on OSX. We need to wait for his response. I remember because the `Del` key issue was reported by him.

zmwangx:
Just tried, I can reproduce this on macOS 10.13.6 with ncurses 6.1, on my Mid-2015 rMBP 15'''s builtin keyboard (U.S. layout), with the delete key. I don't have my Extended Magic Keyboard with me because I'm traveling.

jarun:
@codeliveroil can you please make the patch specific to OS X?

codeliveroil:
Sure but what about this fix from my commit:
`- Go to the 'end' label if matches() fails (via regcomp()) when hitting backspace.`

Did you intend to `continue` or should it `goto end` if matches fails:
https://github.com/jarun/nnn/blob/master/nnn.c#L1080?

jarun:
Another note, can you please use the debug version and check why we are getting garbage `a^?` in case of the last `Del` press?

Use `make debug` and the debug logs would be in `/tmp/nnndbg`.

PR #119, opened 2018-10-19
Honour CPPFLAGS

reinerh:
Preprocessor flags are passed in CPPFLAGS.
Debian uses it to pass hardening options (-D_FORTIFY_SOURCE)
to the compiler.


jarun:
Thank you!

PR #123, opened 2018-11-02
Map BS (8) and DEL (127) key codes to SEL_BACK as well

detunized:
On some systems backspace key is producing the BS (8) code and on some others DEL (127). KEY_BACKSPACE is mapped to 263 which doesn't seem to work everywhere. I think there's no harm in supporting all these options out of the box.


jarun:
Thank you!

Issue #128, opened 2018-11-15
BLK_SHIFT may be wrong in some platforms


fengyichui:
In *sys/stat.h*, the macro `S_BLKSIZE` is the size of a block and its value not always 512 (eg: 1024 in Cygwin).
So the better way is to use `S_BLKSIZE`  instead of `BLK_SHIFT`.

jarun:
Fixed at #128.

jarun:
@fengyichui would you be interested in adding the package `nnn` to Cygwin?

Packaging docs: https://cygwin.com/packaging-contributors-guide.html

PR #129, opened 2018-11-15
Fix BLK_SHIFT may be wrong in some platforms #128

fengyichui:
Fix BLK_SHIFT may be wrong in some platforms

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3399,9 +3399,6 @@ int main(int argc, char *argv[])
 	char *ipath = NULL;
 	int opt;
 
-	// Get platform block shift
-	BLK_SHIFT = ffs(S_BLKSIZE) - 1;
-
 	/* Confirm we are in a terminal */
 	if (!isatty(0) || !isatty(1)) {
 		fprintf(stderr, "stdin or stdout is not a tty\n");
@@ -3413,6 +3410,7 @@ int main(int argc, char *argv[])
 		case 'S':
 			cfg.blkorder = 1;
 			nftw_fn = sum_bsizes;
+			BLK_SHIFT = ffs(S_BLKSIZE) - 1;
 			break;
 		case 'l':
 			cfg.showdetail = 0;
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -417,6 +417,18 @@ static uchar crc8fast(uchar const message[], size_t n)
 	return remainder;
 }
 
+/* Get platform block shift */
+static int get_blk_shift(void)
```
jarun: Please use `ffs()` or the appropriate one from the family (`man ffs`) based on the size required.

jarun: We already use it in the code.

fengyichui: oh, it's best. 👍

fengyichui: I want to use `ctz` to do ‘Count Tailing Zeros’, but I can't find the API.

src/nnn.c
```diff
@@ -3399,6 +3399,9 @@ int main(int argc, char *argv[])
 	char *ipath = NULL;
 	int opt;
 
+	// Get platform block shift
+	BLK_SHIFT = ffs(S_BLKSIZE) - 1;
```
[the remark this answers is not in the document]
fengyichui: But `-S` option need  to initialize it.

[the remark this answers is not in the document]
fengyichui: Put it into `-S` case ?

jarun:
Thank you!

jarun:
@fengyichui unfortunately this breaks on Termux where I couldn't find `S_BLKSIZE` defined in `sys/stat.h`.

I have raised https://github.com/termux/termux-app/issues/925.

fengyichui:
@jarun We can add the following code to the front of `nnn.c`.

```c
#ifndef S_BLKSIZE
#define S_BLKSIZE 512
#endif
```

fengyichui:
@jarun We can refer to the source code of [ncdu](https://dev.yorhel.nl/ncdu).

jarun:
I have already added this at https://github.com/jarun/nnn/commit/362fd8ce78abe0115788b327eb16f4f9751506e7.

PR #131, opened 2018-11-22
Annotate ^Y/DSUSP issue on BSDs caused by `stty`

Gumnos:
On at least FreeBSD & OpenBSD, typing `^Y` will send a `DSUSP` signal, suspending the process. Annotate the issue along with a solution to allow `^Y` to work on these.

There may be similar issues on NetBSD or possibly on OSX, but I don't have either available for testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,6 +504,10 @@ Enable volume management in your DE file manager and set removable drives or med
 
 `nnn` might not handle keypresses correctly when used with tmux (see issue #104 for more details). Set `TERM=xterm-256color` to address it.
 
+#### BSD terminal issue
+
+By default in OpenBSD & FreeBSD, `stty` maps `^Y` to `DSUSP`. This means that typing `^Y` will suspend `nnn` as if you typed `^Z` (you can bring `nnn` back to the foreground by issuing `fg`) instead of entering multi-copy mode. You can check this with `stty -a`. If it includes the text "`dsusp = ^Y`", issuing `stty dsusp undef` will disable this `DSUSP` and let `nnn` receive the `^Y` instead.
+
 #### WHY FORK?
 
 `nnn` was initially forked from [noice](http://git.2f30.org/noice/) but is significantly [different](https://github.com/jarun/nnn/wiki/nnn-vs.-noice) today. I chose to fork because:


jarun:
Thank you for this valuable contribution!

jarun:
I have added <kbd>Y</kbd> as an additional selection shortcut at 5df231947392475ced257a6ba5e4cec3c3fbb5a9.

jarun:
@Gumnos what does `nnn` use as the opener in OpenBSD? Is it `open(1)` or `xdg-open`?

Gumnos:
`xdg-open` is available as part of the `xdg-utils` package which can
be installed with

    $ doas pkg_add xdg-utils

if needed.

That said, I tend to just use it in the terminal or via SSH so I
don't really poke at the X related features.

jarun:
So is `open(1)` the default on OpenBSD?

jarun:
Thank you so much!

jarun:
@Gumnos Can you confirm if https://github.com/jarun/nnn/issues/143 is an issue on OpenBSD?

jarun:
or the BSD family in general?

gadkadosh:
This annotation is no longer visible in the readme as far as I can tell, yet the issue is still there, and Y is no longer an alternative to ^Y. It took me some time to realize what was going on on MacOS. I can confirm it does the same as in FreeBSD, sends `DSUSP` signal unless undefined.

jarun:
The How to section has moved here: https://github.com/jarun/nnn/wiki/How-to

It's linked twice from the README. But I see what you mean. I will add a troubleshooting section and pull the _annoying_ stuff back to README. ;)

jarun:
@gadkadosh Done!

gadkadosh:
Great! I would suggest adding MacOS, probably quite popular these days. Any chance to get again an alternative key binding?

jarun:
> I would suggest adding MacOS

Please raise a PR. I am not on iOS and my description may not be up to the mark.

> Any chance to get again an alternative key binding?

Why not? Which one would you prefer? Something without `^` would be great so we meet both ends.

jarun:
I have added key <kbd>K</kbd> to toggle selection.

PR #138, opened 2018-11-24
Add neovim support

KillTheMule:
Hey!

I've added neovim support to your plugin script. I had to move some things around, but nothing fancy. I've manually tested that it still works on vim/gvim, so all should be good.

(e) Ah yeah, I've also added printing an error if the tempfile isn't readable. I'd say that's a serious error (not supposed to happen), and should be reported to the user.

### DIFF
--- a/scripts/vim-plugin/nnn-picker.vim
+++ b/scripts/vim-plugin/nnn-picker.vim
@@ -1,4 +1,4 @@
-" vim plugin to use nnn as a file picker
+" vim/neovim plugin to use nnn as a file picker
 " Closely follows and inspired by the vim_file_chooser plugin for ranger.
 "
 " Author: Arun Prakash Jana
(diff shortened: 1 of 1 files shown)


jarun:
Can't thank you enough! :+1:

mcchrish:
@KillTheMule opening a file from nnn doesn't open in neovim but opens to the default program for that file type. e.g. opening a shell file opens it in xcode.

KillTheMule:
@mcchrish Are you sure you're using it correctly? I thought so at first, but then observed the same behavior with vim, so it must be intentional. Did you press `<Enter>` on a file? That opens it in the default viewer alright, both in vim and nvim. Select the file in selection mode and quit nnn, that opens the file in nvim (just as is the behavior in vim).

mcchrish:
@KillTheMule sorry about that. I was using an old version nnn that doesn't have the selection mode. Just updated and works now. 👍

jarun:
> edit: agree. I think pressing `<enter>` seems more intuitive.

I can add a patch which would change the behaviour so when you press `Enter` it picks and quits.

So, normal quit would

- either copy selection if it exists (this is useful as it supports multiple selection)
- or do nothing

KillTheMule:
Sounds about right. I'd also suggest adding a key for "whatever I selected, just quit without passing back anything". Maybe `Q` is the right for that.

jarun:
I think it's becoming unnecessarily complex from this point. Please share a simpler workflow if you have.

KillTheMule:
Oh, don't worry about my workflow, I don't have one since I'm not using nnn. I just threw out suggestions because of how I imagined others might be using it from (neo)vim, but feel free to totally ignore them (especially considering that you have more to consider than how one might use this from (neo)vim).

PR #142, opened 2018-11-26
Replace single dot with source for fish shell

josalmi:
Single dot has been deprecated and can cause issues with users on newer version: https://github.com/fish-shell/fish-shell/issues/310


jarun:
Thank you!

Issue #143, opened 2018-11-27
MAC issue: cp, mv, rm don't work - xargs: illegal option -- d


kjnez:
`xargs:
 illegal option -- d
 usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
              [-L number] [-n number [-x]] [-P maxprocs] [-s size]
              [utility [argument ...]]
`
The above happens when I try to copy, move or delete a file. I'm using version 2.1 installed on a mac using Homebrew.

jarun:
The code was tested on Ubuntu where options `-a` and `-d` are available: http://manpages.ubuntu.com/manpages/bionic/man1/xargs.1.html

But I see they are not on OS X: https://ss64.com/osx/xargs.html

I think we need to do the following here:

- Have a check for __APPLE__ and it's defined
  - `remove -d`
  - use cat to pipe our file

Would it be possible for you to send in a patch?

zmwangx:
I'm afraid looking at immediate surrounding code doesn't give me an idea of what the code is doing (e.g., I'm not sure what this `g_cppath` is, and why both `-0` and `-d '\n'` are used, which are ostensibly conflicting), so it's hard for me to cook up a patch quickly.

In general, using coreutils or GNU's utilities as basis for cross-platform code is a very bad idea because they have a lot more options (I love coreutils, but still). Better look at FreeBSD/OpenBSD too, or even stick to POSIX. Here are my scripts for opening FreeBSD/OpenBSD/POSIX man pages:

`freebsdman`:

```sh
#!/usr/bin/env bash
if [[ "$1" == [12345678] ]]; then
    section=$1
    shift
fi

for topic; do
    open "https://www.freebsd.org/cgi/man.cgi?query=$topic${section:+&sektion=$section}"
done
```

https://www.freebsd.org/cgi/man.cgi?query=xargs

`openbsdman`:

```sh
#!/usr/bin/env bash
if [[ "$1" == [12345678] ]]; then
    section=$1
    shift
fi

for topic; do
    if [[ -n $section ]]; then
        open "http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man$section/$topic.$section"
    else
        open "http://www.openbsd.org/cgi-bin/man.cgi?query=$topic"
    fi
done
```

https://man.openbsd.org/xargs

`posixman`:

```sh
#!/usr/bin/env bash
for utility; do
[...]

PR #145, opened 2018-11-27
Fix for `rm` not working on macOS (macOS doesn't support -I)

detunized:
I used `-i` instead, which could be annoying, since it always prompts (-I prompts for 3 or more files).

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -364,6 +364,13 @@ static char * const utils[] = {
 #define STR_COPY_ID 4
 #define STR_DATE_ID 5
 
+/* rm args */
+#ifdef __APPLE__
+#define RM_ARGS "-ir" // macOS doesn't support -I
+#else
+#define RM_ARGS "-Ir"
+#endif
+
 static const char messages[][16] = {
 	"nftw failed",
 	"HOME not set",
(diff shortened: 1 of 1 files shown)


jarun:
Please rebase with the latest branch. Please confirm if `-a` works.

I am noticing a problem with cat: the interactive mode doesn't wait for user input.

jarun:
Please look into https://github.com/jarun/nnn/issues/143#issuecomment-442115556 as well.

I think the APPLE case should be within APPLE macro. The options differ from Ubuntu and I think it's better to keep these platform specific.

jarun:
I meant to say there are other issues still remaining. Please test latest master and you will notice those.

A summary is available here: https://github.com/jarun/nnn/issues/143#issuecomment-442115556

jarun:
I have changed `-I` to `-i` for all platforms.

The only issue that remains is the `-t` option for `cp` and `mv`. Please see how `cp` and `mv` can work without that option on OS X. I do not have any means to test on Mac.

jarun:
Relevant commit for you reference: 361c18064e26fa3028dda81258acb4ed5aa7c486.

jarun:
Mac related issues should be fixed by 2a4cc85659f247704811da02338a8137908522b8.
I could eliminate `-t` as well. Please test and confirm.

PR #158, opened 2018-11-30
Add Solus to README

davidjharder:
`nnn ` is now packaged for Solus.


davidjharder:
@jarun: I will take a look. Some are already packaged.

jarun:
OK, cool!

PR #162, opened 2018-12-12
Add a snap

evandandrea:
Not wanting to leave nnn out, I also made a snap of it :) Instructions for building and testing it are the same as on the other PR I made:
https://github.com/jarun/bcal/pull/22

This one is a "classic" snap, meaning it has full access to the filesystem. As such, the first upload of it would need to be reviewed and approved with an accompanying explanation ([for example](https://forum.snapcraft.io/t/classic-confinement-request-for-valgrind/8744)). Happy to help with that.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,27 @@
+name: nnn
+version: git
+summary: The fastest terminal file manager ever written
+description: |
+  nnn is probably the fastest and most resource-sensitive file manager you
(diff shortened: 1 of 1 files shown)


jarun:
Would it be possible to handle this as a part of snapcrafters? e.g.: https://github.com/snapcrafters/googler

jarun:
I am closing this with (and the other PR) sincere apologies. There are users who asked for a universal format but I already find myself in a very tight schedule. PS if you can do something about my suggestion above.

PR #166, opened 2018-12-16
Makefile portability tweaks

sjmulder:
Very similar to https://github.com/jarun/bcal/pull/25

Note regarding curses: the BSD (& macOS) built in curses generally support all you need hence **n**curses is not actually needed. pkgsrc can detect the correct curses to use and I’ve used `{CFLAGS,LDLIBS}_CURSES` in the environment to pass those on.

The package itself needs to go through review first; I’ll PR Makefile when it’s live.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,10 +33,10 @@ all: $(BIN)
 $(SRC): src/nnn.h
 
 $(BIN): $(SRC)
-	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $^ $(LDLIBS)
+	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
 debug: $(SRC)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,10 +33,10 @@ all: $(BIN)
 $(SRC): src/nnn.h
 
 $(BIN): $(SRC)
-	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o $@ $^ $(LDLIBS)
+	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $^ $(LDLIBS)
```
jarun: `$CPPFLAGS` was added by https://github.com/jarun/nnn/pull/119.

sjmulder: Ah, I confused CXXFLAGS (for C++) and CPPFLAGS (for the C preprocessor). Will charge it.

PR #169, opened 2018-12-17
describe that homebrew is installer for macOS

avelino:
Hardly a person to use macOS and this familiar with homebrew search by homebrew


jarun:
Please note the tagline of the project and it popularity: https://github.com/Homebrew/brew

That would be all.

PR #178, opened 2019-01-02
Document how to install fish quitcd function

QasimK:
It was not clear that the name of the file should match the name of the function.

---

I really like this file browser. I have tried a few in the past but they never clicked for me, sometimes because they were a little too difficult to get to grips with. `nnn`, on the other hand, was very intuitive. The keyboard shortcuts are clear, once I figured out how to open help :smile:. I am also using this on my Raspberry Pi so I love the low-resource design behind it.

### DIFF
--- a/scripts/quitcd/quitcd.fish
+++ b/scripts/quitcd/quitcd.fish
@@ -1,3 +1,6 @@
+# Rename this file to match the name of the function
+# e.g. ~/.config/fish/functions/n.fish
+
 export NNN_TMPFILE="/tmp/nnn"
 
 function n --description 'support nnn quit and change directory'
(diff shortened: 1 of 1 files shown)


scripts/quitcd/quitcd.fish
```diff
@@ -1,3 +1,6 @@
+# Rename this file to match the name of the function
+# e.g. ~/.config/fish/functions/n.fish
+
```
jarun: Please add:

    or, add the lines to the `config.fish` file.

QasimK: Done!

jarun:
Thank you!

PR #191, opened 2019-01-15
Fix warning of implicit declaration of function 'strcasestr'

szlin:
According to manpage of strcasestr [1], the macro function needs to be added.

[1] https://linux.die.net/man/3/strcasestr

Signed-off-by: SZ Lin (林上智) <szlin@debian.org>

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -29,6 +29,9 @@
  */
 
 #ifdef __linux__
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
 #if defined(__arm__) || defined(__i386__)
 #define _FILE_OFFSET_BITS 64 /* Support large files on 32-bit */
 #endif


jarun:
Travis fails. I guess you'll have to check (`#ifndef`) if `_GNU_SOURCE` is already defined. And if not, then define it.

src/nnn.c
```diff
@@ -28,6 +28,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#define _GNU_SOURCE
```
jarun: This may be required only for Linux (but not on macOS/*BSD).

Or even better if we can put it under `#ifdef __GLIBC__`

jarun: Also, we need to check if it's already defined or not.

szlin: Thanks for the review, I've updated the PR.

jarun:
Thank you!

PR #204, opened 2019-02-05
Clarify the difference between copy and select

breisa:
The intermixed use of 'copied' and 'selected' is quite confusing to me. This pull request trys to make things more clear. For example the displayed message 'x files copied' looks like files had been copied although they only got added to the selection list. Additionally it is now shown if the file got selected/copied when cherry-picking with <kbd>^K</kbd>.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3493,7 +3493,8 @@ nochange:
 				if (copier)
 					spawn(copier, NULL, NULL, NULL, F_NOTRACE);
 			}
-			printmsg(newpath);
+			mvprintw(LINES - 1, 0, "selected %s%s\n",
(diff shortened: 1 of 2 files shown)


jarun:
Thank you so much! There's no great software without great documentation. :+1:

src/nnn.c
```diff
@@ -3493,7 +3493,8 @@ static void browse(char *ipath)
 				if (copier)
 					spawn(copier, NULL, NULL, NULL, F_NOTRACE);
 			}
-			printmsg(newpath);
+			mvprintw(LINES - 1, 0, "selected %s%s\n",
```
jarun: I'll keep this as it is. I think the documentation should suffice. The idea is to just flash the filename.

PR #226, opened 2019-03-17
Added a user-script to send selected files using kde-connect.

juacq97:
The script uses `kdeconnect-cli` to send the files to the linked device.
Uses the `.nnncp` file to `cat` the selected files.
You can get system notifications using `notify-send`. This is disabled
by default.

### DIFF
--- a/user-scripts/kdeconnect
+++ b/user-scripts/kdeconnect
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+# Description: Send the selected files to your Android device using kdeconnect-cli. You must have installed and configured kdeconnect both on the Android device and on the PC.
+#
+# Shell: Bash
+# Author: juacq97
+
+id=$(kdeconnect-cli -a --id-only | awk '{print $1}')
+if [ $(find ~/.nnncp) ]
+then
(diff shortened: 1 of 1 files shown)


user-scripts/kdeconnect
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+# Description: Send the selected files to your Android device using kdeconnect-cli. You must have installed and configured kdeconnect both on the Android device and on the PC.
+#
+# Shell: Bash
```
jarun: If this is bash-specific, please change the first line to

    #!/usr/bin/env bash

jarun:
Also, please add this to the table here: https://github.com/jarun/nnn/blob/master/user-scripts/README.md

with a short description (e.g. _send selected files to KDEConnect_).

jarun:
Please rename the script to kdeconnect.sh and make it executable.

jarun:
Merged. Thank you for your contribution!

jarun:
@juacq97 would it be possible for you to make the script POSIX compliant?

I have made some changes to some other scripts at commit b6c66a78c01384ea1cc255c5c16db4a436136109 you can refer to. I don't have KDE so thought of requesting you.

PR #237, opened 2019-03-25
Recognize End/Home keys for editing

dbotw:


src/nnn.c
```diff
@@ -1745,6 +1745,12 @@ static char *xreadline(char *prefill, char *prompt)
 				}
 			} else {
 				switch (*ch) {
+				case KEY_END:
```
jarun: Please place these below KEY_DC handling. The probability of using immediate edit/arrow keys is more than jumping to beginning/end.

jarun: Also please ensure the behavior remains the same as when you press <kbd>^A</kbd>and  <kbd>^E</kbd> in filter mode. (I tested with current change, looks fine).

jarun:
Note for self: update wiki: https://github.com/jarun/nnn/wiki/hacking-nnn#prompt-shortcuts

jarun:
Thank you!

PR #244, opened 2019-04-17
Added two scripts to change the wallpaper

juacq97:
The one named `nitrogen` allow to change the wallpaper using
nitrogen. The one named `pywal` allows to change the wallpaper and the
terminal color scheme using `pywal`.
Both of them need to put the cursor over an image, there's no need to
select the file with space.


jarun:
Please rename the file `nitrogen` to `nwal` (short and sweet).

user-scripts/README.md
```diff
@@ -8,9 +8,11 @@
 | imgur | Upload an image file to imgur |
 | kdeconnect | Send selected files to Android device using kdeconnect-cli |
 | ndiff | File and directory diff for selection using vimdiff |
+| nitrogen | Set the selected image as wallpaper using nitrogen |
```
jarun: Please rename the file to `nwal`. Note that the entry would move below `nlaunch` alphabetically with that.

user-scripts/nitrogen
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
jarun: Please try to make it work with just `sh`? We try to stay `bash`-independent.

user-scripts/pywal
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
jarun: Please try to make it work with just `sh`.

user-scripts/nitrogen
```diff
@@ -0,0 +1,15 @@
...
+
+# Description: Set the selected image as wallpaper using nitrogen. Just put the cursor on a image and run the script. 
+#
+# Shell: POSIX Compliant
+# Author: juacq97
+
+if ! [ -z "$1" ]; then
+    if [ $(mimetype --output-format %m $1 | awk -F '/' '{print $1}') == "image" ]; then
+	nitrogen --set-zoom-fill --save $1
+	notify-send -a "nnn" "Wallpaper changed!"
```
jarun: We need to comment out notify-send in line 11 and line 13 as it's not platform-independent. Please follow the `kdeconnect` script.

user-scripts/pywal
```diff
@@ -0,0 +1,15 @@
...
+
+# Description: Set the selected image as wallpaper using pywal and change the terminal color schemes. Just put the cursor on a image and run the script. 
+#
+# Shell: POSIX Compliant
+# Author: juacq97
+
+if ! [ -z "$1" ]; then
+    if [ $(mimetype --output-format %m $1 | awk -F '/' '{print $1}') == "image" ]; then
+	wal -i $1
+	notify-send -a "nnn" "Wallpaper changed!"
```
jarun: Same goes for notify-sends int his script.

user-scripts/nwal
```diff
@@ -0,0 +1,15 @@
...
+#
+# Shell: POSIX Compliant
+# Author: juacq97
+
+if ! [ -z "$1" ]; then
+    if [ $(mimetype --output-format %m $1 | awk -F '/' '{print $1}') == "image" ]; then
+	nitrogen --set-zoom-fill --save $1
+	notify-send -a "nnn" "Wallpaper changed!"
+    else
+	notify-send -a "nnn" "You need to select a image"
```
jarun: You need to select **an** image.

user-scripts/pywal
```diff
@@ -0,0 +1,15 @@
...
+    else
+	notify-send -a "nnn" "You need to select a image"
```
jarun: You need to select **an** image.

user-scripts/nwal
jarun: How about the message "No image selected"

juacq97: "No image selected" sounds better

jarun: Cool. Please use that string.

jarun:

user-scripts/nwal
juacq97: Sorry for the misspellings, English is not my native language and I miss some things

jarun: No problem at all!

jarun:
If you have the time, can you see if it's possible to come up with a script that can show images using `ueberzug` (single instance) and then copy the name of the last file opened in the viewer to the system clipboard?

People want the image preview capability but I would like to offload the capability to a distinct script rather than within `nnn`.

jarun:
There is a similar project for `vifm`: https://github.com/cirala/vifmimg

See if that helps.

juacq97:
I'll see what I can do, if I get a functional script I'll do a pull request

jarun:
That's very kind of you. Thanks!

PR #245, opened 2019-04-17
Fixed the problem related to notify-send

juacq97:
I just realized that I forgot to comment the `else` in both scripts. I fixed it, sorry for the inconvenience.


jarun:
I have already pushed this. :)
Please check master.

PR #248, opened 2019-04-20
Differentiate half-page Ctrl+U/D and pageful PageDN/UP movement keys

elig0n:


jarun:

PR #253, opened 2019-05-08
Added mouse support

bruce-hill:
This change adds mouse support for:
- Click on a context number to go to that context
- Click on a file or folder's line to move the cursor to that file
- Double click on a file or folder's line to open it

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -871,6 +871,7 @@ static bool initcurses(void)
 	nonl();
 	//intrflush(stdscr, FALSE);
 	keypad(stdscr, TRUE);
+	mousemask(BUTTON1_CLICKED | BUTTON1_DOUBLE_CLICKED, NULL);
 	curs_set(FALSE); /* Hide cursor */
 	start_color();
 	use_default_colors();
@@ -3034,6 +3035,7 @@ static void browse(char *ipath)
 	bool dir_changed = FALSE;
 	struct stat sb;
 	char *path, *lastdir, *lastname, *dir, *tmp;
+	MEVENT event;
 
 	atexit(dentfree);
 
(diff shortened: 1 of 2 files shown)


jarun:
I will review and test it. Thank you!

jarun:
After opening 1 or 2 files, the double click is not responsive. Any ideas?

jarun:
After opening 1 or 2 media files with double click, mouse clicks stop responding. Any ideas? I am using xfce4-terminal in drop-down mode.

jarun:
@bruce-hill with mouse support selection wouldn't work. Can we enable mouse support using a config var i.e., on a need basis?

bruce-hill:
> Can we enable mouse support using a config var i.e., on a need basis?

I don't really see much value to disabling mouse support with a config var. If mouse support is disabled, you can't click on stuff, but "not clicking on stuff" is something you can do anyway if you have the mouse code enabled and it's not working as desired.

I was digging around a little and found this documentation that might explain why you're seeing buggy behavior with xfce that I'm not seeing with my terminals:

> **Bugs**
>Mouse events under xterm will not in fact be ignored during cooked mode, if they have been enabled by mousemask. Instead, the xterm mouse report sequence will appear in the string read.
>
>Mouse events under xterm will not be detected correctly in a window with its keypad bit off, since they are interpreted as a variety of function key. Your terminfo description should have kmous set to "\E[M" (the beginning of the response from xterm for mouse clicks). Other values for kmous are permitted, but under the same assumption, i.e., it is the beginning of the response.
>
>Because there are no standard terminal responses that would serve to identify terminals which support the xterm mouse protocol, ncurses assumes that if your $TERM environment variable contains "xterm", or kmous is defined in the terminal description, then the terminal may send mouse events.

(from: https://linux.die.net/man/3/mousemask )

jarun:
Cool! Thanks for the details. I am able to select if I press <kbd>Shift</kbd> so my problem is gone.

PR #254, opened 2019-05-08
Fix CLI mode prompt to have "y/N" (instead of "y/Y")

jmoudrik:
Original prompt had a typo.


jmoudrik:
It seems that all the prompts have the same structure; I'll open an issue, this seems like my misunderstanding.

jarun:
The prompts accept y and Y as yes, anything else is a no.

PR #255, opened 2019-05-08
Make yes/no prompts have the "y/N" structure.

jmoudrik:
`nnn` yes / no prompts have currently the structure of "y/Y". E.g. :
```
r = get_input("overwrite? [y/Y]");
if (r != 'y' && r != 'Y') {
// do the stuff
```

This is non-intuitive (to me at least), as imo usually:
- the "/" is understood as xor, while here the 2 choices separated by the slash are the same
- the capital letter usually marks default that gets triggered upon hitting enter.

And thus "y/Y" can be understood as (this is how I understood it at first, and what triggered this PR), type "y" for yes and "Y" for no (default choice).

This PR changes this to
"y/N"
instead to make it more aligned with the intuition above.

With the PR, the logic is however not fully explicit any more (== does not list all the keys required for making the yes), but I do not think this is worse problem than the potential ambiguity above.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3812,7 +3812,7 @@ nochange:
 
 			/* Confirm if app is CLI or GUI */
 			if (sel == SEL_OPENWITH) {
-				r = get_input("cli mode? [y/Y]");
+				r = get_input("cli mode? [y/N]");
 				(r == 'y' || r == 'Y') ? (r = F_CLI)
 						       : (r = F_NOWAIT | F_NOTRACE | F_MULTI);
 			}
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -594,7 +594,7 @@ static void xdelay(void)
 
 static char confirm_force(void)
 {
-	int r = get_input("use force? [y/Y]");
+	int r = get_input("use force? [y/N]");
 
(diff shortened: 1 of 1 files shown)


jarun:
Instead of this, make it "y/Y confirms".

jarun:
Leave it as is. No need to waste few bytes in this case. "y/Y" indicates reasonably that pressing y or Y confirms.

jmoudrik:
@jarun in that case, I think that at least "yY" would still be more clear than the current option; and save a few bytes. Agreed?

jmoudrik:
It is mundane, agreed, but the ux can be better here. But I definitely - as well as you - do not want to spend too much time discussing "y/N" or "yY" or "y/Y"; it is your call, I do not have more arguments to add.

I have plenty other work I wanna be spending my time on; and currently, I do not have a need for more of it. I installed `nnn` just today (and the yes/no stuff is the first thing that sticked out after 2mins of using it), if I end up using it, I will definitely share any further suggestions / open PR's for stuff I customize / improve.

jarun:
@jmoudrik this has been taken care of in commit a4c38d85477ff889b2aa863e2a5b18c72dd266fb.

PR #259, opened 2019-05-09
Adding scrolloff option

bruce-hill:
In the style of vim's `scrolloff` option, this change makes it so nnn will keep the active item a minimum of `SCROLLOFF` lines from the top and bottom of the screen, instead of always at the center. (The old behavior can be replicated by setting `SCROLLOFF` to a large number like 9999) This means the scroll position is tracked separately from the current item, which has the added benefit of simplifying the mouse click and line-rendering logic a bit.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3170,9 +3170,9 @@ nochange:
 				// Single click just selects, double click also opens
 				if (event.bstate != BUTTON1_DOUBLE_CLICKED)
 					break;
-				// fallthrough
+				// fallthrough to select the file
 			} else
-				goto nochange;
+				goto nochange; // fallthrough
 		case SEL_NAV_IN: // fallthrough
 		case SEL_GOIN:
 			/* Cannot descend in empty directories */
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -133,6 +133,7 @@
 #define DOT_FILTER_LEN 7
 #define ASCII_MAX 128
 #define EXEC_ARGS_MAX 8
+#define SCROLLOFF 5
 
 /* Entry flags */
 #define DIR_OR_LINK_TO_DIR 0x1
@@ -272,7 +273,7 @@ static context g_ctx[CTX_MAX] __attribute__ ((aligned));
 
 static struct entry *dents;
 static char *pnamebuf, *pcopybuf;
-static int ndents, cur, total_dents = ENTRY_INCR;
+static int ndents, cur, curscroll, total_dents = ENTRY_INCR;
 static int xlines, xcols;
 static uint idle;
 static uint idletimeout, copybufpos, copybuflen;
@@ -1655,6 +1656,8 @@ end:
 	if (*ch != '\t')
 		g_ctx[cfg.curctx].c_fltr[0] = g_ctx[cfg.curctx].c_fltr[1] = '\0';
 
+	curscroll = MAX(0, cur - SCROLLOFF);
+
 	curs_set(FALSE);
 	settimeout();
 
@@ -2882,6 +2885,8 @@ static void populate(char *path, char *lastname)
 		cur = 0;
 	else
 		cur = dentfind(lastname, ndents);
+
+	curscroll = MAX(0, cur - SCROLLOFF);
 }
 
 static void redraw(char *path)
@@ -2891,7 +2896,8 @@ static void redraw(char *path)
 
 	int ncols = (xcols <= PATH_MAX) ? xcols : PATH_MAX;
 	int lastln = xlines;
-	int nlines = MIN(lastln - 4, ndents), i, attrs;
+	int scrolloff = MIN(SCROLLOFF, (xlines-4)>>1);
+	int i, attrs;
 	char buf[12];
 	char c;
 
@@ -2899,6 +2905,13 @@ static void redraw(char *path)
 
 	/* Clear screen */
 	erase();
+
+	if (ndents <= xlines - 4)
+		curscroll = 0;
+	else if (cur < curscroll + scrolloff)
+		curscroll = MAX(0, cur - scrolloff);
+	else if (cur > curscroll + (xlines - 4) - scrolloff - 1)
+		curscroll = MIN(ndents - (xlines - 4), cur - (xlines - 4) + scrolloff + 1);
 
 #ifdef DIR_LIMITED_COPY
 	if (cfg.copymode)
@@ -2965,18 +2978,8 @@ static void redraw(char *path)
 	}
 
 	/* Print listing */
-	if (cur < (nlines >> 1)) {
-		for (i = 0; i < nlines; ++i)
-			printptr(&dents[i], i == cur, ncols);
-	} else if (cur >= ndents - (nlines >> 1)) {
-		for (i = ndents - nlines; i < ndents; ++i)
-			printptr(&dents[i], i == cur, ncols);
-	} else {
-		const int odd = ISODD(nlines);
-
-		nlines >>= 1;
-		for (i = cur - nlines; i < cur + nlines + odd; ++i)
-			printptr(&dents[i], i == cur, ncols);
+	for (i = curscroll; i < ndents && i < curscroll + (xlines - 4); ++i) {
+		printptr(&dents[i], i == cur, ncols);
 	}
 
 	/* Must reset e.g. no files in dir */
@@ -3157,10 +3160,7 @@ nochange:
 
 			// Handle clicking on a file:
 			if (2 <= event.y && event.y < xlines - 2) {
-				// Get index of the first file listed on-screen:
-				r = MAX(0, MIN(cur-((xlines-4)>>1), ndents-(xlines-4)));
-				// Add the mouse click position to get the clicked file:
-				r += event.y - 2;
+				r = curscroll + (event.y - 2);
 
 				if (r >= ndents)
 					goto nochange;
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2017,8 +2017,8 @@ static char *coolsize(off_t size)
 {
 	static const char * const U = "BKMGTPEZY";
 	static char size_buf[12]; /* Buffer to hold human readable size */
-	static off_t rem;
-	static int i;
+	off_t rem;
+	int i;
 
 	rem = i = 0;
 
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -2891,7 +2896,8 @@ static void redraw(char *path)
 
 	int ncols = (xcols <= PATH_MAX) ? xcols : PATH_MAX;
 	int lastln = xlines;
-	int nlines = MIN(lastln - 4, ndents), i, attrs;
+	int scrolloff = MIN(SCROLLOFF, (xlines-4)>>1);
```
jarun: Please use spaces between token:

    (xlines-4)>>1


    (xlines - 4) >> 1

src/nnn.c
```diff
@@ -3170,6 +3170,7 @@ static void browse(char *ipath)
 				// Single click just selects, double click also opens
 				if (event.bstate != BUTTON1_DOUBLE_CLICKED)
 					break;
+				// fallthrough to select the file
```
jarun: This comment is not required. Please rebase to the latest master.

src/nnn.c
```diff
@@ -2900,6 +2906,13 @@ static void redraw(char *path)
 	/* Clear screen */
 	erase();
 
+	if (ndents <= xlines - 4)
```
jarun: Please pre-calculate `(xlines - 4)` in a tmp var (`entries`???) and use it.

src/nnn.c
```diff
@@ -3274,21 +3275,21 @@ static void browse(char *ipath)
...
-			break;
-		case SEL_PGUP:
-			if (cur > 0)
-				cur -= MIN((xlines - 4), cur);
-			break;
+		case SEL_PGDN: // fallthrough
 		case SEL_CTRL_D:
-			if (cur < ndents - 1)
-				cur += MIN((xlines - 4) / 2, ndents - 1 - cur);
+			r = sel == SEL_PGDN ? (xlines - 4) - 1 : (xlines - 4) / 2;
```
jarun: Please pre-calculate `(xlines - 4)` and use it.

src/nnn.c
bruce-hill: Sure, I like "onscreen", since it's the number of onscreen files.

jarun:
Works like a charm! Thank you for the marvelous idea. :+1: Much more comfortable than having to jump to the middle now and then. :)

src/nnn.c
jarun: Please use a tmp like `onscreen` here as well so we can avoid so many `xlines - 4`.

src/nnn.c
```diff
@@ -3274,21 +3274,21 @@ static void browse(char *ipath)
...
+			curscroll = MIN(ndents - (xlines - 4), curscroll + r);
+			cur = (curscroll == ndents - (xlines - 4)) ? cur + r :
+				curscroll + MIN(SCROLLOFF, (xlines - 4) >> 1);
+			cur = MIN(ndents - 1, cur);
 			break;
+		case SEL_PGUP: // fallthrough
 		case SEL_CTRL_U:
-			if (cur > 0)
-				cur -= MIN((xlines - 4) / 2, cur);
+			r = sel == SEL_PGUP ? (xlines - 4) - 1 : (xlines - 4) / 2;
```
jarun: `(xlines - 4) / 2`

should be:

`onscreen >> 1`

jarun:
In the meantime, wold you like to join as a collaborator to `nnn`? I have been working on it more or less alone for the past 2.5 years. Would love to have some company. :)

jarun:
I have added 2 more review comments. Please check.

src/nnn.c
```diff
@@ -3274,21 +3275,21 @@ static void browse(char *ipath)
...
-			break;
-		case SEL_PGUP:
-			if (cur > 0)
-				cur -= MIN((xlines - 4), cur);
-			break;
+		case SEL_PGDN: // fallthrough
 		case SEL_CTRL_D:
-			if (cur < ndents - 1)
-				cur += MIN((xlines - 4) / 2, ndents - 1 - cur);
+			r = sel == SEL_PGDN ? onscreen - 1 : onscreen >> 1;
```
jarun: Please assign `onscreen = xlines - 4;` here.

src/nnn.c
```diff
@@ -3274,21 +3275,21 @@ static void browse(char *ipath)
...
+			curscroll = MIN(ndents - onscreen, curscroll + r);
+			cur = (curscroll == ndents - onscreen) ? cur + r :
+				curscroll + MIN(SCROLLOFF, onscreen >> 1);
+			cur = MIN(ndents - 1, cur);
 			break;
+		case SEL_PGUP: // fallthrough
 		case SEL_CTRL_U:
-			if (cur > 0)
-				cur -= MIN((xlines - 4) / 2, cur);
+			r = sel == SEL_PGUP ? onscreen - 1 : onscreen >> 1;
```
jarun: Please assign `onscreen = xlines - 4;` here.

src/nnn.c
```diff
@@ -3104,6 +3107,7 @@ static void browse(char *ipath)
 
 	while (1) {
 		redraw(path);
+        onscreen = xlines - 4;
```
jarun: We can move this to the 2 relevant blocks instead of calculating each time.

src/nnn.c
bruce-hill: I know >> can sometimes be faster than /, but I thing the speed difference is negligible (and maybe nonexistant if the compiler is smart) and IMO "x/2" is a bit easier to read as "half of x" than "x>>1". I don't have strong feelings about it, so I changed it to ">>" for consistency with the rest of the code.

jarun: :+1: in fact most of the modern compilers should do this already. However, we support a lot of primitive stuff so it doesn't harm.

The notion of performance is kind of relative. A negligible boost on an i7 can be a significant boost on the Pi, for example.

jarun:
Noticing one issue with this change when you have many entries (e.g. in `/usr/bin`):

`^D` which is supposed to take you half a page down, takes you to th e middle entry + 5.

`PgDn` take you to the `last + 5` element.

Can we make `^D` take to the middle element but use scrolloff and `PgDn` take to the `last + 1` element but use scrolloff?

The same should hold for `^U` and `PgUp` as well.

src/nnn.c
bruce-hill: I think it's a bit of a premature optimization. I strongly suspect that the performance of nnn will always be strongly dominated by filesystem operations, followed by user/screen IO, followed by memory accesses, with integer math ops taking up many orders of magnitude less time, especially since there aren't any calculation-heavy inner loops. (Though you should always profile instead of speculating like I'm doing now if you want to optimize effectively!) There's not really much harm with using `>>`, but it makes the code an iota less readable, which can lead to bugs. `>>` has some gotchas like lower operator precedence than `/` or `+`, and `>>1` behaves differently from `/2` for negative numbers.

bruce-hill:
Right now, the way it works is `PgDn` moves the scroll so that the bottom on-screen item moves to the top of the screen, and it moves the cursor the minimum amount necessary to keep it onscreen and within the scrolloff, and `^d` moves the scroll so the bottom on-screen item moves to the middle of the screen (and the middle item to the top of the screen). I modeled this behavior after the way vim does it.

> Can we make `^D` take to the middle element but use scrolloff and `PgDn` take to the `last + 1` element but use scrolloff?

Something to consider is that SCROLLOFF could be 9999, or the window could be only 13 lines tall. For `PgDn`, that would mean the `bottom+1` item would go to the middle of the screen, advancing at half a page at a time. Not a dealbreaker, but it seems less good. And for `^D`, that would mean no movement (the middle item is the same as the `top+scrolloff` item).

Another possibility would be keeping the cursor's relative screen position constant during the scrolling (instead of being dragged along the edge of the screen), or allowing the cursor to be scrolled off-screen and then have it jump to the nearest on-screen position when it's next moved (this might be messier for bookkeeping).

jarun:
I think we can just retain the current behavior of PgDn, ^D etc.

The problem I see is - with PdDn the user will never see 5 items which breaks the intended behavior of the key. It also makes our definition of PgDn, ^D very complex.

bruce-hill:
> The problem I see is - with PdDn the user will never see 5 items which breaks the intended behavior of the key.

All items _appear_ on screen, but the cursor can jump 5 items past the last on-screen item.

> It also makes our definition of PgDn, ^D very complex.

I think the definition is pretty simple: PgDn scrolls the visible files by one page, and ^D scrolls the visible files by half a page.

One possible alternative would be to temporarily allow the cursor to be inside the scrolloff area, so PgDn would make this (assume scrolloff=2):

      AAA
      BBB
    >>CCC
      DDD
      EEE

transition to this state:

    >>EEE
      FFF
      GGG
      HHH
      III

And allow the cursor to be in the scrolloff area without adjusting the scroll position as long as it doesn't go further into the scrolloff (e.g. pressing `j` would move the cursor, but not the scroll, but pressing `k` would move the scroll upward either 1 line, or enough to put the cursor within the scrolloff). This makes the logic a bit more complex, but it would preserve the invariant that PgDn/^D never moves the cursor to an item that isn't currently on-screen.

jarun:
Yes, all the items appear. What I am trying to say is:

when I open `/usr/bin`, the last item is `apt-sortpkgs`. The immediate next item is `ar` which is not shown. When I press `PgDn` I expect the cursor to be at `ar` because that makes one page scroll. However, my cursor goes to `arepack` which is 5 items below `ar`.

A user would expect PgDn to work as he expects and he may miss that the item next to `apt-sortpkgs` was `ar`.

I tested the patch by retaining the regular PgDn/^D etc. behavior and it seems to work fine and as expected. What problem do you see if the original behavior is retained?

bruce-hill:
> I tested the patch by retaining the regular PgDn/^D etc. behavior and it seems to work fine and as expected. What problem do you see if the original behavior is retained?

In the original behavior, if the cursor is at the top of the page (+scrolloff), and you press PgDn, the cursor will jump +1 pagefull of results down (or +.5 for ^D), but because the scroll moves the minimum amount necessary to keep the cursor within the scrolloff, that would mean the cursor would go to the bottom of the screen (-scrolloff) and only `2*scrolloff` new rows of files would appear on screen. It's not too bad (easy to understand), but it kinda violates the expectation that PgDn brings a full new page worth of files into view. Functionally, that only matters when you PgDn and the cursor isn't at the bottom of the screen already (or PgUp and it's not at the top), so it's just a minor issue and if you hit PgDn repeatedly, all the subsequent presses bring a full screen's worth of new files. But if you want to toggle back and forth between two pages of results, PgDn PgUp will mostly hop the cursor around instead of toggling between two screenfuls of files.

jarun:
> PgDn PgUp will mostly hop the cursor around instead of toggling between two screenfuls of files.

True, and I see it's consistent i.e., it hops between the same files so nothing is skipped.

But I see an issue even if I retain the original behavior of PgDn PgUp: in y above example, when you press PgDn, `ar` is selected and it is the 5th item from the bottom (so scroll down has only scrolled 5 items). I think `ar` should be the 5th item from the top.

jarun:
I think let's have the behavior as it is now. If users come back with issues we can address them. Personally I am not seeing any issue after using it for a while. maybe just a matter of getting accustomed.

jarun:
Thank you!

jarun:
I am seeing a new issue with this.

- Enter a directory with more than a screenful of files
- roll to visit a file towards the end
- change the sort order to time
- change the sort order to normal
- that file becomes the 6th file (scrolloff) from top and not all the files are shown

As the number of files are more than a screenful, I was expecting to see a screenful of files.

In our earlier behaviour, when you change the sort order, _a screenful of files is shown_ and the selected file would not necessarily be shown at the middle (which was the _scrolloff_). We should retain that behaviour.

PR #260, opened 2019-05-10
Alternate scrolling rules.

bruce-hill:
This is an alternate set of rules for scrolling. I think it's a bit of an improvement. It allows the scrolloff rule to be temporarily violated in order to put the cursor on the bottom on-screen file, while also bringing in a full new page of files and the cursor will never jump to an item that is not on-screen. Once the cursor is inside the margin, it can move towards the center without triggering a scroll, but it will trigger a scroll if you attempt to move deeper into the scroll margin. (Try: open a large directory, hit PgDn, then fiddle with j/k to see how it behaves)

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -472,6 +472,7 @@ static void redraw(char *path);
 static int spawn(char *file, char *arg1, char *arg2, const char *dir, uchar flag);
 static int (*nftw_fn)(const char *fpath, const struct stat *sb, int typeflag, struct FTW *ftwbuf);
 static int dentfind(const char *fname, int n);
+static void move_cursor(int target, int ignore_scrolloff);
 
 /* Functions */
 
@@ -1656,7 +1657,7 @@ end:
 	if (*ch != '\t')
 		g_ctx[cfg.curctx].c_fltr[0] = g_ctx[cfg.curctx].c_fltr[1] = '\0';
 
-	curscroll = MAX(0, cur - SCROLLOFF);
+	move_cursor(cur, 0);
 
 	curs_set(FALSE);
 	settimeout();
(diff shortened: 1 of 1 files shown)


jarun:
I checked vim scrolloff behavior in the meantime with the last commit. I think our current behavior aligns with vim. Do you think we need more improvements?

jarun:
This might seem confusing to some. Also, when I press ^D the last item is selected in the page and comes to the center.

bruce-hill:
> I checked vim scrolloff behavior in the meantime with the last commit. I think our current behavior aligns with vim. Do you think we need more improvements?

No, I think the existing behavior is good enough. This is just incrementally better.

> Also, when I press ^D the last item is selected in the page and comes to the center.

That's the goal. PgDn and ^D both move the cursor to the last item on screen, then scroll down the relevant amount (1 or .5 pages). Some other options would be: select the middle file, then scroll up (so it's at the top); or select the `middle+scrolloff` file, so it's at `top+scrolloff`. It felt a little weird to violate the scrolloff rule when it's not necessary, and I liked having PgDn and ^D both move the cursor to the bottom item and then scrolling the relevant amount. I'm open to suggestions though (including "just keep the old behavior" if you think it's better).

jarun:
I think the current behaviour is simpler to understand for people familiar with vim scroloff in particular and regular users in general. I will keep this ER open for now. Once we are done with the next release and get some feedback we can re-visit this item.

jarun:
Looks like this PR addresses https://github.com/jarun/nnn/pull/259#issuecomment-493685934.

I will merge it so we have time to test it before next release.

jarun:
Thank you!

PR #281, opened 2019-06-12
Don't ignore fwrite errors

sjmulder:
Fixes warning on CentOS 6:

    src/nnn.c:754: warning: ignoring return value of 'fwrite', declared with attribute warn_unused_result


jarun:
Thank you!

PR #282, opened 2019-06-12
Use a separate wide char buffer

sjmulder:
Reinterpreting the char buffer as a wchar_t buffer is a breach of
aliasing rules, so have two buffers.

CentOS 6 warning:

    src/nnn.c:1770: warning: dereferencing type-punned pointer will break strict-aliasing rules


jarun:
We had to handle this little differently so I took care of this. Thank you! :+1:

sjmulder:
Ough, I only just noticed the (wrong!) global buffer was returned at the end. Good solution, but no check on malloc?

jarun:
> but no check on malloc

I have to add it. I missed it in the second iteration.

jarun:
Done!

Issue #238, opened 2019-04-04
How do I move files between directories?


alxpettit:
I've read the man page and the `README.md`, and I know how to select files and how to apply actions to file selections like cut & copy. But after that, how do I paste the files in a new location?

jarun:
The workflow is:

- Select files
- Go to target directory
- Press <kbd>P</kbd> to _copy_ selection to new location
- Press <kbd>V</kbd> to _move_ selection to new location

jarun:
Please refer to the program help where it's documented.

alxpettit:
Are you talking about the man page? I did look at that before. :/

Re-reading the `SELECTION MODE` section, I can see how it tries to explain this, but it really wasn't clear to me at all given that it doesn't follow the workflow of what I'm used to. Anyway, thanks for answering! Hopefully anyone who's confused will find this thread useful.

jarun:
You'll find the shortcuts here: https://github.com/jarun/nnn#keyboard-shortcuts

or in the in-program help.

The workflow is explained here: https://github.com/jarun/nnn#selection

It's true `nnn` workflows are different in some cases. If you think the workflow is not very clear can you raise a PR that adds more clarity?

PR #284, opened 2019-06-16
Add clarification for usage of moving and copying selections

anitagraham:
Clarification on moving/copying as requested in #238


jarun:
Thank you so much!

PR #286, opened 2019-06-19
Added md5sum plugin

ath3:
Plugin to create and verify md5 checksums


plugins/md5sum
```diff
@@ -0,0 +1,36 @@
...
+resp=f
+
+if [ -s "$selection" ]; then
+    echo -n "work with selection (s) or current file (f) [default=f]: "
+    read resp
+fi
+
+if [ "$resp" = "s" ]; then
+    arr=$(cat $selection | tr '\0' '\n')
+    { read -r file; } <<< "$arr"
```
jarun: This is Bash, not sh.

plugins/md5sum
```diff
@@ -0,0 +1,34 @@
...
+resp=f
+
+if [ -s "$selection" ]; then
+    echo -n "work with selection (s) or current file (f) [default=f]: "
+    read resp
+fi
+
+if [ "$resp" = "s" ]; then
+    file=$(basename "$(cat $selection | tr '\0' '\n' | head -n 1)").md5
+    cat "$selection" | sed 's|'"$PWD/"'||g' | xargs -0 -i md5sum {} > "$file"
```
jarun: You know shell scripts much better than me! :+1: 

Care to modify ndiff and splitjoin plugins to be POSIX-compliant? I would really appreciate that.

jarun:
I would also suggest to rename the plugin to `checksum` and ask the user whether to calculate md5 or sha256 checksum. sha256 is quite popular and would help a lot of users (including me ;)).

ath3:
I removed the "-o" parameter from xargs, because with that parameter this plugin was not working for me.
This means that we will get the warning message "Input is not from a terminal", but it works.

Since POSIX doesnt have process substitution, the directory comparison now lists the directory contents to two temporary files, and those files will be compared.

jarun:
> ndiff

No problem! Thank you!

plugins/checksum
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env sh
+
+# Description: Create and verify checksums
+#
+# If selection is used: it will generate one file containing the checksums with file names
+#                       [and with paths if they are in another directory]
+#                       The filename will be checksum_timestamp.checksum_type
+# If file is used: if the file is a checksum, it does the verification
+#                  if the file is not a checksum, it will be created from it
```
jarun: "it will be created from it" - @ath3 what does this mean?

jarun: OK. Got it!

PR #294, opened 2019-06-26
Set CP in Makefile

tanyaionova:
$(CP) is not set by GNU Make
Seems like [1] is a full list of what GNU Make defines implicitly
[1] https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html


jarun:
Thank you!

Issue #295, opened 2019-06-26
Duplicate code


tanyaionova:
Seems like selecting opener is duplicated.
Line numbers from `ge93abcf`:
`nnn.c:4555-4557` is exact copy of `nnn.c:4640-4642`:

```c
/* Get custom opener, if set */
opener = xgetenv(env_cfg[NNN_OPENER], utils[OPENER]);
DPRINTF_S(opener);
```
I am not sure, but I believe the second copy is redundant as they both are in the same scope and opener is not changed between them.

jarun:
Closed by PR #296.

jarun:
Thanks for catching this! :+1:

PR #296, opened 2019-06-26
Remove duplicate code (#295)

tanyaionova:
These three lines are exact copy of 4555-4557.
As opener is not changed in between, we can delete these


jarun:
Thank you!

PR #297, opened 2019-06-27
Add shuffle to mocplay

ath3:
If shuffle is on and we add huge number of files it gets really slow,
The reason is that we add each file separately,  otherwise mocp sorts files passed as arguments after -a.


jarun:
Thank you!

PR #298, opened 2019-06-27
Remove unneeded check from checksum plugin

ath3:
The check is not needed because we exit after the checksum verification.


jarun:
Thank you!

PR #301, opened 2019-07-04
Copy user input to a dedicated buffer

0xACE:
I was having trouble creating archives. After some debugging i found that tmp was pointing to g_buf which would eventually be used by other functions.

runfile seemed to be a potential candidate for buffer in this case.

In the future I would recommend refactoring of the code because it is very confusing where the variables will be used and what is using them.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4029,6 +4029,9 @@ nochange:
 				goto nochange;
 			}
 
+			xstrlcpy(runfile, tmp, NAME_MAX);
+			tmp = runfile;
+
 			/* Confirm if app is CLI or GUI */
 			if (sel == SEL_OPENWITH) {
 				r = get_input("cli mode? [y/Y confirms]");
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4029,8 +4029,8 @@ nochange:
 				goto nochange;
 			}
 
-			xstrlcpy(runfile, tmp, NAME_MAX);
-			tmp = runfile;
+			xstrlcpy(rundir, tmp, PATH_MAX);
+			tmp = rundir;
 
 			/* Confirm if app is CLI or GUI */
 			if (sel == SEL_OPENWITH) {


jarun:
I see `getutil()` is the problem. We use the variable `newpath` as a tmp buffer. We can localize the copy as this is the only case where `tmp` can be overwritten.

I will take care of this. Thanks a lot for finding the defect.

jarun:
In fact there was another issue due to similar reasons. The best would be not to use `g_buf` in `getutil()` as we do not need to retain the result.

jarun:
Please test the patch and confirm it works.

0xACE:
I tried to create another archive with selection after applying your change and it is not working. After further debugging I noticed that in `archive_selection()` we are overwriting `g_buf` again via `snprintf()`. As `snprintf()` is filling `g_buf` the input variable `archive` is being overwritten as it also points to `g_buf`.

In my local copy I have re-applied this patch. Contrary to your recommendation of using `newpath`, I'm sticking to `rundir` because it represents the size of `g_buf` better than `runfile` (although I would argue `runfile` should suffice).

The reason I'm using `run(dir|file)` is because I can't find occurances of `run(dir|file)` being used after that point, whereas `newpath` is going to be overwritten in `mkpath()`.

I think the way `g_buf` and various other variables is being shared is very "spaghetti" and it doesn't make the program any better nor does it make it easy to develop it. Maybe the user input deserves its own

    char userinput[CMD_LEN_MAX] __attribute__ ((aligned));

Not sure that it solves anything but it should help...

0xACE:
That error is complaining that `browse()` is too complicated. From my earlier reading `runfile` wasn't as deeply nested compared to `rundir` which I guess what triggered it this time. But really the code could use some refactoring.

Also I'm thinking of applying support for extracting a archive `bare`, meaning without a subdir first. Would you be interested in it? It could require me to reconstruct `handle_archive()` (Also that code is very confusing its' doing some hacks to add/remove `-f` because in `BSDTAR` it means `--file` and in `atool` it means `--force`).

Sry for clottering this topic but I'm on the run atm and can't find the time to set up a structured conversation. If you use any other means of communication it could make our dialog easier...

jarun:
rundir and runfile has different use cases. I wouldn't want to touch them for this.

I checked the number of places g_buf is used now and it's limited. g_buf is used as the scratch buffer.

I am thinking why not have a use and throw buffer inside archive_selection and use that instead of g_buf?

jarun:
> Also I'm thinking of applying support for extracting a archive bare

I always extract my archives inside a directory which already has 100s of log files. Extracting bare doesn't help.

jarun:
For now I have added a local buf.

I think refactoring browse() is a good idea. It's a big one.

0xACE:
> I checked the number of places g_buf is used now and it's limited. g_buf is used as the scratch buffer.

Do you have a tool that does this? Because I manually read the code to find out where which variables were used. I'm running a very much vanilla build of vim, basically only the `vim-surround` plugin (and `vim nnn` ofcourse :P). And browse the code isn't the best/worst experience...

> I am thinking why not have a use and throw buffer inside archive_selection and use that instead of g_buf?

Genius!


> > Also I'm thinking of applying support for extracting a archive bare
> 
> I always extract my archives inside a directory which already has 100s of log files. Extracting bare doesn't help.

Alright, no problem I'll maintain that on my local copy. My version prompts for `bare` and then another one for `force`, not sure I'll ever use it, but it wasn't much work to add...

I also made `archive_selection()` not use the absolute path, on my local copy of `nnn` it strips down parent directories in relation to the `CWD` of `nnn` so if i select `~/nnn` and `~/nnn-local` in my `$HOME` dir and create the arcive from selection in that folder the resulting archive will only list `nnn` and `nnn-local` as the root directories and not the full `$HOME/nnn-{,-local}`. Basically the current `path/` is stripped from the `selections` (this work is delegated to `sed` before `xargs` sees the selection but awk or some other tools could practically do the same). I find this useful when cherry picking certain files in a folder and putting them in a zip...

jarun:
> Do you have a tool that does this

No, I just searched `g_buf` and cycled through the instances.

> not sure I'll ever use it, but it wasn't much work to add...

Exactly the reason I make some design choices to limit the options we provide to users. Most of them may not sure it, and those who would can hack the tool anyway.

While we are discussing this, `nnn` gives first priority to noobs. Simplicity is the key to make more newbies (I am talking *nix newbies) introduced to the terminal. We want to _advertise_ that the terminal isn't a place to remember commands or keep typing for productivity with an endless number of choices. This is one of the driving reasons I keep scrutinizing feature suggestions somewhat harshly.

Take a look at what we have achieved so far - a damn easy but extremely powerful (for regular users) file manager with minimal config. We can keep adding conditional features but if you ask me, if a new tool has more than a dozen settings you have to fool around with, it doesn't remain simple anymore.

Yesterday, I made archiving independent of atool and bsdtar. (un)zip, tar and gz/bz2 support is present by default on most modern distros. If atool/bsdtar are not available, we'll use those. 1 less dependency.

0xACE:
> > Do you have a tool that does this
> 
> No, I just searched `g_buf` and cycled through the instances.

Fair enough.

> > not sure I'll ever use it, but it wasn't much work to add...
> 
> Exactly the reason I make some design choices to limit the options we provide to users. Most of them may not use it, and those who would can hack the tool anyway.

That's true.

> While we are discussing this, `nnn` gives first priority to noobs. Simplicity is the key to make more newbies (I am talking *nix newbies) introduced to the terminal. We want to _advertise_ that the terminal isn't a place to remember commands or keep typing for productivity with an endless number of choices. This is one of the driving reasons I keep scrutinizing feature suggestions somewhat harshly.

I agree.

> Take a look at what we have achieved so far - a damn easy but extremely powerful (for regular users) file manager with minimal config. We can keep adding conditional features but if you ask me, if a new tool has more than a dozen settings you have to fool around with, it doesn't remain simple anymore.
> 
> We also give desktop integration more priority. The terminal has many choices. But if we are being honest, without the GUI, none other than a dev or a sysadmin is going to touch a system. It's dreadful. The terminal should be touched only when someone realizes it's much more easier to achieve something in the terminal. So we provide an in-place-renamer and wait for that realization to come. It also keeps the learning curve simple. A newbie learns when he sees the need. We just provide him enough opportunity to explore that alley. But it should be a cakewalk, not a one full of confusing choices.

I moved from a graphical file browser `thunar` to `nnn` because `thunar` crashes on my pc way too often. Basically the only thing I'm missing is thumbnails of images/movies in `nnn` but I can implement that with `sixel` but I feel like it's not necessary, I just bound `T` to open `CWD` in thunar instead for such moments. But I do agree with your ideas. Btw I did read the design philosophy and do agree with it. I guess the issue is that I'm moving `nnn` into a more important role in my setup. One favorite workflow of `nnn` is the ease of use and the low RAM usage (my current machine has practicaly allocated most of the RAM at all times) to write a patch for `nnn` I hit the following keys: `,nGC`

Which will:
1. `,n` - go to my `nnn` directory (I changed the bookmarks to 20 so I keep a lot of bookmarks...)
2. `G` - highlight the last item on my list which is `z-EDIT`
3. `C` - Run `z-EDIT` which has:

```
#!/usr/bin/env bash
vim src/nnn.{c,h} src/local.h +bn +bn +bp +bp
# the +bn +bp combo is to make all files load and
# let me use ^ to jump between files correctly and
# not warn me when :q for having multiple unloaded files
```

> Yesterday, I made archiving independent of atool and bsdtar. (un)zip, tar and gz/bz2 support is present by default on most modern distros. If atool/bsdtar are not available, we'll use those. 1 less dependency.

That is great news! I only have `atool` on 2 of my machines. and I usually the common archive/compression formats so this change is a very big plus for me.

jarun:
> Basically the only thing I'm missing is thumbnails of images/movies

Check out the plugins thumb and viuimg.

> let me use ^ to jump between files correctly

Neat trick!

> That is great news!

Made me happy too! ;)

PR #302, opened 2019-07-08
Fix quitcd snippets for 'zsh' and 'bash'

fokinpv:
`NNN_TMPFILE` variable should be declared before  calling`nnn`.


jarun:
Can you make the change for quitcd.fish as well?

fokinpv:
Fish is updated

jarun:
Thank you!

PR #305, opened 2019-07-11
Find a parent when CWD disappears

0xACE:
The previous behaviour would exit nnn when the CWD disappeared.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3301,6 +3301,9 @@ nochange:
 		if (access(path, F_OK)) {
 			DPRINTF_S("dir deleted or moved");
 
+			/* Save history */
+			xstrlcpy(lastname, xbasename(path), NAME_MAX + 1);
+
 			xstrlcpy(newpath, path, PATH_MAX);
 			while (true) {
 				dir = visit_parent(path, newpath, &presel);
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3297,11 +3297,36 @@ nochange:
 		if (getppid() == 1)
 			_exit(0);
 
-		/* Check if CWD is deleted to  avoid hang, bad idea */
+		/* Check if CWD is deleted and find a existing parent */
 		if (access(path, F_OK)) {
 			DPRINTF_S("dir deleted or moved");
-			cfg.badexit = 1;
-			return;
+
+			/* Save last working directory */
+			xstrlcpy(lastdir, path, PATH_MAX);
+
+			/* Save history */
+			xstrlcpy(lastname, xbasename(path), NAME_MAX + 1);
+
+			xstrlcpy(newpath, path, PATH_MAX);
+			while (true) {
+				dir = visit_parent(path, newpath, &presel);
+				if (istopdir(path) || istopdir(newpath)) {
+					if (!dir)
+						dir = dirname(newpath);
+					break;
+				} else if (!dir) {
+					xstrlcpy(path, newpath, PATH_MAX);
+					continue;
+				}
+				break;
+			}
+
+			xstrlcpy(path, dir, PATH_MAX);
+
+			setdirwatch();
+			mvprintw(xlines - 1, 0, "folder disappeared\n");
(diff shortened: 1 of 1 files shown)


0xACE:
The failed tests seems to be because of syntax complaints. It's very unclear which syntax style you use, but I found `} else if (...) {` used elsewhere so I imitated it. Personally I wouldn't use that either and just imitated the existing style.

Otherwise I have tested the code. The code also makes the assumption that `/` is always going to be an accessible path. If there are situations where `/` isn't a accessible path then I don't know what the program should do, but you could check for it with `isatopdir() && access() == -1` to detect this situation.

src/nnn.c
```diff
@@ -3297,11 +3297,36 @@ static void browse(char *ipath)
...
+					xstrlcpy(path, newpath, PATH_MAX);
+					continue;
+				}
+				break;
+			}
+
+			xstrlcpy(path, dir, PATH_MAX);
+
+			setdirwatch();
+			mvprintw(xlines - 1, 0, "folder disappeared\n");
```
jarun: Please use the message: "cannot access directory"

src/nnn.c
```diff
@@ -3297,11 +3297,36 @@ static void browse(char *ipath)
...
+			xstrlcpy(lastname, xbasename(path), NAME_MAX + 1);
+
+			xstrlcpy(newpath, path, PATH_MAX);
+			while (true) {
+				dir = visit_parent(path, newpath, &presel);
+				if (istopdir(path) || istopdir(newpath)) {
+					if (!dir)
+						dir = dirname(newpath);
+					break;
+				} else if (!dir) {
```
jarun: Please remove `else if` and make it a separate `if` clause. (You don't need `else if` after a `break`).

jarun: This is also the reason CI fails.

src/nnn.c
```diff
@@ -3297,24 +3297,19 @@ static void browse(char *ipath)
...
-		/* Check if CWD is deleted and find a existing parent */
+		/* Check if CWD is deleted and find an existing parent */
 		if (access(path, F_OK)) {
 			DPRINTF_S("dir deleted or moved");
 
-			/* Save last working directory */
-			xstrlcpy(lastdir, path, PATH_MAX);
-
-			/* Save history */
-			xstrlcpy(lastname, xbasename(path), NAME_MAX + 1);
```
jarun: Please keep the lastname. The directory can stay selected in parent even if the user doesn't have permissions anymore.

jarun: If the dir is deleted, we select the top entry anyway.

0xACE:
You were quick to respond :).

All your comments has been addressed.

jarun:
> You were quick to respond :).

Thanks to a reasonably powerful Android phone and Termux! ;)

Please take a look at the scroll support with ncurses6. It would be a great value addition for our users on Termux. I am one of them! ;)

jarun:
Frankly speaking, I am kind of obsessed by Termux! (That's a compliment for you @fornwall ;))

@0xACE if you have any queries regarding the Termux environment @fornwall is the goto guy.

0xACE:
> Please take a look at the scroll support with ncurses6. It would be a great value addition for our users on Termux. I am one of them! ;)

I am also a termux user, i addressed your comments from the phone and didn't push the changes until i reached a computer, termux is super nice when I'm on the run.

Regarding termux scroll support I would suggest a specific github issue so that the problem can be tracked from 1 place. As, i would also argue that in termux a "click" should be on the "highlighted" item rather then where the user clicks as it's hard to click on the exact row...

Yeah, frankly from termux I really wish they had this fixed sooner: https://github.com/termux/termux-api/pull/233

My early guess is that it's just a matter of implementing the code, and not a very termux specific problem...

jarun:
> Regarding termux scroll support I would suggest a specific github issue

We will continue the discussion in the ToDo list for now.

PR #309, opened 2019-07-16
stop readline from overwriting LINES

0xACE:
Turns out I'm a ducking liar, I guess I found some time to do it now afterall.

The problem was that readline would completely block LINES from updating
after prompting the user. I'm not entirely sure why this happened, but
at least this patch fixes the problem.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4783,7 +4783,9 @@ int main(int argc, char *argv[])
 	setlocale(LC_ALL, "");
 
 #ifndef NORL
+#if RL_READLINE_VERSION >= 0x0603
 	rl_change_environment = 0;
+#endif
 	/* Bind TAB to cycling */
 	rl_variable_bind("completion-ignore-case", "on");
 #ifdef __linux__
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4783,6 +4783,7 @@ int main(int argc, char *argv[])
 	setlocale(LC_ALL, "");
 
 #ifndef NORL
+	rl_change_environment = 0;
 	/* Bind TAB to cycling */
 	rl_variable_bind("completion-ignore-case", "on");
 #ifdef __linux__
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4784,6 +4784,7 @@ int main(int argc, char *argv[])
 
 #ifndef NORL
 #if RL_READLINE_VERSION >= 0x0603
+	/* readline would overwrite the WINCH signal hook */
 	rl_change_environment = 0;
 #endif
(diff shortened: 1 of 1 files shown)


0xACE:
I tried looking for when `rl_change_environment` was introduced but couldn't find it. so I guess this patch doesn't pass as is.

jarun:
It was added in libreadline 6.3: http://upstream.rosalinux.ru/compat_reports/readline/6.2_to_6.3/abi_compat_report.html

src/nnn.c
```diff
@@ -4783,6 +4783,7 @@ int main(int argc, char *argv[])
 	setlocale(LC_ALL, "");
 
 #ifndef NORL
+	rl_change_environment = 0;
```
jarun: Please add a comment on what it does and in which scenario this is useful.

0xACE:
> It was added in libreadline 6.3: http://upstream.rosalinux.ru/compat_reports/readline/6.2_to_6.3/abi_compat_report.html

Wow that was a nice website, I should figure out how that works, I had to `git clone` the official repo and grep through the history...

Anyhow I added a simple comment

jarun:
BTW, if you want to add a SIGWINCH handler please feel free to do so. Just ensure the places in which we have the *LINES variables work with that.

jarun:
Thank you!

0xACE:
> BTW, if you want to add a SIGWINCH handler please feel free to do so. Just ensure the places in which we have the *LINES variables work with that.

Yeah, before i found out about `rl_change_environment` I was going to use `SIGWINCH` but this seemed easier for now. But for now that's maybe something for the future, it is not that urgent to me anymore.

jarun:
>  it is not that urgent to me anymore

Understood. But please consider the importance from user perspective as well. I do not use a tiling window manager. So I am not very sure of the best solution.

0xACE:
> > it is not that urgent to me anymore
> 
> Understood. But please consider the importance from user perspective as well. I do not use a tiling window manager. So I am not very sure of the best solution.

Yeah I sound a bit harsh there don't I. I spent time researching it, and the general consensus was that the most portable approach was to use a `SIGWINCH` method as you can always implement it yourself, while `KEY_RESIZE` depends on your local `curses` to have it implemented.

The biggest deterrence for me was that I would likely break something if I use the`SIGWINCH` because I believe it's sort of what made `readline` break it for us in the first place. `rl_change_environment` makes it so `readline` doesn't hook onto `SIGWINCH` leaving our `curses` to deal with it without bugging out....

jarun:
OK got it! Thanks! We revisit if needed after the next release.

PR #310, opened 2019-07-16
Use KEY_RESIZE when prompting user input

0xACE:
The problem was that a window wouldn't resize while prompting the user
for input. These changes allows the input methods in nnn to resize
properly if KEY_RESIZE is defined.

A more portable solution would be hooking the WINCH signal and update
the xlines value from there along with some resized flag.

In some cases the full window isn't redrawn until the user has finished
the input. This is because in some functions I wasn't sure the current
"path" was available.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -548,6 +548,15 @@ static char *xitoa(uint val)
 	return &ascbuf[++i];
 }
 
+#ifdef KEY_RESIZE
+/* Clear the old prompt */
+static inline void clearoldprompt()
+{
+	move(xlines - 1, 0);
+	clrtoeol();
+}
+#endif
+
 /* Messages show up at the bottom */
 static inline void printmsg(const char *msg)
 {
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -1809,13 +1838,13 @@ static char *xreadline(char *prefill, char *prompt)
...
 
-	getyx(stdscr, y, x);
+	x = getcurx(stdscr);
 	curs_set(TRUE);
 
 	while (1) {
 		buf[len] = ' ';
-		mvaddnwstr(y, x, buf, len + 1);
-		move(y, x + wcswidth(buf, pos));
+		mvaddnwstr(xlines - 1, x, buf, len + 1);
```
jarun: Please explain this change to me,

0xACE: Yeah I found it confusing at first too... I chose solution to avoid changing too many things.

It is very convoluted but it should work like so:

Previously it would first print the `prompt` and then get the x,y coordinates from the cursor position only once, and not update it until the function is re-entered. Since `y` isn't updated the prompt will remain at the same line forever.

The new method updates `xlines` with the latest `LINES` value, ensuring that `xlines` always points to the bottom of the screen. But each time the window is resized, the old `prompt` is first cleared, and then a new `prompt` is printed at the new `xlines`, before the highlighted code is run. As the prompt doesn't change size, the `x` value will still be valid, ensuring that a new and clean `prompt + buffer` is printed at the new location correctly.

I was considering using `clrtoeol()` at the highlighted code, but since the old behaviour wasn't broken I didn't add `clrtoeol()`.

While writing this explination I realised: maybe `xlines = LINES;` could be appended to `clearoldprompt()`.

Also if you aren't happy with the overwriting of `xlines` in my previous version I just kept a separate `y` value, but this application was smaller than that method.

jarun:
Thank you!

PR #312, opened 2019-08-03
Make strip explicit for all targets

rindeal:
Use only the `strip` target fort stripping the binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,6 @@ debug: $(SRC)
 
 norl: $(SRC)
 	$(CC) -DNORL $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o $(BIN) $^ $(LDLIBS)
-	$(STRIP) $(BIN)
 
 install: all
 	$(INSTALL) -m 0755 -d $(DESTDIR)$(PREFIX)/bin


jarun:
> Use only the strip target fort stripping the binary.

And packagers who are working with `sudo make norl install` will not have a stripped binary anymore.

> leftover

What exactly is the problem here? We want released binaries to be stripped. If you want a debug build with debug symbols please use the target `debug`.

rindeal:
The default make target is not stripped, just the `norl` target. For the default target you have a special `strip` target which does nothing but stripping. So in order to strip the release version you have to include the `strip` target anyway (`make all strip install`). I just tried to add some consistency to the stripping situation in the Makefile. I'm sorry I considered the automagic `$(STRIP)` command as a _leftover_, it just didn't come to my mind at that moment that it's put there by your careful design.

jarun:
> For the default target

Ahh OK. I understand.

At this point the existing packager mantainers have handled this based on their choice of readline or no readline version already.

Are you packaging for any new distro and facing some issues because of this? If it's not much of a bother, I would like to leave it be without touching. Hence the question.

rindeal:
Yes, I'm packaging for a distro and we do our own stripping **after** installation. And because I package both versions (`all` and `norl`) I need the Makefile **not** to strip by default in either target. Currently, only the default `all` target satisfies my needs. With the patch I provided any other packager, who will want to strip the binary using the Makefile, can still just add the `strip` target to `make`'s argument list like this:

```
$ make all strip install
```

or 

```
$ make norl strip install
```

jarun:
Makes sense.

I will put a note out to packagers when I release the next version.

For which dstro are you packaging?

rindeal:
Thank you.

> For which dstro are you packaging?

It's a repository of packages for package managers based on [PMS](https://dev.gentoo.org/~ulm/pms/head/pms.html), which are primarily used by [Gentoo](https://www.gentoo.org/) distro.
Homepage of my repository: https://github.com/rindeal/rindeal-ebuild-repo
My package for _nnn_: https://github.com/rindeal/rindeal-ebuild-repo/tree/master/app-misc/nnn

jarun:
Nice work! :+1:

jarun:
When it's complete, please raise a PR to add i to the list of packagers.

jarun:
@rindeal could you please add `nnn` to the "File management" list of Gentoo?

https://wiki.gentoo.org/wiki/Recommended_tools#File_management

`nnn` is available at https://packages.gentoo.org/packages/app-misc/nnn (but you know that already I think)

rindeal:
> @rindeal could you please add `nnn` to the "File management" list of Gentoo?
> 
> https://wiki.gentoo.org/wiki/Recommended_tools#File_management

Ok, done.

> `nnn` is available at https://packages.gentoo.org/packages/app-misc/nnn (but you know that already I think)

I did not know that since I am in no way affiliated with either the Gentoo project or this package in particular. I am merely using Gentoo's package manager to build my own packages (nnn being one of them).

jarun:
Thank you so much! :+1:

Issue #311, opened 2019-08-02
Change archive path(s)?


serussell:
Creating an archive of files deep in a subdirectory, `nnn` roots the files at the home directory; this creates an archive with a deep tree structure.  Is there a way to change where the archive is rooted, say, in the current directory?  Or relative to the where the archive is being created? One of the two following behaviors would be less surprising, to myself, at least:

1) If the archive is being created in the same directory as the files being added, the archive would contain only the files, with no directory structure.

2) Better would be a path relative to where the archive is being created; this would be a super-set of the first case.  E.g., in directory `A`, tag files `a` and `b`.  CD to `..` and create an archive (`f`).  The zip would then contain entries `A/a` and `A/b`.  Without the CD (creating the archive in `A`), the archive would contain `a` and `b` with no paths.

What is surprising -- and I'm struggling to see the value -- is the current behavior, where the archive contains entries with full paths from ${HOME}.

jarun:
This happens when you _select_ some files and archive those, not when you archive the current entry. `nnn` supports selection across directories. While this may look surprising there are 2 good reasons this is done:

- the paths to the directories are preserved when multiple files are selected from multiple paths. This makes it simpler for the user to identify which file came from where on extraction.
- If there are 2 files with the same name from 2 directories there wouldn't be any conflict

So when you pick a file from a deeply located path you get it extracted with a deeply rooted path. `nnn` doesn't add the deeply rooted-ness to the file.

Now that you know this, you can try the following:

- move the folder with the deeply rooted path to a different location and archive it
- select the files you need to a different location and archive them

One can argue why `nnn` doesn't do the conflict resolution to check if file names are same. But that wouldn't help with preserving the directory paths and doing that resolution for 10K files selcted together would be quite intensive.

0xACE:
Hey I've been away but I should mention:

The feature you are requesting has actually been implemented and I'm using it locally. It works by stripping the selected files `fullpath` with `CWD` basically... I don't know what consequences this can have but it fits my use case better... For example what happens when `CWD` doesn't match some of the selected files? In my patch it simply doesn't strip the parent folders that don't match the exact path, so it could have problems with symbolic links...

[...]

PR #313, opened 2019-08-03
Make selection based archives use relative paths

0xACE:
This will create archives relative to your current working directory
rather than a full path when when creating the archive based on your
current selection.

I don't expect this to be adopted to `upstream` but issue #311 seemed to want this and I'm guessing many others expect it to be working this way.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1302,7 +1302,7 @@ static void archive_selection(const char *cmd, const char *archive, const char *
 	char *buf = (char *)malloc(CMD_LEN_MAX * sizeof(char));
 	snprintf(buf, CMD_LEN_MAX,
 #ifdef __linux__
-		 "xargs -0 -a %s %s %s", g_cppath, cmd, archive);
+		"sed -ze 's|^%s/||' '%s' | xargs -0 %s %s", curpath, g_cppath, cmd, archive);
 #else
 		 "cat %s | xargs -0 -o %s %s", g_cppath, cmd, archive);
 #endif
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1303,6 +1303,9 @@ static void archive_selection(const char *cmd, const char *archive, const char *
 	snprintf(buf, CMD_LEN_MAX,
 #ifdef __linux__
 		"sed -ze 's|^%s/||' '%s' | xargs -0 %s %s", curpath, g_cppath, cmd, archive);
+#elif defined __APPLE__
+		"cat '%s' | tr '\\0' '\n' | sed -e 's|^%s/||' | tr '\n' '\\0' | xargs -0 %s %s",
+		g_cppath, curpath, cmd, archive);
 #else
 		 "cat %s | xargs -0 -o %s %s", g_cppath, cmd, archive);
 #endif


jarun:
@ath3 please review this PR. It should take care of both Linux and macOS.

src/nnn.c
```diff
@@ -1302,7 +1302,7 @@ static void archive_selection(const char *cmd, const char *archive, const char *
 	char *buf = (char *)malloc(CMD_LEN_MAX * sizeof(char));
 	snprintf(buf, CMD_LEN_MAX,
 #ifdef __linux__
-		 "xargs -0 -a %s %s %s", g_cppath, cmd, archive);
+		"sed -ze 's|^%s/||' '%s' | xargs -0 %s %s", curpath, g_cppath, cmd, archive);
 #else
 		 "cat %s | xargs -0 -o %s %s", g_cppath, cmd, archive);
```
jarun: Please handle for macOS as well.

0xACE:
I uploaded a version with `MacOS` support, but I don't have a `MacOS` machine to try this with. Feel free to adjust it to your needs.

jarun:
I don't have macOS either.

Would request whoever is here to try it out if he has access to macOS.

serussell:
> Would request whoever is here to try it out if (s)he has access to macOS.

I get a "utility missing" error message when I try to create archives with `nnn` on OSX.

```
4231)~ % uname -a
Darwin PHLLT8526.local 18.5.0 Darwin Kernel Version 18.5.0: Mon Mar 11 20:40:32 PDT 2019; root:xnu-4903.251.3~3/RELEASE_X86_64 x86_64
4232)~ % which zip
/usr/bin/zip
4233)~ % which tar
/usr/bin/tar
4234)~ % zip -v
Copyright (c) 1990-2008 Info-ZIP - Type 'zip "-L"' for software license.
This is Zip 3.0 (July 5th 2008), by Info-ZIP.
...
```

jarun:
Update to master.

0xACE:
>I get a "utility missing" error message when I try to create archives with nnn on OSX.

This patch only introduces `tr` and `sed`. `sed` is the only new reference in the entire codebase, so if it's missing a utility that would be it. Or maybe you are running a old version as jarun said...

serussell:
Just to confirm: this morning, I did a fresh checkout of the branch "relativearchives" from https://github.com/0xACE/nnn; I'm on commit 2baa42b877ca0bc338a1852397ca3a0aa7c30b79.  That's what was the request for OSX testing, right?  And:

```
4245)~/nnn % which tr
/usr/bin/tr
4246)~/nnn % which sed
/usr/bin/sed
```

serussell:
*<sigh>* there was another `nnn` in a different path on OSX, so PEBKAC.

Confirming the patch works as expected on OSX.  The only oddity is that I created `Foo.zip`, and that's the name of the file that got created, but what *actually* got created was a tarball:

```
4257)Documents/architecture % file Foo.zip
Foo.zip: POSIX tar archive
```

0xACE:
> Confirming the patch works as expected on OSX. The only oddity is that I named the archive `Foo.zip`, and that's what got created, but the file type that was created wasn't a zip, but a tarball:
> 
> ```
> 4257)Documents/architecture % file Foo.zip
> Foo.zip: POSIX tar archive
> ```

I would argue the problem resides from where `archive_selection()` is called.

I'm guessing 8e00d74259f98d707e9ba04d982c34d4da66a584 with `get_archive_cmd()` is the culprit where `else if (getutil(utils[BSDTAR]))` should be moved further down the if-else statements...

jarun:
I can reproduce this. And that's strange! The [man page of bsdtar](https://www.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&manpath=freebsd-release-ports) description says:

`... can create tar, pax,	cpio, ar, zip, 7-zip, and shar archives.`

So why would it create an archive with the extension `zip` as `tar`?

jarun:
Ahh, we need to use the option `-a`!

jarun:
Thanks @0xACE! We now have enough time till the next release to get it tested thoroughly.

0xACE:
Yeah, I'm not stressed.

But one concern I have is shouldn't the the `#if define(__APPLE__)` just be the `#else` statement so non  `__APPLE__` ?

I'm not familiar with github interface. But I wrote comments on the affected code above. I figured it didn't ping you or you didn't mind either way.

jarun:
> But one concern I have is shouldn't the the #if define(__APPLE__) just be the #else statement so non __APPLE__ ?

Mostly it would be BSD people and they don't mind non-issues.

0xACE:
>Mostly it would be BSD people and they don't mind non-issues.

Haha, great. I'll maintain a local copy that uses `#else` and annotate why upstream is different

jarun:
I see that FreeBSD, OpenBSD and NetBSD have `sed`, `cat` and `tr` so I think it's safe to include those along with macOS. I am making the case #else and let's see if someone finds an issue. As I mentioned we have time till next release.

jarun:
Pushed at commit 0647c75cd8c75f5377b24b1be2456d1f71f649b3.

0xACE:
> I see that FreeBSD, OpenBSD and NetBSD have `sed`, `cat` and `tr` so I think it's safe to include those along with macOS.

Yes. To clarify: `GNU sed` has the `-z` flag to seperate by lines by `\0`. Howver appels `sed` doesn't seem to have that feature which is why the apple version is using `tr` instead... Essentially if `BSD` also has `sed -z` then the first version would suffice for them...

>I am making the case #else and let's see if someone finds an issue. As I mentioned we have time till next release.

Great! I'll drop my local copy

jarun:
Looks like none of the BSDs have `-z`. Generally they are similar to macOS.

jarun:
I don't understand this: https://wiki.archlinux.org/index.php?title=Nnn&type=revision&diff=578979&oldid=578937

minimalist and lightweight are characteristics of `nnn` and not ads. I would have loved to discuss this with the editor but i don't know the ways of the Arch wiki. Can we do somehting here? I totally believe it's a wrong reversal.

0xACE:
I'm not sure. I wouldn't push my luck. I'm too busy to deal with it this until the later half of september. But I guess it could be discussed on that wiki page... iirc I have seen other tools advertise as "minimalistic and lightweight"

jarun:
I think we can ignore it. Our performance page says it all and I am pushing more content to our wiki from the homepage.

I must say you were right about the proposal to move the hacking nnn page to arch wiki.

PR #316, opened 2019-08-08
Fix keyresize handling of cur in filterentries()

0xACE:
I found some time to look at the bug and found it... I have to stress that it's very difficult to keep track of where variables are used in the current code base...

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -554,6 +554,15 @@ static char *xitoa(uint val)
 	return &ascbuf[++i];
 }
 
+#ifdef KEY_RESIZE
+/* Clear the old prompt */
+static inline void clearoldprompt()
+{
+	move(xlines - 1, 0);
+	clrtoeol();
+}
+#endif
+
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -1716,6 +1738,18 @@ static int filterentries(char *path)
 
 	while ((r = get_wch(ch)) != ERR) {
 		switch (*ch) {
+#ifdef KEY_RESIZE
+		case KEY_RESIZE:
+			clearoldprompt();
+			if (len == 1) {
```
0xACE: This is the culprit if you are wondering what the problem was.

jarun:
> I found some time to look at the bug and found it

Awesome! I will take a look.

> I have to stress that it's very difficult to keep track of where variables are used in the current code base

I think we have discussed this before and I am OK with a code refactor.

jarun:
It's working as expected now. Thank you!

PR #321, opened 2019-08-12
Fix 'batch_rename' moving symlinks

KlzXS:
Replaced paste and tr with awk that only prints the lines that differ. Also there is no need for -n flag on mv as it won't get passed the same source and destination.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1213,13 +1213,7 @@ static bool batch_rename(const char *path)
 	int fd1, fd2, i;
 	uint count = 0, lines = 0;
 	bool dir = FALSE, ret = FALSE;
-	const char renamecmd[] =
-#ifdef __OpenBSD__
-		/* OpenBSD does not have the -n option for mv */
-		"paste -d'\n' %s %s | tr '\n' '\\0' | xargs -0 -n2 mv 2>/dev/null";
-#else
-		"paste -d'\n' %s %s | tr '\n' '\\0' | xargs -0 -n2 mv -n 2>/dev/null";
-#endif
+	const char renamecmd[] = "awk 'FNR==NR{a[NR]=$0} a[FNR]!=$0{printf \"%%s\\0%%s\\0\",a[FNR],$0}' %s %s | xargs -0 -n2 mv 2>/dev/null";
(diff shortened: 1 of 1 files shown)


0xACE:
Beautiful. I haven't tested it yet, but if it's working as intended it's a genius solution. We will see if `awk` is a acceptable dependency.

One thing I would like to bring up is: The other day I added `sed` as a dependency, if we are using `awk` we could potentially change that other `sed` to use `awk` instead... (I don't mind either way)

jarun:
@0xACE `awk` is available in macOS and the *BSDs. So we can definitely have it.

@KlzXS can you also take a look at the function `archive_selection()` that @0xACE is referring to? It would be great if you can apply `awk` there as well.

jarun:
Looks fine in my tests. Thank you!

KlzXS:
It seems I ran into another problem while attempting to create an archive where `getutil()` doesn't recognize that I don't have `atool` installed.

If I'm understanding this correctly it gets the output of `which cmd` and if it's not NULL says it can use the tool. The problem is that my `which` outputs `which: no atool in ${PATH}`.

Perhaps we should look at the return value instead of the output.

jarun:
I see an alternative. `man which` shows:

```
EXIT STATUS
       0      if all specified commands are found and executable

       1      if one or more specified commands is nonexistent or not executable

       2      if an invalid option is specified
```

So we can also check the exit status from the command. Is it the same in your system?

KlzXS:
Mine also returns 1 when it doesn't find the command, but it just ignores invalid options.
I'd say it's safe to treat non-zero exit codes as not found and zero as found.

jarun:
Cool! Looks like *BSD and macOS also behave similarly.

Fixed at commit 8dc706026e89a20a7ec9db9b05e8b680136549ec.

Please check if the `sed` dependency can eb dropped.

jarun:
Please see commit e774634b5babc0c6bb972bdff796f5271671f7dc. I added the inline directive.

KlzXS:
I put in a pull request, but as I've noted there it would be wise to test it on different systems.

jarun:
Sure! Thanks a lot!

KlzXS:
Glad I could help even a little.

PR #322, opened 2019-08-12
Replaced 'sed' with 'awk'

KlzXS:
This should do it, I would just like to see somebody test it some more on different systems.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1299,12 +1299,7 @@ static void archive_selection(const char *cmd, const char *archive, const char *
 {
 	char *buf = (char *)malloc(CMD_LEN_MAX * sizeof(char));
 	snprintf(buf, CMD_LEN_MAX,
-#ifdef __linux__
-		"sed -ze 's|^%s/||' '%s' | xargs -0 %s %s", curpath, g_cppath, cmd, archive);
-#else
-		"cat '%s' | tr '\\0' '\n' | sed -e 's|^%s/||' | tr '\n' '\\0' | xargs -0 %s %s",
-		g_cppath, curpath, cmd, archive);
-#endif
+		"awk 'BEGIN{RS=\"\\0\"} {gsub(\"%s/\", \"\", $0); printf \"%%s\\0\", $0}' %s | xargs -0 %s %s", curpath, g_cppath, cmd, archive);
 	spawn("sh", "-c", buf, curpath, F_NORMAL);
 	free(buf);
 }


jarun:
@ldelossa, @laggardkernel, @sjmulder, @NapoleonWils0n can you please confirm this patch on *BSD?

@danielb2, @sukima, @philFernandez, @relapi can you please confirm this patch on macOS?

sjmulder:
Archiving a selection tested fine in OpenBSD and NetBSD. Can't test in FreeBSD right now.

danielb2:
it compiles. not sure what you want me to do to confirm

KlzXS:
@danielb2 Try creating an archive from a selection, use multiple files.

0xACE:
> it compiles. not sure what you want me to do to confirm

1. Go to a folder
2. `Y` to select all files.
3. hit `f` and create a archive from selection you can name it : example.zip
4. See if the files in `example.zip` were archived with a relative path rather than the full path. e.g:

it should be

```
myfile1
myfile2
myfolder/extrafiles
```

and not

```
/home/danielb2/examplefolder/myfile1
/home/danielb2/examplefolder/myfile2
/home/danielb2/examplefolder/myfolder/extrafiles
```

Essentially see if files are archived with a relative path rather than a full path

danielb2:
no side-effect is noted (nothing happens) and I get this message on the console after exit:

``` bash
[~/t/nnn]$ ./nnn
Usage:
  List:    bsdtar -tf <archive-filename>
  Extract: bsdtar -xf <archive-filename>
  Create:  bsdtar -cf <archive-filename> [filenames...]
  Help:    bsdtar --help
Usage:
  List:    bsdtar -tf <archive-filename>
  Extract: bsdtar -xf <archive-filename>
  Create:  bsdtar -cf <archive-filename> [filenames...]
  Help:    bsdtar --help

```

jarun:
@danielb2 looks like `bsdtar` on macOS does not have the `-a` flag.

Can you please try with this patch?

```
index f0623f3..8cc211c 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1288,11 +1288,11 @@ static void get_archive_cmd(char *cmd, char *archive)
        if (getutil(utils[ATOOL]))
                xstrlcpy(cmd, "atool -a", ARCHIVE_CMD_LEN);
        else if (getutil(utils[BSDTAR]))
-               xstrlcpy(cmd, "bsdtar -acvf", ARCHIVE_CMD_LEN);
+               xstrlcpy(cmd, "bsdtar -cvf", ARCHIVE_CMD_LEN);
        else if (is_suffix(archive, ".zip"))
                xstrlcpy(cmd, "zip -r", ARCHIVE_CMD_LEN);
        else
-               xstrlcpy(cmd, "tar -acvf", ARCHIVE_CMD_LEN);
+               xstrlcpy(cmd, "tar -cvf", ARCHIVE_CMD_LEN);
 }
 
 static void archive_selection(const char *cmd, const char *archive, const char *curpath)
```

Please create an archive with extension `.7z` and ensure from the file details (key `D`) the archive created is indeed a 7-zip archive.

jarun:
Note: On Linux, `bsdtar` creates a tar archive (instead of 7zip) if the `-a` (`--auto-compress`) flag is not used.

danielb2:
Same with mac it seems. `POSIX tar archive`  is what I see

jarun:
If you are saying that without `-a` you are seeing a 7z created as POSIX tar I am not sure why you got the help earlier:

```
[~/t/nnn]$ ./nnn
Usage:
  List:    bsdtar -tf <archive-filename>
  Extract: bsdtar -xf <archive-filename>
  Create:  bsdtar -cf <archive-filename> [filenames...]
  Help:    bsdtar --help
Usage:
  List:    bsdtar -tf <archive-filename>
  Extract: bsdtar -xf <archive-filename>
  Create:  bsdtar -cf <archive-filename> [filenames...]
  Help:    bsdtar --help
```

it essentially means the command did not work. And as I understand from your message the `-a` flag is not the problem.

Please remove my patch then. That's not the issue.

@KlzXS can you please check why the command failed on mac?

jarun:
@danielb2 what do you see with the original implementation (`sed` version)?

```
index f0623f3..fc3f499 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1299,7 +1299,12 @@ static void archive_selection(const char *cmd, const char *archive, const char *
 {
        snprintf(buf, CMD_LEN_MAX,
-               "awk 'BEGIN{RS=\"\\0\"} {gsub(\"%s/\", \"\", $0); printf \"%%s\\0\", $0}' %s | xargs -0 %s %s", curpath, g_cppath, cmd, archive);
+#ifdef __linux__
+               "sed -ze 's|^%s/||' '%s' | xargs -0 %s %s", curpath, g_cppath, cmd, archive);
+#else
+               "cat '%s' | tr '\\0' '\n' | sed -e 's|^%s/||' | tr '\n' '\\0' | xargs -0 %s %s",
+               g_cppath, curpath, cmd, archive);
+#endif
        spawn("sh", "-c", buf, curpath, F_NORMAL);
        free(buf);
 }
```

KlzXS:
I don't have a mac machine, but I could set up a virtualbox tomorrow if I find a more or less recent image.

It might be that it doesn't like the `-a` flag and that it just ignores the extension and makes a tar archive.

danielb2:
> utility missing

this message ^

jarun:
@danielb2 which OS were you testing on now?

danielb2:
i just noticed this laptop has an old version of nnn

2.6 works and gives: `blah.7z: POSIX tar archive`

jarun:
OK. v2.6 uses the `sed` version.

I am more interested in the current patch. Can you confirm with master?

Also, as I understand, you are testing with `bsdtar`, right?

sjmulder:
Ah, I just tested whether an archive was created. Following your steps now, on NetBSD 8:

`Y`, `f`, `y` ("archive from selection?"), `example.com`, return, `q`

I executed this from the root of the source tree, but only got the `misc` directory:

    netbsd8$ LD_LIBRARY_PATH=/usr/pkg/lib ./nnn
    a misc
    a misc/auto-completion
    [...]
    a misc/auto-completion/zsh/_nnn
    a misc/auto-completion/fish/nnn.fish
    netbsd8$

Same for  `example.7z` , `example.tar.gz`. But they do all seem to be the correct file format.

jarun:
@KlzXS looks like only a single path was picked. can you please check?

@sjmulder can you please try the `sed` version?

KlzXS:
That doesn't make any sense. Even if `awk` failed to trim the current path and the entries were left as absolute paths `xargs` still should have received all of them and passed them to `bsdtar`.

I don't see how this could happen if multiple entries were selected. Does this version of `bsdtar` support multiple files as input?

jarun:
Guys, I'm reverting this till we resolve the problem.

jarun:
@sjmulder we need more help from you.

Please confirm the `sed` version. If @KlzXS needs more tests, please help out.

KlzXS:
@jarun I have been able to reproduce the behavior @sjmulder reported and have two fixes.

The problem was that not all implementations of `awk` work nicely with `\0` bytes, but `gawk` (GNU awk) does. So one way to fix this is to specify that it must be `gawk` and the other would be to replace the `awk` in `batch_rename()` with `sed` that does the same thing.

Which one do you prefer?

jarun:
Thanks for the root cause! 

We'll go with `sed` on this. Does the reversal work or we need further changes?

KlzXS:
The thing in `archive_selection()` works fine. As for `batch_rename()` I would revert to what it was before `awk` just insert a call to `sed` between `paste` and `tr`. 

So the full value of `renamecmd` would be `paste -d'\n' %s %s | sed 'N; /^\\(.*\\)\\n\\1$/!p;d' | tr '\n' '\\0' | xargs -0 -n2 mv 2>/dev/null`. Everything in this script is POSIX.

Will you add this?

jarun:
Sure thing! Kindly raise the PR.

sjmulder:
Great work @KlzXS!

jungle-boogie:
FYI, I don't see any issues on amd64 OpenBSD 6.6 beta -current. 
`letssee.tar.gz: gzip compressed data, last modified: Sat Aug 17 07:05:11 2019, from Unix`

```
kern.version=OpenBSD 6.6-beta (GENERIC.MP) #221: Fri Aug 16 16:00:01 MDT 2019
    deraadt@amd64.openbsd.org:/usr/src/sys/arch/amd64/compile/GENERIC.MP
```

jarun:
@jungle-boogie which version did you try? sed/awk?

jungle-boogie:
Whatever is in `nnn` master with commit 8ad1c577e52f16525c1babb40af9cbdf972f3d73

jarun:
I think it's the `sed` version. Thanks!

PR #325, opened 2019-08-15
Replacing last occurance of 'awk'

KlzXS:
I've tested it on Linux, it works, and it should also work on MacOS and BSDs as I only used stuff defined by POSIX this time.


jarun:
Thank you!

Issue #225, opened 2019-03-16
CPU using %100 by nnn


abdyek:
Hi,
I use nnn v2.3 on manjaro linux - i3.
It is running nnn at background when I close it.
https://postimg.cc/s11YZnQ4

jarun:
Please share the exact steps to reproduce.

abdyek:
Ohh.. sorry
Firstly, my english is not good so I am sorry for it
I use vim for programming. Folders of my projects are at /mnt/Depo/Project
I added `export NNN_BMS='d:/mnt/Depo;p:/mnt/Depo/Projects;'` to .bashrc file,
I prefer to go to my project folder using leader key + p and return shell with ! to open vim,
Sometimes I open nnn another terminal window and close it
What can I share info about this issue ?

jarun:
> What can I share info about this issue ?

I couldn't reproduce your issue. So I need definite steps - how did you start `nnn`, what's your terminal emulator and shell. What did you do after starting `nnn`, basically a way to reproduce it on my setup locally so I can analyze the issue.

`nnn` _kills_ itself if the parent process has exited so I have to know the steps when I guess this is not happening.

Also, please try on master and see if you are seeing the issue.

abdyek:
I use i3-wm, gnome-terminal, vim and manjaro linux than
I will explain it step by step
- I opened a new gnome terminal using $mod key + enter
- I started nnn with commant of "nnn"
- I went to my bookmarks with leader key + p (export NNN_BMS='d:/mnt/Depo;p:/mnt/Depo/Projects;')
- I returned shell with ! and started vim using "vim"
- I worked in the terminal for a while
- I opened new another terminal using $mod key + enter
- I started nnn with commant of "nnn"
- I used it and closed it

jarun:
OK.

> I worked in the terminal for a while

did you quit vim and nnn or left vim open?

> I used it and closed it

what is _it_? Used the terminal and closed it directly? Did you quit `nnn`?

jarun:
I have tried to reproduce it using your steps and directly closing the terminal (xfce4-terminal) at the points I noted above. However, I do not see the issue. In both the cases `nnn` quits when I close the terminal directly.

Can you try this on the latest release v2.4?

jarun:
Please revert back when you have more info.

cowabungadude69:
Here's a simple way to recreate it I think.

+ open a terminal
+ start nnn
+ enter shell with !
+ start nnn in that shell
[...]

PR #327, opened 2019-08-19
Fix 100% CPU usage if STDIN closes

KlzXS:
As the title says. We should also update the troubleshooting section of the wiki now that this is no longer a thing.

Remember that other programs might not check for this so if you spawn them as children they can still use 100% CPU. People need fix their programs to not do that.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3447,6 +3447,10 @@ nochange:
 			goto begin;
 		}
 
+		/* If STDIN is no longer a tty (closed) we should exit */
+		if(!cfg.picker && !isatty(STDIN_FILENO))
+			return;
+
 		sel = nextsel(presel);
 		if (presel)
 			presel = 0;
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -3447,6 +3447,10 @@ static void browse(char *ipath)
 			goto begin;
 		}
 
+		/* If STDIN is no longer a tty (closed) we should exit */
+		if(!cfg.picker && !isatty(STDIN_FILENO))
```
jarun: Please reverse the check to:

    if (!isatty(STDIN_FILENO) && !cfg.picker)
as most of the time we are not in picker mode.

Also, please note the space after `if`.

KlzXS: Right! Done.

jarun:
Thank you so much! I will update the Troubleshooting page now.

Fix  #225.

PR #330, opened 2019-09-04
Added simple launcher file for ubuntu

chernikovdmitry:
Just launcher to run nnn in xfce4-terminal with all env vars and aliases from ~/.bashrc & co
Useful if someone wants to lock nnn to Ubuntu Launcher panel and want it run with respect to user local $VISUAL/$EDITOR settings


jarun:
Can you please update the documentation here: https://github.com/jarun/nnn/wiki/Advanced-use-cases#configure-as-default-fm

I would prefer to document this in a single place. The wiki is editable for all.

PR #335, opened 2019-10-03
added trash plugin

superDuperCyberTechno:
For people who find `rm` too scary.

### DIFF
--- a/plugins/trash
+++ b/plugins/trash
@@ -1,0 +1,29 @@
+#!/usr/bin/env sh
+
+# Description: Send a file or a selection of files to the trash can rather than outright delete it/them (rm)
+# Dependencies: trash-cli
+#
+# Shell: POSIX compliant
+# Author: superDuperCyberTechno
+
+#fetch the selection
+selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
+
+#if the selection file is not empty
+if [ -s $selection ]; then
+    #convert to array
+    arr=$(tr '\0' '\n' < "$selection")
+
+    #trash all elements (files) in array
+    for entry in $arr
+    do
+        trash "$entry"
+    done
+
+    #delete the selection file
+    rm ${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
+
+#if there is no selection just trash the currently selected file
+elif ! [ -z "$1" ]; then
+    trash `pwd`/`basename "$1"`
+fi
--- a/plugins/trash
+++ b/plugins/trash
@@ -17,7 +17,7 @@ if [ -s $selection ]; then
     #trash all elements (files) in array
     for entry in $arr
     do
-        trash "$entry"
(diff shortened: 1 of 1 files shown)


jarun:
It's already available. `nnn` works very well with `trash-cli`: https://github.com/jarun/nnn#utility-dependencies

Just export:

    NNN_TRASH=1
see https://github.com/jarun/nnn#configuration

jarun:
We didn't decouple it as a plugin so users can just remove files and selection as they usually do in `nnn`.

superDuperCyberTechno:
Hah! Nice. Ah, well. At least I learned something!

artkpv:
Nice! Yes, I removed 8GB of photos yesterday by accident via `nnn` and spent a day recovering them. Learning... :)

jarun:
Always read the documentation first: https://github.com/jarun/nnn#utility-dependencies

jarun:
@artkpv can you please explain your workflow step by step? I would like to understand the disconnect.

artkpv:
@jarun The most unclear thing to me was (and partially now) is the selection and that there is no progress for mv and cp. I selected those photoes then moved dir then selected something else. At that point I thought photos weren't selected but they were! I press X and it wipes out data. It is really fast. I used to DoubleCommander/TotalCommander/MC workflow were selection is shown at the pane and nothing else is selected. Not that thing here :) Now I double check selection before any operation (pressing M). Probably include some mode for learners? Like warning and a number of selected files/dirs.

jarun:
We show a prompt that goes like "use force?" How about making it more explicit? Like "remove n files in force mode (unrecoverable)". If the user isn't sure he can always cancel, press ^C in the rm prompt to ecit rm. The selection stays so he can cross check.

artkpv:
Sounds good!

jarun:
@artkpv this is implemented at commit 94e4dc51ba7ebcbdaf94dc3bb20c59ee22dc84e5.

Check both the current file and selection workflows.

artkpv:
@jarun That is better! I use trash-cli now. I guess this message won't be displayed as trash can recover files.

jarun:
That's right. This workflow is for people without `trash-cli`.

jarun:
@artkpv can you please confirm with master (specifically commit 4db8a61dea19f3314b9bd035e04eb32a6cc95b26) that trashing works fine with the new xargs changes?

artkpv:
@jarun Well, I use pacman (Arch Linux), just updated nnn to 2.7 (`nnn -v`) and it deletes to Wastebin on KDE. No warning ("fosibly delete ..." ) shown.

jarun:
That's correct behaviour. When you have trash there is no question.

But 2.7 is the earlier release. I wanted to knwo the results from master.

You can download the GitHub master snapshot: https://github.com/jarun/nnn/archive/master.zip

Then compile the program (run `make`), then run `./nnn` and see if trashing is working as expected.

PR #336, opened 2019-10-05
Archives: add support for unrar-free.

Steap:


jarun:
Hi,

- is unrar-free available on *BSD and macOS?
- why not atool?

KlzXS:
I've done a bit of research on this program and I came up with the following:

- It is a dead project that was supposed to be a free (as in freedom) port of RARLAB's unrar
- It doesn't support newer RAR formats
- It's not available on major BSDs, MacOS and Arch Linux; I gave up after not being able to find it in any of these

My opinion is that we shouldn't use `unrar-free`, maybe `unrar`, but not `unrar-free`. If you really want it patch it in yourself.

jarun:
Closing the PR as we would like to avoid this new dependency.

Steap:
Yep, sorry about the noise. I really thought unrar-free was easily available on most distros. I wanted to avoid using atool since it relies on rar(1) which is non-free.

jarun:
No problem!

I am thinking of adding a Wiki section `user patches` which would have these kind of exploratory stuff. Would you like to add this there?

0xACE:
@Steap
This might not be exactly relevant, but I have this in my `~/.bashrc` which I found somewhere on the web, you could easily readjust it and put in in your `$PATH` named as `atool` and adjust it to your liking. Or do the `~/pre-bin` trick i have talked about previously somewhere amongst these `nnn` issues...

```extract () {
   if [ -f "$1" ] ; then
       case "$1" in
           *.tar.bz2)   tar xvjf "$1"    ;;
           *.tar.gz)    tar xvzf "$1"    ;;
           *.bz2)       bunzip2 "$1"     ;;
           *.rar)       unrar x "$1"       ;;
           *.gz)        gunzip "$1"      ;;
           *.tar)       tar xvf "$1"     ;;
           *.tbz2)      tar xvjf "$1"    ;;
           *.tgz)       tar xvzf "$1"    ;;
           *.zip)       unzip "$1"       ;;
           *.Z)         uncompress "$1"  ;;
           *.7z)        7z x "$1"        ;;
           *)           echo -- "don't know how to extract '$1'..." ;;
       esac
   else
       echo -- "'$1' is not a valid file!"
   fi
}
```

Steap:
> No problem!
> 
> I am thinking of adding a Wiki section `user patches` which would have these kind of exploratory stuff. Would you like to add this there (probably more like an indicative patch to open specific archive types in a preferred application)?

I'm not I understand exactly what you mean. Would you want to create some kind of archive for patches that may be useful to people, but not enough to be merged in master?


> @Steap
> This might not be exactly relevant, but I have this in my `~/.bashrc` which I found somewhere on the web, you could easily readjust it and put in in your `$PATH` named as `atool` and adjust it to your liking. Or do the `~/pre-bin` trick i have talked about previously somewhere amongst these `nnn` issues...


Yes, I was thinking of "overwriting" atool to trick nnn into using my own "extract-everything" function. This would not be available to all users, though :/

jarun:
> Would you want to create some kind of archive for patches that may be useful to people, but not enough to be merged in master?

Yes, that's what was in my mind. However, it looks like the script from @0xACE has a better application.

@0xACE how about having your script as a plugin? Users who have this use case can use this plugin. I think that's better than having to install atool or faking it.

0xACE:
It's not really my script, it has been in my `.bashrc` for years...

The idea of overwriting `atool` is `nnn` utilizes it naturally...

Of course it can be made into a plugin or something else. In my own case, I installed `atool` and never bothered with it since. I've been double booked today, and seemingly this weekend, so I'm not sure when I can look closer, but anything you do is practically fine with me.

jarun:
Let's leave it as an atool replacement instead of adding a new plugin.

jarun:
@Steap please take a look at plugin `nuke` (it's a file opener) which you can modify to use your own utilities easily.

PR #345, opened 2019-10-09
Implemented edit selection buffer

KlzXS:
Currently there is no way to unselect something that was wrongly selected other than clearing the whole selection, which at times can get really annoying. The simplest way to solve this was to just allow the user to edit the selection buffer.

As for the keybind, function name and the action name I just chose something so that it has a name. Feel free to suggest other options if you want me to change them.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -889,7 +889,8 @@ static void endselection(void)
 static bool seledit(void)
 {
 	bool ret = FALSE;
-	int fd, i;
+	int fd, lines = 0;
+	ssize_t count;
 
 	if (!selbufpos) {
 		DPRINTF_S("empty selection");
@@ -921,31 +922,35 @@ static bool seledit(void)
 		goto emptyedit;
 	}
 
-	i = read(fd, pselbuf, selbuflen);
+	count = read(fd, pselbuf, selbuflen);
 	close(fd);
 	unlink(g_tmpfpath);
 
-	if (!i) {
+	if (!count) {
 		ret = TRUE;
 		goto emptyedit;
 	}
 
-	if (i < 0) {
+	if (count < 0) {
 		DPRINTF_S("error reading tmp file");
 		goto emptyedit;
 	}
 
 	resetselind();
-	nselected = 0;
-	selbufpos = i;
+	selbufpos = count;
 	/* The last character should be '\n' */
-	pselbuf[--i] = '\0';
-	for (--i; i > 0; --i) {
+	pselbuf[--count] = '\0';
+	for (--count; count > 0; --count) {
 		/* Replace every '\n' that separates two paths */
-		if (pselbuf[i] == '\n' && pselbuf[i+1] == '/') {
-			++nselected;
-			pselbuf[i] = '\0';
+		if (pselbuf[count] == '\n' && pselbuf[count + 1] == '/') {
+			++lines;
+			pselbuf[count] = '\0';
 		}
+	}
+
+	if (lines > nselected) {
+		DPRINTF_S("files added to selection");
+		goto emptyedit;
 	}
 
 	writesel(pselbuf, selbufpos - 1);
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -886,6 +886,58 @@ static void endselection(void)
 	}
 }
 
+static bool seledit(void)
+{
+	int fd, i;
+
+	if (!selbufpos)
+		return FALSE;
+
+	fd = create_tmp_file();
+	if (fd == -1)
+		return FALSE;
+
+	seltofile(fd, NULL);
+	close(fd);
+
+	spawn(editor, g_tmpfpath, NULL, NULL, F_CLI);
+
+	if ((fd = open(g_tmpfpath, O_RDONLY)) == -1){
+		unlink(g_tmpfpath);
+		return FALSE;
+	}
+
+	struct stat sb;
+	fstat(fd, &sb);
+
+	if (sb.st_size > selbuflen){
+		pselbuf = xrealloc(pselbuf, sb.st_size);
+		if(!pselbuf)
+			errexit();
+	}
+
+	i = read(fd, pselbuf, selbuflen);
+	close(fd);
+	unlink(g_tmpfpath);
+
+	resetselind();
+	nselected = 0;
+	selbufpos = i;
+	while(i){
+		if (pselbuf[--i] == '\n'){
+			++nselected;
+			pselbuf[i] = '\0';
+		}
+	}
+
+	if (selbufpos)
+		writesel(pselbuf, selbufpos - 1);
+	else
+		writesel(NULL, 0);
+
+	return TRUE;
+}
+
 static bool selsafe(void)
 {
 	/* Fail if selection file path not generated */
@@ -4139,6 +4191,12 @@ nochange:
 
 			printwait(messages[NONE_SELECTED], &presel);
 			goto nochange;
+		case SEL_SELEDIT:
+			if (!seledit()){
+				printwait(messages[NONE_SELECTED], &presel);
+				goto nochange;
+			}
+			break;
 		case SEL_CP:
 		case SEL_MV:
 		case SEL_RMMUL:
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -888,21 +888,27 @@ static void endselection(void)
 
 static bool seledit(void)
 {
+	bool ret = FALSE;
 	int fd, i;
 
-	if (!selbufpos)
+	if (!selbufpos) {
+		DPRINTF_S("empty selection");
 		return FALSE;
+	}
 
 	fd = create_tmp_file();
-	if (fd == -1)
+	if (fd == -1) {
+		DPRINTF_S("couldn't create tmp file");
 		return FALSE;
+	}
 
 	seltofile(fd, NULL);
 	close(fd);
 
 	spawn(editor, g_tmpfpath, NULL, NULL, F_CLI);
 
-	if ((fd = open(g_tmpfpath, O_RDONLY)) == -1){
+	if ((fd = open(g_tmpfpath, O_RDONLY)) == -1) {
+		DPRINTF_S("couldn't read tmp file");
 		unlink(g_tmpfpath);
 		return FALSE;
 	}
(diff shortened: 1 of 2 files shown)


KlzXS:
CONTROL('G') is already used isn't it. Will change that, just need a suggestion what to bind it to.

jarun:
Awesome! Many thanks for this! I willr eview it soon. Added @0xACE as well.

src/nnn.c
```diff
@@ -886,6 +886,58 @@ static void endselection(void)
...
+
+	if ((fd = open(g_tmpfpath, O_RDONLY)) == -1){
+		unlink(g_tmpfpath);
+		return FALSE;
+	}
+
+	struct stat sb;
+	fstat(fd, &sb);
+
+	if (sb.st_size > selbuflen){
```
jarun: We should not allow this case. Let users remove but not add anything in the file directly. It's dangerous. If user adds, we can simply discard everything and show a message.

Also, can someone add the `NUL` separator easily in the editor? See my comment on line 927.

src/nnn.c
```diff
@@ -886,6 +886,58 @@ static void endselection(void)
...
+	}
+
+	i = read(fd, pselbuf, selbuflen);
+	close(fd);
+	unlink(g_tmpfpath);
+
+	resetselind();
+	nselected = 0;
+	selbufpos = i;
+	while(i){
```
jarun: while (i) {
// add spaces

src/nnn.c
```diff
@@ -886,6 +886,58 @@ static void endselection(void)
...
+		if (pselbuf[--i] == '\n'){
```
jarun: This breaks as soon as there is a `\n` within a filename. And it can be dangerous with `rm`.

How about we use the `.selection` file format? Entries are `NUL`-separated.

src/nnn.c
```diff
@@ -886,6 +886,58 @@ static void endselection(void)
...
+			++nselected;
+			pselbuf[i] = '\0';
+		}
+	}
+
+	if (selbufpos)
```
jarun: If we use the `NUL`-separator this may be redundant.

src/nnn.c
```diff
@@ -4139,6 +4191,12 @@ static void browse(char *ipath)
 
 			printwait(messages[NONE_SELECTED], &presel);
 			goto nochange;
+		case SEL_SELEDIT:
+			if (!seledit()){
+				printwait(messages[NONE_SELECTED], &presel);
```
jarun: `seledit()` may fail due to multiple reasons. I think we can add debug logs and just show a generic message like "edit failed!".

src/nnn.h
```diff
@@ -209,6 +210,8 @@ static struct key bindings[] = {
 	{ 'a',            SEL_SELALL },
 	/* Show list of copied files */
 	{ 'M',            SEL_SELLST },
+	/* Edit selection buffer */
+	{ CONTROL('G'),   SEL_SELEDIT },
```
jarun: `^G` is indeed used. We have `^K  Sel range, clear`. So how about `K` (capital)?

0xACE:
Since this seems urgent I'll review it the next chance I get. But the idea sounds pleasing to me :-D

src/nnn.c
KlzXS: How is that dangerous? If they enter an invalid file an pass it to a program the program should detect that the file doesn't exist. They could still try and fill the buffer with gigabytes of data but that would only crash the program when allocation fails at worst.

I for one have an esay way of entering any unicode character i want and with that also NULL but I realise that that may not be easy for everyone so I don't think that's the best solution. Having newlines in the filename and any other weird character is generally a bad idea. It would already break something like batch rename if that character gets removed. I think we should just not support newlines in filenames and maybe even NULLs, you may even be allowed to have a NULL in the name?

src/nnn.c
KlzXS: The check is for when you empty the buffer completely, as calling `writesel(..., -1)` gives an error.

src/nnn.c
jarun: > How is that dangerous?

The idea is to edit the selection buffer by removing files added to it already... not adding new ones manually. And when I wrote this I have the `NUL`-separator in mind, not the newline-separator.

> Having newlines in the filename and any other weird character is generally a bad idea.
> I think we should just not support newlines in filenames

Believe it or not, macOS ships some. We can't ignore this case.

> It would already break something like batch rename

There are some safeguards:
- we do not support deleting files using batch rename, so nothing would be removed
- we also match the number of lines in source and final list, and exit without changes if they don't match

> you may even be allowed to have a NULL in the name?

That's not possible as it terminates both ASCII and Unicode strings.

jarun: > I don't think that's the best solution.

Staying by the book is always the best solution. We picked the `NUL`-separator specifically because `\n` had issues with files having newline in them.

src/nnn.c
jarun: Got it!

Note: if we use the `NUL`-terminator we wouldn't need to scan through the entries in the file and replace `\n` with `\0`.

src/nnn.c
KlzXS: Editing files with \0 as separator can be weird as it would all be in a single line, if \0 is displayed at all. But I really want this to be a thing so what do you think about inverting \0 and \n, i.e. when writing selection to the tmp file we replace all \n with \0 and vice versa. This way we can display it in multiple lines and people who think putting newlines in filenames can still do that but insert a \0 instead.

Other than that prevent people from adding more files, they can only remove. But what about renaming them? Let them rename them while the total length is less or equal to what it already was, if not clear selection? We can't really check if all the entered paths are valid.

This turned out to be way more complicated that I thought it would be. But I still want it to be a thing.

KlzXS: Renaming as in changing the file that is selected.

jarun: > when writing selection to the tmp file we replace all \n with \0 and vice versa

How do you differentiate with actual end of path in that case? But I think we have a solution.

I think when writing selection to file, we don't add a `NUL` after the last path (please confirm). All other `NUL`-separators should be followed by a `/` immediately. Even if there is a newline (`\n`) at the end of a filename, it can have a `NUL` after it but not a `/`. With this we can differentiate when we are reading back from modified file to buffer. Just convert the `\n`s which are followed by a `/` after them to `\0`. Also ensure that we add a `NUL` after the last path in the selection buffer. What do you think?

> Let them rename them while the total length is less or equal to what it already was, if not clear selection?

This. They are not supposed to rename either and this kind of asserts that we don't expect people to do that.

jarun: > This turned out to be way more complicated that I thought it would be. But I still want it to be a thing.

That's the spirit! When I set out to write a file manager that doesn't make the user wait on a Pi, it seemed easy. Just take a look at some of the algorithms today. ;)

But despair not, once you implement this, you'll become a slightly better programmer.

KlzXS: I don't quite get what you want to do. I wanted to just swap nulls and newlines both when reading and writing to file as then the user will be able to see the buffer as separate lines. But I would need to read your proposal at least one more time and maybe even see it explained slightly "better".

jarun: In brief:

- do not convert anything when writing selection buffer to file
- when reading back to selection buffer _replace only the `\n` which have a `/` after them_ (we store full paths so next file path always starts with a `/`) to `\0`.

jarun: Also confirm if we need to add a `\0` at the end of selection buffer when reading in.

KlzXS: Ah, ok. That is doable. We need that final \0 as string need to be terminated. 

So the final thing would look something like `if((i == selbufpos && pselbuf[i-1] == '\n') || (newline and /)) do stuff;`. I'm currently on mobile so I don't feel like writing it out, but I will rewrite this later tonight and push the commits with all the edits.

jarun: You can loop till the last char as you are doing now (with additional check for the following `/`).

At the end you can add the check for the last char.

KlzXS: I'm looping from behind, but it doesn't really matter. I can make a check for the last character to see if it's `\n` and if it isn't clear the selection.

But I don't know how far we should go with these checks. The only real way to see if the file contains real selection data after the edit would be to parse all the paths and even that is not easy if we have a file located at `/path/to\n/file` as that would be interpreted wrongly by the described method.

So my take is to assume that the last character is indeed `\n`. In my single test if it was not the selection was cleared, I don't really understand why or if it works that way always. I feel like we should just say if you don't enter the thing we expect you to it's UB, but I'll let you decide.

jarun: > The only real way to see if the file contains real selection data after the edit would be to parse all the paths

Let's leave it to the user not to screw up his filesystem.

> So my take is to assume that the last character is indeed \n

The last char would be a EOF (not `\0`, not `\n`.

KlzXS: > The last char would be a EOF (not \0, not \n.

Well yes, but that will no get written into the buffer, only until the last character of the file which should be `\n` from the last path.

jarun: How would a `\n` appear in the last path (if it's not in the file name)? I was suggesting to avoid replacing the last `\0` even if it exists.

src/nnn.c
```diff
@@ -888,54 +888,77 @@ static void endselection(void)
 
 static bool seledit(void)
 {
+	bool ret = FALSE;
 	int fd, i;
```
jarun: Please use `ssize_t` for `i` (as per read() return type). Rename it to `count`.

src/nnn.c
```diff
@@ -888,54 +888,77 @@ static void endselection(void)
...
 	resetselind();
 	nselected = 0;
 	selbufpos = i;
-	while(i){
-		if (pselbuf[--i] == '\n'){
+	/* The last character should be '\n' */
+	pselbuf[--i] = '\0';
+	for (--i; i > 0; --i) {
+		/* Replace every '\n' that separates two paths */
+		if (pselbuf[i] == '\n' && pselbuf[i+1] == '/') {
```
jarun: pselbuf[i+1] -> pselbuf[i + 1]

// space

src/nnn.c
```diff
@@ -888,54 +888,77 @@ static void endselection(void)
...
 			++nselected;
```
jarun: This will tell us if the number of files have increased. Store the value of `nselected` at start of this function and confirm that it's less than or equal to initial value. If it's greater, user has added files. In that case, fail.

jarun:
LGTM. I will test this today.

jarun:
Note for @jarun: add the keybind in show_help(), update README. Update Wiki and man page Selection sections.

jarun:
Thank you!

0xACE:
I tried it, and it seems to be working. I'll update you if I find problems with it. Great work!

Issue #337, opened 2019-10-06
ToDo list


jarun:
Rolled from #332.

### Ready for next release

- run custom (non-shell-interpreted) commands like plugins
- sessions (thanks @annagrram)
- configure _cd-on-quit_ as the default behaviour
- _copy as_ and _move as_ workflow
- edit , flush selection buffer (thanks @KlzXS)
- changed the key to size sort to <kbd>z</kbd>
- ignore events during selection so the `+` symbol is not lost
- mount archives using archivemount
- smoother double click handling
- keybind collision checker (for users setting custom keybinds) (thanks @annagrram)
- show size of file in bytes in status bar in disk usage mode
- pass unresolved path as second argument (`$2`) to plugin
- mechanism for plugins to control active directory
- plugins
  - `getplugs` - a lot of intelligence added to the plugin installer (thanks @KlzXS)
  - `drag-file` & `drop-file`: drag & drop files using dragon
  - `gutenread`: browse, download and read from Project Gutenberg
  - `suedit` - edit file with superuser permissions
  - `fzhist` - fuzzy select commands from history, edit and run
  - `fzcd` - change to a fuzzy-searched directory
  - `pskill` - fuzzy list a process or zombies by name and kill
  - `exetoggle` - toggle executable status of hovered file
  - `treeview` - informative tree output with file permissions and size
  - `checksum` - recursively calculate checksum for files in hovered directory
  - `fzy-open` renamed to `fzopen`
- additional key <kbd>;</kbd> to execute plugin
- more explicit force removal message
- export current file as `$NNN` (instead of `$NN`)
- fix opening files from browser

### Proposed features and tasks (up for grabs)

- [ ] print disk usage and directory snapshot
- [ ] a video exploring `nnn` plugins
- [ ] script to duplicate files in selection

Anything else which would add value (please discuss in this thread).

List of [completed features and tasks](https://github.com/jarun/nnn/wiki/Feature-history).

KlzXS:
I've been meaning to ask, what exactly does "script to duplicate selection in respective directories" mean? Does it mean if I selected some entries in `dirx` that it should select all available entries with the same name in `diry`? Should this be a plugin?

jarun:
@KlzXS the script would ask for a suffix from the user and create duplicates of paths in selection. I don't see a pressing need for it right now.

KlzXS:
Right. I just wanted to clear up the wording a bit, I found it to be a bit ambiguous.

[...]

PR #348, opened 2019-10-09
Implement a basic keybinding collision checker. (ToDo list #337)

annagrram:


jarun:
Wow! This was quick! Thanks for the PR. I will this review shortly.

Note for @jarun: Please document this in the Wiki Advanced...

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+} bucket_t, *pbucket_t;
+
+static bool is_collision_in_bucket(pbucket_t buckets, int sym);
+static void free_buckets(pbucket_t buckets);
+
+int main() {
+    bucket_t buckets[BUCKETS_ARRAY_SIZE] = {0};
+    bool collision_detected = FALSE;
+
+    for (int i = 0; i < sizeof(bindings) / sizeof(struct key); ++i) {
```
jarun: Please declare and initialize `i` at the top of the function. Compiler warns.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+        pbucket_t curr_bucket = &buckets[curr_sym % BUCKETS_ARRAY_SIZE];
+
+        if (is_collision_in_bucket(curr_bucket, curr_sym)) {
+            printf("Collision of key %s detected\n", keyname(curr_sym));
+
+            collision_detected = TRUE;
+        }
+    }
+
+    if (!collision_detected)
```
jarun: I think we can remove this. Notifying only if a collision is detected should be enough. No news is good news.

jarun: Get rid of the boolean too.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+    bucket_t buckets[BUCKETS_ARRAY_SIZE] = {0};
+    bool collision_detected = FALSE;
+
+    for (int i = 0; i < sizeof(bindings) / sizeof(struct key); ++i) {
+        int curr_sym = bindings[i].sym;
+        pbucket_t curr_bucket = &buckets[curr_sym % BUCKETS_ARRAY_SIZE];
+
+        if (is_collision_in_bucket(curr_bucket, curr_sym)) {
+            printf("Collision of key %s detected\n", keyname(curr_sym));
+
```
jarun: Blank line not required.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+    free_buckets(buckets);
+}
+
+bool is_collision_in_bucket(pbucket_t bucket, int sym) {
+    while (bucket) {
+        if (bucket->sym == sym) {
+            return TRUE;
+        } else if (bucket->sym == 0) {
+            bucket->sym = sym;
+
```
jarun: Blank line not required.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+            break;
+        }
+        else if (!bucket->next) {
```
jarun: Keep the `else if` along with the bracket.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+            bucket->next = (pbucket_t)malloc(sizeof(bucket_t));
+            bucket->next->sym = sym;
+            bucket->next->next = NULL;
+
```
jarun: Redundant bank line.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    if (!collision_detected)
+        printf("No collisions detected\n");
+
+    free_buckets(buckets);
+}
+
+bool is_collision_in_bucket(pbucket_t bucket, int sym) {
```
jarun: Can we implement this as a cute hash (bitmap)? The value of the `key` should be the bit number in the bitmap. Set if not set already. If set, you have a collision.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+            bucket = bucket->next;
+        }
+    }
+
+    return FALSE;
+}
+
+void free_buckets(pbucket_t buckets) {
+    for (int i = 0; i < BUCKETS_ARRAY_SIZE; ++i) {
```
jarun: `int i = 0` goes outside.

misc/keybind-collision-checker/collision.c
jarun: Note: please check the feasibility. I am not sure how far the CONTROL+char combinations can scale to. If we can have a max value (up to) 10000, the bitmap would be ~1.25K. Good enough!

annagrram: Well, that made everything simpler :)
Changed that to a boolean array for more straightforward implementation.

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,20 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "../../src/nnn.h"
+
+int main() {
+    int i;
+
+    // CONTROL(KEY_MAX) == 447
+    bool bitmap[CONTROL(KEY_MAX)] = {FALSE};
```
jarun: bool is unsigned char internally. Can we have a true bitmap where each key is represented by 1 bit (and not byte)?

The bitmap size would be: ((CONTROL(KEY_MAX) + 7) / 8).

misc/keybind-collision-checker/collision.c
```diff
@@ -0,0 +1,20 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "../../src/nnn.h"
```
jarun: When it's ready, we will have it in `nnn.c` and expose an option to test collision.

misc/keybind-collision-checker/collision.c
jarun: Also, on my system:

`KEY_MAX`: 511
`CONTROL(KEY_MAX)` = 447

Probably we should use `KEY_MAX`.

annagrram: I'll check again and change the length if necessary. Thanks for noticing.

I don't think there's a simple way of doing such a big bitmap in plain C.
I guess the easiest way would be to create an array with the correct number of bits and each time to calculate in which cell the bit should be and then calculate which bit should be checked or flipped and do it.

I'm not sure this increase in code complexity is worth the relatively minor difference in memory.
Modern and less than modern system should handle an array of bytes of length 511 quite easily. Especially when the function that creates it is short-lived and the memory isn't occupied for long.

What do you think?

jarun: > Modern and less than modern systems should handle an array of bytes of length 511 quite easily

True. There's not a huge save there.

Can you add it in nnn.c and add a program option?

annagrram: Sure.
Will fix and commit soon.

annagrram: Done.

I chose the flag K, feel free to change it.

jarun:

jarun:
@annagrram could you make some time to add the the patch for coreutils v8.31 to [advcmpv](https://github.com/jarun/advcpmv)?

annagrram:
Sure.
I'll take a look later today.

PR #352, opened 2019-10-11
Added 2 plugins for drag and drop support

0xACE:
These scripts uses https://github.com/mwh/dragon
and curl.

It allows nnn to drag and drop files either to other programs, or to
itself.

Not that I recommend adopting it. But just incase anyone has a need for it, here it is.

It has helped me on multiple occasions, particularly when I want to upload files into the webbrowser via `nnn`.

The shell scripts are not polished and could probably need to be tended by someone who cares more.

### DIFF
--- a/plugins/drag-file
+++ b/plugins/drag-file
@@ -1,0 +1,49 @@
+#!/usr/bin/env sh
+
+# Description: Open a Drag and drop window, to drop files onto other programs
+#
+# Shell: POSIX compliant
+# Author: 0xACE
+
+selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
+resp=f
+all=
+
+dnd()
+{
+    dragon "$@"
+}
+
+function use_all()
+{
+    echo -n "mark --all (a) [default=none]: "
+    read resp
+    if [ "$resp" = "a" ]; then
+        all="--all"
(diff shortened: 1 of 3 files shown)


plugins/drag-file
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env sh
+
+# Description: Open a Drag and drop window, to drop files onto other programs
+#
```
jarun: Please provide some details and link to the dependency.

jarun: Please add notes on this plugin in the table in README.

plugins/drop-file
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env sh
+
+# Description: Provides drag and drop window for files.
+#
+# Files that are dropped will be added to nnn's selection
+# Some webbased files will be downloaded to current directory with curl
```
jarun: web based

plugins/drop-file
```diff
@@ -0,0 +1,39 @@
...
+# and it may overwrite some existing files
+#
+# The user has to mm to clear nnn's selection first
```
jarun: The user has to press mm ...

jarun:
Thank you!

PR #353, opened 2019-10-12
Mouseclicks are no longer delayed

0xACE:
This patch changes the mouseinterval to 0. And handles detection of
doubleclicking events internally.

This allows much faster response times for mouse clicks as nnn no longer
has to wait for ncurses to confirm what type of click it was.

It introduces a new config setting for changing how many milliseconds it
may take for a double click to be registered.

Note: It might change the behavior of `nav-as-you-type`. But I'm not sure what the intended behavior is. It is very hard to track why this new behavior acts differently than the current master. But it can be adjusted if required...

If it isn't clear, the current master delays mouse clicks with 400ms (0.4s!!!) and this patch makes the program respond instantly, just like other CLI tools supporting mouse events...

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -107,6 +107,7 @@
 #define S_BLKSIZE 512 /* S_BLKSIZE is missing on Android NDK (Termux) */
 #endif
 
+#define _ABSSUB(N, M) (((N) <= (M)) ? ((M) - (N)) : ((N) - (M)))
 #define LEN(x) (sizeof(x) / sizeof(*(x)))
 #undef MIN
 #define MIN(x, y) ((x) < (y) ? (x) : (y))
@@ -1015,12 +1016,12 @@ static bool initcurses(mmask_t *oldmask)
 	//intrflush(stdscr, FALSE);
 	keypad(stdscr, TRUE);
 #if NCURSES_MOUSE_VERSION <= 1
-	mousemask(BUTTON1_CLICKED | BUTTON1_DOUBLE_CLICKED, oldmask);
+	mousemask(BUTTON1_PRESSED | BUTTON1_DOUBLE_CLICKED, oldmask);
 #else
-	mousemask(BUTTON1_CLICKED | BUTTON1_DOUBLE_CLICKED | BUTTON4_PRESSED | BUTTON5_PRESSED,
+	mousemask(BUTTON1_PRESSED | BUTTON4_PRESSED | BUTTON5_PRESSED,
 		  oldmask);
 #endif
-	mouseinterval(400);
+	mouseinterval(0);
 	curs_set(FALSE); /* Hide cursor */
 	start_color();
 	use_default_colors();
@@ -3542,6 +3543,8 @@ static void browse(char *ipath)
 	struct stat sb;
 	char *path, *lastdir, *lastname, *dir, *tmp;
 	MEVENT event;
+	struct timespec mousetimings[2] = {{.tv_sec = 0, .tv_nsec = 0}, {.tv_sec = 0, .tv_nsec = 0}};
+	bool currentmouse = 1;
 
 	atexit(dentfree);
 
@@ -3663,7 +3666,7 @@ nochange:
 				goto nochange; // fallthrough
 		case SEL_BACK:
 			/* Handle clicking on a context at the top */
-			if (sel == SEL_CLICK && event.bstate == BUTTON1_CLICKED && event.y == 0) {
+			if (sel == SEL_CLICK && event.bstate == BUTTON1_PRESSED && event.y == 0) {
 				/* Get context from: "[1 2 3 4]..." */
 				r = event.x >> 1;
 
@@ -3715,7 +3718,7 @@ nochange:
 #endif
 
 			/* Toggle filter mode on left click on last 2 lines */
-			if (event.y >= xlines - 2) {
+			if (event.y >= xlines - 2 && event.bstate == BUTTON1_PRESSED) {
 				cfg.filtermode ^= 1;
 				if (cfg.filtermode) {
 					presel = FILTER;
@@ -3731,13 +3734,18 @@ nochange:
 			}
 
 			/* Handle clicking on a file */
-			if (2 <= event.y && event.y <= ndents + 1) {
+			if (2 <= event.y && event.y <= ndents + 1 && event.bstate == BUTTON1_PRESSED) {
 				r = curscroll + (event.y - 2);
 				move_cursor(r, 1);
+				currentmouse ^= 1;
+				clock_gettime(CLOCK_MONOTONIC_RAW, &mousetimings[currentmouse]);
 
 				/*Single click just selects, double click also opens */
-				if (event.bstate != BUTTON1_DOUBLE_CLICKED)
+				if (((_ABSSUB(mousetimings[0].tv_sec, mousetimings[1].tv_sec) << 30)
+				  + (mousetimings[0].tv_nsec - mousetimings[1].tv_nsec))
+					> 400000000)
 					break;
+				mousetimings[currentmouse].tv_sec = 0;
 			} else {
 				if (cfg.filtermode)
 					presel = FILTER;


jarun:
Awesome! I'll test this on Android and confirm. :+1:

jarun:
> It might change the behavior of nav-as-you-type.

What's the change in behaviour?

0xACE:
> What's the change in behaviour?

In master, clicking on bottom 2 lines, starts `nav-as-you-type`. and then clicking on another line exits `nav-as-you-type` until you go into another directory...

In my patch if i don't turn of `nav-as-you-type` the filter bar will flicker back into filter mode each time you click an item. I can look through more closely, but I got a toothbrush in my mouth and I'm about to flirt with the pillow behind me :-D

(If you have an idea of what's going on, it'd be nice to hear. but dw about it, the important part of the patch is atleast done.)

0xACE:
Tried it on my android quickly, it doesn't seem to be working as intended. I'll look at this another time, maybe it was too premature for me to release  it without testing it properly...

jarun:
> I'll look at this another time

No problem! These kind of device-dependent features can't be quick, I understand. Shall we close the PR for now?

jarun:
There are 2 warnings on Wikipedia page:

- The topic of this article may not meet Wikipedia's general notability guideline.
- This article relies too much on references to primary sources.

Can you please take a look?

0xACE:
> No problem! These kind of device-dependent features can't be quick, I understand. Shall we close the PR for now?

I'll try today. If I can't fix it today, it's okay to close it.



> There are 2 warnings on Wikipedia page:
> 
> * The topic of this article may not meet Wikipedia's general notability guideline.
> * This article relies too much on references to primary sources.
> 
> Can you please take a look?

I addressed both those problems as soon as they appeared. The problem now is that I am not the right person to make a judgement on whether it has cleared the "minimal requirements".

What it's mostly missing is external sources, talking about the topic... I would argue the project is "generally notable" as the buzz word of `nnn` is as big as other CLI tools with the same utility... So really just more news articles would be needed. But those warnings were generated before I added those blogposts/newsarticles ...

jarun:
> I addressed both those problems as soon as they appeared. The problem now is that I am not the right person to make a judgement on whether it has cleared the "minimal requirements".

> What it's mostly missing is external sources, talking about the topic... I would argue the project is "generally notable" as the buzz word of nnn is as big as other CLI tools with the same utility... So really just more news articles would be needed. But those warnings were generated before I added those blogposts/newsarticles ...

Got it! In the project page we have several links to articles which discuss `nnn`.

0xACE:
I looked closer into this android problem.

My best guess is that android/touch devices will not reveal the `BUTTON1_PRESSED` state and instead directly goes into `BUTTON1_CLICKED`. What we can do is have a special `#ifdefine TOUCH_BASED_DEVICE` which translates `if (event.bstate == BUTTON1_CLICKED) event.bstate = BUTTON1_PRESSED;`

I have looked at the source code of other CLI tools and they:
1. Do not support touch devices as they only cover `BUTTON1_PRESSED`
2. Have their own implementation to interpret mouse buttons (basically not even using `ncurses`)

On a normal computer with `mouseinterval(0)` afaik: `BUTTON1_CLICKED` shouldn't happen unless you have a debouncing problem with your mouse, so it's safe to stick with `BUTTON1_PRESSED`. We could optionally add a check for `BUTTON1_CLICK` too, and it should virtually not have problems in normal use, but still the problem is as mentioned in the other paragraph: Touch devices hides `BUTTON1_PRESSED`.

Of course additional logic can be used to adress this but my recommendation is the `#ifdefine` approach as it would work fine on practically any machine. It would even work on a regular linux machine, so the `#ifdefine` is just there incase someone has a debouncing issue with the mouse (which is really something software cannot support tbh).


Are you okay with me overwriting `BUTTON1_CLICKED` into `BUTTON1_PRESSED`? Because that is really the issue at hand.

jarun:
To be frank, I am unable to understand the behaviour of any of this click, press stuff without actually trying it. Please update the patch to best possible experience and allow me a few days to test it hands-on.

0xACE:
I pushed a small change for touch device support.

jarun:
@0xACE kindly remove the WIP label when you want me to test it.

0xACE:
> @0xACE kindly remove the WIP label when you want me to test it.

Well the patch is working as is. There are only 2 remaining problems with it:

1. (**TL;DR:** consider this to be solved, as it is not a realistic problem) `BUTTON1_CLICKED` is being immediately translated into `BUTTON1_PRESSED`. Personally for this tool I think it is acceptable to do this. It is unlikely this will cause problems for anyone. One scenario where I can see this being a problem is if your mouse button is broken: allowing you to press the mouse button twice within `0ms`. I did some quick measures and i guess roughly `166ms` is the average double click speed for me, and microsoft seems to allow `400-500ms` for double clicking by default. so this point should pass if you ask me.
2. This patch exits out of `nav-as-you-type` if you click on any file. Previously it would return to `nav-as-you-type` as soon as you moved into another folder. I'm not sure this is a problem, but it can be fixed. I'm just not sure where the problem resides. if you want me to look further I will resolve this too.

TL;DR: Is it important for me to retain the original behavior of returning to `nav-as-you-type` when you click into another folder while previously having `nav-as-you-type` activated? Because that's the only difference between current master and this patch.

jarun:
> if you want me to look further I will resolve this too.

Please do. Take your time. I use _nav-as-you-type_ all the time.

0xACE:
Problem solved. The patch is working as intended.

All it does is make the mouse more responsive in `nnn` from `400ms` down to `0ms`.

src/nnn.c
```diff
@@ -3705,13 +3707,18 @@ static void browse(char *ipath)
...
-			if (2 <= event.y && event.y <= ndents + 1) {
+			if (2 <= event.y && event.y <= ndents + 1 && event.bstate == BUTTON1_PRESSED) {
 				r = curscroll + (event.y - 2);
 				move_cursor(r, 1);
+				currentmouse ^= 1;
+				clock_gettime(CLOCK_MONOTONIC_RAW, &mousetimings[currentmouse]);
 
 				/*Single click just selects, double click also opens */
-				if (event.bstate != BUTTON1_DOUBLE_CLICKED)
+				if (labs( (mousetimings[0].tv_sec * 1000000000 + mousetimings[0].tv_nsec)
```
jarun: Can we do the math with a power of 2 constant (maybe a close approximation of the number) so we can use shifts?

jarun: can we try the following?

- set `DOUBLECLICK_TIMING` as `1000000 * DOUBLECLICK_TIMING` and remove the division
- have a macro for #define instead of invoking math lib: `#define ABS(N) ((N<0)?(-N):(N))`
- use a close approximate power of 2 instead of `1000000000` to facilitate shift

src/nnn.h
```diff
@@ -32,6 +32,8 @@
 
 #include <curses.h>
 
+/* Double click timing in milliseconds */
+#define DOUBLECLICK_TIMING 400
```
jarun: We can have this in `nnn.c` instead of `nnn.h`.

src/nnn.c
jarun: I think shift by 30 should be good enough:

    2^30 = 1073741824

Also, (a + b) - (c + d) = (a - c) + (b - d). So we can simplify:

    labs( (mousetimings[0].tv_sec * 1000000000 + mousetimings[0].tv_nsec) - (mousetimings[1].tv_sec * 1000000000 + mousetimings[1].tv_nsec)) / 1000000 > DOUBLECLICK_TIMING;

to

```
#define _ABSSUB(N, M) ((N <= M) ? (M - N) : (N - M))
#define DOUBLECLICK_TIMING_NS 400000000

(_ABSSUB(mousetimings[0].tv_sec, mousetimings[1].tv_sec)) << 30 + (mousetimings[0].tv_nsec - mousetimings[1].tv_nsec)) > DOUBLECLICK_TIMING_NS;
```

0xACE:
I applied the changes you suggested. Everything should work as expected.

The CI error isn't related to this patch, it stems from `master`

jarun:
I have tested it on 2 devices. Worked fine so far.

jarun:
Thank you!

jungle-boogie:
Something in this PR broke building `nnn` on OpenBSD.
See https://github.com/jarun/nnn/issues/361

0xACE:
Hurray! I guess either:

1. clock_gettime needs to be changed (`CLOCK_MONOTONIC_RAW`)...
2. the mousemask is broken (which ncurses version?)
3. mouseinterval needs to be 1 (wild guess if the others are not the problem)

jarun:
A little late to the party, but welcome @0xACE! ;)

PR #354, opened 2019-10-12
Check for alternative name for drag/drop exec

0xACE:
KDE dragon player collides with the name of the drag/drop dependency. On
archlinux the binary is renamed to `dragon-drag-and-drop`.

This change tries `dragon-drag-and-drop` before defaulting to `dragon`.

### DIFF
--- a/plugins/drag-file
+++ b/plugins/drag-file
@@ -12,7 +12,11 @@ all=
 
 dnd()
 {
-    dragon "$@"
+    if which dragon-drag-and-drop; then
+	dragon-drag-and-drop "$@"
+    else
+	dragon "$@"
+    fi
 }
 
 function use_all()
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

0xACE:
TL;DR[1]:
**0xACE commented on 2019-10-11 22:51 wrote:**
>Any reason as to why `dragon` is being renamed to `dragon-drag-and-drop` ?

**Rhinoceros commented on 2019-10-11 22:57 wrote:**
>@0xACE Yes, to avoid conflict with `dragon` from the `Extra` repository.
>
>https://www.archlinux.org/packages/extra/x86_64/dragon/

[1] in the AUR comment section: https://aur.archlinux.org/packages/dragon-drag-and-drop/

PR #356, opened 2019-10-13
Added copy/move-as (copy/move and rename)

KlzXS:
Benefits and potential use cases:

- When copying/moving files and renaming them saves a lot of keypresses (selecting the copied/moved files in new directory, hitting `r`, and confirming rename selection)
- Can copy/move files with same name from different directories into a single directory all at once, maybe you want to add a suffix or prefix to each

Feel free to suggest better keybinds and names for things.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1284,10 +1284,21 @@ static bool cpmv_rename(const char *path, const char *cmd)
 	if ((fd = create_tmp_file()) == -1)
 		return ret;
 
-	if (!selbufpos)
-		goto finish;
+	/* selsafe() returned TRUE for this to be called */
+	if (!selbufpos) {
+		snprintf(buf, sizeof(buf), "cat %s | tr '\\0' '\\n' > %s", g_selpath, g_tmpfpath);
+		spawn("sh", "-c", buf, NULL, F_NORMAL | F_CMD);
 
-	seltofile(fd, &count);
+		while ((i = read(fd, buf, sizeof(buf))) > 0)
+			while(i)
+				count += (buf[--i] == '\n');
+
+		if (!count)
+			goto finish;
+	} else {
+		seltofile(fd, &count);
+	}
+
 	close(fd);
 
 	snprintf(buf, sizeof(buf), formatcmd, g_tmpfpath);
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -383,7 +383,7 @@ static char mv[] = "mvg -gi";
 #define STR_TMPFILE 3
 #define NONE_SELECTED 4
 #define UTIL_MISSING 5
-#define MOUNT_FAILED 6
+#define OPERATION_FAILED 6
 
 static const char * const messages[] = {
 	"no traversal",
@@ -392,7 +392,7 @@ static const char * const messages[] = {
 	"/.nnnXXXXXX",
 	"0 selected",
 	"missing dep",
-	"mount failed",
+	"failed!",
 };
 
 /* Supported configuration environment variables */
@@ -1271,7 +1271,7 @@ static void xrm(char *path)
 	}
 }
 
-static bool mvcp_rename(const char *path, const char *cmd)
+static bool cpmv_rename(const char *path, const char *cmd)
 {
 	int fd, i;
 	uint count = 0, lines = 0;
@@ -1308,7 +1308,6 @@ static bool mvcp_rename(const char *path, const char *cmd)
 	DPRINTF_U(count);
 	DPRINTF_U(lines);
 
-	/* What happens to this and batch_rename() if given filenames with '\n' in them? */
 	if (2 * count != lines) {
 		DPRINTF_S("cannot delete files");
 		goto finish;
@@ -2886,7 +2885,7 @@ static bool archive_mount(char *name, char *path, char *newpath, int *presel)
 	DPRINTF_S(name);
 	DPRINTF_S(newpath);
 	if (spawn(cmd, name, newpath, path, F_NORMAL)) {
-		printwait(messages[MOUNT_FAILED], presel);
+		printwait(messages[OPERATION_FAILED], presel);
 		return FALSE;
 	}
 
@@ -2928,7 +2927,7 @@ static bool sshfs_mount(char *newpath, int *presel)
 
 	/* Connect to remote */
 	if (spawn(env, tmp, newpath, NULL, flag)) {
-		printwait(messages[MOUNT_FAILED], presel);
+		printwait(messages[OPERATION_FAILED], presel);
 		return FALSE;
 	}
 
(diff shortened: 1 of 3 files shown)


src/nnn.c
```diff
@@ -4306,15 +4362,28 @@ static void browse(char *ipath)
 			case SEL_MV:
 				mvstr(g_buf);
 				break;
+			case SEL_CPAS:
+				if (!mvcp_rename(path, cp)) {
```
jarun: Please rename the function to `cpmv_rename`

src/nnn.c
```diff
@@ -4306,15 +4362,28 @@ static void browse(char *ipath)
...
+					printwait("copy and rename failed!", &presel);
```
jarun: How about adding "failed!" to struct `messages` and using it? User already knows the operation he tried.

src/nnn.h
```diff
@@ -217,6 +219,10 @@ static struct key bindings[] = {
 	{ 'P',            SEL_CP },
 	/* Move from selection buffer */
 	{ 'V',            SEL_MV },
+	/* Copyfrom selection buffer and rename */
+	{ 'w',            SEL_CPAS },
```
jarun: Please add to the new keys to help screen and README.

src/nnn.c
```diff
@@ -1271,6 +1271,60 @@ static void xrm(char *path)
...
+	bool ret = FALSE;
+	const char formatcmd[] = "sed -i 's|^\\(\\(.*/\\)\\(.*\\)$\\)|#\\1\\n\\3|' %s";
+	const char renamecmd[] =
+		"sed 's|^\\([^#][^/]\\?.*\\)$|%s/\\1|;s|^#\\(/.*\\)$|\\1|' %s | tr '\\n' '\\0' | xargs -0 -o -n2 %s";
+	char buf[sizeof(renamecmd) + sizeof(cmd) + (PATH_MAX << 1)];
+
+	if ((fd = create_tmp_file()) == -1)
+		return ret;
+
+	if (!selbufpos)
```
jarun: Is this for selection only?

src/nnn.c
KlzXS: Will do! Should I also replace other occurrences of such messages in the file? There are few of them.

src/nnn.c
KlzXS: No. I just used `batch_rename()` as a template since it is really similar to what I wanted to do. Will fix later today.

jarun: Please see if it's possible to refactor and have a single API (with some function

src/nnn.c
jarun: That would be great!

src/nnn.c
```diff
@@ -1271,6 +1271,60 @@ static void xrm(char *path)
...
+		while (i)
+			lines += (buf[--i] == '\n');
+
+	if (i < 0)
+		goto finish;
+
+	DPRINTF_U(count);
+	DPRINTF_U(lines);
+
+	/* What happens to this and batch_rename() if given filenames with '\n' in them? */
```
jarun: None of these are destructive so I wouldn't be much worried. Prpblem starts when there is a chance of the file to get deleted.

src/nnn.c
KlzXS: Do you want me to also replace `MOUNT_FAILED` with a generic `OPERATION_FAILED`?

jarun: Yep! Anything failed, just `failed!` ;)

src/nnn.c
KlzXS: It would just format it wrong and treat the two halves of the file as two separate files and it would just not copy it, it's just a weird nop. So I guess you're right.

src/nnn.c
jarun: Do you think it's possible to refactor?

jarun: On second thoughts, let's keep these separate units.

KlzXS: I think it can be done, if we add a parameter or two and add a few more `if`s here and there I think that cp, mv, rename (fancy mv) and rm can all be invoked by a single function. I'm not going to be able to do it for some time and I'm not even sure how worth the effort it is, but it can be done.

Also why are there 2 files in `batch_rename`, `foriginal` seem to only be there for the original number of lines, but the same selection is written to the other file. We could easily remove that one.

jarun: > I'm not even sure how worth the effort it is

It's not I guess.

> Also why are there 2 files in batch_rename, foriginal seem to only be there for the original number of lines, but the same selection is written to the other file. We could easily remove that one.

If you can simplify, please do that as well.

Maybe post that you may have more insight into having a single function. But again, it's OK either way.

KlzXS: No I'm just blind, you need both files for `batch_rename()` as is.

I'm currently not feeling like rewriting everything into one function so I'll leave it for some other time maybe.

jarun: No problem!

jarun: @KlzXS 

> I just used batch_rename() as a template since it is really similar to what I wanted to do. Will fix later today.

Is this done?

KlzXS: Yes it is. Give it a quick look as I didn't see a nicer way to copy over the contensts of the selection file.

jarun: okies!

src/nnn.c
```diff
@@ -1271,6 +1271,70 @@ static void xrm(char *path)
...
+
+	if ((fd = create_tmp_file()) == -1)
+		return ret;
+
+	/* selsafe() returned TRUE for this to be called */
+	if (!selbufpos) {
+		snprintf(buf, sizeof(buf), "cat %s | tr '\\0' '\\n' > %s", g_selpath, g_tmpfpath);
+		spawn("sh", "-c", buf, NULL, F_NORMAL | F_CMD);
+
+		while ((i = read(fd, buf, sizeof(buf))) > 0)
```
jarun: Can we have a func like linesinfile()? We have this double loop in 3 places.

jarun:
CI is failing. Please check.

KlzXS:
It's complaining about using the globals cp and mv. No idea why.

jarun:
Compiles fine locally for me!

KlzXS:
So does for me. This isn't the first time I've seen it fail for no good reason.

src/nnn.c
KlzXS: The function would have to return `int` as read can fail so that we can catch that.

jarun: > That might be better.

👍

annagrram:
Note that it fails on Mac.
If you check the code, you see the variables are guarded with an #ifdef __linux__, but your usage of them is not.
So, they do not exist to mac compilers, and thus the compilers get angry for a good reason :)

jarun:
Good catch @annagrram!

src/nnn.c
jarun: @KlzXS in that case, please take care of the options used on Mac as well.

jarun:
@KlzXS in that case, please take care of the options used on Mac as well.

jarun:
I'll make the remaining changes.

jarun:
Thank you for this. I like the new workflow!

PR #357, opened 2019-10-13
Remove -Wno-unused-parameter and do cleanup

annagrram:
Not truly necessary, but feels cleaner.
Disabling warnings is kind of bad practice.
If a parameter is genuinely not used with a reason for it to be there, be explicit about it locally.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -200,6 +200,11 @@ typedef struct {
 	char *val;
 } kv;
 
+typedef struct {
+	const regex_t *regex;
+	const char *str;
+} fltrexp_t;
+
 /*
  * Settings
  * NOTE: update default values if changing order
@@ -477,6 +482,8 @@ static bool getutil(char *util);
 
 static void sigint_handler(int sig)
 {
+	(void) sig;
+
 	interrupted = TRUE;
 }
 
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -3057,6 +3068,9 @@ static int sum_bsizes(const char *fpath, const struct stat *sb, int typeflag, st
 
 static int sum_sizes(const char *fpath, const struct stat *sb, int typeflag, struct FTW *ftwbuf)
 {
+	(void) fpath;
```
jarun: How does the compiler treat these? NOOP?

annagrram: It's a standard way (without using compiler-specific attributes) to tell the compiler that a parameter is not used.
But, if you look at the assembly, it does add a single MOV instruction for each "(void) var" line. I don't really know why.

annagrram: Apparently those MOVs are there just because of the parameters themselves (even without the void lines), so maybe it's a parameter-passing optimization of the compiler.

jarun:

annagrram:
Thanks for accepting my OCD :P

0xACE:
Hmm, I don't recall seeing warnings about this. But nvm, `nnn` functions tend to be small anyhow.

3d71uj

annagrram:
@0xACE You haven't seen warnings about this because it was compiled with the warning suppress flag.
Anyway, you can't fix everything in one day (or the browse function in a 100), but if there's something that can be done with a better practice quite easily, why not?

PR #358, opened 2019-10-13
unmount() ask for target if not found

0xACE:
unmount() now properly asks for which target to unmount if it cannot
automatically deduce which folder to unmount.

Btw, the automatic deduction of detecting unmount folder doesn't work nice in:
1. if you are hovering over a `*.zip` which you have previously mounted, you have to go to the mountfolder and then unmount it, rather than `u` over the zip file.
2. if you happen to press `u` while hover over a file which shares the same name as a mountpoint, it will automatically `unmount()` the mountpoint which collides

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2904,6 +2904,8 @@ static bool unmount(char *name, char *newpath, int *presel)
 	static bool found = FALSE;
 	char *tmp = name;
 	struct stat sb, psb;
+	bool child = false;
+	bool parent = false;
 
 	/* On Ubuntu it's fusermount */
 	if (!found && !getutil(cmd)) {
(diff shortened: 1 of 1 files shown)


jarun:
> if you are hovering over a *.zip which you have previously mounted, you have to go to the mountfolder and then unmount it, rather than u over the zip file.

We take you to the mountpoint when we mount. So just visit parent and unmount. Why would you go back to zip file? Also, this is documented.

> if you happen to press u while hover over a file which shares the same name as a mountpoint, it will automatically unmount() the mountpoint which collides

How come? It checks if the hovered file is a mountpoint and only then will it unmount.

jarun:
Please fix the CircleCI failure (garbage value issue)

0xACE:
> We take you to the mountpoint when we mount. So just visit parent and unmount. Why would you go back to zip file? Also, this is documented.

Fair enough.

> How come? It checks if the hovered file is a mountpoint and only then will it unmount.

1. `touch ~/myexample`
2. `c myexample` (to mount myexample)
3. goto `~`
4. hover ontop of `myexample` file you created
5. press `u` and it will automatically unmount `~/.config/nnn/myexample` because `umount()` only looks at file name and automatically prepends `cfgdir`.


I looked at the CI failures, they are really left overs from the previous code, ill take a look at now

0xACE:
Fixed the problem. The code could be made cleaner, but I didn't want to change what you already had, so I just imprinted my fix on top of the existing functionality...

jarun:
> press u and it will automatically unmount ~/.config/nnn/myexample because umount() only looks at file name and automatically prepends cfgdir.

Does your patch fix this? I am not really digging deep as I am working on `bcal` right now.

0xACE:
> Does your patch fix this? I am not really digging deep as I am working on `bcal` right now.

No, I didn't want to impede on your setup. But it can easily be fixed if you want...

jarun:
> But it can easily be fixed if you want...

Yes, please!

0xACE:
It should be good to go. the CI shouldn't fail on this...

jarun:
Thank you so much!

PR #360, opened 2019-10-15
Implementing sessions support

annagrram:
Opening a pull request for you to test the feature as it progresses and give your input on what's already implemented (if it's not implemented, it might be before "work in progress" label is lifted).
Moreover, some more cleanups are due too.

Thoughts to self:
There are all kinds of weird magic numbers and string shenanigans that need to be cleaned somewhen.


src/nnn.h
```diff
@@ -272,4 +274,6 @@ static struct key bindings[] = {
 	{ 'Q',            SEL_QUIT },
 	{ CONTROL('Q'),   SEL_QUIT },
 	{ KEY_MOUSE,      SEL_CLICK },
+	{ CONTROL('W'),   SEL_SAVE_SESSION },
+	{ CONTROL('S'),   SEL_LOAD_SESSION },
```
jarun: This would have issues in different platforms as it's stop in stty.

src/nnn.c
```diff
@@ -4817,6 +4841,10 @@ static void browse(char *ipath)
 				}
 			}
 			return;
+        case SEL_SAVE_SESSION:
+        case SEL_LOAD_SESSION:
```
jarun: Instead of load on demand, let's have a program switch to load session. The reason is - say the user has 2 contexts open and he has 4 contexts saved. It wouldn't be a delightful experience for him.

src/nnn.c
```diff
@@ -2628,6 +2629,29 @@ static void savecurctx(settings *curcfg, char *path, char *curname, int r /* nex
 	*curcfg = cfg;
 }
 
+static void session_manager(int op)
+{
+    char *session_name = xreadline("", "Session name: ");
```
jarun: `session name` (small s)

src/nnn.c
```diff
@@ -2628,6 +2629,29 @@ static void savecurctx(settings *curcfg, char *path, char *curname, int r /* nex
...
+        printmsg("Failed to save session");
+        return;
+    }
+
+    if (op == SEL_SAVE_SESSION) {
+        fwrite(&cfg, sizeof(cfg), 1, fsession);
+        fwrite(g_ctx, sizeof(context), CTX_MAX, fsession);
+    } else { // SEL_LOAD_SESSION
+        fread(&cfg, sizeof(cfg), 1, fsession);
+        fread(g_ctx, sizeof(context), CTX_MAX, fsession);
```
jarun: what happens if a path is removed when session is loaded? Is it handed gracefully already?

src/nnn.h
annagrram: I forgot about that thing.
Do you have any suggestions? It's tough finding somewhat meaningful keybindings.

jarun: See my other comments, We may not need load on demand.

src/nnn.c
annagrram: I planned on adding a program switch for loading a session, but I still think it might be useful to be able to switch between sessions on-the-fly.
What do you think?

src/nnn.c
annagrram: No problem.
The sessions are just serialization and deserialization of the `cfg` and `g_ctx` variables.
Once a session is loaded (the state is restored), there's no need for that file any longer.

src/nnn.c
jarun: I am just worried about what happens to the existing contexts. Having dealt wituh users for years, the more power you give them, the more demanding they become. As soon as you tell them you can load and store up to 4 contexts, they will want to have the current contexts intact, and then load 4 more. I would love to block that by workflow than to have to tell we are limited by the number of contexts.

src/nnn.c
jarun: I was speaking of the directory path in a stored context. What happens if the path is missing when the session is loaded.

Regarding the session file, how many can we have? How does the user delete a session file?

src/nnn.c
annagrram: Well, they can have them intact; they only need to save them as a session.
We can also make a "last session" session that will automagically be created when a new session is loaded.
But, I get what you mean.
If you prefer to keep that just as a program switch, I might keep myself a patch around for dynamic loading for me (because I still find it useful ^_^).

src/nnn.c
annagrram: Haven't handled that problem yet.
It's just a file in a documented location, so they can delete it from there.
Later on (whether in the first or second iteration), we'll add a more refined session management control into the program itself (or just a jump to the sessions directory and use nnn to manage the files?).

src/nnn.c
jarun: Please allow me some time to think on this.

annagrram:
Take your time. 

It will take a while before it is ready for a merge. 

________________________________

src/nnn.c
annagrram: Take your time.
It will be a little while before it is ready for the first merge iteration.

src/nnn.c
annagrram: I tested it, and apparently, you did a great job of handling it already.
It just switches to a parent that exists.
Is it acceptable behaviour, or do you want me to set the invalid context as inactive?

src/nnn.c
jarun: Noting down another case: 2 contexts active and 2 contexts in a saved session.

jarun: Automatic session saving may not help:

- load one session, current session gets saved auto (1)
- load another session, session saved in (1) gets overwritten

src/nnn.c
jarun: > Is it acceptable behaviour

Of course!

src/nnn.c
annagrram: I didn't say it will save the entire session-swapping history — just the last one. If you accidentally loaded a session onto it, so you could restore — a kind of temporary session.

annagrram:
@jarun Do we want to add selections as part of the session?

src/nnn.c
jarun: OK. We'll have to document this. We need a section on sessions in the contexts page anyway. We will have both runtime loading and program option in that case.

jarun:
> Do we want to add selections as part of the session?

Not at all.

src/nnn.c
```diff
@@ -2633,6 +2645,97 @@ static void savecurctx(settings *curcfg, char *path, char *curname, int r /* nex
...
+
+	mkpath(sessiondir, session_name, session_path);
+
+	FILE *fsession = fopen(session_path, "wb");
+	if (!fsession) {
+		printwait("failed to open session file", presel);
+		return;
+	}
+
+	if ((fwrite(&header, sizeof(header), 1, fsession) != 1)
```
jarun: Instead of this, why not write the data in binary mode to the session file?

annagrram: It is written in binary mode.

annagrram:
@jarun, Can we ask people to help to test this?

A lot is going on. But I think it's about ready for a testing phase.

I've found many weird cases along the way that I had to handle and might have missed some.

Moreover, extra input on the workflow, use cases and missing essential features will be valuable and much appreciated.

nnn.1
```diff
@@ -10,6 +10,7 @@
 .Op Ar -b key
 .Op Ar -c
 .Op Ar -d
+.Op Ar -e session
```
jarun: Please add the option in program cmdline help:

    -s name load session by name

src/nnn.h
```diff
@@ -272,4 +275,7 @@ static struct key bindings[] = {
 	{ 'Q',            SEL_QUIT },
 	{ CONTROL('Q'),   SEL_QUIT },
 	{ KEY_MOUSE,      SEL_CLICK },
+	{ CONTROL('W'),   SEL_SAVE_SESSION },
```
jarun: I think we don't need the control chars here.

Also how about having a single key for session management? say `v` and then you can ask for a keypress to 's'ave/'l'oad/'r'estore and continue based on user's choice. See the `n` (new) keybind workflow.

jarun: We need to update keybind help.

Speaking of the single-key thing, now I think we could also do the same for copy/move as... our primary is copy/move/del. The as part could have an additional keypress. I mean, I am going to press`?` and lookup anyway. ;) @KlzXS what do you think?

jarun:
I still need to test this.

src/nnn.h
annagrram: Interesting idea.
I'll make that change.

nnn.1
annagrram: Do you mean the auto-completion files?

jarun: That as well plus the `show_help()` function.

jarun:
> Can we ask people to help to test this?

Of course! I'm going to test it myself.

@0xACE and @KlzXS can you guys help with this once the single keybind session management is in?

KlzXS:
I won't be able to help or contribute in any way until the 25th.

jarun:
@KlzXS thanks for letting us know. We have enough time to next release so you can always test post merge.

0xACE:
I'm busy until the end of the month, and possibly further...

I tried following the development lately but honestly I haven't understood the changes, I'll have to catch up first...

src/nnn.c
```diff
@@ -3080,6 +3212,7 @@ static void show_help(const char *path)
 		"1MISC\n"
 	       "9! ^]  Shell  =  Launch  C  Execute entry\n"
 	       "9R ^V  Pick plugin   :K xK  Execute plugin K\n"
+		   "cU  Manage session\n"
```
jarun: Pull the "=  Launch below" so we have 2 per line.

src/nnn.c
```diff
@@ -2633,6 +2644,127 @@ static void savecurctx(settings *curcfg, char *path, char *curname, int r /* nex
...
+
+	session_name = !last_session ? xreadline("", "session name: ") : "@";
+	if (session_name[0] != '\0')
+		mkpath(sessiondir, session_name, session_path);
+	else
+		return;
+
+	FILE *fsession = fopen(session_path, "wb");
+	if (!fsession) {
+		printwait("failed to open session file", presel);
```
jarun: return the status and if it's a failure `goto nochange`. We flash the failure msg for a few seconds and then clear.

To test, start in _nav-as-you-type_ mode, press `U`, then `l`, then input some invalid session name. This should be the behaviour (goto nochange) for any error we show from the cases in `browse()`.

src/nnn.c
```diff
@@ -2633,6 +2644,127 @@ static void savecurctx(settings *curcfg, char *path, char *curname, int r /* nex
 	*curcfg = cfg;
 }
 
+static void save_session(const char *mark, int *presel, bool last_session)
```
jarun: Do we save/need to save the config as well?

src/nnn.c
jarun: You have to add this to the readme as well.

src/nnn.c
```diff
@@ -254,6 +255,15 @@ typedef struct {
 	uint color; /* Color code for directories */
 } context;
 
+typedef struct {
+	size_t version;
```
jarun: Please use shorter names:

version -> ver
path_length -> pathlen

etc.

jarun: pathln even better.

src/nnn.c
```diff
@@ -4929,6 +5082,7 @@ static void usage(void)
 		" -b key  open bookmark key\n"
 		" -c      cli-only opener\n"
 		" -d      detail mode\n"
+		" -e      load session by name\n"
```
jarun: -e name
and in other places as well to be consistent. name instead of session because of length.

src/nnn.c
annagrram: We are saving the config, contexts and pinned directory.

jarun: Dropped the pinned dir. It was always meant to be a runtime thing.

src/nnn.c
annagrram: I don't understand what you mean.
There is no need to return status from `save_session` because it goes to `nochange` anyhow.
`load_session` does return state and goes to `nochange` on failure, but it doesn't use `printwait` because when loading with the program flag, it loads the session because getting into the `browser`'s main loop.

I tested it on my machine and it works fine. However, I didn't use _nav_as_you_type_ because I don't know how do I press `U` without it navigating to U.

Can you please elaborate further?

src/nnn.c
annagrram: Ok.
I'll remove it and update the man page

src/nnn.c
annagrram: > Pull the "= Launch below" so we have 2 per line.

What do you mean exactly? We have an odd number of entries.

src/nnn.c
jarun: Just add a `presel = MSGWAIT;` if `load_session()` fails in case `SEL_SESSIONS:`.

src/nnn.c
```diff
@@ -2633,6 +2644,127 @@ static void savecurctx(settings *curcfg, char *path, char *curname, int r /* nex
...
+	status = _SUCCESS;
+
+END:
+	fclose(fsession);
+
+	if (status == _FAILURE)
+		printwait("failed to write session data", presel);
+}
+
+static int load_session(const char *session_name, char *mark
```
jarun: I think return type bool will suffice.

src/nnn.c
jarun: Add a bool return type and show a message if save fails.

src/nnn.c
jarun:
```
         ! ^]  Shell             C  Execute entry
         R ^V  Pick plugin   :K xK  Execute plugin K
           cU  Manage session    =  Launch
            c  SSHFS mount       u  Unmount
           ^P  Prompt/run cmd    L  Lock
```

src/nnn.c
annagrram: I'll make all the changes later today, as I don't have much time now.

jarun: No hurry!

jarun:
Thank you for session support! :+1:

annagrram:
It was fun working on it :)

jarun:
> It was fun working on it :)

Very good work!

annagrram:
Thanks!

And don't forget to bump the header's version to 1 before the release freezes the first file structure of a session file.

jarun:
> And don't forget...

Not happening.

I'll raise an issue so I remember. :)

jarun:
BTW, why not use the program version instead?

annagrram:
> BTW, why not use the program version instead?

Because the session version represents the session file structure.
It is written to the session file and matched when read.
That way if we change it in the future, it won't try to read old ones and make a mess.
If we had used the program version, each version would have rendered all the session files created with previous versions unusable.

jarun:
> If we had used the program version, each version would have rendered all the session files created with previous versions unusable.

Right! Didn't think of it.

0xACE:
Hey, sorry for not taking part in this discussion, but I've been busy...

I just rebased my local branch to `origin/master` now and i noticed when I use the new `SEL_SESSIONS` it will involve a folder `~/.config/nnn/sessions`.

1. My question is, how will it deal with a `sshfs` mount called "sessions" or a archive mount called "sessions"?

I previously encountered this issue when `archivemount` was introduced, and on my local branch i prepend "zz-" before any archive mountpoints to not collide with my `sshfs` mounts, and personally make sure none of my `sshfs` mounts start with "zz-". Though that was only a quick hack to not further complicate the codebase...

2. What about a "unnamed" default session for quick save/restore? (to allow a user to just hit `Us<ENTER>` or `Ul<ENTER>`.)

annagrram:
>  it will involve a folder ~/.config/nnn/sessions

It involves a folder in the nnn config directory like everything else in nnn. I think that's kind of legitimate.

> What about a "unnamed" default session for quick save/restore? (to allow a user to just hit Us<ENTER> or Ul<ENTER>

Well, you have a Ur<ENTER> for restore from an "unnamed" session. To save to it you have to use its "name" @. But we can change the entering no name for save automatically saves to that session.

0xACE:
> It involves a folder in the nnn config directory like everything else in nnn. I think that's kind of legitimate.

Yes, I understand what you mean. But imho it's a design flaw with `nnn`. imagine if mount points were located in `~/.config/nnn/mnt/` then we wouldn't have this potential collision. I guess this problem isn't relevant to this PR, but I figured it would be worth to bring it up... In my personal case it would be very likely to have name collissions with archives and some of my `sshfs` mounts, hence i prepend `zz-` on all archive mount points (it's really a cheap hack)...
 
> Well, you have a Ur for restore from an "unnamed" session. To save to it you have to use its "name" @. But we can change the entering no name for save automatically saves to that session.

Good idea, I could use `@` for now, but really I feel like it should default to either `@` or a `default` (because `@` gets implicitly overwritten by other actions) when no name is provided. I will be happy with this even if nothing gets changed...

PR #362, opened 2019-10-15
Fix strftime() buffer overflow

sjmulder:
If the formatted string is longer than 17 characters the string won't
be null terminated, causing gibberish to be printed on the status line
and possibly worse.

This occurs with my current locale in which a date is printed as such:

`2019-okt.-15 19:26`

Which is 18 characters.


jarun:
Thank you!

PR #364, opened 2019-10-19
Implement plugins control of nnn + plugins

annagrram:
As we discussed, a way to control `nnn` from plugins + a helper script + example plugins.


plugins/.nnn-plugin-helper
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env sh
+
+# Description: Helper script for plugins
```
jarun: Please add some more details.

jarun: As this is a dot file, please double check that the download and update works fine.

src/nnn.c
```diff
@@ -427,9 +430,10 @@ static const char * const messages[] = {
 #define NNN_CONTEXT_COLORS 2
 #define NNN_IDLE_TIMEOUT 3
 #define NNN_COPIER 4
-#define NNNLVL 5 /* strings end here */
-#define NNN_USE_EDITOR 6 /* flags begin here */
-#define NNN_TRASH 7
+#define NNNLVL 5
+#define NNN_PIPE 6 /* strings end here */
```
jarun: Please ensure the help screen shows correctly.

src/nnn.c
```diff
@@ -4100,6 +4115,23 @@ static void browse(char *ipath, const char *session)
 				if (cfg.runplugin) {
 					if (!run_selected_plugin(path, dents[cur].name, newpath, rundir, runfile, lastname))
 						continue;
+
+					if (g_buf[0] != '\0' && g_buf[1] != '\0') {
```
jarun: Please refactor and move this out of `browse()`.

src/nnn.c
```diff
@@ -5475,6 +5509,15 @@ int main(int argc, char *argv[])
 	/* Get the clipboard copier, if set */
 	copier = getenv(env_cfg[NNN_COPIER]);
 
+	/* Create pipe for plugin control */
```
jarun: Please re-factor. We should not add start-up delay for plugins which may not even be used.

annagrram: I can't know which plugins use this feature, so I initialise it the first time a plugin is called.

src/nnn.c
annagrram: Seems fine for me (`show_help`)

plugins/.nnn-plugin-helper
annagrram: Good catch!
Fixed

annagrram:
Found a small bug. I will fix shortly

annagrram:
> Found a small bug. I will fix shortly

Done

plugins/.nnn-plugin-helper
annagrram: > Please add some more details.

I have a comment explaining the primary function there. There isn't much more there.
I need to think of a better Description overall if the current one isn't enough

annagrram:
I set up an openbsd and freebsd virtual machines to make sure the changes work over there.
Unfortunately, there are bugs (also, the makefile doesn't work. But that's a different story).
That's why I added the work in progress tag

src/nnn.c
jarun: Yes, that's what I wanted. Not in `main()` please.

jarun: Also, please see if you need the global to know initialization done or not. I see you create a pipe. Would it suffice to check if the pipe exists or not? or something in that line...

annagrram: We can check whether the pipe exists, but that would require more tests to be robust (type of file, permissions).
The global is a simple and much faster solution for the cost of a boolean.

If you still prefer not to have that variable, I'll change it.

jarun: No, that's just fine. I was just sharing ideas. You must have noticed by now that I rarely go through line by line in reviews to save time. It's more of a design and functional review. The rest can always be fixed by the dev.

annagrram: Yeah. I noticed.

Sharing ideas is excellent and much appreciated!

I'm explaining my views on the matter, but as it's your project, I can always adjust to the things you prefer :)

Sometimes it's down to personal coding style and preferences, and one needs to adjust for the project at hand.

jarun: > Sometimes it's down to personal coding style and preferences, and one needs to adjust for the project at hand.

Not necessarily. I am open to newer and smarter ways to do things. That's also the reason the utilities I maintain exist. So if you be patient and explain why something is better (or preferred) I will always make an effort to understand.

annagrram: I noticed. That's one of the reasons I stayed beyond the first commit or two. You seem friendly, responsive, and open to new ideas.
I'm always patient and explain my reasoning if people are willing to listen.

jarun:
Please update the plugin README so users know how to use/extend this powerful mechanism.

annagrram:
> Please update the plugin README so users know how to use/extend this powerful mechanism.

I will update the plugins writing section to include all the necessary information for writing plugins (input arguments, accessing selections, and changing directories in nnn).
It should be ready later today (now it's too late to write docs ZZZzzzz).

jarun:
No hurry!

src/nnn.c
```diff
@@ -342,6 +342,11 @@ static char g_buf[CMD_LEN_MAX] __attribute__ ((aligned));
 /* Buffer to store tmp file path to show selection, file stats and help */
 static char g_tmpfpath[TMP_LEN_MAX] __attribute__ ((aligned));
 
+/* Buffer to store plugins control pipe location */
+static char g_pipepath[TMP_LEN_MAX] __attribute__ ((aligned));
+
+static bool g_is_plugin_control_initialized = FALSE;
```
jarun: Use a shorter var name: `g_plinit` maybe.

Later, even if we want to do any other kind of initialization that would be covered.

src/nnn.c
```diff
@@ -3292,24 +3299,67 @@ static void show_help(const char *path)
 	unlink(g_tmpfpath);
 }
 
-static bool run_selected_plugin(char *path, const char *file, char *newpath, char *rundir, char *runfile, char *lastname)
+static bool initialize_plugin_control() {
```
jarun: How about `init_plctrl` or `plctrl_init`?

src/nnn.c
```diff
@@ -3292,24 +3299,67 @@ static void show_help(const char *path)
 	unlink(g_tmpfpath);
 }
 
-static bool run_selected_plugin(char *path, const char *file, char *newpath, char *rundir, char *runfile, char *lastname)
+static bool plctrl_init() {
```
jarun: Bracket starts in next line.

annagrram: Oops.

plugins/README.md
```diff
@@ -11,9 +11,12 @@ The currently available plugins are listed below.
 | --- | --- | --- | --- |
 | boom | sh | [moc](http://moc.daper.net/) | Play random music from dir |
 | dups | sh | find, md5sum,<br>sort uniq xargs | List non-empty duplicate files in current dir |
+| cd | sh | - | Change to arbitrary directory |
+| cd-to-clipboard | sh | xsel | Change to directory in clipboard |
 | checksum | sh | md5sum,<br>sha256sum | Create and verify checksums |
 | drag-file | sh | [dragon](https://github.com/mwh/dragon) | Drag and drop files from nnn |
 | drop-file | sh | [dragon](https://github.com/mwh/dragon) | Drag and drop files into nnn |
+| fzf | sh | fzf | Change to the directory of a file selected by fzf |
```
jarun: Couldn't fzy do this?

jarun: And if we must, name `fzfcd`.

plugins/README.md
```diff
@@ -36,6 +39,7 @@ The currently available plugins are listed below.
 | picker | sh | nnn | Pick files and list one per line (to pipe) |
 | pywal | sh | pywal | Set image as wallpaper, change terminal colorscheme |
 | readit | sh | pdftotext, mpv,<br>pico2wave | Read a PDF or text file aloud |
+| resolve-link-dir | sh | - | Change to the resolve directory of a link |
```
jarun: Please use a shorter name.

plugins/README.md
annagrram: Please check the change.
I hope I've found a compromise.

plugins/README.md
annagrram: Done

plugins/README.md
```diff
@@ -11,9 +11,12 @@ The currently available plugins are listed below.
 | --- | --- | --- | --- |
 | boom | sh | [moc](http://moc.daper.net/) | Play random music from dir |
 | dups | sh | find, md5sum,<br>sort uniq xargs | List non-empty duplicate files in current dir |
+| cd | sh | - | Change to arbitrary directory |
+| cd-to-clipboard | sh | xsel | Change to directory in clipboard |
 | checksum | sh | md5sum,<br>sha256sum | Create and verify checksums |
 | drag-file | sh | [dragon](https://github.com/mwh/dragon) | Drag and drop files from nnn |
 | drop-file | sh | [dragon](https://github.com/mwh/dragon) | Drag and drop files into nnn |
+| fzcd | sh | fzy/fzf (optional fd) | Change to the directory of a file/directory selected by fzy/fzf |
```
jarun: fzy/fzf (optional fd)

this may be long. Add a break (see dups)

plugins/README.md
```diff
@@ -36,6 +39,7 @@ The currently available plugins are listed below.
 | picker | sh | nnn | Pick files and list one per line (to pipe) |
 | pywal | sh | pywal | Set image as wallpaper, change terminal colorscheme |
 | readit | sh | pdftotext, mpv,<br>pico2wave | Read a PDF or text file aloud |
+| lnkcd | sh | - | Change to the resolve directory of a link |
```
jarun: please order alphabetically.

jarun: The plugin name is `lncd`.

jarun:
@0xACE can you please test the changes?

plugins/README.md
annagrram: Note to self: Be more organised during edits  😪

jarun:
Finally, I could get some real look into this.

Tell me, for a utility that intends to replace `cd`, how does it sound to have 4 plugins (cd, cd-to-clipboard (why such a big name with hyphens???), fzcd, lncd) that do `cd`? :)

plugins/README.md
jarun: Looks good to me.

annagrram:
> Tell me, for a utility that intends to replace cd, how does it sound to have 4 plugins (cd, cd-to-clipboard (why such a big name with hyphens???), fzcd, lncd) that do cd? :)

Well, eventually, they change a directory, but each one works differently and adds a different capability.

> I would suggest keep only one which which seems the most probable use-case and have the others as indicative code in the plugin README.

Ok. Which one would you like to keep?
Except the `lncd` that was a demo implementation for the feature @0xACE suggested, which I, personally, won't use much, I think I'll use the rest quite often.

jarun:
> Ok. Which one would you like to keep?

Given the number of lines, retain `fzcd`. Now that you have a section for plugin control, add the other 3 as indicative example code with a brief on top of each indicating what they are for.

I don't want to overwhelm a new user entering the plugins dir with the number of those.

> I like descriptive names and code because I don't like comments.

At least in this case we don't have a problem as will are having the code as a sample.

I also have a question on `getplugs` now that you say you'll have some custom plugins of yourself. Does `getplugs` overwrite any custom non-repo plugins from the user?

annagrram:
> Given the number of lines, retain fzcd. Now that you have a section for plugin control, add the other 3 as indicative example code with a brief on top of each indicating what they are for.

Ok.

> I don't want to overwhelm a new user entering the plugins dir with the number of those.

What about a mechanism that installs just wanted plugins and not everything?

> I also have a question on getplugs now that you say you'll have some custom plugins of yourself. Does getplugs overwrite any custom non-repo plugins from the user?

It will override any customisation made by the user of provided plugins, but will not override/delete any custom plugins written by the user.

jarun:
> What about a mechanism that installs just wanted plugins and not everything?

The current one is simpler so we should retain this for users who don't know which ones they want. We can definitely have another mechanism where they ca cherry-pick.

> It will override any customisation made by the user of provided plugins, but will not override/delete any custom plugins written by the user.

That's how it's written. It will take a backup so the overwrite part doesn't look so dangerous to me.

jarun:
Thank you!

src/nnn.c
```diff
@@ -342,6 +342,11 @@ static char g_buf[CMD_LEN_MAX] __attribute__ ((aligned));
 /* Buffer to store tmp file path to show selection, file stats and help */
 static char g_tmpfpath[TMP_LEN_MAX] __attribute__ ((aligned));
 
+/* Buffer to store plugins control pipe location */
+static char g_pipepath[TMP_LEN_MAX] __attribute__ ((aligned));
```
jarun: I think malloc() -ing it the first time required would have been better.

src/nnn.c
```diff
@@ -3292,24 +3299,68 @@ static void show_help(const char *path)
...
+		return _FAILURE;
+
+	setenv(env_cfg[NNN_PIPE], g_pipepath, TRUE);
+
+	return _SUCCESS;
+}
+
+static bool run_selected_plugin(char **path, const char *file, char *newpath, char *rundir, char *runfile, char **lastname, char **lastdir)
 {
+	if (!g_plinit) {
```
jarun: Do you think adding a config var to init plugin control would be useful? Not all users may be using these plugins.

src/nnn.c
jarun: This is just fine.

src/nnn.c
jarun: On second thoughts, we are good with this.

0xACE:
Same as before been busy, but I have a few questions:

1. Why doesn't `SEL_PLUGKEY` run `run_selected_plugin()`? Took me a while to figure out why it complained about "No pipe file found."

2. Using this script from within context 1:
```
#!/usr/bin/env sh
. $(dirname $0)/.nnn-plugin-helper
nnn_cd "$(dirname $(readlink -fn $1))"
```
and then choosing `context 1` results in no change when control returns to `nnn`. Why can't we change the currently running context?

annagrram:
> Why doesn't SEL_PLUGKEY run run_selected_plugin()? Took me a while to figure out why it complained about "No pipe file found."

No clue what that is. I'll take a look.

> and then choosing context 1 results in no change when control returns to nnn. Why can't we change the currently running context?

What do you mean no change? Do you get any errors?
Currently running context is called 0 in this scenario. It's the default one (if you just hit Return as indicated in the message). If it wasn't clear enough, do you have any suggestions on how to make it clearer (It is written in the directions of nnn_cd when it asks to choose a context and in the documentation)

0xACE:
> No clue what that is. I'll take a look.

Note you have to setup `NNN_PLUG='f:fzcd' nnn` and then launch it using `xf`

> What do you mean no change? Do you get any errors?

No errors, it just returns back to `nnn` like nothing happened without a error message. I just expected that when I enter `1` it should change context `1`, but it yields no change, `nnn` stays in the directory it was originally in... Everything else seems to work fine with it though.

> Currently running context is called 0 in this scenario. It's the default one (if you just hit Return as indicated in the message). If it wasn't clear enough, do you have any suggestions on how to make it clearer (It is written in the directions of nnn_cd when it asks to choose a context and in the documentation)

Ah, sry i was flooded with tasks when I tried out the features and I missed the text... Imho it's too verbose, but I guess it's fine as it is considering that it is showcasing the features `nnn` provides. Personally I would be unlikely to ever perform a plugin on another context, it'd make more sense to run the task in the intended context to me... After reading your explanation I patched out the prompt for choosing context in my local `.nnn-plugin-helper`. It's fine by me if you do whatever you like...

annagrram:
> No errors, it just returns back to nnn like nothing happened without a error message. I just expected that when I enter 1 it should change context 1, but it yields no change, nnn stays in the directory it was originally in... Everything else seems to work fine with it though.

Nice catch!
There's a bug that if you choose an explicit context that happens to be the current context, it fails.
I'll fix it shorty.
Thank!

PR #368, opened 2019-10-23
Fix plugin control bug from #364

annagrram:
@0xACE This should address the issues you raised.
When you have time, please give it a try.

Thanks!

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3396,7 +3396,7 @@ static bool run_selected_plugin(char **path, const char *file, char *newpath, ch
 	if (len > 1) {
 		int ctx = g_buf[0] - '0';
 
-		if (ctx == 0) {
+		if (ctx == 0 || ctx == cfg.curctx + 1) {
 			xstrlcpy(*lastdir, *path, PATH_MAX);
 			xstrlcpy(*path, g_buf + 1, PATH_MAX);
 		} else if (ctx >= 1 && ctx <= CTX_MAX) {
@@ -4935,15 +4935,18 @@ nochange:
 					if (!tmp)
 						goto nochange;
 
-					mkpath(plugindir, tmp, newpath);
-					if (ndents)
-						spawn(newpath, dents[cur].name, path, path, F_NORMAL);
-					else
-						spawn(newpath, NULL, path, path, F_NORMAL);
+					xstrlcpy(rundir, path, PATH_MAX);
+					xstrlcpy(path, plugindir, PATH_MAX);
+					if (ndents) {
+						if (!run_selected_plugin(&path, tmp, newpath, rundir, dents[cur].name, &lastname, &lastdir))
+							goto nochange;
+					} else {
+						if (!run_selected_plugin(&path, tmp, newpath, rundir, NULL, &lastname, &lastdir))
+							goto nochange;
+					}
 
-					if (cfg.filtermode)
-						presel = FILTER;
-					goto nochange;
+					setdirwatch();
+					goto begin;
 				}
 
 				cfg.runplugin ^= 1;


annagrram:
Found some problems.
Will write back when they are fixed :)
Sorry about that.

src/nnn.c
```diff
@@ -4935,15 +4935,18 @@ static void browse(char *ipath, const char *session)
...
 
-					mkpath(plugindir, tmp, newpath);
-					if (ndents)
-						spawn(newpath, dents[cur].name, path, path, F_NORMAL);
-					else
-						spawn(newpath, NULL, path, path, F_NORMAL);
+					xstrlcpy(rundir, path, PATH_MAX);
+					xstrlcpy(path, plugindir, PATH_MAX);
+					if (ndents) {
+						if (!run_selected_plugin(&path, tmp, newpath, rundir, dents[cur].name, &lastname, &lastdir))
```
jarun: (ndents ? dents[cur].name : NULL) will work as a function param too. The less we duplicate code, the better.

annagrram: True. Thanks!

> The less we duplicate code, the better.

That's the reason I'm still waiting with the fix.
I'm trying to figure out how to do it without code duplication.
I'm a bit busy with something else currently, but a push should be available somewhen today.

jarun: No hurry. :+1:

jarun: General question: when using all 4 contexts, how does the `cd` behave? Does it open in the same context from which plugin is run or tries to open a new one?

jarun: I got it.

jarun:
@annagrram can you please update the PR so we can close it?

annagrram:
Yeah. I haven't forgotten about it.
Will update it this week (hopefully in the next day or two)

jarun:
@annagrram I have incorporated it at commit 53a2e4c56058db01dbe7bcbaeb37a9352fc73959.

bimlas:
Hi!

I'm new here and first of all I like to say thanks for the tool! It really clips into Unix philosophy. :+1:

Thanks for the fix the `fzcd` plugin works, but how do I remove the "Choose context 1-4 (blank for current):" question? Even though I choose the context once, it will ask me again when I use the plugin for the next time. Also, I can only use the plugin in context 1, it does not work in context 2.

jarun:
> it will ask me again when I use the plugin for the next time

That's as per design because you may choose to run the plugin from any context. It doesn't retain the history.

> Also, I can only use the plugin in context 1, it does not work in context 2.

I am assuming you are on master. I can't reproduce this. Please provide the steps.

jarun:
OK, I can reproduce it now.

jarun:
@annagrram please fix this. Invoking a plugin directly with key from any context other than the first one doesn't work.

Also, please check what happens when you do the following:

remove the plugin and use the hotkey. Does `nnn` return to the original directory? Asking this because I see that you modify `path` before calling `run_selected_plugin`. I don't think we should even be doing this because we need to retain these when there's a failure to return to correct path.

0xACE:
> Thanks for the fix the `fzcd` plugin works, but how do I remove the "Choose context 1-4 (blank for current):" question?

Me personally I have changed `.nnn-plugin-helper` to remove these lines:

```
     if [ -n "$2" ]; then
         context=$2
     else
         echo -n "Choose context 1-4 (blank for current): "
         read context
     fi
```

Which will then default to the current context.

jarun:
Does it work in other contexts still?

jarun:
> Does it work in other contexts still?

I have fixed it now.

I see another issue: the containing directory is not opening correctly.

annagrram:
> @annagrram please fix this. Invoking a plugin directly with key from any context other than the first one doesn't work.
> 
> Also, please check what happens when you do the following:
> 
> remove the plugin and use the hotkey. Does `nnn` return to the original directory? Asking this because I see that you modify `path` before calling `run_selected_plugin`. I don't think we should even be doing this because we need to retain these when there's a failure to return to correct path.

Sorry it took me forever. Difficult month.

I've just pushed a fix, and then I saw you already closed the PR.

jarun:
I've taken care of it. No problem!

Better now?

annagrram:
> I've taken care of it. No problem!
> 
> Better now?

Yeah. Thanks!

Things are starting to improve.

jarun:
Great! :+1:

bimlas:
> I have fixed it now.

I tried, everything works.

> I have changed .nnn-plugin-helper on my local copy to remove these lines ...
> Does it work in other contexts still?

Yes, the modification works fine, I can use the plugin without any problems in any context.

jarun:
Awesome! Thanks for confirming! :+1:

Issue #370, opened 2019-11-01
Error on Termux when installing plugins


jarun:
While the plugins are installed, 2 errors are thrown on termux. Applies to the `suedit` and `fzcd` commands as well.

```
u0_a117@localhost ~/G/n/plugins> ./getplugs 
./getplugs: 16: [: 1: unexpected operator
./getplugs: 18: [: 1: unexpected operator
'nnn-master/plugins/.nnn-plugin-helper' -> './plugins/.nnn-plugin-helper'
'nnn-master/plugins/README.md' -> './plugins/README.md'
'nnn-master/plugins/boom' -> './plugins/boom'
'nnn-master/plugins/checksum' -> './plugins/checksum'
'nnn-master/plugins/drag-file' -> './plugins/drag-file'
'nnn-master/plugins/drop-file' -> './plugins/drop-file'
'nnn-master/plugins/dups' -> './plugins/dups'
'nnn-master/plugins/fzcd' -> './plugins/fzcd'
'nnn-master/plugins/fzy-open' -> './plugins/fzy-open'
'nnn-master/plugins/getplugs' -> './plugins/getplugs'
'nnn-master/plugins/gutenread' -> './plugins/gutenread'
'nnn-master/plugins/hexview' -> './plugins/hexview'
'nnn-master/plugins/imgresize' -> './plugins/imgresize'
'nnn-master/plugins/imgur' -> './plugins/imgur'
'nnn-master/plugins/ipinfo' -> './plugins/ipinfo'
'nnn-master/plugins/kdeconnect' -> './plugins/kdeconnect'
'nnn-master/plugins/mediainf' -> './plugins/mediainf'
'nnn-master/plugins/moclyrics' -> './plugins/moclyrics'
'nnn-master/plugins/mocplay' -> './plugins/mocplay'
'nnn-master/plugins/ndiff' -> './plugins/ndiff'
'nnn-master/plugins/nmount' -> './plugins/nmount'
'nnn-master/plugins/notes' -> './plugins/notes'
'nnn-master/plugins/nwal' -> './plugins/nwal'
'nnn-master/plugins/oldbigfile' -> './plugins/oldbigfile'
'nnn-master/plugins/organize' -> './plugins/organize'
'nnn-master/plugins/pastebin' -> './plugins/pastebin'
'nnn-master/plugins/pdfview' -> './plugins/pdfview'
'nnn-master/plugins/picker' -> './plugins/picker'
[...]

PR #371, opened 2019-11-01
Fix bad getplugs script (#370)

annagrram:
Novice mistake. Sorry.

### DIFF
--- a/plugins/getplugs
+++ b/plugins/getplugs
@@ -13,9 +13,9 @@ is_cmd_exists () {
     echo $?
 }
 
-if [ "$(is_cmd_exists sudo)" == "0" ]; then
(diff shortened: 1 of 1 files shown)


jarun:
> Sorry.

It's OK. :+1: Definitely nothing to be sorry about. ;)

I think this is required in `suedit` and `fzcd` plugins as well.

annagrram:
> I think this is required in suedit and fzcd plugins as well.

They are already set appropriately.

I don't write shell scripts too often, so when I wrote that part in getplugs, I probably just ran on autopilot and missed that.

jarun:
> I don't write shell scripts too often

Don't worry. I didn't even know how to fix it without googling so I assumed you'll have a better shot at it. I seldom write too complex scripts myself.

jarun:
Thank you!

PR #374, opened 2019-11-03
Fix dragdrop stderr causing issues

0xACE:
For w/e reason the output from `which` causes problems with nnn
rendering. This commit silences the problematic lines.


jarun:
Thank you!

eugene55:
why not use command -v?
Look at description here: https://github.com/koalaman/shellcheck/wiki/SC2230

PR #382, opened 2019-11-15
Fix dragdrop not working in selection mode

0xACE:
xargs couldn't call function `dnd()` and I couldn't find a portable way
to do it, so I replaced it with a variable which I guess should work
fine.


jarun:
Thank you!

PR #385, opened 2019-11-17
getpulgs now asks before overwriting

KlzXS:
Now it asks before overwriting existing modified plugins and also doesn't copy plugins if they are already there. The output of `cp` looks a bit different now, but I don't think that matters.

I just now noticed that there's a space missing at `prompt`, I'll add it in with any other request if there are any.

### DIFF
--- a/plugins/getplugs
+++ b/plugins/getplugs
@@ -18,7 +18,7 @@ merge () {
 	cp -vRf $1 $2
 }
 
-prompt() {
+prompt () {
 	echo "Plugin $1 already exists and is different."
 	echo -n "Keep (k), merge (m), overwrite (o) [default: k]? "
 	read operation
--- a/plugins/getplugs
+++ b/plugins/getplugs
@@ -13,6 +13,25 @@ is_cmd_exists () {
     echo $?
 }
 
+merge () {
+	vimdiff $1 $2
+	cp -vRf $1 $2
+}
+
+prompt() {
+	echo "Plugin $1 already exists and is different."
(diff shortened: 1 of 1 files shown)


jarun:
> and also doesn't copy plugins if they are already there

and the contents are same?

>  I'll add it in with any other request if there are any.

Please update this patch. Don't wait.

KlzXS:
It won't copy it over if the contents are exactly the same, correct.

jarun:
Thank you so much!

jarun:
@KlzXS there is an issue: if I select to merge and then just press ":qa" in vimdiff to quit (without any changes), the new file from the repo overwrites the earlier file. Please fix.

jarun:
Here is the relevant output:

```
Keep (k), merge (m), overwrite (o) [default: k]? m
2 files to edit
'imgresize' -> '../../plugins/imgresize'
```

jarun:
Overwrite happens all the time if merge is chosen, no matter what.

In case it helps, I am modifying the file on the RHS (the existing one).

jarun:
I have fixed it at commit 472564fd7f1245617a792f20cdf43ca47e4164e5.

KlzXS:
Forgive me for not paying enough attention to what I was doing at the time. For some reason I thought it was better to modify the downloaded file with your changes and then place that in the plugins directory. :disappointed:

jarun:
> Forgive me for not paying enough attention to what I was doing at the time.

No problem at all! The new capabilities are awesome!

PR #388, opened 2019-11-21
Don't use xargs options -I -o and -a

KlzXS:


jarun:
Works like a charm! Simply awesome @KlzXS, simply awesome!

I envy your scripting skills!!!

KlzXS:
Glad I could bring some improvement. Thanks for such kind words!

jarun:
@KlzXS I mean it.

I believe we can also drop our dependency on findutils with this change. As we can now literally work with any `xargs` implementation. Please confirm. I'll update the deps table accordingly.

KlzXS:
Of the findutils we only use `xargs` right? And we only ever use `-0` and `-n` which I personally think any implementation should have. *But* POSIX doesn't specify it needs to have `-0`.

I also just realised that I used `-i` for `sed` to simplify the command in one place, which is also not in the standard.

I honestly don't really care if we leave it as it is, most systems are already covered with what we have right now. I think it's ok to drop findutils.

jarun:
Thanks for the details. We will drop it and see if someone complains. We have a very clean code now.

KlzXS:
>We have a very clean code now.

I think there is still a lot of space for improvement but that's a task for another time.

jarun:
> I think there is still a lot of space for improvement but that's a task for another time.

I'm always up for cleanups. The lesser the code we need to maintain the better.

jarun:
One of the main reasons behind the plugin framework. The core does it's job, rest is done by respective utilities.

0xACE:
> I envy your scripting skills!!!

I'd like to plug [shellcheck](https://github.com/koalaman/shellcheck), and I strongly advocate we use it in the future. Most (i.e. all) of our plugins have issues that `shellcheck` points out, issues that we should address.

Other than that great work :D

jarun:
@0xACE it's a good proposal! Can you please take it up when you have the time? I can add a line item in the ToDo list.

jarun:
I have added a line item and started with some easy ones. You'll have to pitch in definitely. I'll let you know once I am done. Spare some time.

PR #393, opened 2019-12-03
Plugin qmv

jlneder:
It should work as expected.

### DIFF
--- a/plugins/renamer
+++ b/plugins/renamer
@@ -1,0 +1,41 @@
+#!/usr/bin/env sh
+
+# Description: Batch rename selection or current directory with qmv
+#
+# Notes:
+# - Try to mimic current batch rename functionality but with correct
+#   handling of edge cases by qmv or vidir.
+#	Qmv opens with hidden files if no selection is used. Selected
+#	directories are shown.
+#   Vidir don't show directories nor hidden files.
+#
+# Shell: POSIX compliant
+# Author: José Neder
+
+selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
+
+if command -v qmv >/dev/null 2>&1; then
+	batchrenamesel="qmv -fdo -da"
+	batchrename="qmv -fdo -a"
+elif command -v vidir >/dev/null 2>&1; then
+	batchrenamesel="vidir"
+	batchrename="vidir"
+else
+    printf "there is not batchrename program installed."
+    exit
+fi
+
+if [ -s "$selection" ]; then
+    printf "rename selection? "
+    read -r resp
+fi
+
+if [ "$resp" = "y" ]; then
+    # -o flag is necessary for interative editors
+    xargs -o -0 $batchrenamesel < "$selection"
+elif [ ! "$(LC_ALL=C ls -a)" = ".
+.." ]; then
+	# On older systems that don't have ls -A
+    $batchrename
+fi
+
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -14,6 +14,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 
 | Plugin (a-z) | Description | Lang | Deps |
 | --- | --- | --- | --- |
+| renamer | Batch rename files with qmv or vidir | sh | [renameutils](https://www.nongnu.org/renameutils/) or [moreutils](https://joeyh.name/code/moreutils/) |
 | boom | Play random music from dir | sh | [moc](http://moc.daper.net/) |
 | dups | List non-empty duplicate files in current dir | sh | find, md5sum,<br>sort uniq xargs |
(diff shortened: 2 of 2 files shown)


plugins/qmv
```diff
@@ -0,0 +1,29 @@
...
+# - Try to mimic current batch rename functionality but with correct
+#   handling of edge cases by qmv.
+#
+# Shell: POSIX compliant
+# Author: José Neder
+
+selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
+echo $selection
+
+if [ ! -s "$selection" ] && [ -z "$1" ]; then
```
jarun: I think we can get rid of this initial check. Please consider the following:

```
if selection exists:
    ask "rename selection?"

if resp = 'y'
    rename selection
    exit

if current dir is not empty:
    qmv $PWD
```

Note: "$1" is the hovered file or dir. I think the more natural use case is visit the directory and issue batch rename.

plugins/qmv
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
+
+# Description: Batch rename selection or current directory with qmv
+#
+# Notes:
+# - Try to mimic current batch rename functionality but with correct
```
jarun: `vidir` is probably a better candidate with numbering and all.

jarun:
In addition to review comments, CI fails because POSIX `read` has no `-n`.

plugins/qmv
```diff
@@ -0,0 +1,29 @@
...
+
+selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
+echo $selection
+
+if [ ! -s "$selection" ] && [ -z "$1" ]; then
+   exit
+fi
+
+if [ -s "$selection" ]; then
+    printf "rename sel? "
```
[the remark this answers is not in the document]
jlneder: in nnn say rename sel, I only put the same string.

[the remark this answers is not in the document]
jarun: I guessed that. `nnn` strives to save bytes. With scripts we don't have any such restrictions.

plugins/qmv
jlneder: got it

plugins/qmv
```diff
@@ -0,0 +1,29 @@
...
+    read -r -n1 resp
```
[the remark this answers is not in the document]
jlneder: I didn't think of it. Ok

jlneder:
I push just push a version with the changes requested. I open qmv with hidden files now  but it would be better if I can somehow know if nnn is showing hidden files or not. Ideas?

jarun:
I'll check and confirm.

Also, please consider using vidir than qmv. For newbies, vidir is much more cleaner than qmv (even the columns don't show properly in qmv and tht really put me off) .

jlneder:
Are you talking about dual-column mode?
You should run qmv with the same tabsize that your editor:
for ex.
> qmv -otabsize=4 
FYI I'm using it in destination only mode, so, no improperly columns here.
But I'm changing the name to batchrename and generalizing the plugin to take qmv or vidir. Give me a few minutes.

jlneder:
I just force push the changes. Tell me if it needs another change or fix.

jarun:
If hidden files are shown in `nnn` (using `.`) we take all hidden in rename otherwise only non-hidden.

Yes, both vidir and qmv is good.

jarun:
Please `git mv` the plugin to `renamer` and update the table.

jlneder:
Well if you can tell me a way to know if hidden files are shown from the plugin, I would love to add it.
I just changed the name from batchrename to renamer.

jarun:
Please add an var on top that users can set if they want to see hidden also. By default leave it without hidden.

In general, we expect users to customize plugins and use as they need. A lot of work has gone in to detect user modified plugins and give options to keep, overwrite, merge etc.

plugins/renamer
```diff
@@ -0,0 +1,41 @@
...
+    exit
+fi
+
+if [ -s "$selection" ]; then
+    printf "rename selection? "
+    read -r resp
+fi
+
+if [ "$resp" = "y" ]; then
+    # -o flag is necessary for interative editors
```
jarun: interative -> interactive

plugins/renamer
```diff
@@ -0,0 +1,41 @@
...
+    xargs -o -0 $batchrenamesel < "$selection"
+elif [ ! "$(LC_ALL=C ls -a)" = ".
```
jarun: Please drop these hacky checks and keep the code clean.

jarun:
I'll take care of the changes. Thanks!

jarun:
Works like a charm!

plugins/renamer
```diff
@@ -0,0 +1,41 @@
...
+.." ]; then
```
jarun: @jlneder why is the newline necessary here?

PR #396, opened 2019-12-04
Fix always archiving current file

KlzXS:
Sorry if I messed up the formatting, was going by what `git diff` showed. My editor doesn't display it aligned.

Are there any other prompts that where changed but not the logic? I think we should make this part of 2.8, we don't want to break functionality that was previously there.


jarun:
The other one is `MSG_RENAME_OPTS`. We are covered there.

jarun:
Thanks @KlzXS!

Issue #400, opened 2019-12-10
Selection issues with version 2.8.1


jarun:
Received the following issue reports in mail:

I found four selection issues in 2.8.1:

--

1. In case I select files with Space, the number of selected files
is shown within the status bar. But when I de-select them again
the plus sign is removed but the number in the status bar is
increased instead of decreased.

--

2. If I now hit K to flush/edit my selection list I get "failed!"
instead of the expected "0 selected" in my status bar and if I
hit M /something/ is listed. As far as I can see it's the last
(but now unselected!) item of the former selection list. Using
current master the whole (but now unselected list) is shown.

--

3. On a just started nnn (without a selection) pressing X brings
the expected "0 selected", but after using nnn for a while (I
am still looking for a way to reliably trigger it) I got the
question "forcibly remove 0 file(s) (unrecoverable)?" and
when I later quit nnn I saw the formerly hidden error

   rm: cannot remove 'sh': No such file or directory

--

4. When first applying a filter and then hitting a key to operate
on the (still empty) selection list, after the "0 selected" my
filter is removed. Same (or similar) with the following:

  - filter for "a"    [/] [a] [Esc]
  - filter for "b"    [/] [b] [Esc]
  - filter for "c"    [/] [c] [Esc]

If I now (accidently) press the slash twice while trying to
also filter for "d" (i.e. [/] [/] [d]) I get "invalid key"
and my filter is also reset.

jarun:
1. Seems to be fixed in master. I can't reproduce. Commit 3b2b6f2b46325bc74e3dd202fb000760b0a55fe9 may be relevant.

2.
> If I now hit K to flush/edit my selection list I get "failed!" instead of the expected "0 selected"

The message will now be "0 selected"

> Using current master the whole (but now unselected list) is shown.

We do not update the selection buffer but we write to the selection file when we select files. This is done so users who are working on multi-nnn environments like tmux do not have to flush the buffer every time. This issue was happening as we were not clearing the file when no file was selected after deselection.

3. 

> rm: cannot remove 'sh': No such file or directory

Same reason as 2.

4. 

> after the "0 selected" my filter is removed.

Expected. nnn supports filter on filtered entries and the last filter will be cleared because you have run an operation on the selection list.

> If I now (accidently) press the slash twice while trying to also filter for "d" (i.e. [/] [/] [d]) I get "invalid key"

<kbd>/</kbd> works as an additional lead key when filter is on - https://github.com/jarun/nnn/wiki#prompts

[...]

PR #401, opened 2019-12-10
Fix not properly checking if file is selected

KlzXS:
This solves the issues described in #400.

> In case I select files with Space, the number of selected files
is shown within the status bar. But when I de-select them again
the plus sign is removed but the number in the status bar is
increased instead of decreased.


jarun:
Thank you!

Issue #386, opened 2019-11-17
ToDo list


jarun:
Rolled from #337.

### For next release

- [x] all keybinds and options reviewed by the team, finalized and frozen
- [x] auto archive handling by extension (see config `NNN_ARCHIVE`)
- [x] reverse sort (keys <kbd>R</kbd> and <kbd>^T</kbd>)
- [X] sort only filtered entries (instead of refresh)
- [x] Lead key simplified to Bookmark key (<kbd>b</kbd> or <kbd>^/</kbd>)
- [x] improved help screen readability
- [x] plugin `nuke`: sample opener (CLI-only by default)
- [x] internal plugin `.cbcp`: copy selection to system clipboard
- [x] internal plugin `.ntfy`: show notis on cp, mv, rm completion
- [x] plugin `autojump`: navigate using autojump
- [x] plugin `upload` (previously `transfer`), uses https://file.io
- [x] prefix `-` to skip dir refresh after running (cmd as) plugin
- [x] suffix `*` to skip confirmation after running cmd as plugin
- [x] all fuzzy plugins modified to support both `fzf` and `fzy`
- [x] indicate range selection mode with `*`
- [x] list keys at bookmark and plugin key prompts
- [x] jump to pinned directory like bookmark (Bookmark key followed by <kbd>,</kbd>)
- [x] toggle executable (key <kbd>*</kbd>)
- [x] show mime along with file details
- [x] retain filter when <kbd>Up</kbd> and <kbd>Down</kbd> are pressed (across contexts)
- [x] remove option `-f` to run filter as prompt on prompt key (can be disruptive)
- [x] remove redundant plugin key <kbd>:</kbd> (alternatives more convenient)
- [x] remove redundant first file key <kbd>Lead'</kbd> (alternative more convenient)
- [x] option `-x`: enable notis and copy selection to system clipboard
- [x] option `-g`: regex filters (substring filter is default now)
- [x] option `-Q`: quit program without confirmation
- [x] option `-s`: load session (earlier `-e`)
- [x] option `-n`: start in nav-as-you-type mode (earlier `-i`)
- [x] option `-v`: version sort (earlier `-n`)
- [x] option `-V`: show program version (earlier `-v`)
- [x] option `-A`: disable dir auto-select (earlier `-t`)
- [x] key <kbd>v</kbd>: runtime version sort
- [x] key <kbd>s</kbd>: manage session
[...]

PR #403, opened 2019-12-14
Get nnn compile and run on Haiku

annagrram:
I got a bit of free time so I decided to take something small from #386.

I don't use Haiku, but got interested in it, so here is a small nudge in the right direction.
Code now compiles and runs.
Still need to implement eventing and test everything else :)

Maybe more people with more familiarity with Haiku can jump in and help.


jarun:
This is great! I would love the eventing to work on Haiku so please take your time.

annagrram:
Added support for filesystem monitoring.

Haiku's API is C++, so had to write it in C++ and provide interop to C.

Also, had to change the Makefile to detect Haiku and make some more compilations.
I'm pretty sure it's not the cleanest way, but unfortunately I'm not a Makefile witch. However, it works :)

Again, more tests are necessary of course.

jarun:
@KlzXS is there a same way to handle the Makefile?

How about a Haiku-specific Makefile? @scottmc and @waddlesplash is that acceptable? Or you have a build system that requires the file name to be Makefile?

KlzXS:
I'll take a look at all this tomorrow, or later tonight if I find some time. Probably should also setup a basic VM.

jarun:
@annagrram from the CI failure it looks like we should specify the files we need clang-tidy-8 to scan - nnn.c, nnn.h, dbg.h. We can skip the Haiku dir. I don't think once stable we'll touch that frequently.

annagrram:
> @annagrram from the CI failure it looks like we should specify the files we need clang-tidy-8 to scan - nnn.c, nnn.h, dbg.h. We can skip the Haiku dir. I don't think once stable we'll touch that frequently.

Yeah, I saw that.
I need to check where to edit the config and how, because I'm not currently familiar with the CI configs

jarun:
The relevant line is: https://github.com/jarun/nnn/blob/master/.circleci/config.yml#L52

annagrram:
> The relevant line is: https://github.com/jarun/nnn/blob/master/.circleci/config.yml#L52

I also had to change `.travis.yml`, but thanks for the tip in the right direction :)

jarun:
Most of my projects use only CircleCI now. `nnn` uses Travis because of native macOS compilation. I keep forgetting Travis.

annagrram:
@jarun You can also find a basic recipe here: https://gist.github.com/annagrram/c9c7230c6a9fed931b019dfd99aab083

It's a gist and not a file in misc currently because, I guess, a more permanent URI is needed and we need a release to also provide checksums (without them, the config file of the porter system needs to be edited to allow unsafe sources).

annagrram:
@jarun Maybe we can ask people in the thread you opened in haikuports to give nnn a try, and they could report any migration issues back.
With the recipe, it's a breeze to get all the dependencies and build, and I can also provide a built package.

jarun:
That's kind of indefinite and uncertain. I have kind of approached Haiku users on GitHub and reddit and am appalled by their level of interest. No wonder they are building an OS for years.

I will test it myself next week. :+1: 

I won't lie. I wasn't aware of the C++ thing. With the amount of effort and time you have put in this now I kind of wish that I had redacted that line item and saved you from this ordeal.

annagrram:
> I won't lie. I wasn't aware of the C++ thing. With the amount of effort and time you have put in this now I kind of wish that I had redacted that line item and saved you from this ordeal.

Don't worry about it.
It was a fun experience :)

If you don't feel like adding this to the project because of the C++ thing, that's OK. You shouldn't take the time spent into consideration for that.

Moreover, we can also export a patch file from the commits and add them to the recipe.
That way, there would still be a recipe for installing `nnn` on Haiku, but without maintaining C++ code in the project itself.

jarun:
> Moreover, we can also export a patch file from the commits and add them to the recipe.

I will think about it. Let me get to this next week.

annagrram:
@jarun I created a fork of haikuports and pushed a new branch named `nnn-recipe` with a recipe based on v2.8.1.
It can be used as-is, without changing the code tree of the project. (Cherry-picked the commits from this PR against the version tag and created a patchset)

If you want, I can open a PR for it to be added to haikuports.

You can check it out at https://github.com/annagrram/haikuports

jarun:
This looks very good!

How about keeping the changes in `nnn.c` in our repo? That way we will not have merge issues (hopefully) in `nnn.c` and the version dep will also be loosely coupled. It's within the macro so I don't see any harm in doing that.

annagrram:
Even though it's within the macro, it's bizarre having code in the codebase the reference unknown header files, functions and an opaque structure.
I would advise against it.

I think it should be either in the codebase or entirely outside.
We can always rebase against new versions and hopefully, nothing will change drastically in the monitoring area between versions.

jarun:
> We can always rebase against new versions

Our code changes frequently and that would be an issue. We can't keep doing this on every release.

> it's bizarre having code in the codebase the reference unknown header files, functions and an opaque structure

Let's have the code that we need in nnn.c in the file and have the link to the downstream patch (+details) under the first instance where we use the HAIKU macro. That should be clear enough for people to know where to go to for Haiku code.

KlzXS:
Sorry for not being active the last few days, didn't quite realize how long it would take me to finish up what I had to do.

Where are we with this now? What all needs to be done? I can see that it doesn't detect symlinks.

jarun:
No worries. No hurry to next release. My goal for the next release is a stable release.

@annagrram will point you...

annagrram:
> Where are we with this now? What all needs to be done? I can see that it doesn't detect symlinks.

Interesting. I don't see a reason why it won't detect it, but I'll test it and see what kind of event symlinks emit.

> @annagrram will point you...

Currently, there's a standalone recipe against v2.8.1 that apply patches and build a package against that source release of that version.

For the next versions, we need to incorporate the code of `nnn.c` into the tree, so any future changes to the filesystem monitoring areas will be done as a group with the code for Haiku.
But, because the integration with Haiku requires C++ code, we decided that the interop code will remain as a patch in the HaikuPorts tree.

To summarize, we are in a position where the basics work but still require testing.
And, we need to rearrange the code and patches a little bit.

jarun:
> And, we need to rearrange the code and patches a little bit.

@annagrram please do this before you go. Probably I and @KlzXS can test and fix stuff during your absence and get it ready to merge.

annagrram:
@jarun Do you want to also add the necessary Makefile changes to the tree, or keep them as a patch?

jarun:
> Makefile changes to the tree

Is it possible to have a separate Makefile for Haiku in the patch? That way we don't need to go through merge issues.

annagrram:
> Is it possible to have a separate Makefile for Haiku in the patch? That way we don't need to go through merge issues.

Yeah. I can separate them but it won't happen today I think.
I decided to take the laptop with me to finish things on the flight, so I might get it sorted then.
Meanwhile, I pushed the code for nnn.c here and you can find the recipe and the patches for v2.8.1 in the link I shared above.
You can use it to get the Makefile and the interop files necessary to build against master.
I'll make a clean recipe and patch later.

jarun:
No problem! Have a great vacation and lots of fun!

jarun:
@0xACE @KlzXS @maximbaz what are you guys up to during the time?

maximbaz:
I'm starting a new job in January, I am planning to spend some time during holidays to backup data / move to a different laptop and do some organizational work, really not sure if I will have the time to look into this PR...

jarun:
Cool! All the best! Don't worry about the PR. We got it covered.

0xACE:
I'm going to have a crazy fun time with family and friends. Hoping to do some light traveling too :D. 

Other than that my availability for `nnn` is quite limited. I work on it when I find the time for it. But honestly, since I started using it `nnn` has tacked on a lot of features, and it was practically complete when I started using it.

jarun:
I'm visiting hometown too! At the airport right now. Pretty excited.

KlzXS:
Have fun at your various travels!

I should be mostly available, but probably not 100% throughout January.

jarun:
@KlzXS thank you!

Can you please share the commands to install deps on Haiku?
I install devel:libreadline and devel:ncurses through pkgman but it still can't find the `readline/history.h` file.

Also it doesn't understand O_NORL=1.

KlzXS:
Did you try rebooting the system? That magically fixed it for me.

jarun:
OK, it understands the option. I was adding a `-` extra.

jarun:
I could compile it. Hurray!

jarun:
- I can see symlinks, even for broken target.
- Events work, I could see a new file created from another terminal show up.

I see the colors in the context numbers, but both directory and file entries are in black.

KlzXS:
> I can see symlinks, even for broken target.

I just tried it again, for symlinks to files there is the `@` suffix, but symlinks to folders appear just as regular folders with size of `0B`. Try looking at `/bin` that should be a symlink to whatever Haiku uses as `/usr/bin`.

jarun:
OK, got it.

I see another issue:

`make O_NORL=1 strip install` fails with

`cannot create regular file '/usr/local/bin/nnn': Invalid Argument`

jarun:
Could you make vim work? I installed it (pkgman says v8.1 is already installed) but it's says command not found.

KlzXS:
It was doing all that for me as well. Was late at night so I just rage shut it off and in the morning it worked fine :man_shrugging: 

As for the color, ANSI escape sequences work, and context numbers are colored but directory entries are not.

jarun:
I think I found the issue. Let me reboot. ;)

jarun:
yayyyy, vim works after reboot!!!

KlzXS:
> cannot create regular file '/usr/local/bin/nnn': Invalid Argument

Didn't see this. Haiku doesn't have `/usr` at all, I'm not quite sure where it installs programs to as when I try to write to (as root since I'm too lazy to create another user) `/bin` it fails complaining that it's a read-only filesystem. We'll need to change that.

jarun:
The following patch fixes both the dir color and link to dir issues:

```
index 33ac941..1326d3b 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3765,10 +3765,9 @@ static int dentfill(char *path, struct entry **dents)
                goto exit;
 
 #ifdef __sun
-       if (cfg.blkorder) { /* no d_type */
+       flags = AT_SYMLINK_NOFOLLOW; /* no d_type */
 #else
        if (cfg.blkorder || dp->d_type == DT_UNKNOWN) {
-#endif
                /*
                 * Optimization added for filesystems which support dirent.d_type
                 * see readdir(3)
@@ -3778,6 +3777,7 @@ static int dentfill(char *path, struct entry **dents)
                 */
                flags = AT_SYMLINK_NOFOLLOW;
        }
+#endif
 
        do {
                namep = dp->d_name;
```

I will check this is. Applies to all platforms without `dp->type` (like SUN).

If you can figure how to install to `/bin` please share the patch here. I am not sure if @annagrram already has a way to do this. She must have installed it.

annagrram:
> If you can figure how to install to /bin please share the patch here. I am not sure if @annagrram already has a way to do this. She must have installed it.

As was mentioned before, Haiku doesn't have the /usr directory.
The Makefile patch + HaikuPorter should make it install in the correct places.
If you create a package (hpkg file), just move the package file to ~/config/packages.
Otherwise, non-packaged apps can be installed in ~/config/non-packaged or for system-wide /system/non-packaged

KlzXS:
I guess `make install` should install it as a non-package system-wide as that's what it already does. How are we handling the Makefile? Do we have single one with a bunch of ifs? Do we have a separate one for Haiku? Or do we make a configure script which would set the variables to appropriate values for the given system?

annagrram:
> I guess make install should install it as a non-package system-wide

I guess so too.

> How are we handling the Makefile? Do we have single one with a bunch of ifs? Do we have a separate one for Haiku? Or do we make a configure script which would set the variables to appropriate values for the given system?

Currently, I have a patch for the Haiku recipe that adds all the necessary bits for Haiku and sets the install directories appropriately when run with HaikuPorter to create a Haiku package (using the environment variables of the build system for that).

I don't know how we want to handle the Makefile beside that.
We can have a separate one for Haiku or add an if statement in the current Makefile, but that would require some other reorganisations of the Makefile.

jarun:
> We can have a separate one for Haiku

Can we do it this way? Have a Makefile for Haiku within the Haiku directory and before compiling for Haiku move it to the parent from the recipe control?

src/nnn.c
```diff
@@ -3765,7 +3789,7 @@ static int dentfill(char *path, struct entry **dents)
 	if (!dp)
 		goto exit;
 
-#ifdef __sun
+#if defined(__sun) || defined(__HAIKU__)
```
jarun: Please adapt to commit b1207d54ea6fc96fa7244aa5c0f32f264862afc6.

annagrram:
You don't have to move the Makefile to the parent directory. You just need to pass it to the make command with the right flag.
I'll take care of it, but it might be only next week.

I will also update the PR to be fast merged after the recent commits.

jarun:
> You don't have to move the Makefile to the parent directory. You just need to pass it to the make command with the right flag.

True! Bottom line - don't touch the current makefile. ;)

annagrram:
`touch Makefile` :P

annagrram:
@jarun A few questions:
Do you want to haiku folder to be in the misc folder of the tree?
If so, do you want it to contain just the recipe and patchset or the individual files?
Or do you want the patchset and recipe to exist just in HaikuPorts?

How should we manage the Haiku changes apart from `nnn.c`?

jarun:
> Do you want to haiku folder to be in the misc folder of the tree?

This and the individual files would be good! Then we don't have to look at patches to fix problems.

annagrram:
New make file and tree organisation are done.

I ended up touching the current Makefile, but I promise it's for a good cause (so you won't have to edit too many files before a release)

Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION = 2.8.1
+include VERSION.mk
```
jarun: This file should be added in DISTFILES list as well.

jarun:
I knew you would touch the Makefile. ;)

Makefile
annagrram: Done

annagrram:
Should we add `misc/haiku` do dist all the time?
Otherwise we'll need to maintain patch for HaikuPorts.

jarun:
> Otherwise we'll need to maintain patch for HaikuPorts.

Whatever helps with least maintenance.

annagrram:
On second thought, it doesn't seem necessary.
If everything looks in order, you can merge.

jarun:
okies!

jarun:
Thank you!

jarun:
Please submit to haiku after next release.

annagrram:
Already submitted the backport I made for v2.8.1 to test the recipe making.
All the code is pretty much the same.
I'll keep it updated for next releases.

For now, it still hasn't been merged. Let's see how long does it take.

Begasus:
Current PR for nnn has been tested and working on both 32bit and 64bit Haiku, thanks for your work @annagrram ! :+1:

misc/haiku/Makefile
```diff
@@ -0,0 +1,117 @@
+include VERSION.mk
+
+PREFIX ?= /boot/system/non-packaged
+MANPREFIX ?= /boot/system/non-packaged/documentation/man
```
Begasus: Could this work too?
```diff
-MANPREFIX ?= /boot/system/non-packaged/documentation/man
+MANPREFIX ?= $(PREFIX)/documentation/man
```

Begasus: This way you should only set PREFIX=$prefix in the recipe for Haiku

annagrram: Good Idea.
I'll take care of it.

Begasus:
@jarun any thought on moving "misc/haiku/Makefile" to the root directory as "Makefile.haiku" or would it be overkill (checked and working here) :)

Begasus:
Mentioned PR on haikuports from @annagrram : https://github.com/haikuports/haikuports/pull/4499

jarun:
> to the root directory as "Makefile.haiku"

I would like to keep it in the separate directory as it wouldn't be useful for non-Haiku users.

Begasus:
No problem, already talked it over with @annagrram :)

Issue #405, opened 2019-12-16
Show indicator of pending input in status bar


maximbaz:
Some commands require pressing two keys, and I think it would make sense to let `nnn` show in status bar if `nnn` is waiting for the _second_ keypress, because the meaning of this second keypress depends on whether `nnn` is listening on a key to finish some action.

For example, `;` marks the start of selecting a plugin, and the next keypress will select the plugin. If you have a plugin that say jumps to a different directory on `j`, pressing `j` after `;` will not move your cursor down, but will move you to a completely different directory. Right now just by looking at `nnn` window you don't know what the next `j` keypress will do, will it move your cursor down or will it execute a plugin. It would be nice if there was some indicator, some `[plugin]` perhaps, showing that the next keypress determines which plugin to run.

A good example perhaps would be the `m` keypress. The first time you press `m` you mark the start of selection (and in status bar `[0:]` becomes `[1:]`), and a second `m` keypress finishes it. So if you see `[1:]`, you know that the next `m` will finish a selection. Except this also is not perfect, because if you already have a selection, in the status bar you will see `[1:4]`, and even after you press `m` one more time to start the selection again, status bar will continue showing `[1:4]`. 

This could be improved as well I think, maybe the second number could become `*` after you pressed `m` for the first time, because technically we don't know the number of selected files until user finishes the selection?

So the very first `m` will change from `[0:]` to `[1:*]`, the second `m` changes to `[1:3]`, the third `m` changes to `[1:*]`, the fourth `m` changes to `[1:6]` and so forth?

jarun:
> If you have a plugin that say jumps to a different directory on j, pressing j after ;

There's no way someone can press `;` without understanding what it does. Why would you press `;` and then press `j` thinking it will go to next entry? If you have added a plugin with `j` in your rc file you know very well what the sequence `;j` does.

> Right now just by looking at nnn window you don't know what the next j keypress will do

But if I pressed a `;` just now I should remember that.

> So the very first m will change from [0:] to [1:*], the second m changes to [1:3], the third m changes to [1:*], the fourth m changes to [1:6] and so forth?

We do show a popup that says "range selection on". Do you want the code to become a spaghetti?

jarun:
In a user can't remember what he is doing a fraction of a second back, trust me, `nnn` is not for him/her. We are not writing this for him/her.

jarun:
BTW, could you test the haiku port from https://github.com/jarun/nnn/pull/403?

[...]

PR #406, opened 2019-12-16
Indicate when range selection and plugin picker are waiting for input

maximbaz:
Use `[1:*]` as indicator that range selection is in progress.

Removed `range sel on` popup because it is useless now.

When `;` is pressed, it will show a prompt with all available plugins.

Fixes: #405

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3432,6 +3432,14 @@ static void printkv(kv *kvarr, FILE *fp, uchar max)
 		fprintf(fp, " %c: %s\n", (char)kvarr[i].key, kvarr[i].val);
 }
 
+static void sprintkv(kv *kvarr, char *buf, uchar max)
+{
+	uchar i = 0;
+
+	for (; i < max && kvarr[i].key; ++i)
+		buf += snprintf(buf, CMD_LEN_MAX, " %c=%s", (char)kvarr[i].key, kvarr[i].val);
+}
+
 /*
  * The help string tokens (each line) start with a HEX value
  * which indicates the number of spaces to print before the
@@ -5180,10 +5188,15 @@ nochange:
 				}
 
 				if (sel == SEL_PLUGKEY) {
+					xstrlcpy(g_buf, "pick plugin:", CMD_LEN_MAX);
+					sprintkv(plug, g_buf + strlen(g_buf), PLUGIN_MAX);
+					printprompt(g_buf);
 					r = get_input(NULL);
 					tmp = get_kv_val(plug, NULL, r, PLUGIN_MAX, FALSE);
-					if (!tmp)
+					if (!tmp) {
+						clearprompt();
 						goto nochange;
+					}
 
 					if (tmp[0] == '_' && tmp[1]) {
 						xstrlcpy(newpath, ++tmp, PATH_MAX);
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -307,7 +307,6 @@ static context g_ctx[CTX_MAX] __attribute__ ((aligned));
 static int ndents, cur, curscroll, total_dents = ENTRY_INCR;
 static int xlines, xcols;
 static int nselected;
-static bool rangesel;
 static uint idle;
 static uint idletimeout, selbufpos, lastappendpos, selbuflen;
 static char *bmstr;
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -307,6 +307,7 @@ static context g_ctx[CTX_MAX] __attribute__ ((aligned));
 static int ndents, cur, curscroll, total_dents = ENTRY_INCR;
 static int xlines, xcols;
 static int nselected;
+static bool rangesel;
```
jarun: Please add the bool at the end (line 335) and intialize to FALSE to clearly indicate what we are doing.

src/nnn.c
```diff
@@ -4137,7 +4136,7 @@ static void redraw(char *path)
 			c = cfg.apparentsz ? 'a' : 'd';
 
 			mvprintw(lastln, 0, "%d/%d [%d:%s] %cu:%s free:%s files:%lu %lldB %s",
-				 cur + 1, ndents, cfg.selmode, (nselected ?  xitoa(nselected) : ""),
+				 cur + 1, ndents, cfg.selmode, (rangesel ? "*" : nselected ? xitoa(nselected) : ""),
```
jarun: Please add brackets around `nselected ? xitoa(nselected) : ""` for clarity.
Also put (rangesel ? "*" : (nselected ? xitoa(nselected) : "")) in the next line (mind the tabs).

Same in line 4152.

jarun:
I have added some notes in the issue for the leader key indication. Please check. I think that could also be a part of this PR.

src/nnn.c
```diff
@@ -451,15 +452,14 @@ static char * const utils[] = {
...
-#define MSG_RANGE_SEL_ON 32
-#define MSG_DIR_CHANGED 33
-#define MSG_0_FILES 34
-#define MSG_EXISTS 35
-#define MSG_FEW_COLOUMNS 36
-#define MSG_REMOTE_OPTS 37
-#define MSG_RCLONE_DELAY 38
-#define MSG_APP_NAME 39
-#define MSG_ARCHIVE_OPTS 40
+#define MSG_DIR_CHANGED 32
```
jarun: Disable `rangesel` where we are using this macro. This is for dir-limited copy and we should keep the code consistent.

maximbaz:
I didn't quite understand what you meant by the bookmark button and how it is related to plugins picker, could you please have a look at the last commit and give me a hint what I am missing?

src/nnn.c
maximbaz: No change is needed here, I preserved the behavior of how it works on `master`, if you start rangesel, go in a subfolder, return to the parent folder and press `m` again, rangesel will complete successfully. But if you start rangesel, go in a subfolder and press `m` there, then you will see a prompt that directory changed and rangesel was cancelled.

jarun: > then you will see a prompt that directory changed and rangesel was cancelled

But we are not setting `rangesel = FALSE`. That's what I meant by disabling.

jarun:
> I didn't quite understand what you meant by the bookmark button

No problem. I'll take care of it.

jarun:
Thank you!

src/nnn.c
maximbaz: I'm still trying to understand how this works, but what I've tried so far is to change the macro as following (you are talking about this macro, right?):

```c
#define setdirwatch() (cfg.filtermode ? (presel = FILTER) : ((dir_changed = TRUE) && (rangesel = FALSE)))
```

But this breaks the behavior I explained above, if I start rangesel with `m`, go in a subfolder and back, then press `m` again, it will not finish selection...

jarun:
> I didn't quite understand what you meant by the bookmark button

Now I know why. The changes relate to plugin key and I was talking about the Leader key. I think we should also show something when someone presses the leader key.

src/nnn.c
jarun: Yes, you are right. I see that `rangesel` is already FALSE wen we reach the condition so we don't need to set it to FALSE explicitly.

jarun:
I think instead fo showing the key:plugin pair we should just show the plugin keys separated by space.

This would also give a uniform behaviour with bookmarks (when we are waiting after the Lead key.

jarun:
The reason is when you have too many shortcuts (we support 15) and users can't scroll right and left, they'll complain.

maximbaz:
Both consistency with bookmarks and possible screen overflow are good arguments, I agree to only show keys separated by space 👍

jarun:
Please raise the PR. Let's finish it off.

Also we can start the prompt with `plugin keys:` and `bookmark keys:` to distinguish the two. Add MSG_,... macros for these 2.

maximbaz:
I will work on it 👍 Might finish tomorrow as it's 00:30am here, hehe 🙂

jarun:
Sure thing!

Issue #407, opened 2019-12-16
Help screen is hard for perception


tim77:
#### BUG REPORTS

Hello. This help screen [1] is hard for perception. Would be nice to move second block to right a little bit. Also would be nice to colorize hotkeys for better readability. 

#### Environment details (Put `x` in the checkbox along with the information)

[x] Operating System: Fedora
[x] Desktop Environment: i3wm/GNOME
[x] Terminal Emulator: gnome-terminal/alacritty
[x] Shell: fish

![DeepinScreenshot_select-area_20191216192531](https://user-images.githubusercontent.com/5614476/70928816-7ae15080-203a-11ea-87d0-8e2c2bc0cc80.png)

jarun:
Take a look at master. I recently cleaned it up a bit. Let me know if you have some suggestions to move around the entries further without having to shift a full column (that would add a number of bytes).

tim77:
Version from master [2] looks definitely better. But TBH still think that nice to have colorized or bold hotkeys for better distinguish. Maybe this would be enough without shift a full column.

[2]
![DeepinScreenshot_select-area_20191216213416](https://user-images.githubusercontent.com/5614476/70937483-082da080-204d-11ea-9769-88226b1d962e.png)

maximbaz:
@jarun have you considered using `\t` to shift column? If you just go with a simple way of adding a new column of `\t` chars, it would at least not grow the binary by that much as it would if you added a column of say 8 spaces.

If the size is still a concern, you could expand `\t` yourself (say to `8` spaces) (so that `\t` has a predictable width, not dependent on users' PAGER) in `show_help()`, replacing the existing alignment with spaces - this way, you would actually _decrease_ the current binary size by quite a few bytes I think.

maximbaz:
I completely agree that the current help is not very readable, most of all I dislike that different lines have different number of columns (most have two, but some have three).

ignatenkobrain:
You can use libsmartcols (from util-linux) to make proper table printing...

https://karelzak.blogspot.com/2014/05/libsmartcols-pretty-output-for-everyone.html

maximbaz:
Maybe a silly idea, but it was easy to implement, and imho already helps a lot, and also doesn't increase the binary size which seems to be a concern. https://github.com/jarun/nnn/pull/408

jarun:
> You can use libsmartcols (from util-linux) to make proper table printing...

No new deps please.

jarun:
[...]

PR #408, opened 2019-12-16
Make help page more readable

maximbaz:
Proof of concept to make help page more readable by considering `helpstr` as a format string.

This also reduces the binary size from `92kb` to `88kb`.

I still think we should get rid of the third column (e.g. move `^G  QuitCD` to its own line), but before that, let me know what you think of the approach in general.

help

Reference: #407


jarun:
This is damn cool! :+1:

jarun:
`^G` can be where it is instead of a new line I guess. It's still completely readable.

maximbaz:
Glad you like it! 😄 It would help a lot to have consistent number of columns, `^G` is readable but when I'm quickly eyeballing the list and trying to see which options are available, I every time miss some shortcuts, because it feels to be a single entry `Q ^Q  Quit ^G  QuitCD`... 

Now that I know that we will accept this approach in general, let me finalize this a bit by trying to use the same number of columns everywhere, and you will have another look 👍

jarun:
OK, sure!

maximbaz:
Consistent 2-columns design, plus I expanded `extn` and `del` abbreviations as it felt better for some reason, still `88kb` binary size - what do you think about the below?

help2

jarun:
Yes, we are good. Did you update the PR?

maximbaz:
Yes, this is ready for code review and if all good merging 👍

jarun:
Thank you!

0xACE:
https://github.com/jarun/nnn/blob/7b064801471bcafd3d144e11dcafc5ced69905a7/src/nnn.c#L3499

I'd like to say thanks for this line in particular. it makes it easier for me to maintain my personal `helppatches`

previously as that ending bracket was on the same line as the last item. I had to frequently handle merge collissions to append my own helppatches, I guess I won't have to anymore :D

PR #409, opened 2019-12-17
Show prompt for bookmark selection

maximbaz:
Follow-up to #406 

May be easier to review by commits.

Things I want to particularly bring your attention to:

1. I noticed there are 3 unused `MSG_`, I cleaned them up, let me know if this was wrong.

2. We used to error-out when pressing non-existent bookmark key after `,` - this made sense when it was non-interactive, but now that we are showing `bookmark keys: a b c d` it may be intentional that user presses something else to cancel operation. In this case I no longer show error, just silently clear prompt. This is exactly the same behavior as pressing non-existent plugin key after `;`.

> - if you press a bookmark key where the bookmark destination is the same as current dir. You need to clearprompt() here.

This seems to work well out of the box, prompt is cleared even if I choose bookmark destination as the current dir 👍 

> - bookmark destination cannot be accessed. Add a printwait() with "MSG_ACCESS" in this case. it is missing today.

Added 👍

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4620,7 +4620,7 @@ nochange:
 			}
 
 			if (!get_kv_val(bookmark, newpath, fd, BM_MAX, TRUE)) {
-				printwait(messages[MSG_INVBM_KEY], &presel);
+				clearprompt();
 				goto nochange;
 			}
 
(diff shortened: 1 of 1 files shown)


jarun:
I'll try it out! Thanks for the PR! Time to sleep now. ;)

jarun:
Thank you!

PR #411, opened 2019-12-17
Gracefully abort prompts on Esc

maximbaz:
Follow-up to https://github.com/jarun/nnn/commit/8d3c24d19fd6a69b63df1d7ce9625838b4c065a0

I like the informative messages, but I want to have an ability to gracefully cancel prompt without error messages.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4580,6 +4580,9 @@ nochange:
 			}
 
 			switch (fd) {
+			case 27:
+				clearprompt();
+				goto nochange;
 			case '~': // fallthrough
 			case '`': // fallthrough
 			case '-': // fallthrough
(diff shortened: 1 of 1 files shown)


jarun:
These are not prompts. These are well-documented 2 key sequences which you want to break in by adding an escape option. Let's try to avoid such fine tunings. You press the first key of a 2 key sequence, then you decide not to, then you press escape (so still 2 keys) just to avoid an informative delay. I this that's too far fetched.

jarun:
I think there's a middle ground. Reduce the delay for these 2 cases.

maximbaz:
I can agree with your point of not fine-tuning this much to some very specific use-cases, but I think reducing the delay is not good because it can actually be more distracting/confusing to users to see some message flashing that they have no time to even read. Let's keep it like it is on master, at least it's consistent.

PR #414, opened 2019-12-17
Apply printf trick to more places to try to reduce bin size

maximbaz:
@jarun can you test if this impacts your binary size at all? Logically it should, but for some reason how ever I compile I cannot get even close to 60kb...


jarun:
No change in size. And it makes the code difficult to maintain. Probably we should look into reduction of functions, redundant logic, lines of code etc.

maximbaz:
yep, okay, thanks, I'll reopen if I come up with something meaningful

PR #415, opened 2019-12-19
Show 'invalid key' prompt in more places

maximbaz:
`c -> a` now says "invalid key"

`o -> a` now says "invalid key"

`o -> m` on a text file now says "failed!"


jarun:
:+1:

PR #417, opened 2019-12-20
Improvements for distribution

### LINKED ISSUE #416: Improvements for distribution
Since I'm also a packager, I'd like to suggest some improvements for the Makefile and distribution practices in general 🙂 

1. I can see you PGP sign your commits, would you consider providing signatures (`*.sig`) for release assets? That way users can have more trust that application sources did come from you. For Arch Linux specifically I would only want the signature for the source tarball file, since that's what I use to compile the app.

    image

    I can see you create releases using Circle CI, and you probably wouldn't want to upload your private key to them, so I imagine this is something you would have to do locally. I can take care of everything, update Makefile, so all you would have to do is to run `make sign` on each release and upload the resulting `*.sig` file(s) to Github, question is if you would agree to provide PGP signatures?

2. `make dist` is preparing tarball based on the current folder, a better way of doing so is to let `git` prepare the tarball itself. First, you tag the repo with `$VERSION` (say it is `v2.8.1` tag), and then run the following command to generate the tarball:

    ```
    git archive -o nnn-$(VERSION).tar.gz --format tar.gz --prefix=nnn-$(VERSION)/ $(VERSION)
    ```

    This generates tarball that is guaranteed to be based on a git tag (more trustworthy as no chance for accidentally packaging files with local changes), and it is guaranteed to be bit by bit identical to the archive that Github prepares for you in this link:

    image

    In fact, I think there is no reason to provide your custom tarball at all, just let everyone download the tarball already prepared by Github!

    This is needed if you will agree to provide PGP signatures, then you can do the following in `make sign`:

    ```
    git archive -o nnn-$(VERSION).tar.gz --format tar.gz --prefix=nnn-$(VERSION)/ $(VERSION)
    gpg --detach-sign nnn-$(VERSION).tar.gz
    rm -f nnn-$(VERSION).tar.gz
    ```

maximbaz:
Let's start on #416, I would need your feedback to better understand your current workflow, because I want to integrate the signing process as seamlessly as possible in your current release procedure.

My current assumptions:

- Releases are done via CircleCI, all you do is push a tag, and then `package-and-publish` is triggered and it creates Github release with all the assets.
- `make dist` is not used whatsoever (I will delete it).
- You will not upload your PGP private key to CircleCI, so signing has to happen locally on your machine, and uploaded to Github release independently.

I updated CircleCI not to generate sources tarball, we need to remember to notify other package maintainers that they need to switch to the source tarball generated by Github itself:

```diff
- https://github.com/jarun/nnn/releases/download/v2.8.1/nnn-v2.8.1.tar.gz
+ https://github.com/jarun/nnn/archive/v2.8.1.tar.gz
```

I created `make sign` that assumes the repo is already tagged with `v$(VERSION)`, it uses Git command to generate the exact same bit-by-bit identical tarball, signs it and removes the tarball.

You will need to upload the `nnn-$(VERSION).tar.gz.sig` file to Github afterwards. Would you like me to try to find some ways to automate this? We could maybe use `github.com/tcnksm/ghr` if you are okay with installing that locally?

To test, run `make sign` today, download the [source tarball](https://github.com/jarun/nnn/archive/v2.8.1.tar.gz) from Github, and run:

```
gpg --verify ./nnn-2.8.1.tar.gz.sig /path/to/downloaded/nnn-2.8.1.tar.gz
```

It should report `Good signature`.

I made `make sign` only prepare signature for the source tarball, because it's easy to replicate creation of this tarball locally without downloading anything, and because that's the only signature I personally need for Arch Linux 😉. If you'd like to sign the rest of the release assets, we would need to download stuff from Github release, sign them, and then upload the signatures. This might be an overkill for now, but let me know what you prefer.

Fixes #416. So happy that you agreed to provide signatures! 🎉

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,16 +64,10 @@ jobs:
             # Clean up
             rm -rf ./.git
             rm -rf ./dist/*
-            # Pack source
-            cd ..
-            tar -czf ${CIRCLE_PROJECT_REPONAME}-${CIRCLE_TAG}.tar.gz ${CIRCLE_PROJECT_REPONAME}
-            cd ${CIRCLE_PROJECT_REPONAME}
             # Use latest installed python3 from pyenv
             export PYENV_VERSION="$(pyenv versions | grep -Po '\b3\.\d+\.\d+' | tail -1)"
             pip install packagecore
             packagecore -c misc/packagecore/packagecore.yaml -o ./dist/ ${CIRCLE_TAG#v}
-            # Move source pack to dist
-            mv ../${CIRCLE_PROJECT_REPONAME}-${CIRCLE_TAG}.tar.gz dist/
 
       - run:
           name: "publish to GitHub"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,10 +64,14 @@ jobs:
             # Clean up
             rm -rf ./.git
             rm -rf ./dist/*
+            # Pack source
+            git archive -o ../${CIRCLE_PROJECT_REPONAME}-${CIRCLE_TAG}.tar.gz --format tar.gz --prefix=${CIRCLE_PROJECT_REPONAME}-${CIRCLE_TAG#v}/ ${CIRCLE_TAG}
             # Use latest installed python3 from pyenv
             export PYENV_VERSION="$(pyenv versions | grep -Po '\b3\.\d+\.\d+' | tail -1)"
             pip install packagecore
             packagecore -c misc/packagecore/packagecore.yaml -o ./dist/ ${CIRCLE_TAG#v}
+            # Move source pack to dist
+            mv ../${CIRCLE_PROJECT_REPONAME}-${CIRCLE_TAG}.tar.gz dist/
 
       - run:
           name: "publish to GitHub"
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,6 @@ CFLAGS += $(CFLAGS_CURSES)
 
 LDLIBS += $(LDLIBS_CURSES)
 
-DISTFILES = src nnn.1 Makefile README.md LICENSE
 SRC = src/nnn.c
 HEADERS = src/nnn.h
 BIN = nnn
@@ -85,16 +84,14 @@ uninstall:
 strip: $(BIN)
 	$(STRIP) $^
 
-dist:
-	mkdir -p nnn-$(VERSION)
-	$(CP) -r $(DISTFILES) nnn-$(VERSION)
-	tar -cf nnn-$(VERSION).tar nnn-$(VERSION)
-	gzip nnn-$(VERSION).tar
-	$(RM) -r nnn-$(VERSION)
+sign:
+	git archive -o nnn-$(VERSION).tar.gz --format tar.gz --prefix=nnn-$(VERSION)/ v$(VERSION)
+	gpg --detach-sign nnn-$(VERSION).tar.gz
+	rm -f nnn-$(VERSION).tar.gz
 
 clean:
-	$(RM) -f $(BIN) nnn-$(VERSION).tar.gz
+	$(RM) -f $(BIN) *.sig
 
 skip: ;
 
-.PHONY: all debug install uninstall strip dist clean
+.PHONY: all debug install uninstall strip sign clean
--- a/Makefile
+++ b/Makefile
@@ -97,10 +97,10 @@ sign:
 	gpg --detach-sign --yes nnn-$(VERSION).tar.gz
 	rm -f nnn-$(VERSION).tar.gz
 
-	$(eval TOKEN=$(shell cat /tmp/access_token))
 	$(eval ID=$(shell curl -s 'https://api.github.com/repos/jarun/nnn/releases/tags/v$(VERSION)' | jq .id))
 	curl -XPOST 'https://uploads.github.com/repos/jarun/nnn/releases/$(ID)/assets?name=nnn-$(VERSION).tar.gz.sig' \
-	    -H 'Authorization: token $(TOKEN)' -H 'Content-Type: application/pgp-signature' --upload-file nnn-$(VERSION).tar.gz.sig
+	    -H 'Authorization: token $(NNN_SIG_UPLOAD_TOKEN)' -H 'Content-Type: application/pgp-signature' \
+	    --upload-file nnn-$(VERSION).tar.gz.sig
 
 clean:
 	$(RM) -f $(BIN) nnn-$(VERSION).tar.gz *.sig
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ CFLAGS += $(CFLAGS_CURSES)
 
 LDLIBS += $(LDLIBS_CURSES)
 
+DISTFILES = src nnn.1 Makefile README.md LICENSE
 SRC = src/nnn.c
 HEADERS = src/nnn.h
 BIN = nnn
@@ -84,6 +85,13 @@ uninstall:
 strip: $(BIN)
 	$(STRIP) $^
 
+dist:
+	mkdir -p nnn-$(VERSION)
+	$(CP) -r $(DISTFILES) nnn-$(VERSION)
+	tar -cf nnn-$(VERSION).tar nnn-$(VERSION)
+	gzip nnn-$(VERSION).tar
+	$(RM) -r nnn-$(VERSION)
+
 sign:
 	git archive -o nnn-$(VERSION).tar.gz --format tar.gz --prefix=nnn-$(VERSION)/ v$(VERSION)
 	gpg --detach-sign nnn-$(VERSION).tar.gz
@@ -94,4 +102,4 @@ clean:
 
 skip: ;
 
-.PHONY: all debug install uninstall strip sign clean
+.PHONY: all debug install uninstall strip dist sign clean
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ sign:
 	rm -f nnn-$(VERSION).tar.gz
 
 clean:
-	$(RM) -f $(BIN) *.sig
+	$(RM) -f $(BIN) nnn-$(VERSION).tar.gz *.sig
 
 skip: ;
 
--- a/Makefile
+++ b/Makefile
@@ -94,8 +94,13 @@ dist:
 
 sign:
 	git archive -o nnn-$(VERSION).tar.gz --format tar.gz --prefix=nnn-$(VERSION)/ v$(VERSION)
-	gpg --detach-sign nnn-$(VERSION).tar.gz
+	gpg --detach-sign --yes nnn-$(VERSION).tar.gz
 	rm -f nnn-$(VERSION).tar.gz
+
+	$(eval TOKEN=$(shell cat /tmp/access_token))
+	$(eval ID=$(shell curl -s 'https://api.github.com/repos/jarun/nnn/releases/tags/v$(VERSION)' | jq .id))
+	curl -XPOST 'https://uploads.github.com/repos/jarun/nnn/releases/$(ID)/assets?name=nnn-$(VERSION).tar.gz.sig' \
+	    -H 'Authorization: token $(TOKEN)' -H 'Content-Type: application/pgp-signature' --upload-file nnn-$(VERSION).tar.gz.sig
 
 clean:
 	$(RM) -f $(BIN) nnn-$(VERSION).tar.gz *.sig


jarun:
> I updated CircleCI not to generate sources tarball, we need to remember to notify other package maintainers that they need to switch to the source tarball generated by Github itself

This can't happen. `nnn` is available in too many distributions now and it's difficult to reach them. Please keep the location same.

> make dist is not used whatsoever (I will delete it).

It's for local usage. Do not delete.

> so signing has to happen locally on your machine, and uploaded to Github release independently.

What does this mean?

> Would you like me to try to find some ways to automate this?

Yes. I don't want to do anything outside the normal flow.

> if you are okay with installing that locally?

No, I am not OK with that.

jarun:
To see what we do today: https://github.com/jarun/nnn/wiki/Release-checklist

jarun:
Also https://github.com/easybuilders/easybuild-easyconfigs/issues/5151

maximbaz:
> >  so signing has to happen locally on your machine, and uploaded to Github release independently.
>
> What does this mean?

If we can't make CircleCI create the signature (because CircleCI does not have your PGP private key), then signature needs to be create where your private key is present - on your computer. So CircleCI will create all release assets _except_ the signature and upload them to Github release, and locally you will run `make sign` to create the signature and also upload it to the same Github release. So it will be step 11 in "Release checklist", run `make sign`. Makes sense and sounds good?

> > if you are okay with installing that locally?
> 
> No, I am not OK with that

OK, I'll have a look, I'm almost sure it's possible to achieve with a plain `curl` command provided that you will save Github access token in some file / environment variable for the script to read.

> Also https://github.com/easybuilders/easybuild-easyconfigs/issues/5151

This is outdated, I don't have a reference right now, but since recently git archives are guaranteed to be reproducible.

jarun:
That's the reason we started generating our own archives.

jarun:
Also the other reason is, if we move away from github tomorrow we'll have to get rid of these deps too.

jarun:
See if we can use our generated package. Also once done, please udpaet the release checklist with the steps.

maximbaz:
> > I updated CircleCI not to generate sources tarball, we need to remember to notify other package maintainers that they need to switch to the source tarball generated by Github itself
> 
> This can't happen. nnn is available in too many distributions now and it's difficult to reach them. Please keep the location same.

Gotcha, makes sense. In this case I simply changed the way archive is generated, now this is exactly the same Git archive as Github is generating, so even if you switch away from Github, the archive and its checksums will remain the same, and your PGP signature will remain valid.


> > make dist is not used whatsoever (I will delete it).
> 
> It's for local usage. Do not delete.

Ah, I see! I returned it, and kept unchanged, it does not generate git archive (just in case you use `make dist` _before_ tagging repo).

----


Will have a look at how to upload PGP signature to github release without any dependencies after a short break 😉

maximbaz:
Could you please have another look? I made `make sign` also upload the signature to Github release, but let me know where you want to store access token, in a file or in an environment variable? For now it's hardcoded to `/tmp/access_token` in line 100.

To create a new token, go here, select `public_repo` and create:

https://github.com/settings/tokens/new

jarun:
Does all this mean no one other than me can make a release?

maximbaz:
From a packager's point of view it's not a problem at all, I can create a list of PGP keys that are allowed to sign sources. Let's just document in README whose keys can be used to sign the tarball. I can update README in this PR, should I only put your key for the moment?

jarun:
@maximbaz if the answer to my question is "yes" then it's a serious issue. We can't create tokens for everyone. The packages we provide are ad-hoc and we always suggest the distro repositories to have the latest versions available.

When you suggested the idea I was thinking of having checksum available along with the packages like this: https://ubuntu.com/download/desktop/thank-you?country=IN&version=18.04.3&architecture=amd64

to avoid any tampering with the auto-generated packages. However, now I see a tight coupling of the packaging using access token, private keys and so on. As soon as I do that there's additional imposed responsibility and I can be traced back from the packages.

It's becoming a overkill in the name of security and trust.

maximbaz:
The answer is "no", and token is only needed if you want to upload the signature to Github via a script, whoever doesn't want to use the script doesn't need to generate tokens and can just drag-n-drop the signature via Github UI.

However, from the looks of it, it is only you who are creating releases anyway, isn't it?

jarun:
> However, from the looks of it, it is only you who are creating releases anyway, isn't it?

For now yes. But may not be not true forever. And I decline to sign packages with my personal keys either.

maximbaz:
> And I decline to sign packages with my personal keys either.

Why so, if I may ask? 🙂 You sign commits with your personal key it seems?

To reiterate, there is no restriction how many people can sign packages, in principal everyone can do it, we just need to list in README whose signatures should be considered as _trusted_.

jarun:
> You sign commits with your personal key it seems?

The code is plaintext. You can read it and verify yourself.

Also let me put it this way: do I have the right to decide what I sign and what I don't with a digital signature that can be traced back to my personal system?

@zmwangx I need a second opinion on this. Is it worth adding this complexity to the distribution of a open source project package? The code is always available for audits. As per my understanding if we can provide the packages and their checksums from two different severs it's reduces the chance that both are compromised.

jarun:
Additional note: I am also concerned about legal. The LICENSE protects the code and devs by it's clause of no guarantee. However, the generated binary is not. If I sign the binary today and it malfunctions on a system, would I be legally vulnerable to the point that some stricter law (say, in the EU) holds me responsible for the harm done by the package?

zmwangx:
Speaking from my experience of being a package manager maintainer and a casual observer of a few more package managers (general purpose & language specific), I'd say the percentage of users who bother to check PGP signatures is probably <0.01% and those users are either building from source or installing from their trusted repo instead of downloading binaries from GitHub release pages. So I personally wouldn't bother...

maximbaz:
Signing the package does not make you any more accountable for the package quality than claimed by the license, don't worry, the license is what matters and it already claims that there is no liability and warranty for the provided software, so that's it 🙂 

The key does not even need to be linked to any personal identifying information of yours, you can create a brand new PGP key for a fake name; all we need is the key ID and put this ID in the README so that users can verify the signature against that key if they want to.

Deploying release artifacts to two different servers does not help users (nobody will bother to download the same archive from two servers and compare them). Hashsums help ensuring that download was not corrupted, but it doesn't protect against man-in-the-moddle attacks as there is no guarantee that hashsums were not compromised as well (especially if they are generated on the same CircleCI server).

PGP signature provides the same check against corrupted downloads, but additionally provides protection against man-in-the-middle attacks, because attackers cannot compromise the signature.

Basically having a signature file will help users be certain that their `nnn` was built from source code that one of the nnn authors confirmed was correct sources that were not tampered with.

Given that it will only take you extra 10 seconds to run one more command after every release (every 2 months?), I personally think it's a worthy improvement.


> I'd say the percentage of users who bother to check PGP signatures is probably <0.01% and those users are either building from source or installing from their trusted repo instead of downloading binaries from GitHub release pages. So I personally wouldn't bother...

What about all the users who install nnn via OS packages, like brew, or pacman? All of these tools support verifying gpg signatures, this impacts thousands of users.

zmwangx:
brew does not support verifying gpg signatures. As a former maintainer I'm pretty sure...

maximbaz:
Oh, really? Ok, sorry, I am mostly familiar with Arch 🙂 But many others do support them, I'm sure.

jarun:
> nobody will bother to download the same archive from two servers and compare them

not the same archive. the package and the checksum.

> but it doesn't protect against man-in-the-moddle attacks as there is no guarantee that hashsums were not compromised as well (especially if they are generated on the same CircleCI server)

We can have a script that downloads the files and generates the checksum locally.

> Basically having a signature file will help users be certain that their nnn was built from source code that one of the nnn authors confirmed was correct sources that were not tampered with.
> PGP signature provides the same check against corrupted downloads, but additionally provides protection against man-in-the-middle attacks, because attackers cannot compromise the signature.

Instead of the signature I would like to go by the checksum generated locally.

> What about all the users who install nnn via OS packages, like brew, or pacman? All of these tools support verifying gpg signatures, this impacts thousands of users.

Those packages installed by users through package managers are generated downstream from source. If the maintainer is willing (s)he can sign. I am concerned about the packages we are auto-generating in each release. And I think when you mentioned _distribution_, you specifically meant the source code from the eprspective of a maintainer.

jarun:
> Currently they trust package maintainers, but if you provide gpg signature, they will be trusting you instead.

Something I really don't care about given I am not involved in generating the package that goes to the user.

maximbaz:
I find it surprising, if you are not involved in generating packages that go to thousands of users, I would think you would care _very much_ that these thousands of users don't risk of getting a malicious software that pretends to be a `nnn` tool written by `jarun`?

zmwangx:
> Ok, sorry, I am mostly familiar with Arch 🙂 But many others do support them, I'm sure.

I did add a clarification "excluding automated checks by distro package managers". I think we're talking about signing artifacts on the GitHub release page, which users have to manually verify, no? (To be clear, I'm not convinced signatures from random upstream developers provide any tangible benefits not provided by a SHA-256 checksum; IIRC this was the Homebrew team's majority opinion too. I can also cite other debates, but this is mostly beside the point, so let's focus on manual verification.) Unless you want to create an AUR package pulling in a binary artifact...

Again I don't think users who download from the GitHub release page (mostly as a last resort, I suppose) instead of compiling from source would be manually verifying the signature.

> I find it surprising, if you are not involved in generating packages that go to thousands of users, I would think you would care very much that these thousands of users don't risk of getting a malicious software that pretends to be a nnn tool written by jarun?

First I would be genuinely surprised if thousands of users are downloading binary artifacts from the GitHub release page; secondly if https://github.com is MITM'ed for them, the fingerprint you publish on README is MITM'ed too, so it's kinda pointless. There's no IRL identity verification here, WoT doesn't work.

jarun:
> if you are not involved in generating packages that go to thousands of users

What goes from this repo is the source code. The package that a malicious maintainer is generating can include something that's malicious. Why should I take responsibility for that? There is no way for me to sign that package.

What I wanted to guard are the auto-generated non-source installation packages.

>  I would think you would care very much that these thousands of users don't risk of getting a malicious software that pretends to be a nnn tool written by jarun?

If you are testing, packaging and uploading a package in your package manager where's my involvement? What you are asking for here is a signed source archive. Let's say you have it. But still you can tamer and upload something that's not `nnn`, right?

jarun:
> Unless you want to create an AUR package pulling in a binary artifact...

Exactly my point. Thanks for wording it.

maximbaz:
> Unless you want to create an AUR package pulling in a binary artifact...

I have a package that downloads sources tarball from Github, and uses it to compile the binary. If I had a PGP signature for this sources tarball, I would be able to verify that `@jarun`, the author of the software, confirms that it is valid non-tampered sources archive. If I don't have the signature, I can be less sure that I'm building `nnn` from the "good" sources because they could have been tampered in transit.

> if https://github.com is MITM'ed for them, the fingerprint you publish on README is MITM'ed too, so it's kinda pointless.

For example Arch package will have this key fingerprint as part of PKGBUILD; if key fingerprint suddenly changes, it will not be blindly updated, someone (e.g. package maintainer) will go to Github and try to figure out why the key has changed. It's not pointless 🙂 

> But still you can tamer and upload something that's not nnn, right?

Correct, there are some measures against that too. But for the moment I'm assuming the package maintainer is not malicious person here, but this package maintainer can be man-in-the-middled, and a PGP signature will help protecting against this.

zmwangx:
To make sure I don't sound completely dismissive of package signing: I heard that Debian Developers actually meet in person to exchange keys when they onboard; now that's something. You're trusting them and their gatekeeping, and the signatures are proofs that they vetted the software (although you still have to make sure the key you initially downloaded from debian.org isn't compromised). Meanwhile, random upstreams publishing random signatures -- mostly meaningless.

> For example Arch package will have this key fingerprint as part of PKGBUILD; if key fingerprint suddenly changes, it will not be blindly updated, someone (e.g. package maintainer) will go to Github and try to figure out why the key has changed. It's not pointless 🙂

And that's somehow not already protected by a SHA-256 checksum?

zmwangx:
> Meanwhile, random upstreams publishing random signatures -- mostly meaningless.

To expand on that, this is like PGP support on PyPI. Python developers paying some attention to the packaging story should know how well that worked...

maximbaz:
> And that's somehow not already protected by a SHA-256 checksum?

Of course not, SHA-256 sum will be different for every release, there is no way to spot if a new release is compromised because someone guessed @jarun's Github password and created a new release with malicious nnn code, and submitted a new SHA-256 checksum for this malicious source code.

Now if we all knew that @jarun signs his releases with a specific PGP key with a certain ID, if someone guesses his Github password and uploads a new release with malicious code, they cannot provide PGP signature from the same key ID - at best they can create a new key, with different ID, and this will raise questions from everybody.

maximbaz:
I don't need to meet @jarun in person, whoever has access to @jarun's github account _right now_ can post here "my PGP key id is: XXXX", and I will take a note of it. If someone later comes and tells us that PGP key id changed, yes we will all be suspicious, and we will try to figure out what's going on. But if nobody will come and try to change key ID, and months go by, we will just keep trusting this exact PGP key, because it keeps signing the good code.

zmwangx:
I would hope downstream packagers do more diligence to vet changes than assuming @jarun released it, hence good.

jarun:
> because someone guessed @jarun's Github password

I think we are getting paranoid here. Just FYI, I use a 2-step and if that's compromised, we would have much more to worry about than `nnn`.

> they cannot provide PGP signature from the same key ID

Wrong! What if my laptop is also compromised?

maximbaz:
> I would hope downstream packagers do more diligence to vet changes than assuming @jarun released it, hence good.

What do you expect from them, that they read the entire codebase before packaging every release? 🙂

zmwangx:
I do try to read commit log when I update packages. I assume Debian Developers probably do that too, but that's just an assumption.

jarun:
> What do you expect from them, that they read the entire codebase before packaging every release?

I think we are doing rounds on who will take responsibility of a package gone bad - maintainer or packager.

zmwangx:
I think as an invited outsider I've said enough.

I do believe maintainer sustainability trumps security theater; a burned out maintainer isn't going to create any code, secure or compromised (maybe that's the goal?). People who are genuinely concerned can choose to use something else that meets their opsec standard.

maximbaz:
Yes, let's summarize 🙂 

The point is to find a good balance between security and usability. 

The point of proposal in this PR is that there is a very specific step that will take you 10 seconds at most, that you can run once per every new release (i.e. once every 2 months if I read history correctly), and it will provide us with all these benefits regardless of how good your github password is 😉 Yes it will not be a 100% proof (your point of compromising your laptop), but in my mind it will be improvement over the current state.

You decide now 👍

jarun:
For storing the token, please use an env variable: `NNN_SIG_UPLOAD_TOKEN`.

jarun:
I think we have discussed some good points. Let me think about it.

jarun:
OK, let's have this for your peace of mind. Please update the release checklist in wiki as well.

jarun:

maximbaz:
Awesome! Could you please run "make sign" to create and upload the signature for the last release? I'm already away from computer, but I want to test tomorrow that everything works as expected and thus is ready for the next release - let me know if something is not working or making you trouble on your end too!

jarun:
Sure! Give me a few minutes.

jarun:
I have uploaded the signature.

jarun:
Updated the release checklist as well. So we are good here.

maximbaz:
Good morning, thanks, I confirm that the signature is passing my checks and all is good, thanks for doing this and updating the wiki 👍 😉

jarun:
Awesome! We are all good then!

PR #418, opened 2019-12-21
Document how to verify PGP signature

maximbaz:
Follow-up to #417 

I'd like to have the list of PGP keys that can sign releases be listed in README. Right now this list contains only your key.


jarun:
Please add this to wiki under developer section.

PR #423, opened 2019-12-29
Make showsel print warning when nothing is selected

maximbaz:
Pressing `y` after starting `nnn` will print `0 selected` instead of silently failing.


jarun:
Thank you!

PR #428, opened 2020-01-11
Fix: BSD expr syntax error on empty $NNNLVL

JanneSalokoski:
MacOS uses BSD expr, instead of GNU expr. This causes a syntax error any time nnn is quit, if $NNNLVL is nothing. I added a test if $NNNLVL is defined to prevent the syntax error.


jarun:
How about the bash and zsh scripts?

jarun:
I've taken care of all the scripts in 3c69e6ccac5fece5089d816de7a36fc1eb3f74e6.

Thanks for the note!

PR #429, opened 2020-01-11
optionally append to log

ghistes:
Hi,

this is just for me to learn the mechanics of creating pull-requests.

It's just a few lines to optionally append to the log-file, rather than overwriting it every time (I perfer it that way).

I don't really care whether or not you want this, but please let me know if there is any problem with the pull-request as such.

### DIFF
--- a/src/dbg.h
+++ b/src/dbg.h
@@ -49,7 +49,11 @@ static int xprintf(int fd, const char *fmt, ...)
 
 static int enabledbg(void)
 {
-	FILE *fp = fopen("/tmp/nnndbg", "w");
+        /* overwrite or append to previous log via environment */
+        int keep_log = getenv("NNN_DEBUG_KEEP_LOG") ? 1 : 0;
+        char *log_file_mode = keep_log ? "a" : "w"; 
+
+	FILE *fp = fopen("/tmp/nnndbg", log_file_mode);
 
 	if (!fp) {
 		perror("dbg(1)");
@@ -68,7 +72,14 @@ static int enabledbg(void)
 		return -1;
 	}
 
-	return 0;
+        if(keep_log) {
+           /* seperate the runs with a blank line if the file is not empty */
+           if(lseek(DEBUG_FD, 0, SEEK_CUR)) {  
+             write(DEBUG_FD, "\n", 1);
+           }
+        }
+
+        return 0;
 }
 
 static void disabledbg(void)
--- a/src/dbg.h
+++ b/src/dbg.h
@@ -58,7 +58,7 @@ static int enabledbg(void)
 	if (!fp) {
 		perror("dbg(1)");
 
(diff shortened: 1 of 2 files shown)


src/dbg.h
```diff
@@ -49,12 +49,16 @@ static int xprintf(int fd, const char *fmt, ...)
 
 static int enabledbg(void)
 {
-	FILE *fp = fopen("/tmp/nnndbg", "w");
+        /* overwrite or append to previous log via environment */
+        int keep_log = getenv("NNN_DEBUG_KEEP_LOG") ? 1 : 0;
+        char *log_file_mode = keep_log ? "a" : "w";
```
KlzXS: Indentation is messed up. Please use tabs.

src/dbg.h
```diff
@@ -68,7 +72,14 @@ static int enabledbg(void)
...
 
-	return 0;
+        if(keep_log) {
+           /* seperate the runs with a blank line if the file is not empty */
+           if(lseek(DEBUG_FD, 0, SEEK_CUR)) {  
+             write(DEBUG_FD, "\n", 1);
+           }
+        }
+
+        return 0;
```
KlzXS: Please fix Indentation.

src/dbg.h
```diff
@@ -68,7 +72,14 @@ static int enabledbg(void)
 		return -1;
 	}
 
-	return 0;
+        if(keep_log) {
```
KlzXS: Use `if (keep_log) {`

src/dbg.h
```diff
@@ -68,7 +72,14 @@ static int enabledbg(void)
...
+           /* seperate the runs with a blank line if the file is not empty */
+           if(lseek(DEBUG_FD, 0, SEEK_CUR)) {
```
KlzXS: Use `if lseek(DEBUG_FD, 0, SEEK_CUR)) {`. Mind the space.

jarun:
How does appending help? To compare what happened in the last instance and current instance we will need to jump around. However, we can put our plugin `diffs` to good use ;). For a similar flow the logs would be similar and the diff would be very useful.

So i would suggest adding a timestamp to the log file. We can go to `/tmp` (or wherever the debug log is generated) and sort y time to get the logs from past runs in the correct order.

Also, if you do this and generate varying log names, when in debug mode, please dump the log filename to stderr (before entering curses mode) so we can just open the path in vim and start debugging.

jarun:
Also, please rebase your view. It is outdated and conflicts with current code. `nnn` code changes very fast. Always rebase before raising a PR.

jarun:
Also the timestamp should preferably in YYYYMMDDHHSS for simple sorting.

jarun:
@ghistes any progress on this?

jarun:
Closing as no response.

PR #431, opened 2020-01-12
Fast redraw on up/down key navigation

annagrram:


jarun:
I think there is some issue with long names. See the images long_names{1,2,3}.jpg I attached for 3 consecutive files. Note that in the current code we do some adjustments in ncols (below your code in redraw()) before sending it to printptr. I think you also need to do the same and rendering will be fine.

long_names1
long_names2
long_names3

jarun:
With xfce4-terminal and `nnn` in nav-as-you-type mode, if I open a new tab and close it, the line `str [keys /\], ic [key :]` shows twice. The reason is, when two tabs are opened, the tab pane is shown. When you close one, it is gone. The remaining tab is resized. So I think we need to do a full refresh when the screen is resized.

Hint: we handle the resize today in the code within `KEY_RESIZE` macro.

jarun:
Also, we may have to disable it when the filtering happens because only entries matching the filter stay in the list, rest are cleared. I think the best is to tell redraw() to enable or disable fast redraw by adding a param and disable it when we enter filterentries() and enable it when we exit filterentries().

Note: I didn't see the patch yet but I am hoping you did it ONLY for Up and Down (or k and j). No need to optimize anything else.

jarun:
Image of the last problem. Forgot in last comment.
nav_as_you_type

annagrram:
There's a reason I like backend and hate frontend; Rendering is the devil.

I'll take care of what you mentioned.

> Note: I didn't see the patch yet but I am hoping you did it ONLY for Up and Down (or k and j). No need to optimize anything else.

It happens to work for `g` and `G` as well, but I didn't optimize for `PgUp/PgDn/CtrlU/CtrlD`

jarun:
Also, just to make sure, we should show the `+` before selected as we do today.

And if filter is on, when you press up/down the filter is gone from the status bar and shows the file info. The filter info line (last but one, I added it last night) vanishes too I think.

BTW, the change is much more localized than I anticipated. Thanks for the nice little change with a huge performance boost.

jarun:
It actually works miracles when you are connected over ssh on a weak network.

jarun:
> It happens to work for g and G as well, but I didn't optimize for PgUp/PgDn/CtrlU/CtrlD

If there are more elements than the screen can render, doesn't g or G need to redraw new screenful of entries from the top or bottom respectively? I mean it may not be affecting only 2 lines. Just make sure. It would probably work when the first sorted element is at the top of the screen and you press g to jump to it (similarly for the last sorted element at the bottom of the screen).

annagrram:
Of course, `g` and `G` will do a full redraw when there are more entries than the screen can display, but will fast-draw otherwise.

> BTW, the change is much more localized than I anticipated. Thanks for the nice little change with a huge performance boost.

That's why it takes me some time to start implementing such things. I'm trying to study the relevant code and make as few and localised changes as I can get.
I'm glad it's useful :)

I'll take a look at the filtering soon.

annagrram:
> And if filter is on, when you press up/down the filter is gone from the status bar and shows the file info. The filter info line (last but one, I added it last night) vanishes too I think.

This also happens against `master`, unless I missed something.

The rest should be ok I think. Please give it a go.

jarun:
Yes, that is the expected behaviour. If it's that way, it's fine.

annagrram:
> Yes, that is the expected behaviour. If it's that way, it's fine.

It is.

src/nnn.c
```diff
@@ -337,6 +337,7 @@ static kv plug[PLUGIN_MAX];
 static uchar g_tmpfplen;
 static uchar blk_shift = BLK_SHIFT_512;
 static regex_t archive_re;
+static bool is_move_op = FALSE;
```
jarun: This can go to `g_states` so we truly use 1 bit.

src/nnn.c
```diff
@@ -4291,6 +4299,39 @@ static void redraw(char *path)
 	int i;
 	char *ptr = path;
 
+	// Fast redraw
```
jarun: Is it possible to have it as a function? If not without having 4 params, just leave as it is. But I think it may work with `ncols` and `path` passed.

src/nnn.c
```diff
@@ -4291,6 +4299,39 @@ static void redraw(char *path)
...
+	if (is_move_op && last_curscroll == curscroll) {
+		is_move_op = FALSE;
```
jarun: Do we need to clear `is_move_op` when it's set and `last_curscroll != curscroll`?

src/nnn.c
```diff
@@ -4291,6 +4299,39 @@ static void redraw(char *path)
...
+		printptr(&dents[cur], ncols, true);
+
+		/* Must reset e.g. no files in dir */
+		if (cfg.dircolor) {
+			attroff(COLOR_PAIR(cfg.curctx + 1) | A_BOLD);
+			cfg.dircolor = 0;
+		}
+
+		statusbar(path);
+		return;
```
jarun: Please append the following point in Readme -> Features -> Resource sensitive:

    Fast redraws to avoid full screen refresh

Or, however you want to word in in a short description. :)

src/nnn.c
```diff
@@ -4291,6 +4299,39 @@ static void redraw(char *path)
 	int i;
 	char *ptr = path;
 
+	// Fast redraw
+	if (is_move_op && last_curscroll == curscroll) {
+		is_move_op = FALSE;
+		/* Calculate the number of cols available to print entry name */
```
jarun: If this adjustment is the same as in the other place, please use an API. We should avoid duplicate code if easily avoidable.

jarun:
And I want it now: https://www.youtube.com/watch?v=hFDcoX7s6rE

I'll take care of the trivia. Thank you! :+1:

src/nnn.c
```diff
@@ -4279,6 +4285,8 @@ static void statusbar(char *path)
 			 ((g_states & STATE_RANGESEL) ? "*" : (nselected ? xitoa(nselected) : "")),
 			 sort, buf, get_lsperms(pent->mode), coolsize(pent->size), ptr, fname);
 	}
+
+	clrtoeol();
```
jarun: I think all out clrtoeol() can be replaced with addch('\n')... at least I don't see any issues during testing.

annagrram:
> I'll take care of the trivia. Thank you! +1

Sorry. I was a bit busy :)

Welcome.

jarun:
No problem! This works great when I connect to my Android from the laptop!

annagrram:
Do you want to take care of your inputs or me?
I've already made some of them before noticing you merged.

jarun:
Ahh, I have already pushed those in. Can you please review 917123c40099c734a7bd2e68896d17d30baf3f10?

PR #433, opened 2020-01-13
order toggle: show invalid key prompt

maximbaz:
For consistency with other places where we ask for second key (like plugin selection)

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5034,6 +5034,7 @@ nochange:
 					break;
 				default:
 					cfg.filtermode ? presel = FILTER : statusbar(path);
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -5034,6 +5034,7 @@ static void browse(char *ipath, const char *session)
 					break;
 				default:
 					cfg.filtermode ? presel = FILTER : statusbar(path);
+					printwait(messages[MSG_INVALID_KEY], &presel);
```
jarun: In that case no need to call `statusbar()`.

```
if (cfg.filtermode)
    presel = FILTER;
printwait(messages[MSG_INVALID_KEY], &presel);
```

Issue #422, opened 2019-12-29
Sane keybinds and switches


jarun:
I spent a lot of time today on having some sanity in our keybinds & program switches. Came up with commit 146da5911d8273ce302fb70ddf443cb577f9def4.

Keybinds: https://github.com/jarun/nnn/wiki/Usage#keyboard-and-mouse

Options: https://github.com/jarun/nnn/wiki/Usage#command-line-options

I need all of you to review this and suggest any further changes or discuss if you have anything in mind. From the next release these keys will be fixed.

annagrram:
I have some job-interview tasks to do. So I'll be a bit busy in the next couple of days.

jarun:
No problem! We have time for this.

All the best!!!

maximbaz:
One idea: if `/` is filter, it feels very intuitive that `^/` would be some sort of "extended filter" mode, which is "nav-as-you-type" in `nnn`. I would therefore suggest to keep `b` for bookmarks and move `^/` to "nav-as-you-type".

maximbaz:
I would recommend changing `N  Context N` to `1 2 3 4  Context`, because (unless I'm mistaken?) it is hardcoded that we can have only 4 contexts, and the current help message misleads you into thinking that you must press `N`, while in fact you must press `1` or `2` or `3` or `4`.

jarun:
> I would therefore suggest to keep b for bookmarks and move ^/ to "nav-as-you-type"

I did that in an intermediate check-in. There are several points against it:

- I need to provide a ctrl shortcut for people who use the nav-as-you-type mode all the time to pick bookmarks. The 2 keys in `^B` are too far from each other. it's too I was having difficulties myself.
- Users are already used to these keys.
- `^N` goes well with _Nav-as-you-type toggle_.
- The option to start in _nav-as-you-type` is also `-n`.

> and the current help message misleads you into thinking that you must press N

We can have `1-4 Context 1-4`

maximbaz:
I was thinking `b ^B` for bookmarks and `^N ^/` for nav-as-you-type, but I see your reasoning for `^B` being difficult to press... 🤔  OK I don't have a better suggestion at the moment.

> We can have 1-4 Context 1-4

👍

jarun:
> We can have 1-4 Context 1-4

Commit c1f168c7cbbf87a8f1acc4ba88e14d68da556392

maximbaz:
[...]

PR #436, opened 2020-01-13
Propose new shortcuts

maximbaz:
My proposal for our discussion in #422 

My objectives:

1. Have Ctrl shortcuts so that nav-as-you-type is as powerful as normal mode.
1. Have Ctrl shortcuts use the same letter as non-Ctrl shortcuts, so that users need to remember less shortcuts (e.g. if shortcut is `p`, the Ctrl shortcut should be `^P`).

With this change:

- `^P, ^V, ^X` are set as twins for `p, v, x`
- `w ^W` is assigned to "Copy/move sel as" as you suggested
- `^O` becomes "Sort toggles" instead of `^T` (rule 2, `o ^O`)
- `^M` becomes "Mark range/clear" instead `^K` (rule 2, `m ^M`)
- `^;` becomes "Plugin" instead `^P` (rule 2, `; ^;`)
- "Delete entry" becomes `^K` (mnemonic: **k**ill file, plus `^K` is quite common for delete line in other apps)
- "Open with..." becomes `^T` (stretched not-really mnemonic: send file **T**O another app) 

I also noticed one more shortcut that requires Shift key: `! Shell`. Although it already has a Ctrl mapping, we could still set an easier shortcut to it? I propose `t` (mnemonic: terminal).


jarun:
> ^M becomes "Mark range/clear" instead ^K (rule 2, m ^M)

`^M` is `Enter`. :dancing_men: It's OK. People will remember because of the ^J-^K combo.

> Open with..." becomes ^T (stretched not-really mnemonic: send file TO another app)

We're losing it. ;) Let `^O` be as it is. It's standard too.
Maybe we should have t and ^T for sort toggles.

> I also noticed one more shortcut that requires Shift key: ! Shell

Please leave the shell shortcut `!` alone.

> ^; becomes "Plugin" instead ^P (rule 2, ; ^;)

I didn't know it worked! This is nice!

Now all we need to find is something for Delete entry.

jarun:
We have the following as per above:

```
 NAVIGATION
         Up k  Up                PgUp ^U  Scroll up
         Dn j  Down              PgDn ^D  Scroll down
         Lt h  Parent            ~ ` @ -  HOME, /, start, last
     Ret Rt l  Open                    '  First file
         g ^A  Top                     .  Toggle hidden
         G ^E  End                     0  Lock terminal
         b ^/  Bookmark key            ,  Pin CWD
          1-4  Context 1-4       (Sh)Tab  Cycle context
            /  Filter                 ^N  Nav-as-you-type toggle
          Esc  Exit prompt            ^L  Redraw/clear prompt
            ?  Help, conf             ^G  QuitCD
         Q ^Q  Quit                    q  Quit context
 FILES
         o ^O  Open with...            n  Create new/link
         f ^F  File details            d  Detail view toggle
           ^R  Rename/dup              r  Batch rename
            z  Archive entry           *  Toggle exe
     Space ^J  (Un)select           m ^K  Mark range/clear
         p ^P  Copy sel here           a  Select all
         v ^V  Move sel here        w ^W  Copy/move sel as
         x ^X  Delete sel             ^K  Delete entry
         t ^T  Sort toggles            e  Edit sel
 MISC
         ; ^;  Plugin                  =  Launch app
         ! ^]  Shell                   ]  Cmd prompt
            s  Manage session          u  Unmount
            c  Connect remote
```

jarun:
We'll do this - x ^X for delete both cur and sel:

- if no sel, del cur (we already show correct msg)
- if no cur, del sel (we already show correct msg)
- if both present, confirm sel or cur and proceed

jarun:
Archive entry should be just Archive. It handles sel too!

jarun:
Similarly with above proposal for Delete, it would be just Delete.

maximbaz:
it sounds good, feel free to take over and do this in master if you have time, because I'll be able to get back to this only in the evening

jarun:
No probs. I'll work on it. Please review when you are back.

jarun:
I have made the changes. tehre's a small problem. In filterentries() if I press `^;` it is captured as `;`. Can you please check?

jarun:
However, `getch()` in nextsel() works fine. Does `get_wch()` have any issue?

jarun:
There is no `^;`... it comes as `;`.

jarun:
I think we can settle with:

D - file details
f ^F - Fire plugin

PR #438, opened 2020-01-14
Catch NULL from malloc()

KlzXS:


src/nnn.c
```diff
@@ -2628,8 +2633,13 @@ static char *get_kv_val(kv *kvarr, char *buf, int key, uchar max, bool path)
 				ssize_t len = strlen(home);
 				ssize_t loclen = strlen(kvarr[r].val);
 
-				if (!buf)
+				if (!buf) {
 					buf = (char *)malloc(len + loclen);
+					if (!buf) {
+						DPRINTF_S("malloc failed!");
```
jarun: Please log `strerror(errno)`.

src/nnn.c
```diff
@@ -1676,6 +1676,11 @@ static void get_archive_cmd(char *cmd, char *archive)
 static void archive_selection(const char *cmd, const char *archive, const char *curpath)
 {
 	char *buf = (char *)malloc(CMD_LEN_MAX * sizeof(char));
+	if (!buf) {
```
jarun: If we look at what we are trying to store in buf, we need to allocate:

strlen(cmd) + strlen(archive) + strlen(curpath) + strlen(g_selpath);

src/nnn.c
```diff
@@ -1676,6 +1676,11 @@ static void get_archive_cmd(char *cmd, char *archive)
...
+		DPRINTF_S("malloc failed!");
+		printmsg(messages[MSG_FAILED]);
```
jarun: please use `printwarn(NULL)`;

src/nnn.c
```diff
@@ -1676,6 +1676,11 @@ static void get_archive_cmd(char *cmd, char *archive)
 static void archive_selection(const char *cmd, const char *archive, const char *curpath)
 {
 	char *buf = (char *)malloc(CMD_LEN_MAX * sizeof(char));
+	if (!buf) {
+		DPRINTF_S("malloc failed!");
```
jarun: please log `strerror(errno)`.

src/nnn.c
KlzXS: You want me so allocate this amount instead?

jarun: Yes,` CMD_LEN_MAX` is insufficient.

KlzXS: You forgot to count the length of the string we snprintf() into, that's an aditional ~70.

jarun:
Thank you!

PR #440, opened 2020-01-15
Release improvements, take 2

maximbaz:
The error `fatal: Not a git repository (or any of the parent directories): .git` was caused because of `rm -rf .git` command in the beginning, I didn't notice it before 🙂 

It's a bit tricky to test CircleCI, but hopefully the rest should work just fine.

Most importantly, I saw you uploaded the signature as part of the release, so thank you very much for that - I'm about to push the release to Arch Linux now 👍


jarun:
No problem! Thank you!

PR #441, opened 2020-01-16
Fixed calculation of capacity on OpenBSD

horrad:
nnn reports on OpenBSD the wrong capacity of volumes.

On OpenBSD we use fragments, so that f_bsize and f_frsize differ. 

With the attached patch nnn reports the correct size on Linux as well as OpenBSD. (The manpage for statvfs on Linux also states that f_frsize ist correct)

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3285,7 +3285,7 @@ static size_t get_fs_info(const char *path, bool type)
 		return 0;
 
 	if (type == CAPACITY)
-		return svb.f_blocks << ffs((int)(svb.f_bsize >> 1));
+		return svb.f_blocks << ffs((int)(svb.f_frsize >> 1));
 
 	return svb.f_bavail << ffs((int)(svb.f_frsize >> 1));
 }


jarun:
Thank you!

jarun:
@horrad I got a crash when trying to du at / with this change (it was trying to calculate /usr). The second time it hung (trying to calculate /proc).

I could not repro this on reboot. I have kept the change for now but we need to investigate this.

jarun:
> The manpage for statvfs on Linux also states that f_frsize ist correct

I don't see this here: http://man7.org/linux/man-pages/man3/statvfs.3.html

jarun:
Also it seems, we don't need to bother about it on Linux: https://stackoverflow.com/questions/54823541/what-do-f-bsize-and-f-frsize-in-struct-statvfs-stand-for

I will revert the patch. I should have taken a deeper look into it. Please re-raise the PR specifically for BSD (only for the flavours where this applies) after reasonable testing.

jarun:
I missed the reference to `f_frsize` in the man page earlier. I was looking into the details section when it's actually in the inline comment. I will re-apply this.

jarun:
Back in commit 679374680fa8b961a30e5af47b779153f32734ad. Thanks for noting this and the PR!

PR #442, opened 2020-01-16
Quit with an error code

maximbaz:
I was thinking `^H` could be mnemonic for "halt"...?

Feedback welcome, especially if you think this shouldn't be a new shortcut but some other approach.

```
 NAVIGATION
         Up k  Up                PgUp ^U  Scroll up
         Dn j  Down              PgDn ^D  Scroll down
         Lt h  Parent            ~ ` @ -  HOME, /, start, last
     Ret Rt l  Open                    '  First file
         g ^A  Top                     .  Toggle hidden
         G ^E  End                     0  Lock terminal
         b ^/  Bookmark key            ,  Pin CWD
          1-4  Context 1-4        (B)Tab  Cycle context
            /  Filter                 ^N  Nav-as-you-type toggle
          Esc  Exit prompt            ^L  Redraw/clear prompt
            ?  Help, conf             ^G  QuitCD
            q  Quit context           ^H  Quit with err
         Q ^Q  Quit
 FILES
         o ^O  Open with...            n  Create new/link
...
```


jarun:
Please use `Q`. Most people won't care about exit code.

maximbaz:
I am okay with that, `Q` and `^Q` still mean exit so we don't "break the contract", but if you care about exit code, use `^Q` for success and `Q` for error.

jarun:
Right!

src/nnn.c
```diff
@@ -6184,7 +6186,7 @@ int main(int argc, char *argv[])
 	if (!initcurses(&mask))
 		return _FAILURE;
 
-	browse(initpath, session);
+	bool success = browse(initpath, session);
```
jarun: We don't need to store the return value in `opt` in line 6198. Remove that and use opt for this so we don't need to have an extra var.

maximbaz: To clarify, do you mean changing this to `opt = browse(initpath, session);` ?

`opt` might get overridden later in line 6200 though...

jarun: That's why I meant... to store the return value. We can lose the curly brackets too. ;)

```
                if (selbufpos)
                        if (seltofile(1, NULL) != (int)(selbufpos))
                                xerror();
```

maximbaz: Gotcha 👍 Well the `if` can be simplified even further if you like, have a look at the last commit :)

src/nnn.c
```diff
@@ -6193,16 +6195,11 @@ int main(int argc, char *argv[])
...
-		if (selbufpos) {
-			opt = seltofile(1, NULL);
-			if (opt != (int)(selbufpos))
-				xerror();
-		}
-	} else if (cfg.picker) {
-		if (selbufpos)
-			writesel(pselbuf, selbufpos - 1);
-	} else if (!cfg.picker && g_selpath)
+	if (cfg.pickraw && selbufpos && seltofile(1, NULL) != (size_t)(selbufpos))
```
jarun: Combining won't work. The modes are to be handled exclusively.

if cfg.pickraw and !selbufpos, we don't want to go into cfg.picker handling.

maximbaz: Oh, right, I see the flaw in my logic.

Fixed now - there are still some braces left, if I remove them compiler prints warnings "suggest explicit braces to avoid ambiguous else".

src/nnn.c
```diff
@@ -6194,15 +6196,14 @@ int main(int argc, char *argv[])
 #endif
 
 	if (cfg.pickraw) {
-		if (selbufpos) {
-			opt = seltofile(1, NULL);
-			if (opt != (int)(selbufpos))
-				xerror();
-		}
-	} else if (cfg.picker) {
+		if (selbufpos && seltofile(1, NULL) != (size_t)(selbufpos))
```
jarun: Please keep the inequality check within ().

src/nnn.c
```diff
@@ -6194,15 +6196,14 @@ int main(int argc, char *argv[])
...
+			xerror();
+	}
```
jarun: } else if

src/nnn.c
```diff
@@ -6194,15 +6196,14 @@ int main(int argc, char *argv[])
...
+	else if (cfg.picker) {
 		if (selbufpos)
 			writesel(pselbuf, selbufpos - 1);
-	} else if (!cfg.picker && g_selpath)
+	}
```
jarun: } else if

src/nnn.c
```diff
@@ -3649,7 +3649,8 @@ static void show_help(const char *path)
 		  "c/  Filter%-17c^N  Nav-as-you-type toggle\n"
 		"aEsc  Exit prompt%-12c^L  Redraw/clear prompt\n"
 		  "c?  Help, conf%-13c^G  QuitCD\n"
-	       "9Q ^Q  Quit%-20cq  Quit context\n"
+	       "cq  Quit context%-12cQ  Quit with err\n"
```
jarun: I think you used tabs. It is tab followed by spaces (or 2 tabs if it aligns). TAB is 8 spaces.

maximbaz: Oh, it's 8, I see now how it's supposed to align! But then there are a few more lines that need fixing, I fixed them all - here's how it looks with the latest commit:

image

maximbaz: Is that what you expected? Sorry, until now I didnt know how you format, I was just copy-pasting the previous line and changing the inner parts of the string only...

jarun: Yes, this looks good!

jarun:
Thank you!

PR #443, opened 2020-01-17
List of files as input

KlzXS:
Usage:
Either pipe command output into `nnn` or redirect a file to `stdin`. The paths must be separated by `\0`.
E.g. `find -size +1M | nnn` or `nnn < list_of_files.txt`

Up to 65,536 paths can be read at a time, and no more than 256M will be read from `stdin`. Reads are done in chunks of 512K.

A tmp directory is created and symlinks to the given paths are created with paths being as compressed as can be i.e. no directories are created that only contain another directory. This directory is removed on exit. Currently the symlinks are not followed for operations like remove and such. I'll do that tomorrow.

I didn't have much quality time to spend on time so it took a _bit_ longer that I would've liked but I haven't forgotten about it. Hopefully the code isn't that horrid.

This is really WIP, and I'm going to stop looking for bugs as I keep finding more and more as I get more and more tired.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1233,7 +1233,10 @@ static void endselection(void)
 	}
 
 	seltofile(fd, NULL);
-	close(fd);
+	if (close(fd)) {
+		DPRINTF_S(strerror(errno));
+		return;
+	}
 
 	snprintf(buf, sizeof(buf), replaceprefixcmd, g_listpath, g_prefixpath, g_tmpfpath);
 	spawn(utils[UTIL_SH_EXEC], buf, NULL, NULL, F_CLI);
@@ -1241,15 +1244,20 @@ static void endselection(void)
 	fd = open(g_tmpfpath, O_RDONLY);
 	if (fd == -1) {
 		DPRINTF_S(strerror(errno));
-		unlink(g_tmpfpath);
+		if (unlink(g_tmpfpath))
+			DPRINTF_S(strerror(errno));
 		return;
 	}
 
 	count = read(fd, pselbuf, selbuflen);
-	close(fd);
-	unlink(g_tmpfpath);
-
 	if (count < 0) {
+		DPRINTF_S(strerror(errno));
+		if (close(fd) || unlink(g_tmpfpath))
+			DPRINTF_S(strerror(errno));
+		return;
+	}
+
+	if (close(fd) || unlink(g_tmpfpath)) {
 		DPRINTF_S(strerror(errno));
 		return;
 	}
@@ -1290,7 +1298,10 @@ static int editselection(void)
 	}
 
 	seltofile(fd, NULL);
-	close(fd);
+	if (close(fd)) {
+		DPRINTF_S(strerror(errno));
+		return -1;
+	}
 
 	/* Save the last modification time */
 	if (stat(g_tmpfpath, &sb)) {
@@ -1324,16 +1335,20 @@ static int editselection(void)
 	}
 
 	count = read(fd, pselbuf, selbuflen);
-	close(fd);
-	unlink(g_tmpfpath);
-
-	if (!count) {
-		ret = 1;
+	if (count < 0) {
+		DPRINTF_S(strerror(errno));
+		if (close(fd) || unlink(g_tmpfpath))
+			DPRINTF_S(strerror(errno));
 		goto emptyedit;
 	}
 
-	if (count < 0) {
+	if (close(fd) || unlink(g_tmpfpath)) {
 		DPRINTF_S(strerror(errno));
+		goto emptyedit;
+	}
+
+	if (!count) {
+		ret = 1;
 		goto emptyedit;
 	}
 
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -940,11 +940,101 @@ static void *xmemrchr(uchar *s, uchar ch, size_t n)
 	return NULL;
 }
 
+static char *common_prefix(const char *s, char *prefix)
+{
+	if (!s || !prefix)
+		return NULL;
+
+	ulong x, y;
+	size_t i, j;
+	size_t len_s = strlen(s), len_prefix = strlen(prefix);
+	size_t len = MIN(len_s, len_prefix);
+	char *tmp;
+
+	for (i = 0; i < len; i += sizeof(ulong)) {
+		if (len - i >= sizeof(ulong)) {
+			x = *((ulong *)(s + i));
+			y = *((ulong *)(prefix + i));
+			if (!(x ^ y))
+				continue;
+		}
+
+		for (j = 0; j < MIN(sizeof(ulong), len - i); ++j) {
+			if (s[i + j] != prefix[i + j]){
+				tmp = xmemrchr((uchar *)prefix, '/', i + j);
+				if (!tmp)
+					return NULL;
+
+				*(tmp != prefix ? tmp : tmp + 1) = '\0';
+
+				return prefix;
+			}
+		}
+	}
+
+	if (len_s < len_prefix || (len_s > len_prefix && s[len_prefix] != '/')) {
+		tmp = xmemrchr((uchar*)prefix, '/', len);
+		if (!tmp)
+			return NULL;
+
+		*(tmp != prefix ? tmp : tmp + 1) = '\0';
+	}
+
+	return prefix;
+}
+
 static char *xbasename(char *path)
 {
 	char *base = xmemrchr((uchar *)path, '/', strlen(path)); // NOLINT
 
 	return base ? base + 1 : path;
+}
+
+static char *xrealpath(const char *path, char *resolved_path, const char *cwd)
+{
+	if (!path || !resolved_path || !cwd)
+		return NULL;
+
+	size_t dst_size = 0, src_size = strlen(path);
+	const char *src, *next;
+	char *dst;
+
+
+	/* Turn relative paths into absolute */
+	if (path[0] != '/')
+		dst_size = xstrlcpy(resolved_path, cwd, strlen(cwd) + 1) - 1;
+	else
+		resolved_path[0] = '\0';
+
+
+	src = path;
+	dst = resolved_path + dst_size;
+	for (next = NULL; next != path + src_size;) {
+		next = strchr(src, '/');
+		if (!next)
+			next = path+src_size;
+
+		if (next - src == 2 && src[0] == '.' && src[1] == '.') {
+			if (dst - resolved_path) {
+				dst = xmemrchr((uchar*)resolved_path, '/', dst-resolved_path);
+				*dst = '\0';
+			}
+		} else if (next - src == 1 && src[0] == '.'); // NOP
+		else if (next - src){
+			*(dst++) = '/';
+			xstrlcpy(dst, src, next - src + 1);
+			dst += next - src;
+		}
+
+		src = next + 1;
+	}
+
+	if (*resolved_path == '\0') {
+		resolved_path[0] = '/';
+		resolved_path[1] = '\0';
+	}
+
+	return resolved_path;
 }
 
 static int create_tmp_file(void)
@@ -5861,6 +5951,146 @@ nochange:
 	}
 }
 
+static char *make_tmp_tree(char **paths, size_t entries, const char *prefix)
+{
+	char *slash;
+	size_t i, len = strlen(prefix);
+	char *tmpdir = malloc(sizeof(char) * (PATH_MAX + 15));
+
+	if (!tmpdir) {
+		DPRINTF_S(strerror(errno));
+		return NULL;
+	}
+
+	xstrlcpy(tmpdir, "/tmp/nnnXXXXXX", 15);
+
+	if (!mkdtemp(tmpdir)) {
+		free(tmpdir);
+
+		DPRINTF_S(strerror(errno));
+		return NULL;
+	}
+
+	for(i = 0; i < entries; ++i){
+		xstrlcpy(tmpdir + 14, paths[i] + len, strlen(paths[i]) + 1);
+
+		slash = xmemrchr((uchar *)tmpdir, '/', strlen(paths[i]) + 15);
+		*slash = '\0';
+
+		xmktree(tmpdir, TRUE);
+
+		*slash = '/';
+		if (symlink(paths[i], tmpdir))
+			DPRINTF_S(strerror(errno));
+	}
+
+	tmpdir[14] = '\0';
+	return tmpdir;
+}
+
+static char *load_input()
+{
+	/* 512 KiB chunk size */
+	ssize_t i, chunk_count = 1, chunk = 512 * 1024, entries = 0;
+	char *input = malloc(sizeof(char) * chunk), *tmpdir;
+	char prefix[PATH_MAX], cwd[PATH_MAX], *next, *prev, **paths;
+	ssize_t input_read, total_read = 0;
+
+	if (!input) {
+		DPRINTF_S(strerror(errno));
+		return NULL;
+	}
+
+	total_read = input_read = read(STDIN_FILENO, input, chunk);
+	while (input_read == chunk && chunk_count < 512) {
+		if (!xrealloc(input, (++chunk_count) * chunk))
+			goto malloc_1;
+
+		input_read = read(STDIN_FILENO, input + (chunk_count - 1) * chunk, chunk);
+		if (input_read < 0) {
+			DPRINTF_S(strerror(errno));
+			goto malloc_1;
+		}
+
+		total_read += input_read;
+	}
+
+	input[total_read] = '\0';
+
+	/* 256MiB already read, enough for 2^16 paths */
+	if (input_read == chunk && chunk_count == 512)
+		goto malloc_1;
+
+	prev = input;
+	while (prev < input + total_read + 1) {
+		next = memchr(prev, '\0', total_read - (prev - input) + 1) + 1;
+
+		if (next - prev == 1) {
+			prev = next;
+			continue;
+		}
+
+		prev = next;
+
+		/* No NULL terminator at the end or limit exceeded */
+		if (++entries > (1 << 16) || !next)
+			goto malloc_1;
+	}
+
+	// allocate the space for all the strings 
+	paths = malloc(sizeof(char *) * entries);
+	if (!paths) {
+		DPRINTF_S(strerror(errno));
+		goto malloc_1;
+	}
+
+	getcwd(cwd, PATH_MAX);
+
+	for (i = 0; i < entries; ++i) {
+		paths[i] = malloc(sizeof(char) * PATH_MAX);
+		if (!paths[i]) {
+			DPRINTF_S(strerror(errno));
+
+			for(--i; i >= 0; --i)
+				free(paths[i]);
+
+			goto malloc_2;
+		}
+	}
+
+	for (i = 0, prev = input; i < entries; ++i) {
+		if (!xrealpath(prev, paths[i], cwd))
+			goto malloc_3;
+
+		next = memchr(prev, '\0', total_read - (prev - input) + 1) + 1;
+		if (next - prev == 1) {
+			prev = next;
+			continue;
+		}
+
+		prev = next;
+	}
+
+	xstrlcpy(prefix, paths[0], strlen(paths[0]) + 1);
+	for (i = 1; i < entries; ++i) {
+		if (!common_prefix(paths[i], prefix))
+			goto malloc_3;
+	}
+
+	tmpdir = make_tmp_tree(paths, entries, prefix);
+	if (tmpdir)
+		return tmpdir;
+
+malloc_3:
+	for(i = 0; i < entries; --i)
+		free(paths[i]);
+malloc_2:
+	free(paths);
+malloc_1:
+	free(input);
+	return NULL;
+}
+
 static void check_key_collision(void)
 {
 	int key;
@@ -6149,8 +6379,19 @@ int main(int argc, char *argv[])
 	}
 
 	/* Confirm we are in a terminal */
-	if (!cfg.picker && !(isatty(STDIN_FILENO) && isatty(STDOUT_FILENO)))
+	if (!cfg.picker && !isatty(STDOUT_FILENO)) {
 		exit(1);
+	}
+
+	/* Now we are in path list mode */
+	if (!cfg.picker && !isatty(STDIN_FILENO)) {
+		initpath = load_input();
+		if (!initpath)
+			return _FAILURE;
+
+		/* We return to tty */
+		dup2(STDOUT_FILENO, STDIN_FILENO);
+	}
 
 #ifdef DBGMODE
 	enabledbg();
@@ -6185,7 +6426,8 @@ int main(int argc, char *argv[])
 		return _FAILURE;
 	}
 
-	if (arg) { /* Open a bookmark directly */
+	if (initpath); // NOP
+	else if (arg) { /* Open a bookmark directly */
 		if (!arg[1]) /* Bookmarks keys are single char */
 			initpath = get_kv_val(bookmark, NULL, *arg, BM_MAX, TRUE);
 
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -871,7 +871,7 @@ static size_t xstrlcpy(char *dest, const char *src, size_t n)
 
 	/*
 	 * To enable -O3 ensure src and dest are 16-byte aligned
-	 * More info: http://www.felixcloutier.com/x86/MOVDQA.html
+	 * More info: https://www.felixcloutier.com/x86/MOVDQA:VMOVDQA32:VMOVDQA64
 	 */
 	if ((n >= LONG_SIZE) && (((ulong)src & _ALIGNMENT_MASK) == 0 &&
 	    ((ulong)dest & _ALIGNMENT_MASK) == 0)) {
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -905,13 +907,103 @@ static void *xmemrchr(uchar *s, uchar ch, size_t n)
...
+
+	ulong x, y;
+	size_t i, j;
+	size_t len_s = strlen(s), len_prefix = strlen(prefix);
+	size_t len = MIN(len_s, len_prefix);
+	char *tmp;
+
+	for (i = 0; i < len; i += sizeof(ulong)) {
+		if (len - i >= sizeof(ulong)) {
+			x = *((ulong *)(s + i));
```
jarun: Please take to tmp `ulong *` pointers to get rid of this. It's compile time but still.

src/nnn.c
```diff
@@ -905,13 +907,103 @@ static void *xmemrchr(uchar *s, uchar ch, size_t n)
...
+	if (!s || !prefix)
+		return NULL;
+
+	ulong x, y;
+	size_t i, j;
+	size_t len_s = strlen(s), len_prefix = strlen(prefix);
+	size_t len = MIN(len_s, len_prefix);
+	char *tmp;
+
+	for (i = 0; i < len; i += sizeof(ulong)) {
```
jarun: You can't do `i += sizeof(ulnog)`. You may cross boundaries. Follow `xstrlcpy()`... the way we calculate the number of `sizeof(ulong)` blocks and check for alignment to avoid crashes due to O3 (SIMD) optimization etc.

jarun:
> The paths must be separated by \0.

How about newline?

KlzXS:
> How about newline?

Now it's my turn to say, what about filenames with newline in them. :)

src/nnn.c
KlzXS: I see. I'll rework it then.

jarun:
> Now it's my turn to say, what about filenames with newline in them. :)

I forgot find has `-print0`.

jarun:
BTW, please confirm how `fzy` works when a file with newline is fed to it. It works with newline separated input so let's see.

jarun:
Both break. So much for popular software.

KlzXS:
Please see if operations made on the symlinks propagate to the targets. 

That is if you move one of the symlinks somewhere the target should get moved not the symlink. Also moving and renaming the target won't update the symlink, do we need that, I think we can do without that?

src/nnn.c
```diff
@@ -354,6 +354,12 @@ static char g_buf[CMD_LEN_MAX] __attribute__ ((aligned));
 /* Buffer to store tmp file path to show selection, file stats and help */
 static char g_tmpfpath[TMP_LEN_MAX] __attribute__ ((aligned));
 
+/* Buffer to store path to tmp directory for listing files from input */
```
jarun: Please allocate both of these variables dynamically. We do not need it during normal run.

Also I would suggest not using a global for these.

src/nnn.c
```diff
@@ -1058,8 +1190,51 @@ static void updateselbuf(const char *path, char *newpath)
...
+	ssize_t count;
+	char buf[sizeof(replaceprefixcmd) + PATH_MAX + (TMP_LEN_MAX << 1)];
+
 	if (cfg.selmode)
 		cfg.selmode = 0;
+
+	if (!*g_listpath)
+		return;
+
+	fd = create_tmp_file();
```
jarun: I think we should have a flag that says we are in list files mode and then we should do this processing. Otherwise we should just set `cfg.selmode` to 0 and return like we do today.

Keep these paths as detached as possible.

src/nnn.c
```diff
@@ -5078,7 +5253,7 @@ static bool browse(char *ipath, const char *session)
 		case SEL_STATS: // fallthrough
 		case SEL_CHMODX:
 			if (ndents) {
-				mkpath(path, dents[cur].name, newpath);
+				mkpath((xstrcmp(path, g_listpath) ? path : g_prefixpath), dents[cur].name, newpath);
```
jarun: Please use a flag to indicate we are in list files path.

jarun: These apply to all places we have touched mkpath.

src/nnn.c
```diff
@@ -5980,15 +6313,31 @@ int main(int argc, char *argv[])
...
-	if (!cfg.picker && !(isatty(STDIN_FILENO) && isatty(STDOUT_FILENO)))
-		exit(1);
-
 #ifdef DBGMODE
 	enabledbg();
 	DPRINTF_S(VERSION);
 #endif
 
+	/* Confirm we are in a terminal */
+	if (!cfg.picker && !isatty(STDOUT_FILENO)) {
```
jarun: We can put both the checks together within `if (!cfg.picker)`.

jarun: I mean `!isatty(STDOUT_FILENO` and then `!isatty(STDIN_FILENO)`

src/nnn.c
KlzXS: > Also I would suggest not using a global for these.

I need it to be global or at least accessible in `main()`, `browse()` and `endselection()`.

src/nnn.c
KlzXS: This already functions like that, if we are in "list mode", let's call it that, we do all the processing beneath the `if (!*g_listpath)` otherwise we return just setting `cfg.selmode`.

The flag won't work I need both the tmp dir path and the common prefix thingy, they can also double as the flag, if they are not `NULL` we are in list mode.

src/nnn.c
KlzXS: Even if we had the flag I would only change the path if we are in the tmp dir, I don't want to follow symlinks in all directories.

src/nnn.c
KlzXS: Ok.

src/nnn.c
jarun: Have the 2 pointers as global. When you enter list mode, malloc in main().

src/nnn.c
jarun: > let's call it that, we do all the processing beneath the if (!*g_listpath) otherwise we return just setting cfg.selmode

Right!

src/nnn.c
jarun: If g_listpath works as the flag, please check for NULL and then do the xstrcmp().

jarun:
@maximbaz can you please test this?

jarun:
@KlzXS if I run `find . -type f -print0 | ./nnn` I don't see the file `nnn` but `./nnn` is in the output.

Also block `du` and `au` in this mode.

maximbaz:
`\0`-separated values make sense 👍 

I like the approach with symlinks.

It seems passing a single file does not work:

```
$ printf 'LICENSE' | ./nnn              # no files
$ printf 'LICENSE\0' | ./nnn            # no files
$ printf 'LICENSE\0CHANGELOG' | ./nnn   # two files as expected
```

Not sure if I'm thinking about this is too early, but some path validation would be nice to have (silently ignore non-existing paths or abort with an error if at least one path is invalid?). For example, `$ printf 'LICENSE\0~/notapath' | ./nnn` will coredump if you try to enter `~` directory 🙂

KlzXS:
> It seems passing a single file does not work:

I'll look into why that doesn't work.

> For example, $ printf 'LICENSE\0~/notapath' | ./nnn will coredump if you try to enter ~ directory 

For me it crashes if I try to enter notpath. I don't know why. I'll put in a validity check. I think silently ignoring is good enough with a message at the bottom saying some of the paths are invalid, this would maybe lead to an empty dir but I think that's okay.

jarun:
Functionality-wise looks good to me. We need to document what works and what doesn't very clearly in wiki.

@maximbaz can you try the latest changes?

src/nnn.c
```diff
@@ -5768,6 +5947,182 @@ static bool browse(char *ipath, const char *session)
...
+	if (!mkdtemp(tmpdir)) {
+		free(tmpdir);
+
+		DPRINTF_S(strerror(errno));
+		return NULL;
+	}
+
+	g_listpath = tmpdir;
+
+	for(i = 0; i < entries; ++i){
```
jarun: for space i

src/nnn.c
```diff
@@ -5768,6 +5947,182 @@ static bool browse(char *ipath, const char *session)
...
+		err = stat(paths[i], &sb);
+		if (err == ENOENT)
+			continue;
+
+		xstrlcpy(tmp + 10, paths[i] + len, strlen(paths[i]) + 1);
```
jarun: Please run checkpatch on this.

    ./checkpatch.pl --no-tree --max-line-length=129 -f ../nnn/src/nnn.c > logs

My checkpatch branch: https://github.com/jarun/checkpatch

KlzXS:
> Functionality-wise looks good to me. 

I'd say it doesn't need anything more. Now it's only about confirming it works for any given input.

jarun:
There are some coding standard issues... spacing etc. Please see my comments inline.

KlzXS:
I ran checkpatch and fixed that, it's in dc8643e

jarun:
I am yet to look into the algorithms. Please allow me some time. I'll get back on that.

src/nnn.c
```diff
@@ -6055,17 +6435,35 @@ int main(int argc, char *argv[])
...
+			exit(1);
+
+		/* Now we are in path list mode */
+		if (!isatty(STDIN_FILENO)) {
+			/* Prefix for temporary files */
+			if (!set_tmp_path())
+				exit(1);
+
+			/* This is the same as g_listpath */
+			initpath = load_input(&startmsg);
```
jarun: startmsg is a fixed one. Add a new bool in g_states and set it. Show the msg in browse() if set. Try to avoid adding heady params, copying strings etc. If we ever need to show a different msg we can extent a single bit to 2.

src/nnn.c
```diff
@@ -930,13 +935,139 @@ static void *xmemrchr(uchar *s, uchar ch, size_t n)
...
+	 * More info: https://www.felixcloutier.com/x86/MOVDQA:VMOVDQA32:VMOVDQA64
+	 */
+	if ((len >= LONG_SIZE) && (((ulong)s & _ALIGNMENT_MASK) == 0
+			       && ((ulong)prefix & _ALIGNMENT_MASK) == 0)) {
+		x = (ulong *)s;
+		y = (ulong *)prefix;
+		blocks = len >> _WSHIFT;
+		len &= LONG_SIZE - 1;
+
+		i = 0;
```
jarun: Initialize i and j at declaration.

src/nnn.c
```diff
@@ -930,13 +935,139 @@ static void *xmemrchr(uchar *s, uchar ch, size_t n)
...
+}
+
 static char *xbasename(char *path)
 {
 	char *base = xmemrchr((uchar *)path, '/', strlen(path)); // NOLINT
 
 	return base ? base + 1 : path;
 }
 
+static char *xrealpath(const char *path, char *resolved_path, const char *cwd)
```
jarun: Why not use `realpath(3)`?

src/nnn.c
```diff
@@ -1076,8 +1207,53 @@ static void updateselbuf(const char *path, char *newpath)
...
+
+	if (count < 0) {
+		DPRINTF_S(strerror(errno));
+		return;
+	}
+
+	selbufpos = count;
+	pselbuf[--count] = '\0';
+	for (--count; count > 0; --count) {
+		if (pselbuf[count] == '\n' && pselbuf[count+1] == '/') {
```
jarun: No need of {} brackets for single statement condition checks.

src/nnn.c
```diff
@@ -5151,7 +5334,11 @@ static bool browse(char *ipath, const char *session)
 		case SEL_STATS: // fallthrough
 		case SEL_CHMODX:
 			if (ndents) {
-				mkpath(path, dents[cur].name, newpath);
+				if (g_listpath && xstrcmp(path, g_listpath) == 0)
```
jarun: (g_listpath && xstrcmp(path, g_listpath) == 0 ? tmp = g_prefixpath : tmp = path;
    mkpath(tmp, ...)

Same at line 5523.

src/nnn.c
```diff
@@ -930,13 +935,139 @@ static void *xmemrchr(uchar *s, uchar ch, size_t n)
...
+
+	/* Only accept non-empty strings */
+	if (*s == '\0' || *prefix == '\0')
+		return NULL;
+
+	ulong *x, *y;
+	size_t i, j, blocks = 0;
+	size_t len_s = strlen(s), len_prefix = strlen(prefix);
+	size_t len = MIN(len_s, len_prefix);
+	const uint _WSHIFT = (LONG_SIZE == 8) ? 3 : 2;
```
jarun: I think we should just have _WSHIFT as a global const:

    static const uint _WSHIFT = (LONG_SIZE == 8) ? 3 : 2;
and use it here as well as in `xstrlcpy()` instead of calculating it again and again.

jarun:
Reviewed partially. I still need time.

src/nnn.c
KlzXS: I think I already explained this once. `realpath()` follows symlinks and resolves the path fully. If the user gives me `a/b/../c` I want to show him `a/c` not `/what/c/was/pointing/to`.

src/nnn.c
```diff
@@ -930,13 +935,139 @@ static void *xmemrchr(uchar *s, uchar ch, size_t n)
...
+	else
+		resolved_path[0] = '\0';
+
+
+	src = path;
+	dst = resolved_path + dst_size;
+	for (next = NULL; next != path + src_size;) {
+		next = strchr(src, '/');
+		if (!next)
+			next = path+src_size;
```
jarun: path + src_size

maximbaz:
Tested functionality, works like a charm :+1:

jarun:
@KlzXS can you please address the review comments?

KlzXS:
Here it is. Anything else?

jarun:
> Anything else?

I'll review again. I am kind of forced to do it in multiple phases because of bandwidth. But don't worry, it would be over by next weekend.

src/nnn.c
```diff
@@ -1076,8 +1206,51 @@ static void updateselbuf(const char *path, char *newpath)
...
+		DPRINTF_S(strerror(errno));
+		unlink(g_tmpfpath);
+		return;
+	}
+
+	count = read(fd, pselbuf, selbuflen);
+	close(fd);
+	unlink(g_tmpfpath);
+
+	if (count < 0) {
```
jarun: close() an unlink() both set errno.

src/nnn.c
```diff
@@ -1150,7 +1323,7 @@ static int editselection(void)
 	}
 
 	if (count < 0) {
-		DPRINTF_S("error reading tmp file");
+		DPRINTF_S(strerror(errno));
```
jarun: same here. close() and unlink() both set errno on failure.

I think we should DPRINTF_S the errno after each of these calls if they fail. And if any one of these fail we should return from here.

src/nnn.c
```diff
@@ -6236,6 +6622,10 @@ int main(int argc, char *argv[])
 
 	opt = browse(initpath, session);
 	mousemask(mask, NULL);
+
+	if (g_listpath)
```
jarun: We also need to free `g_listpath`.

jarun:
Added some more comments.

Now I have 2 functions left: `make_tmp_tree()` and `load_input()` to review.

src/nnn.c
KlzXS: `g_listpath` and `initpath` point to the same thing when in list mode to avoid copying strings.

src/nnn.c
KlzXS: How about this, I handle everything that sets `errno` in `main, common_prefix, xrealpath, editselection, endselection, make_tmp_tree, load_input` now and save the rest of the file for another PR.

jarun: I think it's better to handle current errno in another PR. For now just handle these 2 cases.

KlzXS: > I think we should DPRINTF_S the errno after each of these calls if they fail. And if any one of these fail we should return from here.

is something like 
```
if (close(fd) || unlink(g_tmpfpath) || count < 0) {
    DPRINTF_S(strerror(errno));
    return;
}
```
ok, or do I split it into multiple `if`s? I'll get to this tonight.

jarun: Let's say count is < 0 but we will still proceed with close() and that fails too. We lose the reason count is < 0.
Please check for error after all calls. See if we can use `printwarn()` which prints strerror.

KlzXS: So an if with count < 0 in which we log errno and in it we try to close and unlink logging and error if it occurs. Then another if with close and unlink. Ok

KlzXS: > See if we can use printwarn() which prints strerror.

We can, but that doesn't print to debug file.

jarun: Add both so it shows as well as prints.

src/nnn.c
```diff
@@ -5768,6 +5969,191 @@ static bool browse(char *ipath, const char *session)
...
+		total_read += input_read;
+	}
+
+	input[total_read] = '\0';
+	/* 256MiB already read, enough for 2^16 paths */
+	if (input_read == chunk && chunk_count == 512)
+		goto malloc_1;
+
+	prev = input;
+	while (prev < input + total_read) {
```
jarun: I think we can avoid this loop and count the number of NULs (for entries) in the first loop. When you read each chunk, calculate the number of NULs in the chunk and add to entries.

src/nnn.c
```diff
@@ -5768,6 +5969,191 @@ static bool browse(char *ipath, const char *session)
...
+		/* path limit exceeded */
+		if (++entries > (1 << 16))
+			goto malloc_1;
+	}
+
+	if (!entries)
+		goto malloc_1;
+
+	/* allocate the space for all the strings */
+	paths = malloc(sizeof(char *) * entries);
```
jarun: I think you are allocating memory for character pointers. Why not use offsets? If you are reading in max 2^16, a uint offset per string should be enough to know where it starts.

src/nnn.c
```diff
@@ -5768,6 +5969,191 @@ static bool browse(char *ipath, const char *session)
...
+	}
+
+	/* prev used as tmp variable */
+	prev = getcwd(cwd, PATH_MAX);
+	if (!prev) {
+		goto malloc_2;
+	}
+
+	for (i = 0; i < entries; ++i) {
+		paths[i] = malloc(sizeof(char) * PATH_MAX);
```
jarun: We have already allocated memory for all the strings in line 6043, Why allocate again?

src/nnn.c
```diff
@@ -5768,6 +5969,191 @@ static bool browse(char *ipath, const char *session)
...
+
+			for (--i; i >= 0; --i)
+				free(paths[i]);
+
+			goto malloc_2;
+		}
+	}
+
+	for (i = 0, prev = input; i < entries; ++i) {
+		next = memchr(prev, '\0', total_read - (prev - input) + 1) + 1;
```
jarun: We did this already. Had we saved the offsets at that point AND used the earlier allocated memory, we would not need this.

jarun:
I am done with the review. I think we can optimize `load_input()` as below (allocate memory for all chunks and allocate memory for all offsets using xrealloc()):

for each chunk read
  - xrealloc and read in a chunk
  - scan for NUL char
    - if found xrealloc and save the offset from start of memory
    - increment the num of entries

If you want to avoid handling with offsets, and want to use char pointers for each input file:

for each chunk read
  - xrealloc and read in a chunk
  - scan for NUL char
    - if found xrealloc and save the pointer to that file path
    - increment the num of entries

What I am aiming at is reducing the number of loops (and if possible allocations) we use.

KlzXS:
I did it this way because:

1.  I thought it would be faster to read all the input and then work on it, because cpu optimizing access to memory on the hardware level. This might be a case of premature optimization.
2.  I can't know if a read has read a whole number of paths. I might just so happen that the 512K boundary landed in the middle of some path. That would complicate the logic for handling it. So does an input of mostly NULL bytes back to back.

But I think it's okay to use offsets, or pointers instead of allocating a new region of memory.

jarun:
> But I think it's okay to use offsets, or pointers instead of allocating a new region of memory.

I think using pointers would be simpler. Try that one.

KlzXS:
If we want to do it in-place we also need to allocate (can also be on stack or a free global) a tmp variable for ouput in `xrealpath` as it can't do it's thing in-place. But string copying would still need to happen in the end.

jarun:
Wait wait... we do xrealloc() which may change the start pointer so we cannot have string pointers in first iteration. So store offsets (as an array... we can pre-allcoate this memory if we have a limit on the max number of file paths we will support) and when all are read, allocate the number of string pointers you have read in as entries and set them one by one using the offsets.

1. NUL can be within a chunk, but you will have the offset till the last NUL in the last chunk so after reading in the next chunk and doing xrealloc you can always find the next NUL from the previous offset as you are doing now.

2. Is the last path NUL terminated or is just EOL or some other end of input indicator?

If the last path is also NUL terminated and we do not see a NUL termination, we can always discard the last one by setting the last string pointer to NULL.

If the last path may not be NUL terminated but it's not a complete path to an existing file, stat will fail for the last string pointer so we can discard it.

KlzXS:
Ok, that works. It doesn't need to be NULL terminated, I just add a NULL at the end regardless, but I should move the place where I do that.

But the `xrealpath` thing still stands.

jarun:
Let's optimize on reading in data and figuring out the offsets where the strings start...

Once we have read in, for each path can we re-use the same string pointers?

So we have a string pointer to a path in input chunk, we then call xrealpath(). Can xrealpath() allocate and give back a new string by mallocing the number of bytes required instead of 4096 all the time? We can then set the earlier string pointer to this string returned by xrealpath.

jarun:
When we have processed all the string pointers, we can just discard the chunks we have read in.

KlzXS:
> Can xrealpath() allocate and give back a new string by mallocing the number of bytes required instead of 4096 all the time?

The best I can do is either length of input or length of input plus length of current working dir, based on if it's a relative path or not.

So in total I need some offsets, you said `uint`, isn't it usually `size_t`? And some `char*`.

jarun:
> The best I can do is either length of input or length of input plus length of current working dir, based on if it's a relative path or not.

Should be fine.

> isn't it usually size_t? And some char*

Right!

jarun:
Also please add this feature to the Convenience section under Features in README.... maybe just above _Copy (as), move (as), delete, archive, link selection_. It's an important feature.

jarun:
Looks much better now!

Please let us know when it's ready for testing!

@maximbaz please test the final version. I'll too!

KlzXS:
That's not my build fail. I never touched that function (`xreadline`). I just resolved a merge conflict. And why is it not failing on master if I just merged master into this.

Can someone enlighten me?

jarun:
master is passing. It's a static analysis issue and showing from this PR.

KlzXS:
I know master is passing and this is not, but why. I don't think I've touched anything near `xreadline` or `launch_app`, and that's the call chain it goes down.

I'll need some time to get myself familiar with them then. The build fail shouldn't prevent you from testing the rest.

jarun:
I think you are calling xreadline somewhere and you have passed some buffer which is uninitialized or not NULL.

jarun:
I am also checking it.

jarun:
I think it is complaining about a case where stdscr is NULL. We can ignore it as we check at the beginning when we initialize curses. I will push a patch.

jarun:
@KlzXS please rebase again and see how it goes.

KlzXS:
Now it works. Thank you.

jarun:
Cool! This works as intended.

maximbaz:
In my testing also everything seems to work as expected, nice job!

jarun:
@KlzXS I see a minor issue... the `ignoring invalid paths...` msg would probably show at start frequently and it kills the startup experience. I think we should just drop it.

src/nnn.c
```diff
@@ -5861,6 +6069,175 @@ static bool browse(char *ipath, const char *session)
...
+
+		DPRINTF_S(strerror(errno));
+		return NULL;
+	}
+
+	g_listpath = tmpdir;
+
+	for (i = 0; i < entries; ++i) {
+		err = stat(paths[i], &sb);
+		if (err && errno == ENOENT) {
```
jarun: This happens more frequently than I anticipated, so just :

    if (err && errno == ENOENT && !ignore) {

jarun:
I think we are almost good to go. I suggested a review change and removing that msg... please take a look.

KlzXS:
> would probably show at start frequently and it kills the startup experience

You mean you frequently give it invalid paths? The message is shown only once during startup and then never again.

jarun:
Yes, but that msg has a delay along with it when we do not register keypresses. So I was losing one or two Down keys. @maximbaz and I are working on it. Give as 10/20 mins. Check the review comment meanwhile. It's a minor optimization.

src/nnn.c
KlzXS: That would make it only skip the first invalid path while creating links and dirs for others. I don't get what you want it to do.

jarun:
Keep the msg! I think our fix works.

jarun:
All clear! Just address the last review comment and we are having the most wanted feature of the next release!

I know it's a lot of work. Thank you for pursuing and bearing with my phased reviews.

KlzXS:
> This happens more frequently than I anticipated, so just :
`if (err && errno == ENOENT && !ignore) {`

This comment? As I said I can't add `!ignore`, if I did it would only skip the first invalid path and keep all the other invalid ones.

jarun:
Ahh OKK! I missed that.

jarun:
Thank you so much!

KlzXS:
No problem. I'm happy to help.

PR #444, opened 2020-01-18
fixed a formating error in Plugins README

sdushantha:
Some normal text was indent a little, which caused to be treated as code.


jarun:
Thank you!

Issue #427, opened 2020-01-10
ToDo list


jarun:
Rolled from #386.

### For next release

- [x] program option `-e` replaces `NNN_USE_EDITOR`
- [x] program option `-t` replaces `NNN_IDLE_TIMEOUT`
- [x] use `umount` on macOS to unmount
- [x] confirm and try lazy unmount when regular unmount fails
- [x] quit with error code on <kbd>Q</kbd>
- [x] additional key <kbd>F5</kbd> to toggle hidden
- [x] fix crash on entering empty dir, then Down
- [x] add [`glow`](https://github.com/charmbracelet/glow) as Markdown viewer in `nuke`

### Proposed features and tasks (up for grabs)

- [ ] take list of files as input and show
- [ ] add PCRE code within a macro and add Makefile target to compile with PCRE library ([sample code](https://www.mitchr.me/SS/exampleCode/AUPG/pcre_example.c.html))
- [ ] support pre-defined filters like bookmarks
- [ ] Cygwin package
- [ ] update the `nnn.vim` plugin to show a persistent bar (https://github.com/mcchrish/nnn.vim/issues/46)
- [ ] a video exploring `nnn` plugins

Anything else which would add value (please discuss in this thread).

List of [completed features and tasks](https://github.com/jarun/nnn/wiki/Feature-history).

ghistes:
My absolutely biggest usability problem at the moment is that I search, don't find what I want and press "up" to access the search history.

In nnn this means that you don't access the search history but move up in the file-list, the status bar is displayed again and your previous search-term is gone for good.

So a big gain for me would be for nnn to at least remember the last search item, so you could search, see what you find, move around, and then maybe slightly tweak your search without having to retype everything. It would be even better if maybe the last 5 items or so could be kept - maybe even saved to disk so that they would be accessible again after a restart.

I don't know if the string-search as it currently works is intended to work across directories, because I could live very well with typing "/" to start a search, then if the next char is again a "/" it will be a regex search (which should be case-sensitive as you can build case-insensitivy into the regex if you want it) and if not it's a string.

So "/hubba" would search case-insensitive for "hubba", while "//h[0-9]{2}" would regex-search for a lower-case h followed by 2 digits. If I also wanted to include upper-case "H", I would simply do "//[hH][0-9]{2}"

I would not really need all the powers of pcres, that would be silly in a file manager, but to be able to search for patters of digits and letters in a simple way would be very handy - so I would like to have things like "\d" or "\w" etc.

[...]

PR #447, opened 2020-01-22
Shorten home path in the header to "~".

ruocco:
Hi! I made this little change in "nnn" and I thought that some people
might like it as well. I didn't find any information on the proper way to
contribute, so if this isn't it please ignore this merge request.

---

In the UNIX world "\~" is a synonym for the home folder of the current
user. This commit substitutes the full path of the home for the current
user with the char "\~" in the header.

An user that prefers to see the whole address for the home directory
can do so passing the command line argument "F" or using the "F"
toggle when the software is already running.

### DIFF
--- a/nnn.1
+++ b/nnn.1
@@ -67,6 +67,9 @@ supports the following options:
 .Fl d
         detail mode
 .Pp
+.Fl F
+        show full path for the home directory instead of "~"
+.Pp
 .Fl e
         open text files in $VISUAL (else $EDITOR, fallback vi) [preferably CLI]
 .Pp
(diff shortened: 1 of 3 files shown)


jarun:
Hi, thanks for the PR!

I kept the absolute path to avoid this extra processing. Generally a user knows where (s)he is when within home. To show the tilde we can't strlen, malloc, strstr etc. repeatedly on every redraw. This minimalism differentiates `nnn` from other file managers.

jarun:
If you are interested in contributing, there is a line item in the ToDo list to link to PCRE libs for simpler regex filters. I think it's a good start.

jarun:
ToDo list: https://github.com/jarun/nnn/issues/427

PR #452, opened 2020-01-27
Added more readline bindings

AN3223:


jarun:
Please provide some details on what is being changed. Also CI fails.

AN3223:
> Please provide some details on what is being changed. Also CI fails.
Sorry didn't notice CI, fixed now.

This PR adds some more readline keybinds to the xreadline function:
C-d is the same as Delete
C-f moves the cursor one character forward
C-b moves the cursor one character backward
C-w deletes one word backwards
C-k kills the line before the cursor (opposite of the existing C-u)

jarun:
Thanks for the details. I'll review.

PR #454, opened 2020-01-28
Link librt to fix build with old glibc

### LINKED ISSUE #453: Undefined clock_gettime() on CentOS 6
When built on CentOS 6:

```
cc -I/usr/pkg/include -D_XOPEN_SOURCE_EXTENDED=1 -I/usr/pkg/include/ncurses -pipe -O2 -pipe -O2 -D_FORTIFY_SOURCE=2 -I/usr/pkg/include -D_XOPEN_SOURCE_EXTENDED=1 -I/usr/pkg/include/ncurses -Wall -Wextra   -L/usr/pkg/lib -Wl,-R/usr/pkg/lib -o nnn src/nnn.c -lreadline -lncursesw
/tmp/ccCwUFuM.o: In function `browse':
nnn.c:(.text+0x5430): undefined reference to `clock_gettime'
```

This can be fixed by adding `-lrt`, which that glibc version's man page says is required for the clock functions. I'll send a PR shortly.

sjmulder:
Building on CentOS 6 yielded:

    nnn.c:(.text+0x5430): undefined reference to `clock_gettime'

Older glibc versions require librt to be linked for clock_*.

Fixes #453

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,6 @@ endif
 ifeq ($(O_DEBUG),1)
 	CPPFLAGS += -DDBGMODE
 	CFLAGS += -g
-	LDLIBS += -lrt
 endif
 
 ifeq ($(O_NORL),1)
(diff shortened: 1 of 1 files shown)


jarun:
Why do you want to support CentOS 6 when 8 is already here?

CentOS 6 was released on 2011-07-10. People who couldn't update their OS in this ridiculous amount of time do not need bleeding edge utilities.

CentOS 6 will be EOL in Nov 2020. We haven't supported it so far. We'll not going forward.

PR #455, opened 2020-01-29
Fallbacks for absent NAME_MAX and PATH_MAX

sjmulder:
Certainly NAME_MAX isn't guaranteed to exist and on (some versions of?)
Illumos and SmartOS it doesn't, so provide some reasonably accommodating
fallbacks.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -118,6 +118,18 @@
 #define S_BLKSIZE 512 /* S_BLKSIZE is missing on Android NDK (Termux) */
 #endif
 
+/*
+ * NAME_MAX and PATH_MAX may not exist, e.g. with dirent.c_name being a
+ * flexible array on Illumos. Use somewhat accomodating fallback values.
+ */
+#ifndef NAME_MAX
+#define NAME_MAX 512
+#endif
+
+#ifndef PATH_MAX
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -118,6 +118,18 @@
 #define S_BLKSIZE 512 /* S_BLKSIZE is missing on Android NDK (Termux) */
 #endif
 
+/*
+ * NAME_MAX and PATH_MAX may not exist, e.g. with dirent.c_name being a
+ * flexible array on Illumos. Use somewhat accomodating fallback values.
+ */
+#ifndef NAME_MAX
+#define NAME_MAX 512
```
jarun: Please limit to 255. We use NAME_MAX+1 in code to accommodate this.

src/nnn.c
```diff
@@ -118,6 +118,18 @@
...
+#endif
+
+#ifndef PATH_MAX
+#define PATH_MAX 8191
```
jarun: Please limit to 4096.

jarun:
We define path max to 4096 for gnu hurd  it's not defined. I guess you can remove that now.

jarun:
Thank you!

PR #457, opened 2020-01-30
Compile static bin locally for release

maximbaz:
So now you will just do `make upload-local` and it will do the rest.

Is this something you had in mind?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,10 @@ uninstall:
 strip: $(BIN)
 	$(STRIP) $^
 
+static:
+	make O_STATIC=1 strip
+	mv $(BIN) $(BIN)-static
+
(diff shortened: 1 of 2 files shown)


jarun:
Exactly! Thank you!

jarun:
Oops! I missed something. Can you also package the static binary in tar.gz format with the release version?

maximbaz:
Do you mean to simply wrap the binary in the archive, with no additions? I don't think archive will save any space on the binary... :slightly_smiling_face:

jarun:
> Do you mean to simply wrap the binary in the archive, with no additions?

Yes, an archive is better.

jarun:
I'll take care of it. No problem!

Issue #458, opened 2020-01-31
Failure to copy ownership permissions to remote folder mounted with sshfs


Kabouik:
#### Environment details 

[x] Operating System: Solus 4
[x] Desktop Environment: Budgie Desktop
[x] Terminal Emulator: Tilix and Gnome terminal
[x] Shell: Bash
[ ] Custom desktop opener (if applicable):
[x] Issue exists on `nnn` master: yes

#### Exact steps to reproduce the issue
1. Mount a remote folder using `c`, then `s` for sshfs and type relevant user1@ip and password.
2. Copy local files with permissions `664` to a remote folder owned by `user1` (`nemo` in the example below).

```
nemo@192.168.1.29's password: 
cp: failed to preserve ownership for "./Donovan/The Hurdy Gurdy Man/1.01. Donovan - Hurdy Gurdy Man.mp3": Permission denied
```
The file is copied, but its permissions go from `664` in the local folder to `600` in the remote folder (and parent folders copied with the file go from `775` to `700`). Here are the file properties after it has been pasted in the remote folder:

```
  File: /home/mathieu/.config/nnn/nemo@192.168.1.29/Music/Donovan/The Hurdy Gurdy Man/1.01. Donovan - Hurdy Gurdy Man.mp3
  Size: 7847696         Blocks: 15328      IO Block: 4096   regular file
Device: 2dh/45d Inode: 102         Links: 1
Access: (0600/-rw-------)  Uid: (100000/ UNKNOWN)   Gid: (100000/ UNKNOWN)
Access: 2020-01-31 22:24:23.000000000 +0100
Modify: 2020-01-31 22:24:20.000000000 +0100
Change: 2020-01-31 22:24:20.000000000 +0100
 Birth: -
```

*versus* the original file in the local folder:

```
  File: /media/stockage/Media/MUSIC/Donovan/The Hurdy Gurdy Man/1.01. Donovan - Hurdy Gurdy Man.mp3
  Size: 7847696         Blocks: 15328      IO Block: 4096   regular file
Device: 801h/2049d      Inode: 3933962     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/ mathieu)   Gid: ( 1000/ mathieu)
Access: 2020-01-31 22:24:23.744808304 +0100
Modify: 2020-01-31 22:24:20.730805500 +0100
Change: 2020-01-31 22:24:20.730805500 +0100
 Birth: -
```

jarun:
[...]

PR #460, opened 2020-02-01
fix sshfs not utilizing -o options

0xACE:
This is a quick and dirty hack to potentially fix one issues apparent
in issue #458.

This is mostly untested guess work.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3579,7 +3579,7 @@ static bool archive_mount(char *name, char *path, char *newpath, int *presel)
 
 static bool remote_mount(char *newpath, int *presel)
 {
-	uchar flag = F_CLI;
+	uchar flag = F_MULTI;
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -3579,7 +3579,7 @@ static bool archive_mount(char *name, char *path, char *newpath, int *presel)
 
 static bool remote_mount(char *newpath, int *presel)
 {
-	uchar flag = F_CLI;
+	uchar flag = F_MULTI;
```
jarun: I don't follow. I see:

    #define F_CLI     (F_NORMAL | F_MULTI)

0xACE: Ah, my bad, closing this PR then. I thought `F_CLI` was different from `F_MULTI`. I really have to go to bed now as it's way too late.

PR #461, opened 2020-02-01
Do not swallow keypress after printwait

maximbaz:
>>> but for some reason the presence of that message eats up the ';'
>>
>> I see. That's a problem.
>
> can we fix this? don't eat the key press while popup is shown, but move it down the global key handling pipeline maybe?

This implements this change and seems to work for me, but I'm very uncertain if I'm not breaking anything else by this patch?

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2024,12 +2024,10 @@ static int nextsel(int presel)
 		//DPRINTF_D(c);
 		//DPRINTF_S(keyname(c));
 
-		/* Clear previous filter when manually starting */
-		if (c == FILTER)
-			clearfilter();
-
-		if (presel == MSGWAIT)
+		if (c == ERR && presel == MSGWAIT)
 			c = (cfg.filtermode) ? FILTER : CONTROL('L');
+		else if (c == FILTER) /* Clear previous filter when manually starting */
+			clearfilter();
 	}
 
 	if (c == -1) {


jarun:
How does this affect line 2032? If presel is MSGWAIT, we will still override and lose `c`.

jarun:
I think the problem is we are overriding `c` in line 2032. We shod reassign `c` only if `c` is not equal to 0 or MSGWAIT. A different value of `c` means someone has pressed a key.

maximbaz:
mmm could you show the code, or just push to this branch? I'm not sure what you mean, but I'll test :wink:

jarun:
Does the following patch work for you?

```
index 133660d..4b27a2e 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2024,12 +2024,10 @@ static int nextsel(int presel)
                //DPRINTF_D(c);
                //DPRINTF_S(keyname(c));
 
-               /* Clear previous filter when manually starting */
-               if (c == FILTER)
-                       clearfilter();
-
-               if (presel == MSGWAIT)
+               if (c == ERR && presel == MSGWAIT)
                        c = (cfg.filtermode) ? FILTER : CONTROL('L');
+               else if (c == FILTER) /* Clear previous filter when manually starting */
+                       clearfilter();
        }
 
        if (c == -1) {
```

maximbaz:
Yes! Works like a charm, I pushed it

jarun:
Awesome! Thanks for confirming!

jarun:
It's in!

PR #476, opened 2020-02-14
Add NNN_OPTS environment variable feature

uffejakobsen:
Hi,
This is my suggestion on how to add the NNN_OPTS feature.
/Uffe :-)

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -553,16 +553,18 @@ static const char * const messages[] = {
 };
 
 /* Supported configuration environment variables */
-#define NNN_BMS 0
-#define NNN_PLUG 1
-#define NNN_OPENER 2
-#define NNN_COLORS 3
-#define NNNLVL 4
-#define NNN_PIPE 5
-#define NNN_ARCHIVE 6 /* strings end here */
-#define NNN_TRASH 7 /* flags begin here */
+#define NNN_OPTS 0
+#define NNN_BMS 1
+#define NNN_PLUG 2
+#define NNN_OPENER 3
+#define NNN_COLORS 4
+#define NNNLVL 5
+#define NNN_PIPE 6
+#define NNN_ARCHIVE 7 /* strings end here */
+#define NNN_TRASH 8 /* flags begin here */
 
 static const char * const env_cfg[] = {
+	"NNN_OPTS",
 	"NNN_BMS",
 	"NNN_PLUG",
 	"NNN_OPENER",
@@ -6450,7 +6452,12 @@ int main(int argc, char *argv[])
 	mmask_t mask;
 #endif
 
-	while ((opt = getopt(argc, argv, "aAb:cdeEgHKnop:QrRs:St:vVxh")) != -1) {
+  const char* const env_opts = xgetenv(env_cfg[NNN_OPTS], NULL);
+  int env_opts_idx = (env_opts ? (int)strlen(env_opts) : -1);
+  while ((opt = (--env_opts_idx >= 0 ? env_opts[env_opts_idx] : getopt(argc, argv, "aAb:cdeEgHKnop:QrRs:St:vVxh"))) != -1) {
+		if (env_opts_idx >= 0) {
+			optarg = NULL;
+		}
 		switch (opt) {
 		case 'a':
 			cfg.mtime = 0;
@@ -6495,6 +6502,9 @@ int main(int argc, char *argv[])
 			cfg.nonavopen = 1;
 			break;
 		case 'p':
+			if (!optarg) {
+				break;
+			}
 			cfg.picker = 1;
 			if (optarg[0] == '-' && optarg[1] == '\0')
 				cfg.pickraw = 1;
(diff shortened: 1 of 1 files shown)


uffejakobsen:
Well I can see that the CI validation found some problems - I will do another iteration of the pull request

src/nnn.c
```diff
@@ -6450,7 +6452,18 @@ int main(int argc, char *argv[])
 	mmask_t mask;
 #endif
 
-	while ((opt = getopt(argc, argv, "aAb:cdeEgHKnop:QrRs:St:vVxh")) != -1) {
+  const char* const env_opts = xgetenv(env_cfg[NNN_OPTS], NULL);
+  int env_opts_idx = (env_opts ? (int)strlen(env_opts)-1 : -1);
+  while (true) {
+		if (env_opts_idx >= 0) {
+			opt = env_opts[env_opts_idx--];
+			optarg = NULL;
```
jarun: Where are you setting this?

src/nnn.c
```diff
@@ -6450,7 +6452,18 @@ int main(int argc, char *argv[])
 	mmask_t mask;
 #endif
 
-	while ((opt = getopt(argc, argv, "aAb:cdeEgHKnop:QrRs:St:vVxh")) != -1) {
+  const char* const env_opts = xgetenv(env_cfg[NNN_OPTS], NULL);
+  int env_opts_idx = (env_opts ? (int)strlen(env_opts)-1 : -1);
+  while (true) {
```
jarun: Isn't there a better way to do this without a glaring `while (true)` at the beginning of a program?

and it should be TRUE.

src/nnn.c
```diff
@@ -6524,24 +6540,39 @@ int main(int argc, char *argv[])
...
+			if (!optarg) {
+				break;
+			}
 			idletimeout = atoi(optarg);
 			break;
 		case 'v':
 			namecmpfn = &xstrverscasecmp;
 			break;
 		case 'V':
+			if (env_opts_idx >= 0) {
```
jarun: What is this check for?

jarun:
Huh! There's a very simple way to do that!

```
nn="nnn -n"
eval $nn
```

jarun:
Works in `dash` too!

jarun:
The following works too:

```
nn="nnn -cEnrx $@"
eval $nn /usr
```

src/nnn.c
uffejakobsen: Basically I am processing the NNN_OPTS string like the existing getopt() - by setting opt and optarg - just like getopt()
if env_opts_idx >= 0 then we know that we are in the process of going through NNN_OPTS options and not getopt() options.
The only reason for this check is that I felt that we should skip processing options like 'V' and 'h' in NNN_OPTS

src/nnn.c
uffejakobsen: Sorry, but I'm not really sure what line you are referring (asking) to ?

jarun: line 6460:

    optarg = NULL;

src/nnn.c
uffejakobsen: I can give it a try  :-)

jarun: See this: https://github.com/jarun/nnn/pull/476#issuecomment-586310064

src/nnn.c
uffejakobsen: Ahhh, well that is because NNN_OPTS only accepts simple options (flags) and not those with additional arguments like options 'p', 's' and 't'
So basically this is to keep away from those switch/case entries while processing NNN_OPTS env var.

jarun:
I can see that sending `-` is a problem.

uffejakobsen:
> The following works too:
> 
> ```
> nnnn="nnn -cEnrx -t 900 "$@""
> eval $nn /usr
> ```
Well, seen from my perspective - some config file syncing is still needed for me to just write "nnn" on any system that I SSH to. Where the NNN_OPTS solution just works out of the box :-)

src/nnn.c
jarun: Instead of touching `optarg` why not just check `env_opts_idx >= 0` like you are doing in other places.

src/nnn.c
jarun: Plese remove these checks. If someone wants to check the version of `nnn` installed on the remote (s)he may use these as well.

src/nnn.c
jarun: Basically:

    if (env_opts_idx < 0)
        idletimeout = atoi(optarg);

like this.

jarun:
> The NNN_OPTS solution just works out of the box :-)

Yes, it's much better.

jarun:
I'll take care of polishing it. Thanks for the PR!

src/nnn.c
uffejakobsen: well, the first solution seemed to make the Travis-CI clang-tidy check more happy - that was the only reason :-)

uffejakobsen:
Thanks a lot - cannot get through the day without nnn - and this addition will make life with nnn much better :+1:

jarun:
No problem! This option is indeed a fantastic suggestion.

> cannot get through the day without nnn

Neither can I! ;)

PR #482, opened 2020-03-01
Add right click file selection

leovilok:
This allows to select files by clicking the right mouse button.

IDK if it's even wanted.
I didn't see where (or if ?) mouse actions should be documented.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1458,9 +1458,11 @@ static bool initcurses(void *oldmask)
 	keypad(stdscr, TRUE);
 #ifndef NOMOUSE
 #if NCURSES_MOUSE_VERSION <= 1
-	mousemask(BUTTON1_PRESSED | BUTTON1_DOUBLE_CLICKED, (mmask_t *)oldmask);
+	mousemask(BUTTON1_PRESSED | BUTTON1_DOUBLE_CLICKED | BUTTON3_PRESSED,
+			(mmask_t *)oldmask);
 #else
-	mousemask(BUTTON1_PRESSED | BUTTON4_PRESSED | BUTTON5_PRESSED, (mmask_t *)oldmask);
+	mousemask(BUTTON1_PRESSED | BUTTON3_PRESSED | BUTTON4_PRESSED | BUTTON5_PRESSED,
+			(mmask_t *)oldmask);
 #endif
 	mouseinterval(0);
 #endif
@@ -4952,6 +4954,7 @@ static bool browse(char *ipath, const char *session)
 	MEVENT event;
 	struct timespec mousetimings[2] = {{.tv_sec = 0, .tv_nsec = 0}, {.tv_sec = 0, .tv_nsec = 0} };
 	bool currentmouse = 1;
+	bool rightclicksel = 0;
 #endif
 
 #ifndef DIR_LIMITED_SELECTION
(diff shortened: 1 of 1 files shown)


jarun:
How does it behave in the Termux environment?

Instead of the goto, can we use `presel`?

leovilok:
> How does it behave in the Termux environment?
I can't test that since I don't own a phone.

> Instead of the goto, can we use `presel`?
Yes.

We could use a `break` instead of the `goto` and set `presel` to a new special value that would make `nextsel()` return `SEL_SEL`.
But I'm not sur when the first `if` in the loop is fired, and I wonder if this could overwrite presel in a realistic case.

jarun:
> But I'm not sur when the first if in the loop is fired

function? line number?

leovilok:
In `browse()`, line 5056 in my fork, the first `if` in the big `while(1)`.

jarun:
No effect in Termux. That's sad but this PR is indeed useful for people who use the mouse.

The variable `rightclicksel` is redundant. We can remove it. The following modified patch works fine for me:

```
index ea6b554..42bb227 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -144,6 +144,7 @@
 #define RFILTER '\\'
 #define CASE ':'
 #define MSGWAIT '$'
+#define SELECT ' '
 #define REGEX_MAX 48
 #define ENTRY_INCR 64 /* Number of dir 'entry' structures to allocate per shot */
 #define NAMEBUF_INCR 0x800 /* 64 dir entries at once, avg. 32 chars per filename = 64*32B = 2KB */
@@ -1458,9 +1459,11 @@ static bool initcurses(void *oldmask)
        keypad(stdscr, TRUE);
 #ifndef NOMOUSE
 #if NCURSES_MOUSE_VERSION <= 1
-       mousemask(BUTTON1_PRESSED | BUTTON1_DOUBLE_CLICKED, (mmask_t *)oldmask);
+       mousemask(BUTTON1_PRESSED | BUTTON1_DOUBLE_CLICKED | BUTTON3_PRESSED,
+                       (mmask_t *)oldmask);
 #else
-       mousemask(BUTTON1_PRESSED | BUTTON4_PRESSED | BUTTON5_PRESSED, (mmask_t *)oldmask);
+       mousemask(BUTTON1_PRESSED | BUTTON3_PRESSED | BUTTON4_PRESSED | BUTTON5_PRESSED,
+                       (mmask_t *)oldmask);
 #endif
        mouseinterval(0);
 #endif
@@ -5165,9 +5168,18 @@ nochange:
                        }
 
                        /* Handle clicking on a file */
-                       if (event.y >= 2 && event.y <= ndents + 1 && event.bstate == BUTTON1_PRESSED) {
+                       if (event.y >= 2 && event.y <= ndents + 1 &&
+                                       (event.bstate == BUTTON1_PRESSED ||
+                                        event.bstate == BUTTON3_PRESSED)) {
                                r = curscroll + (event.y - 2);
                                move_cursor(r, 1);
+
+                               /* Handle right click selection */
+                               if (event.bstate == BUTTON3_PRESSED) {
+                                       presel = SELECT;
+                                       goto nochange;
+                               }
+
                                currentmouse ^= 1;
                                clock_gettime(
 #if defined(CLOCK_MONOTONIC_RAW)
```

Please test and update the PR.

jarun:
OK. I see why `rightclicksel` is required. I have updated the documentation.

jarun:
Removed the goto. See commit ad9f834b640832a1f64e4901ade3845ca0971df7.

leovilok:
Thank you!

PR #488, opened 2020-03-06
POSIX compliant batchrename plugin

KlzXS:
Note that `mktemp` is not POSIX.

Please test this, cover some edge cases. Try using filenames with `'`. I wouldn't recommend using it with important files until tested as it does `mv` files around.

### DIFF
--- a/plugins/batchrename
+++ b/plugins/batchrename
@@ -1,22 +1,38 @@
 #!/usr/bin/env sh
 
-# Description: An almost fully POSIX compliant way to batch rename files
+# Description: An almost fully POSIX compliant batch file renamer
 #
 # Shell: POSIX compliant
 # Author: KlzXS
 
-EDITOR="${EDITOR:-vim}"
+EDITOR="${EDITOR:-vi}"
 TMPDIR="${TMPDIR:-/tmp}"
+INCLUDE_HIDDEN="${INCLUDE_HIDDEN:-0}"
 
 selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
 exit_status=0
 
 dst_file=$(mktemp "$TMPDIR/.nnnXXXXXX")
 
-arr=$(tr '\0' '\n' < "$selection" | sed "s/'/'\\\\''/g;s/^\(.*\)$/'\1'/" | tr '\n' ' ')
+if [ -s "$selection" ]; then
+	printf "Rename 'c'urrent / 's'election? "
+	read -r resp
+fi
+
+if [ "$resp" = "s" ]; then
+	arr=$(tr '\0' '\n' < "$selection")
+else
+	if [ "$INCLUDE_HIDDEN" -eq 0 ]; then
+		arr=$(find . ! -name . -prune ! -name ".*" -print | sort)
+	else
+		arr=$(find . ! -name . -prune -print | sort)
+	fi
+fi
+
+printf "%s" "$arr" | awk '{print NR " " $0}' > "$dst_file"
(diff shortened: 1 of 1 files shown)


KlzXS:
CI timed out.

jarun:
Restarted CI.

Can you modify it so it asks whether to use selection (if something is selected) or open entries in current dir?

plugins/batchrename
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env sh
+
+# Description: An almost fully POSIX compliant way to batch rename files
+#
+# Shell: POSIX compliant
+# Author: KlzXS
+
+EDITOR="${EDITOR:-vim}"
```
jarun: Please use `vi`.

plugins/batchrename
```diff
@@ -0,0 +1,168 @@
...
+TMPDIR="${TMPDIR:-/tmp}"
+
+selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
```
jarun: If there's a selection (size of .selection file is on-0) prompt user for selection or current dir.
In case there's no selection (size of .selection file is 0), open the entries in current dir directly.

plugins/batchrename
```diff
@@ -0,0 +1,168 @@
...
+exit_status=0
+
+dst_file=$(mktemp "$TMPDIR/.nnnXXXXXX")
```
jarun: Probably have a switch to include hidden files as well. By default we can go non-hidden only.

plugins/batchrename
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env sh
+
+# Description: An almost fully POSIX compliant way to batch rename files
```
jarun: An almost fully POSIX compliant batch file renamer

KlzXS:
> Can you modify it so it asks whether to use selection (if something is selected) or open entries in current dir?

I though you wanted to setup selection in `batch_rename()` and then call the plugin? I can modify it.

plugins/batchrename
KlzXS: This is in regards to editing the current directory entries? What should the switch be, I don't really like `-h`, that looks too much like "help".

jarun: Not that.

Have a variable at the top like:

    INCLUDE_HIDDEN=0

If user makes it 1, include the hidden files **in the dir** (doesn't apply to selection anyway) as well.

jarun:
> I though you wanted to setup selection in batch_rename() and then call the plugin? I can modify it.

We'll have to make it behave like other plugins. But the key to run this would remain `r`. So select in `nnn` and press `r`. If there's a selection, the script prompts...

jarun:
This is an excellent script BTW. I don't know of any POSIX-compliant _intelligent_ shell script to batch rename!

jarun:
Thank you so much!

@0xACE @maximbaz let's test this!

maximbaz:
Thanks for the ping!

Out of curiosity, why did you decide to move this functionality out of the core and into a plugin?

> We'll have to make it behave like other plugins. But the key to run this would remain r. So select in nnn and press r. If there's a selection, the script prompts...

Just as a heads-up, until now plugins were optional, and for example they are not being shipped as part of distro packages, so unless we address this, thousands of users will require manual intervention after nnn upgrade in order to fix the rename functionality.

jarun:
The core batch rename functionality is limited.

We can still have the core functionality as it is and invoke the plugin if it's available.

maximbaz:
If you want, we can also just start packaging plugins, so that you dont have to maintain two implementations of the same functionality - it should be as easy as adding one line in `install` target in Makefile 👍

jarun:
... that and the current renamer plugin with external deps goes.

Plugins are installed per-user, only if they want. Not planning to install them system-wide.

Issue #489, opened 2020-03-08
batchrenamer issues


jarun:
I integrated the plugin with the following patch:

```diff
diff --git a/plugins/README.md b/plugins/README.md
index b93fbf5..0728fbf 100644
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -27,6 +27,7 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | Plugin (a-z) | Description | Lang | Deps |
 | --- | --- | --- | --- |
 | autojump | Navigate to dir/path (**autojump stores navigation patterns**) | sh | autojump |
+| batchrename | Batch file renamer | sh | mktemp |
 | boom | Play random music from dir | sh | [moc](http://moc.daper.net/) |
 | dups | List non-empty duplicate files in current dir | sh | find, md5sum,<br>sort uniq xargs |
 | chksum | Create and verify checksums | sh | md5sum,<br>sha256sum |
diff --git a/src/nnn.c b/src/nnn.c
index 9633c46..c490e22 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -428,6 +428,7 @@ static uchar g_states;
 #define UTIL_FZY 16
 #define UTIL_NTFY 17
 #define UTIL_CBCP 18
+#define UTIL_BATCHRENAME 19
 
 /* Utilities to open files, run actions */
 static char * const utils[] = {
@@ -466,6 +467,7 @@ static char * const utils[] = {
 	"fzy",
 	".ntfy",
 	".cbcp",
+	"batchrename",
 };
 
 /* Common strings */
@@ -608,16 +610,13 @@ static char mv[] = "mv -i";
 /* Patterns */
 #define P_CPMVFMT 0
 #define P_CPMVRNM 1
-#define P_BATCHRNM 2
-#define P_ARCHIVE 3
-#define P_REPLACE 4
+#define P_ARCHIVE 2
+#define P_REPLACE 3
 
 static const char * const patterns[] = {
 	"sed -i 's|^\\(\\(.*/\\)\\(.*\\)$\\)|#\\1\\n\\3|' %s",
 	"sed 's|^\\([^#][^/]\\?.*\\)$|%s/\\1|;s|^#\\(/.*\\)$|\\1|' "
 		"%s | tr '\\n' '\\0' | xargs -0 -n2 sh -c '%s \"$0\" \"$@\" < /dev/tty'",
[...]

PR #490, opened 2020-03-09
Addressing #489

KlzXS:
Can we keep this open until we close the issue?

### DIFF
--- a/plugins/batchrename
+++ b/plugins/batchrename
@@ -51,7 +51,7 @@ while read -r num name; do
 		exit 1
 	fi
 
-	src=$(eval printf "%s" "\${$num}")
+	src=$(eval printf "%s" "\"\${$num}\"")
 
 	if [ -z "$src" ]; then
 		printf "%s: unknown item number %s\n" "$0" "$num" > /dev/stderr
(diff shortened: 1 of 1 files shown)


jarun:
Sure!

jarun:
I am merging this after good amount of testing. if something new comes up, we'll handle it in new PR.

jarun:
Also, for now, let the prints be on.

KlzXS:
> Also, for now, let the prints be on.

What prints?

jarun:
```
'./nnn' -> './nnn1'
'./nnn.1' -> './nnn.11'
'./plugins' -> './plugins1'
'./plugins' => './plugins1'
'./README.md' -> './README.md1'
```

KlzXS:
vidir called them a feature and I couldn't be bothered to implement a flag for them. Also they are handy as a debug print. So yeah, we can remove them later.

jarun:
```
'./plugins' -> './plugins1'
'./plugins' => './plugins1'
```

what's the difference?

KlzXS:
The one with `=>` prints only if the entry moved was a directory. Not sure why in that case `->` is also printed.

jarun:
I think we can leave it as it is for now.

PR #491, opened 2020-03-10
Add middle click copy

leovilok:
I think it's the most expected middle mouse button behaviour.

### DIFF
--- a/nnn.1
+++ b/nnn.1
@@ -380,9 +380,11 @@ separated by \fI;\fR:
     export NNN_MCLICK='p'
 
     NOTES:
-    1. Only the first character is taken into account.
-    2. Use your shell/editor's quoted insert function (usually \fICtrl+V\fR)
-       to set it to a \fICtrl+key\fR combo.
+    1. You can use the '^\fIkey\fR' syntax to assign it to a \fICtrl+key\fR combo, e.g.:
+
+    export NNN_MCLICK='^A'
+
+    2. Otherwise, only the first character is taken into account.
 .Ed
 .Pp
 \fBnnn:\fR this is a special variable set to the hovered entry before executing
--- a/nnn.1
+++ b/nnn.1
@@ -375,6 +375,16 @@ separated by \fI;\fR:
     export NNN_TRASH=1
 .Ed
 .Pp
+\fBNNN_MIDDLECLICK:\fR key emulated by a middle mouse click.
+.Bd -literal
+    export NNN_MIDDLECLICK='p'
+
+    NOTES:
+    1. Only the first character is taken into account.
+    2. Use your shell/editor's quoted insert function (usually \fICtrl+V\fR)
+       to set it to a \fICtrl+key\fR combo.
+.Ed
+.Pp
 \fBnnn:\fR this is a special variable set to the hovered entry before executing
 a command from the command prompt or spawning a shell.
 .Sh KNOWN ISSUES
--- a/nnn.1
+++ b/nnn.1
@@ -375,9 +375,9 @@ separated by \fI;\fR:
     export NNN_TRASH=1
 .Ed
 .Pp
-\fBNNN_MIDDLECLICK:\fR key emulated by a middle mouse click.
+\fBNNN_MCLICK:\fR key emulated by a middle mouse click.
 .Bd -literal
-    export NNN_MIDDLECLICK='p'
+    export NNN_MCLICK='p'
 
     NOTES:
     1. Only the first character is taken into account.
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -6603,7 +6603,10 @@ int main(int argc, char *argv[])
 #ifndef NOMOUSE
 	mmask_t mask;
 	char *middle_click_env = xgetenv(env_cfg[NNN_MCLICK], "\0");
-	middle_click_key = middle_click_env[0];
+	if (middle_click_env[0] == '^' && middle_click_env[1])
+		middle_click_key = CONTROL(middle_click_env[1]);
+	else
+		middle_click_key = middle_click_env[0];
 #endif
 	const char* const env_opts = xgetenv(env_cfg[NNN_OPTS], NULL);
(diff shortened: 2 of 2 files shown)


jarun:
Why? We already have right click copy in.

jarun:
We already have right click add to to selection in. Why do we have to have middle click copy and not move? Are there any other file managers which supports this? Even if someone mis-clicks a copy would be issued. Why would we want this?

jarun:
How about something much more useful? Like cycle context color?

leovilok:
> We already have right click add to to selection in. Why do we have to have middle click copy and not move?
The idea is to be able to copy files around using only the mouse. Moving the files could work too.

> Are there any other file managers which support this?
I don't think so. I think that graphical file managers tend to open a new tab on middle click, just as browsers. But middle click is used to paste (text) selection everywhere on x11.

> Even if someone mis-clicks a copy would be issued. Why would we want this?
This is indeed an issue, as most misclicks. It behaves just as if you pressed 'p'. I don't know if middle mouse miss click happen that often, but it might be more surprising than a keyboard missclick.

> How about something much more useful? Like cycle context color?
This deserves its own (keyboard) key first.

But middle click could just set `presel` to a custom, configurable, key. So that anyone can compile in the behaviour of any other allready available command.

What do you think about that ?
What would be the default behaviour ?

jarun:
> I don't know if middle mouse miss click happen that often

It happens with me often while scrolling the wheel.

> This deserves its own (keyboard) key first.

It doesn't have and that's something that would make it unique. But I do have an idea here. We already detect clicking on the context number. When left-clicked on the current context number, we can change the color to the next color code. We can have the same behaviour if we press the same context number which you are in.

But would that be confusing? What do you say?

> But middle click could just set presel to a custom, configurable, key.

That's a nice idea. By default we can use an unassigned key. But how would someone assign a Ctrl-key?

leovilok:
> When left-clicked on the current context number, we can change the color to the next color code. We can have the same behaviour if we press the same context number which you are in.
> 
> But would that be confusing? What do you say?
Might be a little confusing and hard to discover, but it's harmless, so I don't think it would be an issue.

> That's a nice idea. By default we can use an unassigned key. But how would someone assign a Ctrl-key?

If it's in `config.h` we can just use the `CONTROL()` macro.
If it's an env variable it can be passed raw in the shell or in an editor, usually with Ctrl+V.

leovilok:
Middle click action is now configurable with the `NNN_MIDDLECLICK` env variable.

jarun:
Can you share a working config example for middle click for me to test?

jarun:
Say for `^R`.

nnn.1
```diff
@@ -375,6 +375,16 @@ separated by \fI;\fR:
     export NNN_TRASH=1
 .Ed
 .Pp
+\fBNNN_MIDDLECLICK:\fR key emulated by a middle mouse click.
+.Bd -literal
+    export NNN_MIDDLECLICK='p'
+
+    NOTES:
+    1. Only the first character is taken into account.
+    2. Use your shell/editor's quoted insert function (usually \fICtrl+V\fR)
```
jarun: We need to put an example. I am not able to figure out what we need to set.

src/nnn.c
```diff
@@ -567,8 +570,9 @@ static const char * const messages[] = {
 #define NNN_COLORS 4
 #define NNNLVL 5
 #define NNN_PIPE 6
-#define NNN_ARCHIVE 7 /* strings end here */
-#define NNN_TRASH 8 /* flags begin here */
+#define NNN_MIDDLECLICK 7
```
jarun: Please use `NNN_MCLICK` for a short and sweet one. We will document it.

leovilok:
> Say for `^R`.

Github won't let me attach the file, probably because of the `^R`, but here it is:
http://ix.io/2dVF
Download it, `source` it, and lanch `./nnn`.

You can do it in the shell too, just type ^V before inserting ^R.

jarun:
Thanks! I'll test this. There are some minor review comments. Please address.

nnn.1
leovilok: Well it looks like

```
export NNN_MIDDLECLICK='^R'
```

But if you input it right, ^R is a actually a single character.

Of course if you use `bash` or `zsh` you can write `export NNN_MIDDLECLICK=$'\01'` for `^A`, `export NNN_MIDDLECLICK=$'\02'` for `^B`, etc.
`^R` is `\022`

jarun:
I think we can do better than this. We can take the standard notation in, like:

    export NNN_MCLICK='^V'
and parse the value to set the key to `'V' ^ 0x40`.

leovilok:
> I think we can do better than this. We can take the standard notation in, like:
> 
> ```
> export NNN_MCLICK='^V'
> ```
> 
> and parse the value to set the key to `'V' ^ 0x40`.

Yes.

nnn.1
jarun: As discussed, we will use the standard notation and parse. Please add 2 examples in the man page, one for `p` (as it is now) another for `^R`.

jarun: Also, please remove the notes.

leovilok: Ah, I didn't see that.

jarun:
Thank you!

leovilok:
Current `CONTROL()` macro only works for uppercase letters, it's basically `c ^ 0x40`.
If I replace it with `c & 0x1f` it works for both uppercase and lowercase letters.

jarun:
Thanks! I'll add it along with a minor man page change. I just merged the PR.

jarun:
Please review commit bf89040ace3105b6efde45ce3a078d5d594e5deb.

jarun:
This is a really nice feature addition!

leovilok:
To a nice file manager!

jarun:
I have updated the wiki configuration and mouse/keyboard sections: https://github.com/jarun/nnn/wiki/Usage#configuration

PR #494, opened 2020-03-13
Add x2sel plugin (reverse .cbcp)

leovilok:
This plugin is based on `.cbcp` and does basically the reverse operation.
It copies system clipboard newline-separated file list to selection.

I need help to test:

- [ ] MacOS's `pbpaste`
- [ ] Termux's `termux-clipboard-get`
- [ ] Windows stuff
- [x] `wl-paste` (done)

I originally intended to call it automatically in `-x` mode like `.cbcp`, but I'm still unsure.

This allows to copy files from e.g. `thunar` and paste them in `nnn`.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -68,6 +68,7 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | upload | Paste text to ix.io, upload binary to file.io | sh | curl, jq, tr |
 | vidthumb | Show video thumbnails in terminal | sh | [ffmpegthumbnailer](https://github.com/dirkvdb/ffmpegthumbnailer),<br>[lsix](https://github.com/hackerb9/lsix) |
 | wall | Set wallpaper or change colorscheme | sh | nitrogen/pywal |
+| x2sel | Copy newline-separated file list from system clipboard to selection | sh | tr, xsel/xclip<br>pbpaste/wl-paste<br>termux-clipboard-get<br>powershell.exe/`/dev/clipboard`|
 
 ## Invoking a plugin
 
--- a/plugins/x2sel
+++ b/plugins/x2sel
@@ -1,0 +1,57 @@
+#!/usr/bin/env sh
+
+# Description: Copy system clipboard newline-separated file list to selection
+# Requires: tr and
+#           xclip/xsel (Linux)
+#           pbpaste (macOS)
+#           termux-clipboard-get (Termux)
+#           powershell (WSL)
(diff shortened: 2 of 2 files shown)


jarun:
Can we rename it to `x2sel`?

leovilok:
> Can we rename it to `x2sel`?

Yes, done.

leovilok:
Tested Wayland's `wl-paste`.

jarun:
Thank you!

jarun:
@leovilok please check if the `tr` method works with file paths having space. I think I noticed some issues sometime back.

leovilok:
@jarun I checked with a filename containing a space: it worked.

Thanks for the merge, but MacOS, Termux, Windows stuff is still untested.

Some of these commands might add a newline at the end for exemple, and I'm not sure if the code still works in that case.

And windows commands are not `.cbcp` counterparts, as `clip.exe` doesn't seems to have an obvious "paste" version. So what I wrote there comes from a basic web search, and without windows access I can't check it at all.

jarun:
> MacOS, Termux, Windows stuff is still untested.

That's fine. Plugins are indicative. Platform-dependent stuff are supposed to be modified by the user to customize to his/her needs.

> Some of these commands might add a newline at the end

No it won't. File paths are strictly NUL-terminated in the selection file. Please fix that. It has to be fixed by us. It's a bad miss. ;)

> So what I wrote there comes from a basic web search, and without windows access I can't check it at all.

This is also fine.

jarun:
Probably you can use `tr` to convert newlines to NUL.

leovilok:
> No it won't. File paths are strictly NUL-terminated in the selection file. Please fix that. It has to be fixed by us. It's a bad miss. ;)

> Probably you can use `tr` to convert newlines to NUL.

Yes, of course, `x2sel` already does that.
But the expected clipboard content is a newline separated path list, as generated by `.cbcp` of any file manager, *without* a terminating newline.
Neither `xsel` nor `xclip` nor `wl-paste` add a newline at the end of the selection, so this is fine. But if some other program adds a `\n` at the end, `tr` turns it into a '\0', so `.selection` now ends with a `\0` and I don't know if it matters.

Of course I can check how clipboard content ends, but at the moment I don't know if it's usefull. But I guess it can be adapted by the user …

jarun:
> Of course I can check how clipboard content ends, but at the moment I don't know if it's usefull. But I guess it can be adapted by the user …

I think we can simply add a check to eliminate any terminating newline.

leovilok:
Well it turns out that command substitution already deletes trailing newlines. So no change is required.

jarun:
Awesome!

PR #495, opened 2020-03-15
Convert batchrename to bash and rename it

KlzXS:


jarun:
Please update the filename in nnn.c as well.

jarun:
The plugin README as well.

jarun:
I'll take care of the above 2.

PR #497, opened 2020-03-20
Add verbose option

KlzXS:
For now let it be environment variables.

How are these options (verbose and showing hidden files) accessed by `nnn`?

### DIFF
--- a/plugins/.nmv
+++ b/plugins/.nmv
@@ -17,6 +17,7 @@
 EDITOR="${EDITOR:-vi}"
 TMPDIR="${TMPDIR:-/tmp}"
 INCLUDE_HIDDEN="${INCLUDE_HIDDEN:-0}"
+VERBOSE="${VERBOSE:-0}"
 
 selection=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection
 exit_status=0
@@ -95,7 +96,9 @@ while read -r num name; do
 			done
 
 			if mv "$name" "$tmp"; then
-				printf "'%s' -> '%s'\n" "$name" "$tmp"
+				if [ "$VERBOSE" -ne 0 ]; then
+					printf "'%s' -> '%s'\n" "$name" "$tmp"
+				fi
 			else
 				printf "%s: failed to rename %s to %s: %s\n" "$0" "$name" "$tmp" "$!" > /dev/stderr
 				exit_status=1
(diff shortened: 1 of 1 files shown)


plugins/.nmv
```diff
@@ -17,6 +17,7 @@
 EDITOR="${EDITOR:-vi}"
 TMPDIR="${TMPDIR:-/tmp}"
 INCLUDE_HIDDEN="${INCLUDE_HIDDEN:-0}"
+VERBOSE="${VERBOSE:-0}"
```
maximbaz: Wondering if the last two variables should be prefixed with `NNN_` ?

jarun: > Wondering if the last two variables should be prefixed with NNN_ ?

Nope. There are not linked to `nnn`.

jarun:
> How are these options (verbose and showing hidden files) accessed by nnn?

If you export VERBOSE at your prompt, you'll get it in the script. `nnn` is not involved.

jarun:
Right now there is no way to send hidden or verbose to the script from `nnn`.

plugins/.nmv
jarun: Right now there's no way to export `nnn` options to scripts.

maximbaz: My thinking was simply to namespace custom variables that are not standard (like EDITOR) and related to this script, otherwise if many scripts all use the same `VERBOSE`, exporting `VERBOSE=1` will affect ALL scripts, and not just this one.

jarun: OK, got it!

In that case we need to visit all scripts. We can prefix with: `NNN_PLUG_`.

Can you pick this up and raise a PR? I would merge this one as it is and you cna modify it as well.

jarun:
Thank you!

jarun:
@KlzXS I think we also need a `-f` option for `rm`. Removing a dir with large number of files gets very tedious otherwise.

plugins/.nmv
maximbaz: I realized that this will break lots of people's configs, and also it's hard for me to judge if a variable is common enough (i.e. if it must be prefixed or not), such as `GUIPLAYER` in `boom` or `BROWSER` and `READER` in `gutenread` - maybe just for me because I'm not using any of the standard plugins.

I'll drop this idea, doesnt make sense to introduce chaos if nobody has ever complained and this issue is theoretical.

jarun: Done!

PR #501, opened 2020-03-31
Add .desktop file for XDG compatible DEs

leovilok:
Install with `make install-desktop`
It now appears in application menus, search, etc. in XDG compatible
desktop environments (like GNOME, KDE, ...).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ VERSION = $(shell grep -m1 VERSION $(SRC) | cut -f 2 -d'"')
 
 PREFIX ?= /usr/local
 MANPREFIX ?= $(PREFIX)/share/man
+DESKTOPPREFIX ?= $(PREFIX)/share/applications
 STRIP ?= strip
 PKG_CONFIG ?= pkg-config
 INSTALL ?= install
@@ -81,6 +82,7 @@ DISTFILES = src nnn.1 Makefile README.md LICENSE
 SRC = src/nnn.c
 HEADERS = src/nnn.h
 BIN = nnn
+DESKTOPFILE = misc/desktop/nnn.desktop
 
 all: $(BIN)
 
@@ -91,6 +93,13 @@ $(BIN): $(SRC) $(HEADERS)
 debug: $(BIN)
 norl: $(BIN)
 noloc: $(BIN)
+
+install-desktop: $(DESKTOPFILE)
+	$(INSTALL) -m 0755 -d $(DESTDIR)$(DESKTOPPREFIX)
+	$(INSTALL) -m 0644 $(DESKTOPFILE) $(DESTDIR)$(DESKTOPPREFIX)
+
+uninstall-desktop:
+	$(RM) $(DESTDIR)$(DESKTOPPREFIX)/$(DESKTOPFILE)
 
 install: all
 	$(INSTALL) -m 0755 -d $(DESTDIR)$(PREFIX)/bin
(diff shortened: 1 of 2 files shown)


leovilok:
We might want to install this by default on Linux/BSDs?

Makefile
```diff
@@ -2,6 +2,7 @@ VERSION = $(shell grep -m1 VERSION $(SRC) | cut -f 2 -d'"')
 
 PREFIX ?= /usr/local
```
jarun: We need to add this to the haiku Makefile as well. See misc/haiku dir.

nnn.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
```
jarun: This goes somewhere in misc dir too (maybe misc/desktop). Otherwise it's an eyesore. Don't litter. ;)

jarun:
See this wiki section: https://github.com/jarun/nnn/wiki/Advanced-use-cases#create-the-nnndesktop-file

Why is the Gnome .desktop file different?

If we have this in (and this single file works for all) we also need to update the above wiki section.

leovilok:
> See this wiki section: https://github.com/jarun/nnn/wiki/Advanced-use-cases#create-the-nnndesktop-file

Argh, I don't know why I could't find this on the wiki ... I dont understand github search ...

> Why is the Gnome .desktop file different?

It just adds GNOME specific stuff, but it seems it's just cosmetic additions nothing really required. (I can't really check on GNOME right now though, installing it is a bit heavy for my confined connexion ^^)

nnn.desktop
leovilok: I put it in misc/ , as there should not be any other .desktop files ...

Makefile
leovilok: I don't know Haiku, does it use XDG specs too?

jarun:
> I dont understand github search

It's garbage. :)

> but it seems it's just cosmetic additions nothing really required

:+1:

Makefile
jarun: @annagrram please confirm this.

nnn.desktop
jarun: Please keep under misc/desktop... it's much easier to link and self-explanatory.

leovilok: done

leovilok:
> See this wiki section: https://github.com/jarun/nnn/wiki/Advanced-use-cases#create-the-nnndesktop-file

Note that when started from a display manager (GDM, ...) the desktop environment might start `nnn` in a terminal without calling any login shell. So if the user exported its `NNN_*` variables in e.g. `.bashrc`, `.zshrc`, `.bash_profile`, and *not* in `.profile` or `.xprofile` it might never be exported.

In the wiki's `.desktop`, `nnn` calls are wrapped to call `bash -l` first in order to source shell specific profile files. But in my opinion if you donc call nnn from a shell you have to expect that shell specific exports are not done.

This should be mentionned in the wiki, though.

0xACE:
> > I dont understand github search
> 
> It's garbage. :)

Makefile
annagrram: Not that I know of

annagrram: But, we can add `nnn` to the Applications menu in a different way if that's what you want

annagrram: Seems like pressing it in the Application menu doesn't work...
Maybe it just doesn't know how to open it in the terminal and it wants it to be a regular desktop app.

jarun: OK, then we spare Haiku this time ;)

jarun:
Thank you!

jarun:
@leovilok could you clean up and update the wii page I mentioned? Please link to this script.

leovilok:
@jarun I updated the wiki page: added a *Desktop integration* section and tried to reorganise what was already there.

jarun:
Thank you!

PR #502, opened 2020-03-31
Added fzfz plugin

nwaywood:
This plugin reads the data from the `z` database file and displays the contents in an `fzf` pane to let the user `cd` to any directory that is indexed by `z`

### DIFF
--- a/plugins/fzfz
+++ b/plugins/fzfz
@@ -10,15 +10,14 @@
 if [ "$(cmd_exists fzf)" -eq "0" ]; then
     if [ -f "$HOME/.z" ]; then
         # I read the data from z's file instead of calling the z command so that the data doesn't need to be processed twice
-        sel=$(cat $HOME/.z | awk -F "|" '{print $1}' | fzf | awk '{$1=$1};1')
+        sel=$(awk -F "|" '{print $1}' "$HOME/.z" | fzf | awk '{$1=$1};1')
         # NOTE: Uncomment this line and comment out the line above if you want to see the weightings of the dir's in the fzf pane
-        # sel=$(cat $HOME/.z | awk -F "|" '{printf "%s %s\n", $2, $1}' | fzf | sed 's/^[0-9,.]* *//' | awk '{$1=$1};1')
+        # sel=$(awk -F "|" '{printf "%s %s\n", $2, $1}' cat "$HOME/.z" | fzf | sed 's/^[0-9,.]* *//' | awk '{$1=$1};1')
     else
         exit 1
     fi
 else
-	exit 1
+    exit 1
 fi
 
 printf "%s" "0$sel" > "$NNN_PIPE"
-
--- a/plugins/fzfz
+++ b/plugins/fzfz
@@ -1,0 +1,24 @@
+#!/usr/bin/env sh
+
+# Description: cd to any dir in the z database using an fzf pane
+#
+# Shell: POSIX compliant
+# Author: Nick Waywood
+
+. "$(dirname "$0")"/.nnn-plugin-helper
+
+if [ "$(cmd_exists fzf)" -eq "0" ]; then
+    if [ -f "$HOME/.z" ]; then
+        # I read the data from z's file instead of calling the z command so that the data doesn't need to be processed twice
+        sel=$(cat $HOME/.z | awk -F "|" '{print $1}' | fzf | awk '{$1=$1};1')
+        # NOTE: Uncomment this line and comment out the line above if you want to see the weightings of the dir's in the fzf pane
+        # sel=$(cat $HOME/.z | awk -F "|" '{printf "%s %s\n", $2, $1}' | fzf | sed 's/^[0-9,.]* *//' | awk '{$1=$1};1')
+    else
+        exit 1
+    fi
+else
+	exit 1
+fi
+
+printf "%s" "0$sel" > "$NNN_PIPE"
+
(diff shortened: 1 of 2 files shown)


jarun:
@KlzXS please take a look.

jarun:
@nwaywood shellcheck fails with:

```
In plugins/fzfz line 13:
        sel=$(cat $HOME/.z | awk -F "|" '{print $1}' | fzf | awk '{$1=$1};1')
                  ^-- SC2086: Double quote to prevent globbing and word splitting.
                  ^-- SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
```

Please fix.

nwaywood:
@jarun fixed the errors 👍

Sorry about that, I don't have shellcheck on my machine.

plugins/fzfz
```diff
@@ -0,0 +1,23 @@
...
+# Author: Nick Waywood
+
+. "$(dirname "$0")"/.nnn-plugin-helper
+
+if [ "$(cmd_exists fzf)" -eq "0" ]; then
+    if [ -f "$HOME/.z" ]; then
+        # I read the data from z's file instead of calling the z command so that the data doesn't need to be processed twice
+        sel=$(awk -F "|" '{print $1}' "$HOME/.z" | fzf | awk '{$1=$1};1')
+        # NOTE: Uncomment this line and comment out the line above if you want to see the weightings of the dir's in the fzf pane
+        # sel=$(awk -F "|" '{printf "%s %s\n", $2, $1}' cat "$HOME/.z" | fzf | sed 's/^[0-9,.]* *//' | awk '{$1=$1};1')
```
KlzXS, line 15: This doesn't work for me, it shows `fzf` with zero entries.

KlzXS: It clipped the line above as well. The line in question is 15, with the note above.

nwaywood: @KlzXS mmm that's weird. Can you confirm that your `z` database is located at `$HOME/.z`?

nwaywood: And could you show me an example line from your `.z` database file?

nwaywood: > It clipped the line above as well. The line in question is 15, with the note above.

I'm not sure what you mean by this, line 15 is a comment.

KlzXS: It's there. Line 13 works, it's just 15.

Here's the plugin folder line:
```
/home/user/.config/nnn/plugins|4|1585724118
```

nwaywood: ah ok I understand now, thanks. When I have a sec I'll double check that line 15 works for me.

nwaywood: ok @KlzXS can you please try it again? It should work now

jarun:
@KlzXS please ping me here when you guys are done.

plugins/fzfz
KlzXS: Now it works.

KlzXS:
Works as expected @jarun.

0xACE:
> Sorry about that, I don't have shellcheck on my machine.

For future reference: https://www.shellcheck.net/

jarun:
Merged. Thanks guys!

jarun:
@nwaywood why no fzy support? In all our plugins we support both. You can choose to keep fzf as default.

jarun:
@nwaywood On follow ups, I expect responses from contributors to be as prompt as they are before PRs are merged.

nwaywood:
> why no fzy support? In all our plugins we support both. You can choose to keep fzf as default.

Sorry @jarun I did see that some of the other plugins support both but I have never used `fzy` before and am not familiar with it's usage... 

Hopefully this plugin will still be useful for `fzf`/`z` users though :)

jarun:
> I did see that some of the other plugins support both but I have never used fzy before and am not familiar with it's usage...

I think in this case you can use it without any option like fzy. It's easy to compile from source as well.

jarun:
> Hopefully this plugin will still be useful for fzf/z users though :)

Plugins are indicative but we try to be remain flexible.

jarun:
In other cases I would have pitched in. Unfortunately, I don't use z database and can't spend time in understanding, configuring and then cleanup. Please confirm if you can add fzy support, otherwise we can revert this change. You can retain the fzfz plugin in your personal fork.

nwaywood:
> I think in this case you can use it without any option like fzy.

Ok in that case I will install `fzy` tonight and try using it the same as `fzf`. If that works I'll add support. If it doesn't though, like you, am not interested in understanding, configuring and cleaning up a tool that I don't use.

Will update you tonight when I try it

jarun:
Thank you for your understanding!

To give you a context on the fzy usage... we actually supported fzy first in `nnn` because it's in C, is tiny and much faster than any other fuzzy tool. However, on popular demand we had to support fzf later (and for good reasons, it has tons of features and is much more configurable).

Now, for most cases fzy just works without any options (fzf does too, but we tend to be fancy with it) so I am kind of persistent that we retain fzy support in all plugins.

On my smartphone (Termux) I now have fzf because I can trim the complete dir name and just list the filenames. But at the same time I intend to keep development direction of `nnn` above the personal preferences of the devs (as long as it doesn't compromise performance or has an easy workaround).

nwaywood:
@jarun added support for `fzy` and opened a new PR https://github.com/jarun/nnn/pull/505

PR #513, opened 2020-04-12
Opinionated xstrlcpy changes

saagarjha:
Two major changes to xstrlcpy. One is that I unvectorized it because every platform I could find does a memcpy at a stride of at least 8 bytes on x86_64, and I couldn't come up with a way to get rid of the `strlen` and remain vectorized in C. (There's really no good way to get them to emit the `pcmpeqb` that we might want I could find; I tried for a while and gave up since compilers can't really understand alignment guarantees…). Also, I changed `xstrlcpy` to `xstrsncpy` because the operation we're performing isn't actually a `strlcpy` so I felt it was less confusing.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -359,7 +359,6 @@ static kv *bookmark;
 static kv *plug;
 static uchar tmpfplen;
 static uchar blk_shift = BLK_SHIFT_512;
-static const uint _WSHIFT = (LONG_SIZE == 8) ? 3 : 2;
 #ifndef NOMOUSE
 static int middle_click_key;
 #endif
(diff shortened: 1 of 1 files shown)


jarun:
Thank you! Definitely much cleaner than the earlier implementation.

jarun:
An unrelated question on https://github.com/jarun/nnn/blob/3bab34e962f3c1b6167afcefb81aa1f1117f7794/src/nnn.c#L760

This is a test and clear loop because in more than 99% of the cases the bitmap (of inodes to detect hard links in case of disk usage calcualtion) will be sparsely populated. I am not sure if memset() checks the byte value and optimize to ensure it does not re-clear bytes which are already clear. What do you say?

saagarjha:
Hmm, I'm not a performance expert so I would defer to benchmarks if you have any, but I would expect that `memset` would the right thing to use here in any case; I don't actually think there is any benefit to avoiding the write. If you knew beforehand which bytes you wanted to clear you could do much better, of course, but here you're doing a read (likely from the cache) anyways so there can't be that much of a difference. I know glibc's `memset` uses AVX 2 if you have it so it should be really fast anyways.

jarun:
OK. Thanks for the insight! :+1:

saagarjha:
Just FYI, in the course of implementing this pull request I did enough research to finish writing [one of the blog posts](https://saagarjha.com/blog/2020/04/12/designing-a-better-strcpy/) on my to-do list. TL;DR: you may want to take a look at `memccpy` as well.

jarun:
Thanks for sharing! Nice write-up too! :)

jarun:
I've updated `xstrsncpy()` to use `memccpy()`. Thanks for the tip!

Also, many thanks for being patient on HN. If you look at `nnn`, even after 3 yrs of development, it's a very busy project. All of us contributing to it have other personal and professional commitments so it gets difficult to prioritize even major improvements at times.

saagarjha:
Unlikely to actually be useful, but you might find it interesting that newer Intel processors actually [seem to perform better on redundant zero stores](https://travisdowns.github.io/blog/2020/05/13/intel-zero-opt.html).

jarun:
Thank you! It would probably have its use in event handling where we do clear the buffer every time. And it's not too small.

PR #524, opened 2020-04-17
nuke: add lowdown as alternative markdown viewer

leovilok:
See https://kristaps.bsd.lv/lowdown/
This would be a fallback to *glow* (it's an equivalent in C).


jarun:
Awesome!

PR #527, opened 2020-04-18
Add nvim as a diff tool for getplugs plugin

akrifari:
On current version of `getplugs`, it doesn't seem support `neovim` as comparison/diff tool. I believe it could be better if we add `neovim` diff tool as well, since `neovim` uses different command than `vim` and `gvim` for comparing files.

https://github.com/jarun/nnn/blob/12df1ab083d9692d97c2bf248000cc5e94a0445d/plugins/getplugs#L17

Here's related error message:

```
Plugin ./nuke already exists and is different.\nKeep (k), merge (m), overwrite (o) [default: k]? m
/home/akrifari/.config/nnn/plugins/getplugs: line 17: vimdiff: command not found
```

### DIFF
--- a/plugins/getplugs
+++ b/plugins/getplugs
@@ -22,7 +22,7 @@ merge () {
 }
 
 prompt () {
-	printf "%s" "Plugin $1 already exists and is different.\n"
+	printf "%s\n" "Plugin $1 already exists and is different."
 	printf "Keep (k), merge (m), overwrite (o) [default: k]? "
 	read -r operation
 
--- a/plugins/getplugs
+++ b/plugins/getplugs
@@ -8,10 +8,10 @@
 CONFIG_DIR=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/
 PLUGIN_DIR=${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins
 
-is_cmd_exists () {
-    which "$1" > /dev/null 2>&1
-    echo $?
-}
+# is_cmd_exists () {
+#     which "$1" > /dev/null 2>&1
+#     echo $?
+# }
 
 merge () {
 	if which nvim >/dev/null 2>&1; then
--- a/plugins/getplugs
+++ b/plugins/getplugs
@@ -14,7 +14,11 @@ is_cmd_exists () {
 }
 
 merge () {
-	vimdiff +0 "$1" "$2"
+	if which nvim >/dev/null 2>&1; then
+		nvim -d "$1" "$2"
+	else
+		vimdiff +0 "$1" "$2"
+	fi
 }
 
 prompt () {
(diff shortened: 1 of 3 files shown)


plugins/getplugs
```diff
@@ -14,7 +14,11 @@ is_cmd_exists () {
 }
 
 merge () {
-	vimdiff +0 "$1" "$2"
+	if which nvim >/dev/null 2>&1; then
```
akrifari, line 17: Should I use`is_cmd_exists` aforementioned or `cmd_exists` in `.nnn-plugin-helper`?

jarun: It's OK not to use any of these. Probably just comment out `is_cmd_out()` in this file as no one is calling it.

plugins/getplugs
```diff
@@ -14,7 +14,11 @@ is_cmd_exists () {
...
+		nvim -d "$1" "$2"
```
jarun, line 18: There are 2 minor items you can take care of in the current PR:

- update the `diffs` plugin to invoke `nvim` if it exists
- update `nuke` to invoke `$EDITOR` instead of vi

akrifari:
Oops, will fix this one

akrifari:
I noticed that this line not produces output as expected due to misplaced `\n`. Should I make a change in current PR too?

https://github.com/jarun/nnn/blob/4b5ecbe8fc1d16f11325baf268bf484043e8f531/plugins/getplugs#L21

Output:
```
Plugin ./nuke already exists and is different.\nKeep (k), merge (m), overwrite (o) [default: k]? 
```

Expected:
```
Plugin ./nuke already exists and is different.
Keep (k), merge (m), overwrite (o) [default: k]? 
```

Suggestion:
```diff
-	printf "%s\n" "Plugin $1 already exists and is different.\n"
+	printf "%s\n" "Plugin $1 already exists and is different."
```
or using different format

```diff
-	printf "%s" "Plugin $1 already exists and is different.\n"
+	printf "%b" "Plugin $1 already exists and is different.\n"
```

jarun:
Yes, it would be good to fix the question as well.

@KlzXS can you please review this one?

jarun:
Something I missed: in `nuke` we also use `less -R` in some places. Would it be possible to have `PAGER` and `less -R` as fallback? (like `EDITOR`)

akrifari:
Working on it

jarun:
Thank you! Looks good to me.

@KlzXS please review.

leovilok:
Note that some commands in `nuke` output ANSI colors/formatting, so if `$PAGER` is not `less -R`, it might not work well with escape sequences.

jarun:
> Note that some commands in nuke output ANSI colors/formatting, so if $PAGER is not less -R, it might not work well with escape sequences.

I think that's fine as plugins are indicative (customize and use) and we have it correctly as the fallback.

jarun:
Thank you!

PR #528, opened 2020-04-19
Add nbak plugin to backup all nnn config

leovilok:
This is WIP, only creating the archive is supported at this time.
This needs more testing.
I don't support `fish` (is it even possible to write fish compatible POSIX shell script?).

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -50,6 +50,7 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | moclyrics | Show lyrics of the track playing in moc | sh | [ddgr](https://github.com/jarun/ddgr), [moc](http://moc.daper.net/) |
 | mocplay | Append (and/or play) selection/dir/file in moc | sh | [moc](http://moc.daper.net/) |
 | mp3conv | Extract audio from multimedia as mp3 | sh | ffmpeg |
+| nbak | Creates an archive in CWD with all `nnn` (shell, dir) config | sh | tar, awk |
 | nmount | Toggle mount status of a device as normal user | sh | pmount, udisks2 |
 | nuke | Sample file opener (CLI-only by default) | sh | _see in-file docs_ |
 | oldbigfile | List large files by access time | sh | find, sort |
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -50,7 +50,7 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | moclyrics | Show lyrics of the track playing in moc | sh | [ddgr](https://github.com/jarun/ddgr), [moc](http://moc.daper.net/) |
 | mocplay | Append (and/or play) selection/dir/file in moc | sh | [moc](http://moc.daper.net/) |
 | mp3conv | Extract audio from multimedia as mp3 | sh | ffmpeg |
-| nbak | Creates an archive in CWD with all `nnn` (shell, dir) config | sh | tar, awk |
+| nbak | Backs up `nnn` config | sh | tar, awk |
 | nmount | Toggle mount status of a device as normal user | sh | pmount, udisks2 |
 | nuke | Sample file opener (CLI-only by default) | sh | _see in-file docs_ |
 | oldbigfile | List large files by access time | sh | find, sort |
--- a/plugins/nbak
+++ b/plugins/nbak
@@ -18,7 +18,7 @@ configdir="${XDG_CONFIG_HOME:-$HOME/.config}/nnn"
 workdir="$PWD"
 
 if [ "$1" = "--show" ] ; then
-	if [ -z "$2" -o ! -r "$2" ] ; then
+	if [ -z "$2" ] || [ ! -r "$2" ] ; then
 		echo "Please provide a backup file"
 		exit 1
 	fi
@@ -45,8 +45,8 @@ env | sed "s/'/'\\\\''/" |\
 case "$shellname" in
 	bash)
 		for name in $nnn_names ; do
-			if [ "$(type -t "$name")" = "function" ] ; then
-				type "$name" | tail -n+2 >> "$shellfile"
+			if [ "$(bash -c "type -t $name")" = "function" ] ; then
+				bash -c "type $name" | tail -n+2 >> "$shellfile"
 			elif alias "$name" >/dev/null 2>&1 ; then
 				alias "$name" >> "$shellfile"
 			fi
--- a/plugins/nbak
+++ b/plugins/nbak
@@ -61,7 +61,9 @@ if [ -d "$configdir" ] ; then
 	cp -r "$configdir" config.d
 fi
 
-tar caf "$workdir/$outfile" ./*
+printf "Saving as '%s' ... " "$workdir/$outfile"
+
+tar caf "$workdir/$outfile" ./* && echo "Done" || echo "Failed"
 
 cd "$workdir" && rm -rf "$tempdir"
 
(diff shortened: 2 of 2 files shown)


leovilok:
So currently this plugin creates a backup archive containing:
* a sourceable shell file named `config.$shellname`, containing:
  - all exorted `NNN_*` env vars
  - definition for each shell alias/function nammed either `n` or `nnn` (only for bash/zsh)
* a copy of `${XDG_CONFIG_HOME:-$HOME/.config}/nnn`, in a dir named `config.d`

@jarun there are a few issues with what you asked for:
1. This plugin is executed by `/usr/bin/env sh`, which might not be the user's interactive shell, where he could have his functions/aliases.
2. You mentionned `fish`, but `fish` is not a POSIX compatible shell, and no one would set his default non interactive `sh` to `fish`.
3. If `nbak --restore filename` only shows the content of `config.{ba,z}sh`, the option is more `--show` than `--restore`, and in that case, what about the config directory content? We could either:
    * simply extract it in a new directory, in the working directory
    * print a recursive diff, comparing it to the current `${XDG_CONFIG_HOME:-$HOME/.config}/nnn`
    * extract it in a temp dir and drop an interactive shell
    * ...

leovilok:
I added a `--show` option.

jarun:
@0xACE please take a look at my idea on this in the project discussion and review this.

jarun:
@leovilok where is this at? Done? In progress? If it's done I can review.

plugins/nbak
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env sh
+
+# Description: Backup of all nnn config
+#
+# Shell: POSIX compliant
+# Author: Léo Villeveygoux
+
+nnn_names="n nnn"
```
jarun, line 8: please use `nnn_aliases`.

plugins/nbak
```diff
@@ -0,0 +1,80 @@
...
+
+shellname="$(basename "$SHELL")"
+
+shellfile="config.sh"
+
+configdir="${XDG_CONFIG_HOME:-$HOME/.config}/nnn"
+
+workdir="$PWD"
+
+if [ "$1" = "--show" ] ; then
```
jarun, line 20: What's this for?

plugins/nbak
```diff
@@ -0,0 +1,80 @@
...
+		;;
+
+	*)
+		echo "Unknown shell, skipping alias/function checking." >&2
+		;;
+esac
+
+# Config dir backup
+if [ -d "$configdir" ] ; then
+	cp -r "$configdir" config.d
```
jarun, line 72: Create one dir, maybe `nnn-$(whoami)@$(hostname)` and copy the nnn dir plus the config.txt within it. Then archive it.

plugins/nbak
```diff
@@ -0,0 +1,80 @@
...
+# Shell: POSIX compliant
+# Author: Léo Villeveygoux
+
+nnn_names="n nnn"
+
+outfile="nnn-$(whoami)@$(hostname).tar.bz2"
+
+shellname="$(basename "$SHELL")"
+
+shellfile="config.sh"
```
jarun, line 14: Just `config.txt`. Let them run their poison on another system manually.

plugins/nbak
leovilok: Done.

plugins/nbak
leovilok: Why am I saving original `$PWD`? Because I like to work in a `mktemp -d` temp dir.

The `--show` option? It's to print out (for easy copy/pasting) the content of the `config.txt` file in a `nbak` generated archive, which contains all `nnn` related environment and shell functions/aliases.

plugins/nbak
leovilok: Now the archive contains a `nnn-$(whoami)@$(hostname)` dir containing all the `nnn` config dir content and the `config.txt` file, at the same level.
Is that what you asked for?

plugins/nbak
leovilok: Changed `config.sh` to `config.txt`.
But do you mean that you don't want shell function/aliases backup?

leovilok:
> @leovilok where is this at? Done? In progress? If it's done I can review.

It's basically done, ignoring the few issues I talked about earlier, mainly:
1. How to restore the config dir (or do we even want to do that)?
2. We don't know if we're executed in the user's main interactive shell, so we might miss functions/aliases definition.

plugins/nbak
jarun: We can skip `--show`, it's a text file.

leovilok: `--show` option removed

> We can skip `--show`, it's a text file.

It's a text file *in an archive*, but it's not really worth is, yes.

leovilok:
I forgot about the FUSE remotes, if one is mounted it will be a mess.

leovilok:
> I have zsh. When I run the plugin I get the following error:
> 
> ```
> $ ./nbak 
> ./nbak: 64: ./nbak: functions: not found
> alias: n not found
> ./nbak: 64: ./nbak: functions: not found
> alias: nnn not found
> Saving as '/home/vaio/nnn-vaio@tux.tar.bz2' ... Done
> ```

Try: `env sh -c 'echo $SHELL'`

If it's `*/zsh` it must mean that zsh has some sort of POSIX mode where non-standard builtins are not available ... I'll do the same as bash anyways.

jarun:
I get:

```
$ env sh -c 'echo $SHELL'
/usr/bin/zsh
```

leovilok:
@jarun try the last commit.
(I use bash and on my system even `zsh -c 'echo "$SHELL"'` prints `/bin/bash`, zsh is tricky)

jarun:
Almost there.

In the top level directory `nnn-vaio@tux` have 2 things - config.txt and `nnn` directory.

plugins/nbak
```diff
@@ -0,0 +1,84 @@
...
+
+if [ ! -d "$tempdir" ]; then
+	echo "Can't create work directory." >&2
+	exit 1
+fi
+
+cd "$tempdir" || exit 1
+
+# Backing up config dir content
+cp -r "$configdir" "$outdir" || mkdir "$outdir" || exit 1
```
jarun, line 39: Please copy the configdir, i.e. the nnn dir. It keeps everything clean.

leovilok: Done.
It's also easier, because I don't have to backup old `config.txt` lying around in `.config/nnn`.

plugins/README.md
```diff
@@ -50,6 +50,7 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | moclyrics | Show lyrics of the track playing in moc | sh | [ddgr](https://github.com/jarun/ddgr), [moc](http://moc.daper.net/) |
 | mocplay | Append (and/or play) selection/dir/file in moc | sh | [moc](http://moc.daper.net/) |
 | mp3conv | Extract audio from multimedia as mp3 | sh | ffmpeg |
+| nbak | Creates an archive in CWD with all `nnn` (shell, dir) config | sh | tar, awk |
```
jarun, line 53: Please make it shorter:

Backs up `nnn` config

jarun:
Works as it should now! :+1:

plugins/README.md
jarun: Or, maybe 'Backup `nnn` configuration'.

leovilok: Done.

jarun:
Thank you!

0xACE:
> @0xACE please take a look at my idea on this in the project discussion and review this.

I've been away for a couple of days, and github notification thing changed. So I'm a bit lost...

Anyhow, what input exactly am I suppose to give here? Just regarding this plugin? or more than this?

I really should get around to getting in sync with master, I'm still on 9cf77708a1e1d9ec94afacd68847dcb6e98e1c27

I'll be reading the other thread you mentioned me in until i understand this situation...

jarun:
@0xACE no problem! I took care of the review.

Issue #529, opened 2020-04-19
alignment of nodes listed in Batch rename


harriott:
When I hit `r` in a folder, in this example a folder of music tracks, with the recently added feature of file-numbering on the left, there is a visual step at file number 10:
```
1 ./01 The Monitor
2 ./03 Click, Click, Click, Click
3 ./04 The Chinatown Bus
4 ./05 Flight 180
5 ./06 Like Castanets
6 ./07 Butterfly Nets
7 ./08 Shrinking Violet
8 ./09 Corazon
9 ./10 Middle Management
10 ./11 Choose Again
11 ./12 The News From Your Bed
```
This staggering means that I can't visually select an entire column in `Vim` (here to add a hypen between track number and title) in one go, but have to do the same operation twice (once for files 1-9, then again for the rest).

While in this case, as I'm using `vim`, and there's a consistency in the start of the file names, I can equally `:%s#\(\d\d\) #\1 - #` to insert my desired hyphen, but, for more general use-cases, **it'd be sweet if you could pad out the first 9 file numbers**.

jarun:
@KlzXS I think this is a nice improvement. Probably the padding needs to be done based on the number of files taken as input. Can you take a look?

KlzXS:
Added the padding in a PR. After editing the padding can freely be removed or added further as the script ignores leading and trailing whitespace on a line.

jarun:
> as the script ignores leading and trailing whitespace on a line

What if the filename has a leading or trailing whitespace OR the user wants to add?

KlzXS:
> What if the filename has a leading or trailing whitespace OR the user wants to add?

As my old professor used to say: that's a sad story. Unless you want to use some other delimiter which I don't think we should.

KlzXS:
Maybe if we used something else instead of `read` but is it really worth it? How names their files with leading and trailing whitespace?

jarun:
> How names their files with leading and trailing whitespace?

I think we don't need to over-engineer either if someone wants to use crypic names for their files. Just add a note in the plugin that space is the delimiter and should not be used as a suffix or prefix.

Issue #522, opened 2020-04-16
rewrite fzopen plugin


akimdi:
#### Environment details (Put `x` in the checkbox along with the information)

[x] Operating System: ArchLinux
[x] Desktop Environment: i3wm
[x] Terminal Emulator: kitty
[x] Shell: zsh 5.8
[ ] Custom desktop opener (if applicable):
[ ] Issue exists on `nnn` master

Is it possible to rewrite the plugin for a more comprehensive search?
Skim contains the same as fzf. In addition it has several improvements.
Mainly difference fuzzy search algorithm.
something like this `xdg-open $(sk -m)`
https://github.com/lotabout/skim#difference-to-fzf

jarun:
@KlzXS please check.

jarun:
Are you suggesting we use skim instead of fzf?

KlzXS:
I think we can add skim to this plugin. I don't know about adding it as a default.
What is the main selling point, the multi-select feature or is it the interactive mode? From what I've gathered that is the only difference besides the "different" algorithm.

akimdi:
Yes. Skim instead of fzf.
Skim includes everything that is in fzf.
Аnd additionally opens up other possibilities.

skim can search by folders and files
fzf searches by only file name

for example
https://i.imgur.com/klFsgsX.png

simple use case
You need to find the info file
there are several such files of the same name, they are different and are located in different folders, but they are called the same.
you forgot the name of the file itself, but remember that it was in folder temp
skim will help here
simply put, this is the difference between fuzzy search skim and fzf
https://youtu.be/iLTlQ2aTUB0

in the current plugin it is done like this
https://github.com/jarun/nnn/blob/c63fb28effb2fbae903928b9b9453974c6e96c9f/plugins/fzopen#L28

I tried to do something like
```xdg-open $ (sk -m)```

it works, but not always
because I don’t know how to escape spaces and other characters

akimdi:
Yes.
multi-select feature and interactive mode is a bonus

jarun:
> Skim instead of fzf.

fzf is way more popular than I thought myself. We actually supported only fzy earlier. And then I noticed fzy author doesn't respond to defects on GitHub. And everyone uses fzf.

You can have skim as fallback but not the default.

> because I don’t know how to escape spaces and other characters

You have to invoke xdg-open which takes only one file. That's what fzopen does, it opens the file. How you'll do is up to you. :)

akimdi:
[...]

PR #531, opened 2020-04-20
Plugin improvements

KlzXS:
Addressing #522 and #529.

### DIFF
--- a/plugins/.nmv
+++ b/plugins/.nmv
@@ -43,7 +43,10 @@ else
 	fi
 fi
 
-printf "%s" "$arr" | awk '{print NR " " $0}' > "$dst_file"
+lines=$(printf "%s\n" "$arr" | wc -l)
+width=${#lines}
+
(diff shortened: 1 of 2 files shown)


plugins/.nmv
```diff
@@ -43,7 +43,10 @@ else
 	fi
 fi
 
-printf "%s" "$arr" | awk '{print NR " " $0}' > "$dst_file"
+lines=$(printf "%s\n" "$arr" | wc -l)
+width=${#lines}
+
+eval "printf \"%s\" \"$arr\" | awk '{printf(\"%${width}d %s\\n\", NR, \$0)}' > \"$dst_file\""
```
leovilok, line 49: Did you add this `eval` just to insert `${width}` in the awk string?
In that case it's not needed. (It will also expand `$arr` too early, I believe).
```suggestion
printf "%s" "$arr" | awk '{printf("%'"${width}"'d %s\n", NR, $0)}' > "$dst_file"
```

KlzXS: Right. Didn't realize I could do that.

jarun:
Thanks guys!

PR #539, opened 2020-04-26
Optimize dups

KlzXS:
There might be more room for improvement but this still brings the running time in my home from several minutes to 5 seconds.


jarun:
Thank you!

Issue #540, opened 2020-04-28
nnn (n³) logo discussion


leovilok:
Here are a few ideas/tries for `nnn`'s logo (n³):
![nnn-logo-ideas](https://user-images.githubusercontent.com/1430862/80529681-17780f80-8998-11ea-9039-dfd128112b3c.png)

Colors are just here to differentiate variations at this point.

Any comment/criticism/new idea is welcome.

jarun:
Second one in the fourth row looks good to me.

jarun:
Can we have an image of the first one in the fourth row without the shadow?

leovilok:
![nnn-logo-ideas-r4c1](https://user-images.githubusercontent.com/1430862/80540380-772ae680-89a9-11ea-913c-f362997921bd.png)
It's similar to the orange one without it's border.

jarun:
Left another suggestion in the discussion room. Only if it's easily doable.

leovilok:
Its easy:
![nnn-logo-ideas-chess](https://user-images.githubusercontent.com/1430862/80542804-ffab8600-89ad-11ea-88be-4ce6228f98a1.png)

jarun:
Wow! All of them look cool!

@maximbaz @KlzXS what do you you think?

maximbaz:
Call me boring, but I personally like the most minimalistic version, an `n³` that isn't embedded into any image or shape.

I have an almost infinite amount of variations of it 😄 

https://fonts.google.com/?preview.text=n%C2%B3&preview.text_type=custom&preview.size=65

jarun:
That's just too many fonts to choose from. ;)

The font used by @leovilok looks good to me.

@maximbaz how does the 5th knight embossed n3 look to you?

jarun:
To me it doesn't look squeezed in and makes a clear impression.

maximbaz:
As said I personally favor the minimalistic version that isn't embodied into anything, but if you like knights, it's fine with me 👍

jarun:
I do see the point of infinite variants. :)

Let's want for more inputs from others.

jarun:
Check out how the n3 looks live in the wiki page: https://github.com/jarun/nnn/wiki

jarun:
@leovilok can we have a single image of the 5th dark knight?

leovilok:
Here is a 128x128px version, but the source is SVG, so I can render it to any resolution.
[...]

PR #541, opened 2020-04-28
Add a simple logo for nnn (n³)

leovilok:
See #540


leovilok:
I also added some white border behind the text so that is still works on dark backgrounds.

jarun:
Thank you!

PR #543, opened 2020-04-29
Add alternative shortcut for redraw

ybc37:
Hey there!

First, thanks for `nnn` :)

I started using `nnn`, but quickly realized a conflict of shortcuts with my config. I use `^H`, `^J`, `^K`, `^L` for navigating tmux/vim panes, so I'm not able to trigger a redraw.

Of course it's my responsibility if I overwrite such a common shortcut like `^L` and I'm aware of adapting `nnn.h` to [change shortcuts](https://github.com/jarun/nnn/wiki/Usage#keyboard-mouse). But I could imagine this is quite a common use case (given the popularity of e.g. [vim-tmux-navigator](https://github.com/christoomey/vim-tmux-navigator/).

I noticed `^J` and `^K` already have alternatives. So I wonder if you're willing to add an alternative shortcut for redraw?

I choose `R` in this PR as it feels natural to me, but it's just an idea :)

### DIFF
--- a/src/nnn.h
+++ b/src/nnn.h
@@ -188,6 +188,7 @@ static struct key bindings[] = {
 	{ CONTROL('T'),   SEL_SORT },
 	/* Redraw window */
 	{ CONTROL('L'),   SEL_REDRAW },
+	{ 'R',            SEL_REDRAW },
 	/* Select current file path */
 	{ CONTROL('J'),   SEL_SEL },
 	{ ' ',            SEL_SEL },


jarun:
^L is kind of ubiquitous. And we have frozen the shortcuts in v2.9 after long discussions (see #422). So we can't add a new alternative for existing ones. We have to understand that every user has his own choices and we can't satisfy the taste of everyone.

However, it's trivial to compile `nnn` so I would suggest you change the shortcuts to your liking in your own fork. The `-k` switch should ease that as well.

toddyamakawa:
@ybc37 you can workaround your shortcut conflicts. e.g.

```
tmux bind-key C-l send-keys C-l
```

Then you can use `prefix ^L` to send `^L`

ybc37:
@jarun Ok, I understand. I just thought this might be useful for others as well and it's "only" an additional alternative. But I understand that there were big efforts to define the shortcuts and that it doesn't make sense to "just" add/change shortcuts after this discussion. Thanks for pointing me to it!

@toddyamakawa Thanks! Looks so obvious now :) Works like expected and I can use the package from the distro again.

jarun:
Thanks guys!

PR #545, opened 2020-05-01
Support $_Z_DATA environment variable

step-:
Support `$_Z_DATA` environment variable per z's documentation https://github.com/rupa/z/blob/dcd5541e7dc37a3a32885565539c6fb1479af26f/z.1#L78.
`$_Z_DATA` allows changing the default location of the z database file.


jarun:
Thank you!

Issue #547, opened 2020-05-02
Having hooks 


sucrecacao:
**Is your feature request related to a problem? Please describe.**

I understand than nnn don't want to implement any kind of preview design. But having to press a key on every files to preview  them can be annoying. Being able to hook a "preview script" on an given event ( in this case change of the current selected file )  could be very handy to preview file dynamically. 

**A clear and concise description of what the problem is.**

Not being able to run a script on a hook, and having to press a key every time to run a script/plugin.

**Describe the solution you'd like**
Follow the git / [task warrior]( https://taskwarrior.org/docs/hooks.html) hook system:  
having a hooks folder which contain hook file. a hook file start with "on-" followed by the hook type. 
For example:
```bash
on-change-current      # called everytime we change the current selected file
on-change-directory   # called when we enter a directory
on-exit                        # called when exiting nnn
```

**A clear and concise description of what you want to happen.**

When I press 'j' or 'k' in nnn and change the current selected file or dir, the script written in the file `hooks/on-change-current` is run with $1 being the new current selected file/dir.

jarun:
> When I press 'j' or 'k' in nnn and change the current selected file or dir, the script written in the file hooks/on-change-current is run with $1 being the new current selected file/dir.

We can't run a script on every keypress.

> But having to press a key on every files to preview them can be annoying.

You can always open a directory of files in an image viewer like `sxiv`. `nnn` is not about image previews using different utilities and hacks to preview different kinds of files.

jarun:
Some more details in the design notes: https://github.com/jarun/nnn/wiki#design

There are several good file managers which already do what you are looking for. Pick one of those.

sucrecacao:
> You can always open a directory of files in an image viewer like sxiv as thumbnails and browse them. nnn is not about image previews using different utilities and hacks for different kinds of files.

> There are several good file managers which already do what you are looking for. Pick one of those.

Hooks is the requested feature, not previews. I just used it as an example. 
  
I used preview has an example because it show how hooks could help user extend nnn to implement popular feature without changing its UNIX orthodoxies. Previews would be handled by external program. 

> We can't run a script on every keypress.

nnn won't run script on every keypress. It would run q script on certain event _only_ if the user create a hook on that particular event.

jarun:
> It would run q script on certain event

which essentially means we will have to add checkpoints in several places and if enabled run the script as well.

jarun:
Can you share some examples which are not achievable by plugins?

[...]

PR #548, opened 2020-05-03
Add an option to print hovered files to a FIFO

leovilok:
This adds an env variable, `NNN_FIFO`, that can be set to a path that
`nnn` will open/create as a FIFO, and where every hovered file's path is
printed. This allows creating external perview/quick open plugins, ...

It can be compiled out with the make variable `O_NOFIFO`.

See some discussion about it in #547

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,10 @@ ifeq ($(O_NOBATCH),1)
 	CPPFLAGS += -DNOBATCH
 endif
 
+ifeq ($(O_NOFIFO),1)
+	CPPFLAGS += -DNOFIFO
+endif
+
 ifeq ($(shell $(PKG_CONFIG) ncursesw && echo 1),1)
 	CFLAGS_CURSES ?= $(shell $(PKG_CONFIG) --cflags ncursesw)
 	LDLIBS_CURSES ?= $(shell $(PKG_CONFIG) --libs   ncursesw)
--- a/nnn.1
+++ b/nnn.1
@@ -427,9 +427,7 @@ separated by \fI;\fR:
 .Bd -literal
     export NNN_FIFO='/tmp/nnn.fifo'
 
-    NOTES:
-    1. If the FIFO file doesn't exist it will be created, but it will never be removed.
-    2. You can actually give a path to a regular file, and get a persistant log.
+    NOTE: If the FIFO file doesn't exist it will be created, but it will never be removed.
 .Ed
 .Pp
 \fBnnn:\fR this is a special variable set to the hovered entry before executing
--- a/nnn.1
+++ b/nnn.1
@@ -423,6 +423,15 @@ separated by \fI;\fR:
 .Pp
 \fBNNN_SEL:\fR absolute path to custom selection file.
 .Pp
+\fBNNN_FIFO:\fR path of a named pipe to write current file path:
+.Bd -literal
+    export NNN_FIFO='/tmp/nnn.fifo'
+
+    NOTES:
+    1. If the FIFO file doesn't exist it will be created, but it will never be removed.
+    2. You can actually give a path to a regular file, and get a persistant log.
+.Ed
+.Pp
 \fBnnn:\fR this is a special variable set to the hovered entry before executing
 a command from the command prompt or spawning a shell.
 .Pp
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4687,14 +4687,15 @@ static void notify_fifo()
 		}
 	}
 
-	static char oldpath[PATH_MAX] = {0};
-	char path[PATH_MAX];
-	size_t len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
+	static char *name = NULL;
 
-	if (xstrcmp(oldpath, path) == 0)
+	if (dents[cur].name == name)
 		return;
 
-	xstrsncpy(oldpath, path, len);
+	name = dents[cur].name;
+
+	char path[PATH_MAX];
+	size_t len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
 
 	path[len - 1] = '\n';
 
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4695,7 +4695,7 @@ static void notify_fifo()
 	name = dents[cur].name;
 
 	char path[PATH_MAX];
-	size_t len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
+	size_t len = mkpath(g_ctx[cfg.curctx].c_path, ndents ? name : "", path);
 
 	path[len - 1] = '\n';
 
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -330,6 +330,9 @@ static context g_ctx[CTX_MAX] __attribute__ ((aligned));
 
 static int ndents, cur, last, curscroll, last_curscroll, total_dents = ENTRY_INCR;
 static int nselected;
+#ifndef NOFIFO
+static int fifofd = -1;
+#endif
 static uint idletimeout, selbufpos, lastappendpos, selbuflen;
 static ushort xlines, xcols;
 static ushort idle;
@@ -350,6 +353,9 @@ static char *prefixpath;
 static char *plugindir;
 static char *sessiondir;
 static char *pnamebuf, *pselbuf;
+#ifndef NOFIFO
+static char *fifopath;
+#endif
 static ull *ihashbmp;
 static struct entry *dents;
 static blkcnt_t ent_blocks;
@@ -4667,6 +4673,39 @@ static void populate(char *path, char *lastname)
 	last_curscroll = -1;
 }
 
+#ifndef NOFIFO
+static void notify_fifo()
+{
+	if (fifofd == -1) {
+		fifofd = open(fifopath, O_WRONLY|O_NONBLOCK);
+		if (fifofd == -1) {
+			if (errno != ENXIO)
+				/* Unexpected error, the FIFO file might have been removed */
+				/* We give up FIFO notification */
+				fifopath = NULL;
+			return;
+		}
+	}
+
+	static char oldpath[PATH_MAX] = {0};
+	char path[PATH_MAX];
+	size_t len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
+
+	if (xstrcmp(oldpath, path) == 0)
+		return;
+
+	xstrsncpy(oldpath, path, len);
+
+	path[len - 1] = '\n';
+
+	ssize_t ret = write(fifofd, path, len);
+
+	if (ret != (ssize_t)len && !(ret == -1 && (errno == EAGAIN || errno == EPIPE))) {
+		DPRINTF_S(strerror(errno));
+	}
+}
+#endif
+
 static void move_cursor(int target, int ignore_scrolloff)
 {
 	int onscreen = xlines - 4; /* Leave top 2 and bottom 2 lines */
@@ -4693,6 +4732,11 @@ static void move_cursor(int target, int ignore_scrolloff)
 	}
 	curscroll = MIN(curscroll, MIN(cur, ndents - onscreen));
 	curscroll = MAX(curscroll, MAX(cur - (onscreen - 1), 0));
+
+#ifndef NOFIFO
+	if (fifopath)
+		notify_fifo();
+#endif
 }
 
 static void handle_screen_move(enum action sel)
@@ -7022,6 +7066,19 @@ int main(int argc, char *argv[])
 
 	DPRINTF_S(getenv("PWD"));
 
+#ifndef NOFIFO
+	/* Create fifo */
+	fifopath = getenv("NNN_FIFO");
+	if (fifopath) {
+		if (mkfifo(fifopath, 0600) != 0 && !(errno == EEXIST && access(fifopath, W_OK) == 0)) {
+			xerror();
+			return _FAILURE;
+		}
+
+		signal(SIGPIPE, SIG_IGN);
+	}
+#endif
+
 #ifdef LINUX_INOTIFY
 	/* Initialize inotify */
 	inotify_fd = inotify_init1(IN_NONBLOCK);
@@ -7143,6 +7200,11 @@ int main(int argc, char *argv[])
 	close(kq);
 #elif defined(HAIKU_NM)
 	haiku_close_nm(haiku_hnd);
+#endif
+
+#ifndef NOFIFO
+	if (fifofd != -1)
+		close(fifofd);
 #endif
 
 	return opt;
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -190,7 +190,7 @@ Usage examples can be found in the Examples section below.
 There are many plugins provided by `nnn` which can be used as examples. Here are a few simple selected examples.
 
 - Show the git log of changes to the particular file along with the code for a quick and easy review.
-   ```sh
+    ```sh
     #!/usr/bin/env sh
     git log -p -- "$1"
     ```
@@ -219,6 +219,40 @@ There are many plugins provided by `nnn` which can be used as examples. Here are
 
     printf "%s" "0c$dir" > "$NNN_PIPE"
     ```
+
+#### Get notified on file hover
+
+If `NNN_FIFO` is set, `nnn` will open it and write every hovered files.
+This can be used in plugins, e.g. to implement file previews.
+
+If a `NNN_FIFO` is set globally, each `nnn` instance will write to it, and a process reading from the pipe will get hovered path from every instance, interleaved.
+
+If you want to prevent this and be sure to have a private pipe to one `nnn` instance, you can unlink (remove) the FIFO file.
+If you opened the FIFO before and you have read from it (so that `nnn` have it opened too), you can still read from it while you don't close it.
+But new `nnn` instances will recreate a new FIFO not linked to the previous one.
+
+Don't forget to fork in the background to avoid blocking `nnn`.
+
+Example (send every hovered file to X selection):
+
+```sh
+#!/usr/bin/env sh
+if [ -z "$NNN_FIFO" ] ; then
+	exit 1
+fi
+
+while read FILE ; do
+	if [ -n "$NNN_FIFO" ] ; then
+		# If you want to remove the FIFO,
+		# don't do it before first read,
+		# nnn won't have it opened yet
+		rm "$NNN_FIFO"
+		NNN_FIFO=
+	fi
+	printf "%s" "$FILE" | xsel
+done < "$NNN_FIFO" &
+disown
+```
 
 ## Contributing plugins
 


leovilok:
It can be used to implement external previewers, e.g. this one based on tabbed: https://gist.github.com/leovilok/173d2b6362c51af53149947fcc62a181

([Worst screencast ever](http://vgx.fr/view.sh-capt.webm))

src/nnn.c
```diff
@@ -6938,6 +6993,22 @@ int main(int argc, char *argv[])
 
 	DPRINTF_S(getenv("PWD"));
 
+#ifndef NOFIFO
+	/* Create fifo */
+	fifopath = getenv("NNN_FIFO");
+	if (fifopath) {
+		if (mkfifo(fifopath, 0600) != 0 && !(errno == EEXIST && access(fifopath, W_OK) == 0)) {
+			xerror();
+			fprintf(stderr, "lel\n");
```
jarun, line 7002: No need to print after `xerror()`. It shows line number.

src/nnn.c
```diff
@@ -7062,5 +7133,10 @@ int main(int argc, char *argv[])
 	haiku_close_nm(haiku_hnd);
 #endif
 
+#ifndef NOFIFO
+	if(fifofd)
```
jarun, line 7137: Space:

    if (fifofd)

src/nnn.c
```diff
@@ -4620,6 +4669,12 @@ static void move_cursor(int target, int ignore_scrolloff)
 	}
 	curscroll = MIN(curscroll, MIN(cur, ndents - onscreen));
 	curscroll = MAX(curscroll, MAX(cur - (onscreen - 1), 0));
+
+#ifndef NOFIFO
+	if (cfg.usefifo) {
```
jarun, line 4674: No need to curly brackets around single liners.

src/nnn.c
```diff
@@ -273,6 +273,9 @@ typedef struct {
 	uint cliopener  : 1;  /* All-CLI app opener */
 	uint waitedit   : 1;  /* For ops that can't be detached, used EDITOR */
 	uint rollover   : 1;  /* Roll over at edges */
+#ifndef NOFIFO
+	uint usefifo    : 1;  /* Write current to FIFO */
```
jarun, line 277: Why this? Use `if (fifofd > 0)`.

src/nnn.c
```diff
@@ -324,6 +327,9 @@ static settings cfg = {
 	0, /* cliopener */
 	0, /* waitedit */
 	1, /* rollover */
+#ifndef NOFIFO
+	0, /* usefifo */
```
jarun, line 331: Remove this as well.

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
 	last_curscroll = -1;
 }
 
+#ifndef NOFIFO
+static void notify_fifo()
+{
+	if (fifofd == -1) {
+		fifofd = open(fifopath, O_WRONLY|O_NONBLOCK);
```
jarun, line 4613: Can't we open it once at the beginning instead of doing it every time?

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
...
+		if (fifofd == -1) {
+			if(errno != ENXIO) {
```
jarun, line 4615: Space.

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
...
+				/* Unexpected error, the FIFO file might have been removed */
+				/* We give up FIFO notification */
+				cfg.usefifo = 0;
+			}
+			return;
+		}
+	}
+	static char oldpath[PATH_MAX + 1] = {0};
```
jarun, line 4623: No need to initialize. See `mkpath()`.

And why store old path? How can you hover twice on a file?

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
...
+
+	char path[PATH_MAX + 1];
```
jarun, line 4625: No need of `+ 1`.

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
...
+
+	int len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
+	path[len++] = '\n';
```
jarun, line 4628: Why add this `\n`?

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
...
+	path[len] = '\0';
+
+	if(xstrcmp(oldpath, path) == 0){
```
jarun, line 4631: Space. Curly brackets not required.

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
...
+		return;
+	}
+	xstrsncpy(oldpath, path, len);
+
+	/* I'm not sure if partial writes can happen on FIFOs with O_NONBLOCK,
+	 * so this might be just to please cppcheck */
+	int wrote = 0, ret = 0;
+	while((ret = write(fifofd, path + wrote, len - wrote)) < len){
```
jarun, line 4639: Space.

src/nnn.c
```diff
@@ -4596,6 +4606,45 @@ static void populate(char *path, char *lastname)
...
+		if(ret == -1){
```
jarun, line 4640: Space. Curly brackets not required.

Makefile
```diff
@@ -57,6 +57,10 @@ ifeq ($(O_NOBATCH),1)
 	CPPFLAGS += -DNOBATCH
 endif
 
+ifeq ($(O_NOFIFO),1)
```
jarun, line 60: I would want it the other way. Enable the code if someone wants to hack enough.

src/nnn.c
jarun: Also, when you do soemthing like this, try the following:

```
path[len] = '\n';
path[++len] = '\0';
```

src/nnn.c
leovilok: ahah, I forgot this debug line :rofl:

src/nnn.c
leovilok: We can't open a FIFO in O_NONBLOCK mode if there is no reader (and without O_NONBLOCK, we block while there is no reader). A workaround exist, where you open the FIFO in RW mode... but I don't think it's cleaner.

Besides, this would make us write lines in the pipe while there is space left, and if a reader connects, it gets a few lines in a bulk (the first files hovered, but not the last ones if the pipe was full). It would be a mess.

src/nnn.c
leovilok: argh, there shound be a clang-format/astyle/indent file somwhere... I'll try to be more carefull.

src/nnn.c
leovilok: it's for the `\n` before the final `\0`.

src/nnn.c
jarun: OK. Thanks for the notes!

src/nnn.c
leovilok: I hook it to cursor `move_cursor`, which is sometimes called twice in a row for the same file (e.g. when entering a directory). I didn't investigate more on why this happens.

src/nnn.c
jarun: Just add it to our `clang-tidy` thing in src/ dir. It's really useful.

src/nnn.c
jarun: You get `len` from `mkpath`. So maybe overwrite the `\0` at the end with `\n` and write exactly the number of bytes you want?

src/nnn.c
leovilok: Path are sent to the FIFO one per line, to be handled by external applications this way.
We could use an '\n' too, but it's just a bit less script friendly.

jarun: I added a note to do this in a simpler way. You can't write code like this in `nnn` and get away with it. ;)

leovilok:
@jarun I'll do the changes tomorrow, it's late here.

jarun:
Good night! No hurry!

src/nnn.c
jarun: There's a `move_cursor()` in `redraw()`. Check it out. Do we need this?

Makefile
jarun: Maybe just leave ti this way for now. Let's see how the performance goes.

src/nnn.c
jarun: Please raise a new review for the clang-tidy thing.

@toddyamakawa would it be possible for you to work on this?

src/nnn.c
jarun: Also, how about on resize?

Makefile
leovilok: If you don't set `NNN_FIFO`, it's just 1 `getenv` and 2 `if`s in the whole program execution + one `if` per `move_cursor`, so this can't be noticeable.
If you use the fonctionnality, there are a few syscalls, but only non blocking ones.

src/nnn.c
leovilok: done

src/nnn.c
leovilok: Now the `\n` replaces the '\0', as suggested (maybe printing '\0' separated strings would be better, to avoid problems with path containing '\n', but it's harder to script).

src/nnn.c
leovilok: I can't use `fifofd` because it might not be open yet, but I now use `fifopath`, and I could remouve this.

src/nnn.c
leovilok: I tried commenting out the `move_cursor()` in `redraw()`: it is very much needed, without it arrow keys don't move, and I even got a segfault.

leovilok: Of course `redraw()` is called on each resize, hence `move_cursor()` too, so `notify_fifo()` is called too, and without the `oldpath` chech, it would print duplicates to the FIFO.

jarun:
@leovilok when this is in, we need a step by step guide to what the user needs to do to achieve what you did in https://github.com/jarun/nnn/pull/548#issuecomment-623043487.

Makefile
jarun: Yes, I realized that later. :+1:

src/nnn.c
```diff
@@ -4664,6 +4668,43 @@ static void populate(char *path, char *lastname)
...
+			if (errno != ENXIO)
+				/* Unexpected error, the FIFO file might have been removed */
+				/* We give up FIFO notification */
+				fifopath = NULL;
+			return;
+		}
+	}
+	static char oldpath[PATH_MAX] = {0};
+
+	char path[PATH_MAX];
```
jarun, line 4686: No need on newlines above var declarations.

src/nnn.c
```diff
@@ -4664,6 +4668,43 @@ static void populate(char *path, char *lastname)
...
+	if (fifofd == -1) {
+		fifofd = open(fifopath, O_WRONLY|O_NONBLOCK);
+		if (fifofd == -1) {
+			if (errno != ENXIO)
+				/* Unexpected error, the FIFO file might have been removed */
+				/* We give up FIFO notification */
+				fifopath = NULL;
+			return;
+		}
+	}
```
jarun, line 4683: Add a newline below this block.

src/nnn.c
```diff
@@ -4664,6 +4668,43 @@ static void populate(char *path, char *lastname)
...
+	if (xstrcmp(oldpath, path) == 0)
+		return;
+
+	xstrsncpy(oldpath, path, len);
+
+	path[len++] = '\n';
+
+	/* I'm not sure if partial writes can happen on FIFOs with O_NONBLOCK,
+	 * so this might be just to please cppcheck */
+	int wrote = 0, ret = 0;
```
jarun, line 4699: Why do we need to track how much data we wrote in `wrote`?

src/nnn.c
```diff
@@ -4664,6 +4668,43 @@ static void populate(char *path, char *lastname)
...
+	char path[PATH_MAX];
+
+	int len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
+
+	if (xstrcmp(oldpath, path) == 0)
+		return;
+
+	xstrsncpy(oldpath, path, len);
+
+	path[len++] = '\n';
```
jarun, line 4695: Is this correct? if we have `/bin/hello`, len is 11.

If we want to overwrite the terminating `\0`, we should have:

    path[len - 1] = '\n'

and then:

    write(fifofd, path, len);

jarun:
Another note: in addition to working example in https://github.com/jarun/nnn/wiki/Basic-use-cases#image-video-pdf we also need to add this to config section in Wiki.

src/nnn.c
leovilok: Oh you're right, `mkpath`'s len count includes the terminating `\0` (if the comment is right), I thought it was like `strlen`.

jarun: > I thought it was like strlen

We are much smarter. ;)

Also, `xstrsncpy` does the same. Probably just go through our functions for a overview whenever you get the time.

src/nnn.c
leovilok: This is to handle partial writes, to go on writing if `write` returns a wrote bytes count lower than `len`, and I need to retry writing the remaining. 

Theorically, this might happen, but in pratice, when writing to a FIFO (and one opened with O_NONBLOCK), I don't see how it could. But `cppcheck` is unhappy if I don't check and use `write`'s return value ... so I implemented the whole partial write handling anyways...

src/nnn.c
jarun:
```
}

static char oldpath[PATH_MAX] = {0};
char path[PATH_MAX];
int len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);

if (xstrcmp(oldpath, path) == 0)
...
```

src/nnn.c
jarun: I think we can do what we do with standard writes:

```
ret = write(fifofd, path, len);
if (ret != len)
{
    DPRINTF_S("Failed to write to FIFO");
}
```

src/nnn.c
```diff
@@ -4664,6 +4668,43 @@ static void populate(char *path, char *lastname)
...
+				/* We give up FIFO notification */
+				fifopath = NULL;
+			return;
+		}
+	}
+	static char oldpath[PATH_MAX] = {0};
+
+	char path[PATH_MAX];
+
+	int len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
```
jarun, line 4688: len should be `size_t`

0xACE:
I realize I'm putting up this question late in the design process but my hope is that this will be a note that may evoke thoughts. So there is no need to respond if you don't feel it to be necessary.

If `NNN_FIFO` is a designated env var: What happens if 2 instances of `nnn` runs while pointing to the same `NNN_FIFO`?

I know that the end user could probably script around this... My idea was that `nnn` creates a unique `NNN_FIFO` and could spawn a preview plugin where `nnn` sets the `NNN_FIFO` for the plugin (basically inheriting).

src/nnn.c
leovilok: I also experimented checking if `cur != last` before printing, but there a a lot of false positives (context, switch, dir enter, filter,... and I'm not sure I got everything).

The only way to be sure it's (not) the same file is to check everything.

leovilok:
> I realize I'm putting up this question late in the design process but my hope is that this will be a note that may evoke thoughts. So there is no need to respond if you don't feel it to be necessary.
> 
> If `NNN_FIFO` is a designated env var: What happens if 2 instances of `nnn` runs while pointing to the same `NNN_FIFO`?

They will both use the same FIFO, and a reader will get lines from both instances interleaved.

> I know that the end user could probably script around this... My idea was that `nnn` creates a unique `NNN_FIFO` and could spawn a preview plugin where `nnn` sets the `NNN_FIFO` for the plugin (basically inheriting).

`NNN_FIFO` is in `nnn`'s env, so it should be inherited to plugins.
If you want to use a single `NNN_FIFO` name, but be sure that it won't be used by anyone else, you can open the file and immediately unlink (rm) it. This won't close the pipe (it still exist in kernel until reader/writers close it), but new named pipes created with the same path will be different objects, not linked to the old FIFO. e.g.:
~~~sh
# in a plugin...
while read FILE ; do
    handle_preview "$FILE"
done < "$NNN_FIFO" &
disown
rm "$NNN_FIFO"
~~~

jarun:
> My idea was that nnn creates a unique NNN_FIFO and could spawn a preview plugin where nnn sets the NNN_FIFO for the plugin (basically inheriting).

The disadvantage is user cannot have an independent script.

But I see the issues with the current approach. Two instance of `nnn` will create issues. We can have an additional program option to turn on/off FIFO writes conditionally to work around that.

Maybe let's think of the workflow. How do we want it to work?

option 1 as independent script:

- user runs external script which listens to FIFO
- user opens `nnn` and browses

option 2 as plugin:

- user runs plugin from within `nnn`
  - the plugin reads FIFO?
- user closes `nnn` after browsing
  - plugin closes?

@leovilok can you clarify on the planned workflow?

jarun:
OK. I see you have clarified. @0xACE does this help?

@leovilok we need detailed info on the usage of `NNN_FIFO` in plugin page (not the wiki).

src/nnn.c
leovilok: Done, but `size_t` is unsigned, so it needs to be cast when compared to write output (which is signed).

src/nnn.c
```diff
@@ -350,6 +350,10 @@ static char *prefixpath;
 static char *plugindir;
 static char *sessiondir;
 static char *pnamebuf, *pselbuf;
+#ifndef NOFIFO
+static char *fifopath;
+static int fifofd = -1;
```
jarun, line 355: Can we place this along with other ints above?

src/nnn.c
```diff
@@ -4664,6 +4670,44 @@ static void populate(char *path, char *lastname)
...
+	char path[PATH_MAX];
+
+	size_t len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
+
+	if (xstrcmp(oldpath, path) == 0)
+		return;
+
+	xstrsncpy(oldpath, path, len);
+
+	path[len-1] = '\n';
```
jarun, line 4697: Space:

    path[len - 1]

src/nnn.c
```diff
@@ -4664,6 +4670,44 @@ static void populate(char *path, char *lastname)
...
+
+	/* I'm not sure if partial writes can happen on FIFOs with O_NONBLOCK,
+	 * so this might be just to please cppcheck */
+	size_t wrote = 0;
```
jarun, line 4701: I had an earlier comment here which is gone now.

Why do you think the while loop is required? I think the following should also work:

```
ssize_t ret = write(fifofd, path, len);

if (ret != (ssize_t)len)
{
    DPRINTF_S(strerror(errnum));
}

File: src/nnn.c (hunk 19 above)
Comment (jarun) [reply src/nnn.c]: True, but let's stick to the signs.

File: src/nnn.c (hunk 22 above)
Comment (leovilok) [reply src/nnn.c:4701]: I still see your old question, here was my answer:
> This is to handle partial writes, to go on writing if `write` returns a wrote bytes count lower than `len`, and I need to retry writing the remaining. 
> 
> Theorically, this might happen, but in pratice, when writing to a FIFO (and one opened with O_NONBLOCK), I don't see how it could. But `cppcheck` is unhappy if I don't check and use `write`'s return value ... so I implemented the whole partial write handling anyways...

But I like your suggestion better, I'll do that.

Comment (jarun) [reply src/nnn.c:4701]: > But I like your suggestion better, I'll do that.

Thanks! We'll figure it with debug logs if we see this problem. But I want to avoid the `while` if possible.

Comment (leovilok) [reply src/nnn.c]: Done.

File: src/nnn.c [hunk 23]
```diff
@@ -4664,6 +4670,39 @@ static void populate(char *path, char *lastname)
...
+		if (fifofd == -1) {
+			if (errno != ENXIO)
+				/* Unexpected error, the FIFO file might have been removed */
+				/* We give up FIFO notification */
+				fifopath = NULL;
+			return;
+		}
+	}
+
+	static char oldpath[PATH_MAX] = {0};
```
jarun, line 4687: Let's see if we really need to retain this. This is a static and we add a full path comparison for this.

Can you check the following:

In `move_cursor`:

```
bool move = (cur != target);
...
if (fifopath && move)
    notify_fifo();
```

0xACE:
I guess he answered it.

My interpretation of his comment and the code is that the user has to provide a `NNN_FIFO`. And to use it independently across multiple instances with the same `NNN_FIFO` path you just delete the file as the kernel will maintain the communication socket.

I guess it solves the problem I had in mind. Thanks for the clarification.

jarun:
OK. These will be documented in the plugins README anyway.

src/nnn.c
leovilok: I also answer to that a bit earlier, here is what I wrote:

> I also experimented checking if `cur != last` before printing, but there a a lot of false positives (context, switch, dir enter, filter,... and I'm not sure I got everything).
> 
> The only way to be sure it's (not) the same file is to check everything.

The issue here is that `cur` is an `int` it's the current file index in the current dir, but if the dir changes but not the file index, `cur` is the same as `target` (and last).

Of course, I can store and check only the directory part, but that's not much gained. But there must be something smarter to do...

leovilok:
@0xACE you got it right, I hope I'll be able to word it well in the plugins README.

src/nnn.c
jarun: > But there must be something smarter to do

We do not store the inode number, otherwise that would solve the issue. But that would increase the memory usage significantly.

jarun: Do you see an issue if we just compare the file name instead of the full path?

jarun: Ok. I think the following wold work:

retain timestamp, filename length and size. So the check could be:

```
if (!namelength same || !timestamp same || !size same)
    proceed to write to FIFO;
```

jarun:
> I hope I'll be able to word it well in the plugins README

We'll pitch in, don't worry. Do this is the current review.

src/nnn.c
leovilok: In the last commit I use filename pointer check. It seems to work well, I could't make it break. What do you think?

Your suggestion could work too, I wonder if there would be collisions in some `/proc` or `/sys` path...

src/nnn.c
```diff
@@ -4684,14 +4684,15 @@ static void notify_fifo()
 		}
 	}
 
-	static char oldpath[PATH_MAX] = {0};
-	char path[PATH_MAX];
-	size_t len = mkpath(g_ctx[cfg.curctx].c_path, dents[cur].name, path);
+	static char *name = NULL;
 
-	if (xstrcmp(oldpath, path) == 0)
+	if (dents[cur].name == name)
```
jarun, line 4689: Good approximation!

jarun:
Please update details on `NNN_FIFO` control here: https://github.com/jarun/nnn/tree/master/plugins#create-your-own-plugins

Also, please add an example.

We also need to update the wiki:

- The new config with footnotes at https://github.com/jarun/nnn/wiki/Usage#configuration
- The sample config for the screencast you shared here: https://github.com/jarun/nnn/wiki/Basic-use-cases#image-video-pdf

nnn.1
```diff
@@ -423,6 +423,15 @@ separated by \fI;\fR:
 .Pp
 \fBNNN_SEL:\fR absolute path to custom selection file.
 .Pp
+\fBNNN_FIFO:\fR path of a named pipe to write current file path:
+.Bd -literal
+    export NNN_FIFO='/tmp/nnn.fifo'
+
+    NOTES:
+    1. If the FIFO file doesn't exist it will be created, but it will never be removed.
+    2. You can actually give a path to a regular file, and get a persistant log.
```
jarun, line 432: I think we can remove the second note.

jarun: Or reword to:

2. A persistent tog is generated if it's a path to a regular file.

jarun: Also, move both NNN_SEL and NNN_FIFO above NNN_LOCKER (below NNN_TRASH).

leovilok: Removed.

leovilok:
> Please update details on `NNN_FIFO` control here: https://github.com/jarun/nnn/tree/master/plugins#create-your-own-plugins
> 
> Also, please add an example.

I added documentation on how to write plugins using this feature.
The example is a simple plugin to send every hovered files to X selection, I might make it a real plugin later.

> We also need to update the wiki:
> 
> * The new config with footnotes at https://github.com/jarun/nnn/wiki/Usage#configuration
> * The sample config for the screencast you shared here: https://github.com/jarun/nnn/wiki/Basic-use-cases#image-video-pdf

Should I edit wiki before this is merged?

I might make a real (basic) preview plugin based on the [view.sh](https://gist.github.com/leovilok/173d2b6362c51af53149947fcc62a181) script I used in the screencast. In that can I could document how to set it up.

leovilok:
@0xACE I forgot that the FIFO is not actually opened by `nnn` before the first read, so you can't remove the FIFO file before that. See the example I added [here](https://github.com/jarun/nnn/pull/548/files#diff-a72b13f48ae94508d8b863f210a60122):
```sh
#!/usr/bin/env sh
if [ -z "$NNN_FIFO" ] ; then
        exit 1
fi

while read FILE ; do
        if [ -n "$NNN_FIFO" ] ; then
                # If you want to remove the FIFO,
                # don't do it before first read,
                # nnn won't have it opened yet
                rm "$NNN_FIFO"
                NNN_FIFO=
        fi
        printf "%s" "$FILE" | xsel
done < "$NNN_FIFO" &
disown
```

(This is a simple plugin that prints every file hovered in `nnn` to X selection)

leovilok:
Rebased on master.

jarun:
Thank you!

jarun:
@leovilok 

Does the view.sh script from https://github.com/jarun/nnn/pull/548#issuecomment-623043487 work as a plugin? If yes, add it as a plugin named `previewer`.

Hell! We'll have a finder and a previewer in 2 consecutive days!

toddyamakawa:
Holy shit, I just tested this out and it's incredible.

I start each of my nnn instances with a unique FIFO and automatically split my tmux pane to create a "preview" pane:
https://github.com/toddyamakawa/bin/blob/master/%2Cnnn#L53..L60
https://github.com/toddyamakawa/bin/blob/master/nnn-preview

I'm pretty sure there's no way other file managers can compete against nnn now.

jarun:
Where's the `preview` script? https://github.com/toddyamakawa/bin/blob/master/nnn-preview#L3

jarun:
Also @toddyamakawa I need a complete step by step guide to configure this along with scripts that will wrkk anywhere (so, probably POSIX sh).

Can you and @leovilok work on this? I also left some questions for @leovilok in the discussion thread to control when `nnn` will write to FIFO and when it will not. Because if no one is reading, there's no reason to keep writing.

jarun:
And it would be great if the scripts are independent of tmux or anything... just `nnn`, the previewer script and the utilities required to view the files.

If you guys want, have a new Wiki page dedicated for preview.

toddyamakawa:
> And it would be great if the scripts are independent of tmux or anything

What should we use for the preview window? `xterm`?

toddyamakawa:
> Where's the `preview` script? https://github.com/toddyamakawa/bin/blob/master/nnn-preview#L3

I've been using my own custom scripts for fzf previews so I reused them for nnn, but I can rewrite my script pretty easily to just make it `cat` the file in the preview window.

jarun:
Yes, we need something that is not heavily dependent on something, can be replaced easily (like preferred image viewers, editors etc.)

PR #549, opened 2020-05-03
browse: fix nnn -e

sgn:
The current code will start editor if the mime doesn't start with
"text/". But, we want the opposite.

Simplify the check by using `strncmp` instead.

We may improve by writing a helper: `start_withs` in the future.

This change also cleans -Wstrict-aliasing on Ubuntu 16.04 LTS


jarun:
Thank you!

PR #552, opened 2020-05-05
Add a tabbed/xembed based file previewer plugin

leovilok:
This plugin is written in bash, because job control is not well
specified in POSIX sh (`jobs` can return anything).

We use [`tabbed`][1] as a [xembed][2] host, to have a single window
owning each previewer window. So each previewer must be a xembed client.
For text previewers, this is not an issue, as there are a lot of
xembed-able terminal emulator (we default to `xterm`, but examples are
provided for `urxvt` and `st`). For other filetypes, this can be
trickier, but a few popular viewers are xembed-able, we use:
 - `mpv`: multimedia player, for video/audio preview
 - `sxiv`: image viewer
 - `zathura`: PDF viewer

How to use:
-----------

First, install the dependencies. Then you need to set a NNN_FIFO path
and set a key for the plugin, then start `nnn`:

    $ NNN_FIFO=/tmp/nnn.fifo NNN_PLUG='p:preview' nnn

Then in `nnn`, press `;p`.

[1]: http://tools.suckless.org/tabbed/
[2]: https://specifications.freedesktop.org/xembed-spec/xembed-spec-latest.html

### DIFF
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -52,39 +52,39 @@ NUKE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/nuke"
 
 
 if which xterm ; then
-	TERMINAL="xterm -into"
+    TERMINAL="xterm -into"
 elif which urxvt ; then
-	TERMINAL="urxvt -embed"
+    TERMINAL="urxvt -embed"
 elif which st ; then
-	TERMINAL="st -w"
+    TERMINAL="st -w"
 else
-	echo "No xembed term found" >&2
+    echo "No xembed term found" >&2
 fi
 
 
 term_nuke () {
-	# $1 -> $XID, $2 -> $FILE
-	$TERMINAL "$1" -e "$NUKE" "$2" &
+    # $1 -> $XID, $2 -> $FILE
+    $TERMINAL "$1" -e "$NUKE" "$2" &
 }
 
 start_tabbed () {
-	FIFO="$(mktemp -u)"
-	mkfifo "$FIFO"
+    FIFO="$(mktemp -u)"
+    mkfifo "$FIFO"
 
-	tabbed > "$FIFO" &
+    tabbed > "$FIFO" &
 
-	jobs # Get rid of the "Completed" entries
+    jobs # Get rid of the "Completed" entries
 
-	TABBEDPID="$(jobs -p %%)"
+    TABBEDPID="$(jobs -p %%)"
 
-	if [ -z "$TABBEDPID" ] ; then
-		echo "Can't start tabbed"
-		exit 1
-	fi
+    if [ -z "$TABBEDPID" ] ; then
+        echo "Can't start tabbed"
+        exit 1
+    fi
 
-	read -r XID < "$FIFO"
+    read -r XID < "$FIFO"
 
-	rm "$FIFO"
+    rm "$FIFO"
 }
 
 get_viewer_pid () {
@@ -92,93 +92,93 @@ get_viewer_pid () {
 }
 
 previewer_loop () {
-	unset -v NNN_FIFO
-	# mute from now
-	exec >/dev/null 2>&1
+    unset -v NNN_FIFO
+    # mute from now
+    exec >/dev/null 2>&1
 
-	MAINWINDOW="$(xdotool getactivewindow)"
+    MAINWINDOW="$(xdotool getactivewindow)"
 
-	start_tabbed
+    start_tabbed
 
-	xdotool windowactivate "$MAINWINDOW"
-	xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &
+    xdotool windowactivate "$MAINWINDOW"
+    xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &
 
         while read -r FILE ; do
 
-		jobs # Get rid of the "Completed" entries
+        jobs # Get rid of the "Completed" entries
 
-		if ! jobs | grep tabbed ; then
-		# Uncomment and remove `break` want to restart tabbed everytime its window is closed
-		#	start_tabbed
-			break
-		fi
+        if ! jobs | grep tabbed ; then
+        # Uncomment and remove `break` want to restart tabbed everytime its window is closed
+        #   start_tabbed
+            break
+        fi
 
-		if [ ! -e "$FILE" ] ; then
-			continue
-		fi
+        if [ ! -e "$FILE" ] ; then
+            continue
+        fi
 
-		if [ -n "$VIEWERPID" ] && jobs -p | grep "$VIEWERPID" ; then
-			kill "$VIEWERPID"
-		fi
+        if [ -n "$VIEWERPID" ] && jobs -p | grep "$VIEWERPID" ; then
+            kill "$VIEWERPID"
+        fi
 
-		MIME="$(file -b --mime-type "$FILE")"
+        MIME="$(file -b --mime-type "$FILE")"
 
-		case "$MIME" in
-			video/*)
-				if which mpv ; then
-					mpv --wid="$XID" "$FILE" &
-				else
-					term_nuke "$XID" "$FILE"
-				fi
-				;;
-			audio/*)
-				if which mpv ; then
-					mpv --force-window --wid="$XID" "$FILE" &
-				else
-					term_nuke "$XID" "$FILE"
-				fi
-				;;
-			image/*)
-				if which sxiv ; then
-					sxiv -e "$XID" "$FILE" &
-				else
-					term_nuke "$XID" "$FILE"
-				fi
-				;;
-			application/pdf)
-				if which zathura ; then
-					zathura -e "$XID" "$FILE" &
-				else
-					term_nuke "$XID" "$FILE"
-				fi
-				;;
-			inode/directory)
-				$TERMINAL "$XID" -e nnn "$FILE" &
-				;;
-			text/*)
-				if [ -x "$NUKE" ] ; then
-					term_nuke "$XID" "$FILE"
-				else
-					# shellcheck disable=SC2086
-					$TERMINAL "$XID" -e $PAGER "$FILE" &
-				fi
-				;;
-			*)
-				if [ -x "$NUKE" ] ; then
-					term_nuke "$XID" "$FILE"
-				else
-					$TERMINAL "$XID" -e sh -c "file '$FILE' | $PAGER -" &
-				fi
-				;;
-		esac
-		get_viewer_pid
-	done
-	kill "$TABBEDPID"
+        case "$MIME" in
+            video/*)
+                if which mpv ; then
+                    mpv --wid="$XID" "$FILE" &
+                else
+                    term_nuke "$XID" "$FILE"
+                fi
+                ;;
+            audio/*)
+                if which mpv ; then
+                    mpv --force-window --wid="$XID" "$FILE" &
+                else
+                    term_nuke "$XID" "$FILE"
+                fi
+                ;;
+            image/*)
+                if which sxiv ; then
+                    sxiv -e "$XID" "$FILE" &
+                else
+                    term_nuke "$XID" "$FILE"
+                fi
+                ;;
+            application/pdf)
+                if which zathura ; then
+                    zathura -e "$XID" "$FILE" &
+                else
+                    term_nuke "$XID" "$FILE"
+                fi
+                ;;
+            inode/directory)
+                $TERMINAL "$XID" -e nnn "$FILE" &
+                ;;
+            text/*)
+                if [ -x "$NUKE" ] ; then
+                    term_nuke "$XID" "$FILE"
+                else
+                    # shellcheck disable=SC2086
+                    $TERMINAL "$XID" -e $PAGER "$FILE" &
+                fi
+                ;;
+            *)
+                if [ -x "$NUKE" ] ; then
+                    term_nuke "$XID" "$FILE"
+                else
+                    $TERMINAL "$XID" -e sh -c "file '$FILE' | $PAGER -" &
+                fi
+                ;;
+        esac
+        get_viewer_pid
+    done
+    kill "$TABBEDPID"
 }
 
 if [ ! -r "$NNN_FIFO" ] ; then
-	echo "Can't read \$NNN_FIFO ('$NNN_FIFO')"
-	exit 1
+    echo "Can't read \$NNN_FIFO ('$NNN_FIFO')"
+    exit 1
 fi
 
 previewer_loop < "$NNN_FIFO" &
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -17,6 +17,7 @@
 #    dependencies, see the script itself
 #  - vim (or any editor/pager really)
 #  - file
+#  - xdotool (optional, to keep main window focused)
 #
 # How to use:
 #  First, install the dependencies. Then you need to set a NNN_FIFO path
@@ -72,7 +73,7 @@ start_tabbed () {
 
 	tabbed > "$FIFO" &
 
-	jobs >/dev/null 2>&1 # Get rid of the "Completed" entries
+	jobs # Get rid of the "Completed" entries
 
 	TABBEDPID="$(jobs -p %%)"
 
@@ -92,10 +93,15 @@ get_viewer_pid () {
 
 previewer_loop () {
 	unset -v NNN_FIFO
-	start_tabbed
-
 	# mute from now
 	exec >/dev/null 2>&1
+
+	MAINWINDOW="$(xdotool getactivewindow)"
+
+	start_tabbed
+
+	xdotool windowactivate "$MAINWINDOW"
+	xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &
 
         while read -r FILE ; do
 
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -1,0 +1,179 @@
+#!/bin/bash
+
+# Description: tabbed/xembed based file previewer
+#
+# Note: This plugin needs a "NNN_FIFO" to work. See man.
+#
+# Shell: Bash (job control is weakly specified in POSIX)
+# 
+# Dependencies:
+#  - tabbed (https://tools.suckless.org/tabbed): xembed host
+#  - xterm (or urxvt or st) : xembed client for text-based preview
+#  - mpv (https://mpv.io): xembed client for video/audio
+#  - sxiv (https://github.com/muennich/sxiv): xembed client for images
+#  - zathura (https://pwmt.org/projects/zathura): xembed client for PDF documents
+#  - nnn's nuke plugin for text preview and fallback (should be in plugins directory)
+#    nuke is a fallback for 'mpv', 'sxiv', and 'zathura', but it has has its own
+#    dependencies, see the script itself
+#  - vim (or any editor/pager really)
+#  - file
+#
+# How to use:
+#  First, install the dependencies. Then you need to set a NNN_FIFO path
+#  and set a key for the plugin, then start `nnn`:
+#
+#    $ NNN_FIFO=/tmp/nnn.fifo nnn
+#
+#  Then in `nnn`, launch the `preview-tabbed` plugin.
+#
+#  If you provide the same NNN_FIFO to all nnn instances, there will be a
+#  single common preview window. I you provide different FIFO path, they
+#  will be independent.
+#
+# How it works:
+#   We use `tabbed` [1] as a xembed [2] host, to have a single window
+#   owning each previewer window. So each previewer must be a xembed client.
+#   For text previewers, this is not an issue, as there are a lot of
+#   xembed-able terminal emulator (we default to `xterm`, but examples are
+#   provided for `urxvt` and `st`). For graphic preview this can be trickier,
+#   but a few popular viewers are xembed-able, we use:
+#     - `mpv`: multimedia player, for video/audio preview
+#     - `sxiv`: image viewer
+#     - `zathura`: PDF viewer
+#     - but we allways fallback to `nuke` plugin
+#
+# [1]: http://tools.suckless.org/tabbed/
+# [2]: https://specifications.freedesktop.org/xembed-spec/xembed-spec-latest.html
+
+PAGER=${PAGER:-"vim -R"}
+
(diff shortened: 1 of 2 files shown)


leovilok:
> Yes, we need something that is not heavily dependent on any specific tool, only utilities those can be replaced easily (like preferred image viewers, editors etc.)

@jarun this is not the path taken by this plugin, where `bash`, `tabbed` (and even `mpv`, `sxiv`, `zathura`, ...) are rather hard dependencies.

There are a few issue related to that:
 1. we will spawn and kill processes, and the command `jobs` (more generally job control) in not well specified in POSIX sh, so everything is shell dependent.
 2. we want every preview to be displayed in the same window, especially in tiling WM, where adding/removing windows can change all the layout. To achieve this I used tabbed and the xembed protocol: tabbed acts as a host windows that allways stays in place, and file viewers get embeded in it. But not every application is xembed capable, hence my little selection, and more dependencies ...

If you follow @toddyamakawa 's path (doing text previews only), I think that you can be more flexible on dependencies.

Note that there are other options too:
 - spawning previewers in their own windows, but forcing their window geometry (I think that more programs support that that xembed spec)
 - use a xterm window for text preview, but still diplay images on it with sixels or `w3mimgdisplay`, *à la* `ranger` (but I like having full featured image/document/multimedia viewers).

toddyamakawa:
> `tabbed`

What is `tabbed` anyway? It sounds like an interesting tool but I can't seem to find any information about it.

leovilok:
@toddyamakawa http://tools.suckless.org/tabbed/
Its primary use is to provide tab support to software that don't support tabs, I believe.

plugins/preview
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
jarun, line 1: Is POSIX `sh` not possible?

Also please rename the file to `previewer-tabbed`.

In short, previewer plugins would start with `previewer-`.

toddyamakawa:
That looks incredibly useful. I'll look into it. Thanks for sharing!

plugins/preview
```diff
@@ -0,0 +1,109 @@
...
+
+# Description: tabbed/xembed based file previewer
+#
+# Note: This plugin needs a "NNN_FIFO" to work. See man.
+#
+# Shell: Bash (job control is weakly specified in POSIX)
+# 
+# Dependencies:
+#  - tabbed : xembed host
```
jarun, line 10: Please link to the site. I didn't know `tabbed` till date.

plugins/preview
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+# Description: tabbed/xembed based file previewer
```
jarun, line 3: Please add this to plugin README table.

jarun:
```
This plugin is written in bash, because job control is not well
specified in POSIX sh (`jobs` can return anything).

We use [`tabbed`][1] as a [xembed][2] host, to have a single window
owning each previewer window. So each previewer must be a xembed client.
For text previewers, this is not an issue, as there are a lot of
xembed-able terminal emulator (we default to `xterm`, but examples are
provided for `urxvt` and `st`). For other filetypes, this can be
trickier, but a few popular viewers are xembed-able, we use:
 - `mpv`: multimedia player, for video/audio preview
 - `sxiv`: image viewer
 - `zathura`: PDF viewer

How to use:
-----------

First, install the dependencies. Then you need to set a NNN_FIFO path
and set a key for the plugin, then start `nnn`:

    $ NNN_FIFO=/tmp/nnn.fifo NNN_PLUG='p:preview' nnn

Then in `nnn`, press `;p`.

[1]: http://tools.suckless.org/tabbed/
[2]: https://specifications.freedesktop.org/xembed-spec/xembed-spec-latest.html
```

This goes to in-file documentation. See `nuke`.

plugins/preview
jarun: > Also please rename the file to previewer-tabbed.

Please rename to `preview-tabbed`.

plugins/preview
leovilok: Added.

plugins/preview
leovilok: Done.

plugins/preview
leovilok: I'm not sure how to do proper job control in POSIX sh.

Renamed to `preview-tabbed`.

leovilok:
I now use `nuke` (in its default text mode):
1. as a fallback if you don't have either mpv or sxiv or zathura installed
2. as the default to preview text files

It's probably a bit slower, and it exits a bit quickly on some filetypes (blacklisted ones, and unknown, where it doesn't open a pager to retain info). But I feel it's still an improvement.

plugins/preview
jarun: > I'm not sure how to do proper job control in POSIX sh.

Dear wizard @KlzXS, is this doable?

KlzXS: I'll check it out.

KlzXS:
Is there no way to preview files in the same window? I just tried this and it tried to open everything I hovered over in a new window stealing focus, which in my opinion makes it a bit annoying to use.

Also is there a way to disable it after starting?

plugins/preview-tabbed
```diff
@@ -0,0 +1,171 @@
...
+#    nuke is a fallback for 'mpv', 'sxiv', and 'zathura', but it has has its own
+#    dependencies, see the script itself
+#  - vim (or any editor/pager really)
+#  - file
+#
+# How to use:
+#  First, install the dependencies. Then you need to set a NNN_FIFO path
+#  and set a key for the plugin, then start `nnn`:
+# 
+#    $ NNN_FIFO=/tmp/nnn.fifo NNN_PLUG='p:preview' nnn
```
jarun, line 25: I think we can omit the NNN_PLUG setting. How to use plugins is not specific to a single plugin.

leovilok:
> Is there no way to preview files in the same window? I just tried this and it tried to open everything I hovered over in a new window stealing focus, which in my opinion makes it a bit annoying to use.

You have to install [tabbed](https://tools.suckless.org/tabbed/), as explained in the in-file docs. I didn't even know that it was possible to use the plugin without tabbed...

> Also is there a way to disable it after starting?

Not currently (well appart from closing `nnn`), but @jarun plans to have a `F` key, blocking FIFO output.

Also, at the moment, I restart `tabbed` every time it's closed, but maybe I should just quit when the tabbed window is closed, since it's a plugin and it can be restarted anyways.

plugins/preview-tabbed
leovilok: Well that was the ELI5 part :wink:
But this might belong more to the preview Wiki section.

jarun: Are you saying this also can't be a plugin?

jarun: > But this might belong more to the preview Wiki section.

I am confused now.

jarun:
> I just tried this and it tried to open everything I hovered over in a new window stealing focus, which in my opinion makes it a bit annoying to use

Same here with `tabbed` installed. We have to find a way around it.

plugins/preview-tabbed
leovilok: Reworded

jarun:
> but maybe I should just quit when the tabbed window is closed, since it's a plugin and it can be restarted anyways.

Yes, this is the expected behaviour.

jarun:
One of the latest 2 pushes breaks the interface. I see this:

/[1]+  Running                 tabbed > "$FIFO"

leovilok:
Now the plugin should not start when tabbed can't be started.

And it should exit when tabbed window is closed.

jarun:
Is this the output from `jobs`? Breaks the UI.

    /[1]+ Running tabbed > "$FIFO"

KlzXS:
It stops now, that's good. But I do have tabbed installed and it still starts in in a separate window.

leovilok:
> Is this the output from `jobs`? Breaks the UI.

Yes, this should be fixed now. I didn't see it because of the immediate window resize.

jarun:
OK. Now we need to think about the focus stealing prevention. Any ideas? I tried some settings on XFCE but to no avail.

leovilok:
> It stops now, that's good. But I do have tabbed installed and it still starts in in a separate window.

@KlzXS 
To be clear: only the first window spawn steals focus?
If it's how it behaves for you it's the same for me, if every file hover steal focus ... then my WM might be nicer to me than yours.

In both cases, this could be fixed with something like xdotools, but this would add dependencies.

KlzXS:
As far as I can tell there is only ever one window, but yes on each hover the focus gets changed to it. I am using i3, that might be why I can't see it respawning the window.

jarun:
On xfce I have 2 windows. :(

leovilok:
@jarun there should be 2 windows: 1 for `nnn` and one for preview.

If you have `xdotool` installed, I added a (quite bruteforce) focus stealing prevention.
Please confirm that it works (I tested xfce, will test i3)

jarun:
> If you have xdotool installed, I added a (quite bruteforce) focus stealing prevention.
Please confirm that it works (I tested xfce, will test i3)

Will try this now.

jarun:
Works like a charm!!!

Please confirm on i3 and we are done!

jarun:
Please replace the TABs with 4 spaces. We follow that for scripts generally.

leovilok:
Works in i3 too aparently, but now it's much harder to close the preview window in i3.

jarun:
> but now it's much harder to close the preview window in i3

Why is that happening?

jarun:
Keep it working for i3 (and other tiling window managers). And keep the lines/conditions needed for XFCE within comments (with instructions). Plugin are indicative. So we are good there. Let the people read and adapt.

leovilok:
The issue is that everytime the preview window is focused, focus get back to the main `nnn` terminal window. And in tiling WM, you typically need to focus a window to close it.

jarun:
OK. As I mentioned above:

Keep it working for i3 (and other tiling window managers). And keep the lines/conditions needed for XFCE within comments (with instructions). Plugin are indicative. So we are good there. Let the people read and adapt.

leovilok:
This new commit adds focus steal prevention timeout, this is not perfect because:
1. some files might take more than 2 secs to open and
2. you can't focus the preview window less than 2 secs after moving cursor in `nnn`...
But I feel like it's a good compromise (and it should work out of the box in XFCE too)

plugins/preview-tabbed
```diff
@@ -101,15 +101,20 @@ previewer_loop () {
...
+    #xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &
 
         while read -r FILE ; do
 
         jobs # Get rid of the "Completed" entries
 
         if ! jobs | grep tabbed ; then
-        # Uncomment and remove `break` want to restart tabbed everytime its window is closed
-        #   start_tabbed
+            # Uncomment and remove `break` want to restart tabbed everytime its window is closed
```
jarun, line 116: I don't think anyone without having prior info will be able to use this.

Let's have 2 plugins:

    preview-fltwm
    preview-regwm

I am sorry but looking at the changes this is no more minimal.

plugins/preview-tabbed
```diff
@@ -0,0 +1,198 @@
...
+                ;;
+        esac
+        get_viewer_pid
+
+        # following lines are not needed with the bruteforce xdotool method
+        ACTIVE_XID="$(xdotool getactivewindow)"
+        if [ $((ACTIVE_XID == XID)) -ne 0 ] ; then
+            xdotool windowactivate "$MAINWINDOW"
+        else
+            timeout 2 xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &
```
jarun, line 186: Maybe the timeout should be a global macro with a meaningful name so it's easily visible:

    XDOTOOL_TIMEOUT
?

leovilok: Done.

plugins/preview-tabbed
leovilok: I deleted the tabbed auto-restart comment.

This plugin works in every case. Do you want a separate plugin to spare the 2 extra xdotool calls per cursor move?

plugins/preview-tabbed
```diff
@@ -0,0 +1,197 @@
...
+
+    xdotool windowactivate "$MAINWINDOW"
+
+    # Bruteforce focus stealing prevention method,
+    # works well in floating window managers like XFCE
+    # but make interaction with the preview window harder
+    # (uncomment to use):
+    #xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &
+
+        while read -r FILE ; do
```
jarun, line 112: 4 spaces extra?

plugins/preview-tabbed
```diff
@@ -0,0 +1,197 @@
...
+                if [ -x "$NUKE" ] ; then
+                    term_nuke "$XID" "$FILE"
+                else
+                    $TERMINAL "$XID" -e sh -c "file '$FILE' | $PAGER -" &
+                fi
+                ;;
+        esac
+        get_viewer_pid
+
+        # following lines are not needed with the bruteforce xdotool method
```
jarun, line 180: How many lines? till line 186? A note on the number of lines would help.

plugins/preview-tabbed
jarun: In that case let's have only one.

plugins/preview-tabbed
jarun: I think this is fine.

plugins/preview-tabbed
```diff
@@ -0,0 +1,197 @@
...
+#   provided for `urxvt` and `st`). For graphic preview this can be trickier,
+#   but a few popular viewers are xembed-able, we use:
+#     - `mpv`: multimedia player, for video/audio preview
+#     - `sxiv`: image viewer
+#     - `zathura`: PDF viewer
+#     - but we allways fallback to `nuke` plugin
+#
+# [1]: http://tools.suckless.org/tabbed/
+# [2]: https://specifications.freedesktop.org/xembed-spec/xembed-spec-latest.html
+
```
jarun, line 48: Please add your name/id as Author as in other plugins.

jarun:
>    Also is there a way to disable it after starting?

> Not currently (well appart from closing nnn), but @jarun plans to have a F key, blocking FIFO output.

@leovilok @KlzXS with the latest revision of the plugin do we need this anymore?

jarun:
I'm merging this. I will make the 2 changes.

KlzXS:
What's "this"? F for blocking pipe? I don't know, I'm indifferent to it.

jarun:
> What's "this"? F for blocking pipe? I don't know, I'm indifferent to it.

OK. Then fine. With the update from @leovilok in the discussion room doesn't look like we need it anyway,

leovilok:
@jarun It's not needed for this plugin anymore

jarun:
We are done then! Thank you very much! It's a great value addition to `nnn`!

jarun:
@leovilok I see one minor problem. When I close `nnn` the plugin does not close. I remember this was working in a draft version...

jarun:
And the next time I start `nnn`, preview starts immediately (without even plugin invoked).

leovilok:
Did you have a second running `nnn` instance, with the same `NNN_FIFO`?

jarun:
No. Only one instance and after I quit the previewer doesn't quit.

jarun:
My script:

```
#!/bin/bash

# Description: tabbed/xembed based file previewer
#
# Note: This plugin needs a "NNN_FIFO" to work. See man.
#
# Shell: Bash (job control is weakly specified in POSIX)
#
# Dependencies:
#  - tabbed (https://tools.suckless.org/tabbed): xembed host
#  - xterm (or urxvt or st) : xembed client for text-based preview
#  - mpv (https://mpv.io): xembed client for video/audio
#  - sxiv (https://github.com/muennich/sxiv): xembed client for images
#  - zathura (https://pwmt.org/projects/zathura): xembed client for PDF documents
#  - nnn's nuke plugin for text preview and fallback (should be in plugins directory)
#    nuke is a fallback for 'mpv', 'sxiv', and 'zathura', but it has has its own
#    dependencies, see the script itself
#  - vim (or any editor/pager really)
#  - file
#  - xdotool (optional, to keep main window focused)
#
# How to use:
#  First, install the dependencies. Then you need to set a NNN_FIFO path
#  and set a key for the plugin, then start `nnn`:
#
#    $ NNN_FIFO=/tmp/nnn.fifo nnn
#
#  Then in `nnn`, launch the `preview-tabbed` plugin.
#
#  If you provide the same NNN_FIFO to all nnn instances, there will be a
#  single common preview window. I you provide different FIFO path, they
#  will be independent.
#
# How it works:
#   We use `tabbed` [1] as a xembed [2] host, to have a single window
#   owning each previewer window. So each previewer must be a xembed client.
#   For text previewers, this is not an issue, as there are a lot of
#   xembed-able terminal emulator (we default to `xterm`, but examples are
#   provided for `urxvt` and `st`). For graphic preview this can be trickier,
#   but a few popular viewers are xembed-able, we use:
#     - `mpv`: multimedia player, for video/audio preview
#     - `sxiv`: image viewer
#     - `zathura`: PDF viewer
#     - but we allways fallback to `nuke` plugin
#
# [1]: http://tools.suckless.org/tabbed/
# [2]: https://specifications.freedesktop.org/xembed-spec/xembed-spec-latest.html
#
# Author: leovilok


XDOTOOL_TIMEOUT=2
PAGER=${PAGER:-"vim -R"}
NUKE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/nuke"


if which xterm >/dev/null 2>&1 ; then
    TERMINAL="xterm -into"
elif which urxvt >/dev/null 2>&1 ; then
    TERMINAL="urxvt -embed"
elif which st >/dev/null 2>&1 ; then
    TERMINAL="st -w"
else
    echo "No xembed term found" >&2
fi


term_nuke () {
    # $1 -> $XID, $2 -> $FILE
    $TERMINAL "$1" -e "$NUKE" "$2" &
}

start_tabbed () {
    FIFO="$(mktemp -u)"
    mkfifo "$FIFO"

    tabbed > "$FIFO" &

    jobs # Get rid of the "Completed" entries

    TABBEDPID="$(jobs -p %%)"

    if [ -z "$TABBEDPID" ] ; then
        echo "Can't start tabbed"
        exit 1
    fi

    read -r XID < "$FIFO"

    rm "$FIFO"
}

get_viewer_pid () {
        VIEWERPID="$(jobs -p %%)"
}

previewer_loop () {
    unset -v NNN_FIFO
    # mute from now
    exec >/dev/null 2>&1

    MAINWINDOW="$(xdotool getactivewindow)"

    start_tabbed

    xdotool windowactivate "$MAINWINDOW"

    # Bruteforce focus stealing prevention method,
    # works well in floating window managers like XFCE
    # but make interaction with the preview window harder
    # (uncomment to use):
    xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &

    while read -r FILE ; do

        jobs # Get rid of the "Completed" entries

        if ! jobs | grep tabbed ; then
            break
        fi

        if [ ! -e "$FILE" ] ; then
            continue
        fi

        if [ -n "$VIEWERPID" ] && jobs -p | grep "$VIEWERPID" ; then
            kill "$VIEWERPID"
        fi

        MIME="$(file -b --mime-type "$FILE")"

        case "$MIME" in
            video/*)
                if which mpv >/dev/null 2>&1 ; then
                    mpv --force-window=immediate --loop-file --wid="$XID" "$FILE" &
                else
                    term_nuke "$XID" "$FILE"
                fi
                ;;
            audio/*)
                if which mpv >/dev/null 2>&1 ; then
                    mpv --force-window=immediate --loop-file --wid="$XID" "$FILE" &
                else
                    term_nuke "$XID" "$FILE"
                fi
                ;;
            image/*)
                if which sxiv >/dev/null 2>&1 ; then
                    sxiv -e "$XID" "$FILE" &
                else
                    term_nuke "$XID" "$FILE"
                fi
                ;;
            application/pdf)
                if which zathura >/dev/null 2>&1 ; then
                    zathura -e "$XID" "$FILE" &
                else
                    term_nuke "$XID" "$FILE"
                fi
                ;;
            inode/directory)
                $TERMINAL "$XID" -e nnn "$FILE" &
                ;;
            text/*)
                if [ -x "$NUKE" ] ; then
                    term_nuke "$XID" "$FILE"
                else
                    # shellcheck disable=SC2086
                    $TERMINAL "$XID" -e $PAGER "$FILE" &
                fi
                ;;
            *)
                if [ -x "$NUKE" ] ; then
                    term_nuke "$XID" "$FILE"
                else
                    $TERMINAL "$XID" -e sh -c "file '$FILE' | $PAGER -" &
                fi
                ;;
        esac
        get_viewer_pid

        # following lines are not needed with the bruteforce xdotool method
        ACTIVE_XID="$(xdotool getactivewindow)"
        if [ $((ACTIVE_XID == XID)) -ne 0 ] ; then
            xdotool windowactivate "$MAINWINDOW"
        else
            timeout "$XDOTOOL_TIMEOUT" xdotool behave "$XID" focus windowactivate "$MAINWINDOW" &
        fi
    done
    kill "$TABBEDPID"
}

if [ ! -r "$NNN_FIFO" ] ; then
    echo "Can't read \$NNN_FIFO ('$NNN_FIFO')"
    exit 1
fi

previewer_loop < "$NNN_FIFO" &
disown
```

leovilok:
@jarun I pushed the fix directly to master. This could have been hard to debug, I'm glad I found out very quicky.

This only happened when you opened a previewer, then a second one. Because when the plugin is spawed the 2nd time, the FIFO have already been opened for writing by `nnn`, and child process inherited this open fd. But they never used this fd, and it stayed open for writing, and never closed. And as the previewer windows exits when the fd closes, it stayed open forever.

If you want to try out the fix with an existing FIFO (used before), I would advise running `fuser $NNN_FIFO` and killing remaining previwers, if there are.

jarun:
It's fixed now! Thank you!

Please try to avoid direct pushes. Try to get it reviewed by at least one.

leovilok:
> Please try to avoid direct pushes. Try to get it reviewed by at least one.

OK, but even for small/obvious fixes like this one?

jarun:
Yes, still. If everyone pushes 8/9 small fixes like these by themselves we'll be unable to keep track.

But yes, I do that myself. The reason is, the team gathered 2 years after lone development (by me) so there was no one to review anyway. And now, old habits die hard ;). At the same time, when I have concerns I always get it reviewed by @KlzXS post push.

PR #554, opened 2020-05-05
Add example xterm preview window

toddyamakawa:
Here is a minimal example for bash/zsh with an xterm dependency.

To use it:

1. Source the `misc/preview/preview.bash_zsh`
2. Run `n`

An `xterm` window will pop up and run the `xterm-preview` plugin. The plugin simply prints the `tree` if it' a directory or the `head` if it's a file. The `xterm` window will automatically close when `nnn` closes.

### DIFF
--- a/misc/preview/preview.bash_zsh
+++ b/misc/preview/preview.bash_zsh
@@ -1,0 +1,39 @@
+n ()
+{
+    # Block nesting of nnn in subshells
+    if [ -n "$NNNLVL" ] && [ "${NNNLVL:-0}" -ge 1 ]; then
+        echo "nnn is already running"
+        return
+    fi
+
+    # The default behaviour is to cd on quit (nnn checks if NNN_TMPFILE is set)
+    # To cd on quit only on ^G, remove the "export" as in:
+    #     NNN_TMPFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.lastd"
+    # NOTE: NNN_TMPFILE is fixed, should not be modified
+    export NNN_TMPFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.lastd"
+
+    # This will create a fifo where all nnn selections will be written to
+    NNN_FIFO="$(mktemp --suffix=-nnn -u)"
+    export NNN_FIFO
+    (umask 077; mkfifo "$NNN_FIFO")
+
+    # Preview command
+    preview_cmd="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/preview"
+
+    # Use `tmux` split as preview
+    if [ -e "${TMUX%%,*}" ]; then
+        tmux split-window -e "NNN_FIFO=$NNN_FIFO" -dh "$preview_cmd"
+
+    # Use `xterm` as a preview window
+    elif (which xterm &> /dev/null); then
+        xterm -e "$preview_cmd" &
+
+    # Unable to find a program to use as a preview window
+    else
+        echo "unable to open preview, please install tmux or xterm"
+    fi
+
+    nnn "$@"
+
+    rm -f "$NNN_FIFO"
+}
--- a/plugins/preview
+++ b/plugins/preview
@@ -1,0 +1,60 @@
+#!/usr/bin/env sh
+
+# #############################################################################
+# Description: Minimal example to preview files and directories
+#              No external dependencies
+#              Should NOT be called by nnn
+#              Automatically exits when the NNN_FIFO closes
+#              Prints a `tree` if directory or `head` if it's a file
+#
+# Shell: POSIX compliant
+# Author: Todd Yamakawa
+# Usage: preview
+#
+# Integration with nnn:
+#   1. Source the setup script
+#         source misc/preview/preview.bash_zsh
+#         # This defines a new function `n`
+#         # `n` will automatically export and open a NNN_FIFO
+#         # This setup script needs `xterm` or `tmux 3.0+`
+#
+#   2. Run `n`
+#
+# Details:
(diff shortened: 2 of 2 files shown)


toddyamakawa:
Wow, I didn't know the CI system checks shell scripts. That's pretty cool because I didn't even know a utility like this existed until this moment. Is there a way I can run these checks before pushing it go GitHub?

plugins/xterm-preview
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env sh
```
jarun, line 1: Please rename this to previewer-xterm

misc/preview/preview.bash_zsh
```diff
@@ -0,0 +1,26 @@
...
+    #     NNN_TMPFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.lastd"
+    # NOTE: NNN_TMPFILE is fixed, should not be modified
+    export NNN_TMPFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.lastd"
+
+    # This will create a fifo where all nnn selections will be written to
+    export NNN_FIFO="$(mktemp --suffix=-nnn -u)"
+    (umask 077; mkfifo "$NNN_FIFO")
+
+    # Open xterm preview window
+    preview="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/xterm-preview"
```
jarun, line 20: Now we support `NNN_SEL` if it's defined. Please add.

jarun:
Yes, you can run [shellcheck](https://www.shellcheck.net/) locally.

misc/preview/preview.bash_zsh
toddyamakawa: My understanding is that the `previewer` window is independent of `NNN_SEL`. I'm not quite sure what to do with `NNN_SEL` in the preview window.

plugins/previewer-xterm
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env sh
```
jarun, line 1: Please follow documentation style used in other plugins.

jarun:
```
Here is a minimal example for bash/zsh with an xterm dependency.

To use it:

    Source the misc/preview/preview.bash_zsh
    Run n

An xterm window will pop up and run the xterm-preview plugin. The plugin simply prints the tree if it' a directory or the head if it's a file. The xterm window will automatically close when nnn closes.
```

Can we have this usage documented in-file? Like in `nuke`?

plugins/previewer-xterm
jarun: Also, please add this to the plugin README table. The entry order is alphabetical.

misc/preview/preview.bash_zsh
jarun: Ahh sorry. I some point I changed the prefix in batch and everything I see starting with `"${XDG_CONFIG_HOME:-$HOME/.config}` registers as we need to add `NNN_SEL` before it. ;)

Ignore it.

leovilok:
Nice script! And with really few dependencies too.

`plugins/previewer-xterm` is not a plugin, is it? at least not one that can be launched from `nnn`, does it belong to that directory? Also it's not tied to `xterm` at all, and you could reuse it in a tmux version. I even think that most of the code could be shared between xterm/tmux versions.

jarun:
Yes, it's not a plugin per se.

@leovilok and @toddyamakawa please start the names with `preview-`. The `previewer-` makes it too long.

plugins/previewer-xterm
```diff
@@ -0,0 +1,32 @@
...
+
+# Check FIFO
+NNN_FIFO=${NNN_FIFO:-$1}
```
jarun, line 4: Please rename to `preview-xterm`. The other prefix is too long. Sorry for the rework.

toddyamakawa:
> Nice script! And with really few dependencies too.

Thanks!

> `plugins/previewer-xterm` is not a plugin, is it? at least not one that can be launched from `nnn`, does it belong to that directory? 

Agree. @jarun, where should I put this file anyways? It's definitely not meant to be invoked directly from `nnn`

> Also it's not tied to `xterm` at all, and you could reuse it in a tmux version. I even think that most of the code could be shared between xterm/tmux versions.

Good point. I could probably make this so it checks if you're already in `tmux` using `[[ -e ${TMUX%%,*} ]]`, so that way if you're in tmux it splits the pane. Otherwise it uses xterm.

So now the question remains is where should I put this file? And what should I name it? This exact same file will be called whether it's using `xterm` or `tmux`.

jarun:
> where should I put this file anyways? It's definitely not meant to be invoked directly from nnn

Add a new Wiki page: Previews

Add both the scripts there with all the documentation.

jarun:
Please add the tmux version as well.

plugins/previewer-xterm
toddyamakawa: Quick clarification, since this isn't a plugin I shouldn't add it to the plugin table. Is that correct?

plugins/previewer-xterm
toddyamakawa: No problem, I think it's important to get this right, so feel free to nitpick as much as possible.

toddyamakawa:
> Add a new Wiki page: Previews
> Add both the scripts there with all the documentation.

Sounds good. Should I abandon this PR and put everything in the Wiki instead?

plugins/previewer-xterm
jarun: Yes, all of the scripts go to Wiki. Nothing will be added to source code.

jarun:
> Sounds good. Should I abandon this PR and put everything in the Wiki instead?

Yes, I am closing it.

PR #555, opened 2020-05-05
Double click only works on same item

0xACE:
This patch disables accidental triggering of double clicking when you
are just single clicking on multiple objects fast.

Btw regarding long hold clicking on termux, on my phone a menu pops up, are you sure you want me to look into long hold clicking on termux?

The reason I'm posting this commit is because I rebased my personal branch on `upstream/master` at the cost of dumping some of my personal patches, just to keep up with this project.

The merge conflicts were severe the past 6months. Lets see what will break in my personal build after this event.

(Sry for posting it here, but I didn't want to start another thread/discussion for this)

EDIT: Note this patch was written in the spur of the moment because I found it annoying as I was testing out `nnn`, but I think it's difficult for this to introduce new problems

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5190,6 +5190,7 @@ static bool browse(char *ipath, const char *session)
 #ifndef NOMOUSE
 	MEVENT event;
 	struct timespec mousetimings[2] = {{.tv_sec = 0, .tv_nsec = 0}, {.tv_sec = 0, .tv_nsec = 0} };
+	int mousedent[2] = {-1, -1};
 	bool currentmouse = 1;
 	bool rightclicksel = 0;
 #endif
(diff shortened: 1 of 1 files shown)


jarun:
> Btw regarding long hold clicking on termux, on my phone a menu pops up, are you sure you want me to look into long hold clicking on termux?

Nope. Leave it.

jarun:
How does this help? I do not use mouse so help me out and Termux is off as I am home. So help me here.

0xACE:
Previously when you single clicked all over the screen:
If 2 sequential clicks where fast enough it would register as a double click.

Say the first click was on `my_document.pdf` at `0.0seconds` and the second click was on `big_mistake.jpg` at `0.1seconds` this would case `big_mistake.jpg` to register as a double click.

The new method:
Double click ONLY registers if you are double clicking on the the same `cur` (basically same `dents[cur]`)

so both mouse clicks have to be on the same item.


This new method is more inline with GUI tools utilizing double click.

jarun:
Awesome! I'll test it.

jarun:
I'll test after it's in! ;)

leovilok:
Oh, I think this got me sometimes, I didn't understand why. Thanks @0xACE !

PR #559, opened 2020-05-06
preview-tui: fix tmux behaviour, del refs to nuke

leovilok:
As reported by @toddyamakawa

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -4,7 +4,7 @@
 #
 # Note: This plugin needs a "NNN_FIFO" to work.
 #
-# Dependencies: tmux or xterm (or set $TERMINAL), file, tree
+# Dependencies: tmux (>=3.0) or xterm (or set $TERMINAL), file, tree
 #
 # How to use:
 #   You need to set a NNN_FIFO path and set a key for the plugin,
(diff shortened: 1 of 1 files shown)


jarun:
Thanks guys! :+1:

jarun:
@leovilok just checked this out! very useful indeed.

A problem I am seeing with this (using tmux): when I start the plugin, the currently hovered file is not shown (but the new tmux pane is created). When I scroll up and down I can see the previews.

jarun:
And that's probably with text files only. Hover on Makefile or README.md, then launch the plugin and you'll probably see it. Works fine when starting hovered on an image.

leovilok:
At the moment, we need a call to `move_cursor` to send the first write to the FIFO. Any resize/actual cursor move should do, there might be a redraw after plugin call, but it's sometimes too early (the plugin didn't open the FIFO yet).

I could pass `$nnn` to the previewer, to make sure the first file is displayed, I'll see if it's not too much code added.

jarun:
The hovered file is always passed to a plugin by design:

https://github.com/jarun/nnn/tree/master/plugins#access-level-of-plugins

jarun:
BTW, the other plugin `preview-tabbed` works as expected.

PR #563, opened 2020-05-09
Allow plugins to write to $NNN_PIPE in background

leovilok:
This is a workaround to the fact that `nnn` reads from the pipe only
*after* plugin returned, so plugins couldn't write more that pipe
capacity to $NNN_PIPE or they would block.

You might want to increase `LIST_FILES_MAX` to test on large file lists.

@jarun @KlzXS please test.

I'm not sure about how I should handle error checking, too.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -87,6 +87,7 @@
 #else
 #include <regex.h>
 #endif
+#include <poll.h>
 #include <signal.h>
 #include <stdarg.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -4245,6 +4246,35 @@ static void readpipe(int fd, char **path, char **lastname, char **lastdir)
 {
 	int r;
 	char ctx, *nextpath = NULL;
+
+	struct pollfd pfd = { .fd = fd, .events = POLLIN };
+
+	/* This will block indefinitely if fd is opened for writiing by someone,
```
jarun, line 4252: What happens if there are no search results and nothing is written to pipe?

leovilok: Then no process have the pipe open for writing, so `poll` puts POLLHUP (or maybe POLLERR) in `pfd.revents`, so we return without atempting any read.

leovilok: I checked with other plugins that don"t use `NNN_PIPE`, and it works.

jarun: Let's see if we can use the method from @zmwangx.

leovilok: yes, I think this is the way to go.

jarun:
@leovilok I ma closing this for now. Let's test master and see if we find some issues with the current implementation.

PR #568, opened 2020-05-12
Make plugins bindable to F* keys

leovilok:
I know that is not a goal for the project, but I leave it here for reference.

This make plugins directly bindable to F1, F2, ... , F12 keys:
~~~sh
NNN_PLUG='t:preview-tui;F3:finder' nnn
~~~

Internally, in the plugin key value store, it binds F1 ... F12 to (char)1 to (char)12, which are not printable.


jarun:
Closing this as we are taking the alternative route to use `Alt`.

PR #569, opened 2020-05-13
Allow launching plugins with Alt + plugin's key

leovilok:
In terminals Alt+*key* is 2 characters : `'\033'` (or Esc, or ^[) and the *key*.
Just setting `'\033'` as a binding for the plugin key makes all plugins available on Alt+*plugin's key*.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2564,7 +2564,14 @@ static int filterentries(char *path, char *lastname)
 #ifndef NOMOUSE
 		case KEY_MOUSE: // fallthrough
 #endif
-		case 27: /* Exit filter mode on Escape */
+		case 27: /* Exit filter mode on Escape and Alt+key */
+			nodelay(getwin(stdout), TRUE);
+			r = get_wch(ch);
+			nodelay(getwin(stdout), FALSE);
+			if (r != ERR) {
+				unget_wch(*ch);
+				*ch = CONTROL('S');
+			}
 			goto end;
 		}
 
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -2776,7 +2783,12 @@ static char *xreadline(const char *prefill, const char *prompt)
 				len -= pos;
 				pos = 0;
 				continue;
-			case 27: /* Exit prompt on Escape */
+			case 27: /* Exit prompt on Escape, but just filter out Alt+key */
```
jarun, line 2786: I think we are good if we don't do anything in `xreadline()`. Can we remove this handling?

leovilok: If we don't do anything any Alt+*key* will get out of xreadline and trigger global keybinding *key*.

jarun: OK. I thought it's pretty weird someone will press an Alt-key combo when providing an archive name. But I see this way we are safer.

src/nnn.c
```diff
@@ -2564,7 +2564,14 @@ static int filterentries(char *path, char *lastname)
 #ifndef NOMOUSE
 		case KEY_MOUSE: // fallthrough
 #endif
-		case 27: /* Exit filter mode on Escape */
+		case 27: /* Exit filter mode on Escape and Alt+key */
+			nodelay(getwin(stdout), TRUE);
```
jarun, line 2568: Can we have a wrapper API for the 3 common lines? We may need it again.

leovilok: Tomorrow ... it's 6:46 am here ...

jarun: I'll take care of it. I'm merging it. Thank you!

jarun:
Thank you!

jarun:
This breaks nav-as-you-type.

nnn -n
Press Esc (prompt no longer removed)

PR #84, opened 2018-02-28
Integer-only coolsize()

arsv:
Quick test to compare the float and the int versions:

https://gist.github.com/arsv/425db2e53f23209d3587ed1c061fa9de

PR #570, opened 2020-05-13
Added support for 10 contexts

0xACE:
10 contexts are more than 4 contexts - more is more better.

Jokes aside, the previous commits were playing around with some bitwise
math. It is strongly advised to adopt this commit.


jarun:
> the previous commits were playing around with some bitwise math

That bitwise math is much faster than modulo by 10.

This commit increases memory consumption for people like me who don't need this. It also litters the code with tons of repetitive cases, overflows our cfg struct and what not!

0xACE:
Fair enough.

> overflows our cfg struct and what not!

Out of curiousity: overflows where/how? I missed that. afaik, c standards ensures uint to be greater than 4 bits... This is the only part I'm interested in if you have the time to reply.

> That bitwise math is much faster than modulo by 10.

True, yet it only works for the case of `CTX_MAX == 2^x` and nothing else, the closes compromise I can think of is 8 contexts.

> It also litters the code with tons of repetitive cases,

This can be alleviated, at the cost of:

1. unreadable code
2. requiring context key bindings to be `1-9` (as in peeking that the char that was read) (drops 10th context since it requires special attention)

Not that I would advice this approach

> This commit increases memory consumption for people like me who don't need this.

What was the idea/expectations when you proposed I raise a PR?

Do you want the patch to compromise between keeping `4` as default and support change with `CTX_MAX` and keybindings if the user so wishes?

Fyi, my time is limited on this project until middle of july. Cheers

leovilok:
> This commit increases memory consumption for people like me who don't need this.

Well to be fair, this has no measurable impact on memory consumption if you don't use additional contexts. Is these tests, base `nnn` even used more memory (but it could have been the opposite):
screenshot

(`/tmp/test` is an empty dir, where I opened both versions using 1 context. `/tmp` is not empty, and I opened 4 context on `/tmp` in both version)

leovilok:
> That bitwise math is much faster than modulo by 10.

This is not an issue either, as:

1. these modulo are in shallow loops, 
2. they work on very small numbers, so they should be optimised anyways,
3. `nnn` spends all its time in syscalls anyways, so that's where optimisations are important for a file manager.

0xACE:
> Well to be fair, this has no measurable impact on memory consumption if you don't use additional contexts. In these tests, base `nnn` even used more memory (but it could have been the opposite):

The memory he is referring to is that the

* `settings` structure is now 2bits longer (honestly don't think the compiler/hardware cares, they usually allocate in chunks)
* 6 more more instances of `struct context`
* 6 more instances in `session_header_t`
* 6 more `cases` for handling `SEL_CTX4-10` (But I wouldn't be surprised if the compiler does some optimizations with these cases)

Again, all of these allocations are most likely not measurable in a normal computer.

> This is not an issue either, as:
> 
>1. these modulo are in shallow loops,
> 
>2. they work on very small numbers, so they should be optimised anyways,
> 
>3. `nnn` spends all its time in syscalls anyways, so that's where optimisations are important for a file manager.

True. Though similiar "optimizations" exist elsewhere in the code, where code readability is sacrificed for premature optimizations...

I'm not sure what hardware jarun is running this on, but it sounds like he runs it on the classic game boy at times... I personally still have `PII` processors, and even they don't measure this difference...

Thanks for your support though, I'd kiss you if it wasn't for the pandemic <3

jarun:
@0xACE and @leovilok we have to understand that this is not a use-case required by most users. 4 contexts are not _too few_.

> What was the idea/expectations when you proposed I raise a PR?

I wanted to have a feel of how it goes. When you raise a PR do not expect it will always be accepted. ;)
10 contexts on the top of the program do NOT look very appealing.

> this has no measurable impact on memory consumption

Each context is 8K+ and the increase in memory usage is majorly by:

    static context g_ctx[CTX_MAX] __attribute__ ((aligned));

> This is not an issue either
> where code readability is sacrificed for premature optimizations...

Whatever optimization we can get, we will get and increasing the number of contexts from 4 to 10 doesn't look like a reason compelling enough to sacrifice that.
Also, 10 is not a power of 2, so yes, using the bitwise operation is faster than a modulo. The compiler optimization you are talking of when dividing by 10 us actually a multiplication by 1/10 but that also requires floating point. That's one of the reasons we moved to integer-only calculation: https://github.com/jarun/nnn/pull/84
Also, there is no reason a file manager cannot optimize the code in areas other than sorting and reading entries in a directory. In `nnn`, we optimize wherever possible.

@0xACE yes, in case of context implementation we have some areas where the code is not generic but that's a different issue and can be worked out easily. For example, if we use a power of 2, it would be much simpler. 8 would be a good choice but I really hate those switch cases with 8 repetitive handlers.

> settings structure is now 2bits longer (honestly don't think the compiler/hardware cares, they usually allocate in chunks)

It actually overshoots 32 bits available in an integer and also breaks the format so loading earlier settings would not work. We also need to think of the impact of these. Sessions are a very important feature.

> this has no measurable impact on memory consumption

The resident memory consumption always shows less in the later instances when you have multiple instances running. I pointed out above which consumption I am talking about. 8K+ per context is high.

jarun:
@0xACE 

> I'm not sure what hardware jarun is running this on, but it sounds like he runs it on the classic game boy at times

I actually run it on a old Pi and everywhere optimization matters, specially when you load a directory of 1000 files. In fact I am least worried about i7s and i9s with 4 GB RAM in them.

leovilok:
> The compiler optimization you are talking of when dividing by 10 is actually a multiplication by 1/10 but that also requires floating point. That's one of the reasons we moved to integer-only calculation

Try pasting the following code in a [compiler explorer](https://godbolt.org/), it doesn't use FP math from `-O1` optimisation flag onward (and a single integer mul) for every architecture/compiler out there:
~~~C
int mod10(int num) {
    return num % 10;
}
~~~

And if you still hate the `%`, I think that in the actual code, it can be easily replaced with something like `n >= 10 ? 0 : n`.

jarun:
C code:

```C
int func(int i)
{
	return i % 10;
}
```

I ran on my system with gcc 7.5.0: `gcc -O3 -S div.c`

See the `imull` in the generated assembly:

```asm
	.file	"div.c"
	.text
	.p2align 4,,15
	.globl	func
	.type	func, @function
func:
.LFB0:
	.cfi_startproc
	movl	%edi, %eax
	movl	$1717986919, %edx
	imull	%edx
	movl	%edx, %eax
	movl	%edi, %edx
	sarl	$2, %eax
	sarl	$31, %edx
	subl	%edx, %eax
	leal	(%rax,%rax,4), %eax
	addl	%eax, %eax
	subl	%eax, %edi
	movl	%edi, %eax
	ret
	.cfi_endproc
.LFE0:
	.size	func, .-func
	.ident	"GCC: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0"
	.section	.note.GNU-stack,"",@progbits
```

From your webpage gcc 10.1:

```asm
func(int):
        movsx   rax, edi
        mov     edx, edi
        imul    rax, rax, 1717986919
        sar     edx, 31
        sar     rax, 34
        sub     eax, edx
        lea     edx, [rax+rax*4]
        mov     eax, edi
        add     edx, edx
        sub     eax, edx
        ret
```

The bitwise math we do, help to rollover. So when with 4 contexts, we get the last 2 bits.

With 10, 14%10 is 4 and there is not bitwise way to get it and we need to invoke multiplication as shown above.

jarun:
Anyway, as I mentioned, I wanted a PR as I wanted to understand what all changes. There will be more changes than we have here. My plan is to revisit this after the next release.

There are a few things I have in mind:

- make it configurable
- make the code generic with number of context as 2^n (e.g. setup masks easily)
- find a better way to handle so many switch cases
- handle top address bar issues (calculate number of columns to show generically, handle too few columns and so on)

So have patience.

jarun:
Also, @0xACE what are the specific use cases where you need more than 4 contexts? I need to understand that. I have never used more than 3 till date in an workflow.

0xACE:
> Also, @0xACE what are the specific use cases where you need more than 4 contexts? I need to understand that. I have never used more than 3 till date in an workflow.

Well one common scenario is when working with some projects:

**Context 1:** Base project directory, for general browsing
**Context 2:** Base note taking project directory
**Context 3:** Article 1 directory
**Context 4:** Article 2 directory
**Context 5:** Project code directory
**Context 6:** Current problem directory
**Context 7:** Books directory (Sacrificed)

Another one is when I'm working with publishing data on my website:

**Context 1:**  Reserved for the webpage directory I'm working on
**Context 2:** Hovering on the rebuild website script (I still have `SEL_EXEC`) (I usually sacrifice this)
**Context 3:** Directory containing data
**Context 4:** Directory containing dependencies (I usually sacrifice this)
**Context 5:** Directory containing second set of data
**Context 6:** Directory containing second set of dependencies (I usually sacrifice this)
**Context 7:** I need to grab a file from a remote server because I'm missing it, can't leave the others because they are essential
**Context 8:** I need to look at a remote files which previously dealt with problems to see the previous solutions
**Context 9**: Basis of operation of the data directory, in case I'm browsing for 1 file, essentially a browsing window, as everything else needs to remain as is.
**Context 10:** Waiting for downloading file to finish, important to remain open, or I'll forget

Don't try to optimize that workflow, those actions happens in the spur of the moment, I'm using the tools I have at my disposal, asking me to bind more bookmarks and plugins is not the solution. Infact it used to be a lot worse, but I made some tools to automate the difficult parts of the dependencies, but in the first workflow I just need that many directories.

**The problem is:** I cannot handle 1 project with 1 instance of `nnn`, and it is a frequent issue for me.

> * make it configurable
> * make the code generic with number of context as 2^n (e.g. setup masks easily)
> * find a better way to handle so many switch cases
> * handle top address bar issues (calculate number of columns to show generically, handle too few columns and so on)
> * how does the top bar look aesthetically

As I said, I'll be busy, and can't touch this again until middle of july, unless I find some free time (earliest next week). But all of these things can be addressed. Though there will be some weird ass `#define` magic going on...


> It actually overshoots 32 bits available in an integer

Is this important, does the cpu handle it as an int32? I'm mostly interested, I don't have time to look at it now... Anyhow why isn't such a unique limitation documented? 

Also, I've overshot it since late last year, and had 0 issues.

I'm running out of time here, I'll see you around, cheers.

jarun:
I got the use case. Thanks!

> Is this important, does the cpu handle it as an int32?

Nothing to do with the CPU. it's an int so we have 32 bits max.

> As I said, I'll be busy, and can't touch this again until middle of july, unless I find some free time (earliest next week).

I will plan these post the next release so we are not is a hurry.

Now that I have a sample, I have enough to think through the design. But it's not happening before next release. I have more on my plate than I can handle ATM.

0xACE:
> Nothing to do with the CPU. it's an int so we have 32 bits max.

Yeah, my question was more in terms of, is the structure even handled as a int32? or does it look at it as a collection of `uints`. I'd investigate this myself, but making a minimal example and investigating it requires time. Again I'm mostly asking because I find this interesting (I find this to be "funny" trivia for C). If you don't care about the topic, you can ignore it: I'll investigate when I find the time.

I wouldn't be surprised if the compiler optimizes away a lot of these value, specially if they never change... Yielding that these values to be hardcoded "ints" rather than a large `int32` with bitwise masking... I feel like this is premature optimization, it's okay to ignore this statement, I'm more interested in the functionality of the struct as is.

Could you recommend any tools for inspecting the code base in great detail for cases such as these? Not sure if I'd dumping the assembly is very read able for a quick peek...

> Now that I have a sample, I have enough to think through the design. But it's not happening before next release. I have more on my plate than I can handle ATM.

No stress, buddy, I'll help when I can.

Also it's not a huge concern, I'm running a different version of `nnn` locally anyhow:

```
 src/ace.h (new)                               | 144 ++++++++++++++++
 src/nnn.c                                     | 697 +++++++++++++++++++++++++++++++++++++++++++++++-------------------------------
 src/nnn.h                                     |  40 ++++-
```

jarun:
> Yeah, my question was more in terms of, is the structure even handled as a int32?

OK, that way. The intention is to limit the cfg in 32 bits. Also, changing a field width affects sessions i.e., old session data would not match with new code.

> I wouldn't be surprised if the compiler optimizes away a lot of these value, specially if they never change... Yielding that these values to be hardcoded "ints" rather than a large int32 with bitwise masking... I feel like this is premature optimization

These are bitfields and so many important information is stored in just 4 bytes.

> Could you recommend any tools for inspecting the code base in great detail for cases such as these? Not sure if I'd dumping the assembly is very read able for a quick peek...

I use vim. You can add nnn.vim, ctags+taglist, minibufexplorer, highlight.vim, bookmark.vim.

> Also it's not a huge concern, I'm running a different version of nnn locally anyhow:

Cool!

jarun:
@0xACE and @leovilok see commit 2f56ef5808b61948ac552e10e4ed5485153988f9.

I think we are generic enough now to have a make option to enable 8 contexts.

leovilok:
@jarun we still lack a few
~~~c
case SEL_CTX5: 
...
~~~
and a few
~~~c
{ '5',            SEL_CTX5 },
...
~~~

jarun:
> we still lack a few

The commit is not intended to implement 8 context support. It's for making the code generic. The SEL_CTXn will go inside a macro.

jarun:
8 contexts implemented at e3d90a98ceb6cbf2bd7ab3b242692b0079a7e15a. I will update the Wiki.

PR #572, opened 2020-05-13
fix #571 reading from nullptr

### LINKED ISSUE #571: Segfault opening a new context on an empty dir

[ ] Operating System: Linux
[ ] Desktop Environment: None
[ ] Terminal Emulator: 
[ ] Shell:
[x] Program options used: None
[x] Configuration options set: Nothing
[x] Issue exists on `nnn` master

Open `nnn` master on an empty dir

~~~sh
mkdir /tmp/empty
nnn /tmp/empty
~~~

Then press `2` to open context 2.
-> Segfault.

0xACE:
Needs to decide what happens if `dst` is a nullptr. Or let SEGFAULT be
the error reporter ;)

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -919,11 +919,20 @@ static void *xrealloc(void *pcur, size_t len)
  */
 static size_t xstrsncpy(char *restrict dst, const char *restrict src, size_t n)
 {
-	char *end = memccpy(dst, src, '\0', n);
 
-	if (!end) {
-		dst[n - 1] = '\0'; // NOLINT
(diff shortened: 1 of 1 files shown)


jarun:
In which scenario do you see this? I thought we had guards in the callers. Not having these checks in the low-level API helps when we are working with paths and file names which are bound to be non-NULL.

jarun:
OK, I see the issue now.

jarun:
Fixed at commit https://github.com/jarun/nnn/commit/124666cb3854730657d490323e01eb1710a9b4f0 with an `ndents` based early check.

PR #574, opened 2020-05-15
preview-tui: Add kitty terminal support

leovilok:
***This is work in progress, it breaks nnn UI for now!***

So I tried implementing kitty remote control into preview-tui, and it almost works ... but there are bugs, that migh be on kitty's side.
1. It breaks `nnn`'s input (you have to press keys multiple time for one to be registered)
2. It shows a wierd backtrace when quitting `nnn` (maybe `kitty @ launch ...` is still running in the background edit: exactly, it can be removed by redirecting stderr to /dev/null)

I fear that kitty is trying to be too smart here, and does wierd thing to the terminal state. Or using remote control in script is just not enough tested. Or I'm missing something.

Anyways, I don't use kitty, so I'm out of this. I'll just leave it there if someone want to try to make it work better.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -4,7 +4,10 @@
 #
 # Note: This plugin needs a "NNN_FIFO" to work.
 #
-# Dependencies: tmux (>=3.0) or xterm (or set $TERMINAL), file, tree
+# Dependencies:
+#   - tmux (>=3.0) or kitty (allow_remote_control on) or xterm (or set $TERMINAL)
+#   - file
+#   - tree
 #
 # Usage:
 #   You need to set a NNN_FIFO path and set a key for the plugin,
@@ -60,6 +63,11 @@ fi
 
 if [ -e "${TMUX%%,*}" ] && [ "$(tmux -V | cut -c6)" -eq 3 ] ; then
     tmux split-window -e "NNN_FIFO=$NNN_FIFO" -e "PREVIEW_MODE=1" -dh "$0" "$1"
+elif [ "$TERM" = "xterm-kitty" ] && kitty @ ls > /dev/null 2>&1 ; then
+    kitty @ launch --no-response\
+          --title "nnn preview" --keep-focus --cwd="$PWD" \
+          --env "NNN_FIFO=$NNN_FIFO" --env "PREVIEW_MODE=1" \
+          "$0" "$1" > /dev/null
 else
     PREVIEW_MODE=1 $TERMINAL -e "$0" "$1" &
 fi


jarun:
The backtrace mentioned in 2:

```
Traceback (most recent call last):
  File "python3.7/runpy.py", line 280, in run_path
  File "python3.7/runpy.py", line 85, in _run_code
  File "kitty/__main__.py", line 121, in <module>
  File "kitty/__main__.py", line 117, in main
O  File "kitty/__main__.py", line 23, in remote_control
  File "kitty/remote_control.py", line 187, in main
  File "kitty/remote_control.py", line 136, in do_io
  File "kitty/remote_control.py", line 115, in simple_recv
OSError: [Errno 5] Input/output error
```

jarun:
I have raised https://github.com/kovidgoyal/kitty/issues/2658 so the kitty devs can take a look.

jarun:
@leovilok I tested with `--no-response` and it works as expected.

Can we improve the kitty and tmux-specific cases some more:

1. Better splitting logic:
- split vertically if w >= h (vertical split is tmux default)
- else split horizontally (horizontal split is kitty default)
2. For image files, can we display them in kitty?

jarun:
Also, there a note from the kitty author on the `--listen-on` option in the kitty thread, please check it out.

jarun:
With the following patch I see only only the first image (if I start preview when already on an image) but no other preview works after that:

```diff
diff --git a/plugins/preview-tui b/plugins/preview-tui
index 3ebf195..16af95f 100755
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -35,6 +35,8 @@ preview_file () {
     elif [ "${mime%%/*}" = "text" ] ; then
         # Print file head
         head -n $lines "$1" | cut -c 1-"$cols"
+    elif [ "${mime%%/*}" = "image" ] ; then
+        kitty +kitten icat "$1"
     else
         # Binary file
         echo "-------- Binary file --------"
@@ -60,6 +62,12 @@ fi
 
 if [ -e "${TMUX%%,*}" ] && [ "$(tmux -V | cut -c6)" -eq 3 ] ; then
     tmux split-window -e "NNN_FIFO=$NNN_FIFO" -e "PREVIEW_MODE=1" -dh "$0" "$1"
+elif [ "$TERM" = "xterm-kitty" ] \
+     && kitty @ launch --no-response \
+        --title "nnn preview" --keep-focus --cwd="$PWD" \
+        --env "NNN_FIFO=$NNN_FIFO" --env "PREVIEW_MODE=1" \
+        "$0" "$1" > /dev/null ; then
+    :
 else
     PREVIEW_MODE=1 $TERMINAL -e "$0" "$1" &
 fi
```

jarun:
@leovilok finally the following worked for me:

```diff
diff --git a/plugins/preview-tui b/plugins/preview-tui
index 3ebf195..9c3b1cd 100755
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -35,6 +35,8 @@ preview_file () {
     elif [ "${mime%%/*}" = "text" ] ; then
         # Print file head
         head -n $lines "$1" | cut -c 1-"$cols"
+    elif [ "${mime%%/*}" = "image" ] ; then
+        kitty +kitten icat --silent --transfer-mode=stream --stdin=no "$1"
     else
         # Binary file
         echo "-------- Binary file --------"
@@ -60,6 +62,12 @@ fi
 
 if [ -e "${TMUX%%,*}" ] && [ "$(tmux -V | cut -c6)" -eq 3 ] ; then
     tmux split-window -e "NNN_FIFO=$NNN_FIFO" -e "PREVIEW_MODE=1" -dh "$0" "$1"
+elif [ "$TERM" = "xterm-kitty" ] \
+     && kitty @ launch --no-response \
+        --title "nnn preview" --keep-focus --cwd="$PWD" \
+        --env "NNN_FIFO=$NNN_FIFO" --env "PREVIEW_MODE=1" \
+        "$0" "$1" > /dev/null ; then
+    :
 else
     PREVIEW_MODE=1 $TERMINAL -e "$0" "$1" &
 fi
```

I think the images get cropped to the preview pane size.

leovilok:
@jarun the issue with --no-response is that it doesn't fail if remote control is not activated.
I my patch, it worked well on kitty with default config (no remote control) because it would fallback to creating a xterm (or $TERMINAL) window.

If we use --no-response, we should find a way of knowing if kitty's remote control is activated.

jarun:
> the issue with --no-response is that it doesn't fail if remote control is not activated

I think the config is a 1-time effort. If things don't work at all without it, we have no other go than to use it. Also, you have documented it clearly so I think we are covered.

leovilok:
With kitty's image preview, I wonder if we could split it to another plugin, like `preview-kitty`.

jarun:
> With kitty's image preview, I wonder if we could split it to another plugin, like preview-kitty.

Exactly my thoughts. Let's have it that way.

leovilok:
I updated the PR with a working kitty version, that check if remote access is available (with the method proposed by the author), but no img support for now.

jarun:
Let's have an independent plugin with the image preview included.

jarun:
You mentioned using rust utilities for a plugin, probably preview-kitty is a good candidate.

jarun:
Closing this as we are having a separate plugin for kitty.

PR #577, opened 2020-05-17
Add a preview-kitty plugin

leovilok:
I didn't exactly now how to format the README, @jarun please feel free to adapt it.

As for the requested vertical/horizontal splits based on window ratio, I think that in kitty it depends on what are the user allowed layout for the window.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -61,6 +61,7 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | pdfread | Read a PDF or text file aloud | sh | pdftotext, mpv,<br>pico2wave |
 | pdfview | View PDF file in `$PAGER` | sh | pdftotext/<br>mupdf-tools |
 | picker | Pick files and list one per line (to pipe) | sh | nnn |
+| preview-kitty | File previewer in a kitty terminal pane | sh | [kitty](https://sw.kovidgoyal.net/kitty/), Optionally:<br>[exa](https://github.com/ogham/exa) [bat](https://github.com/sharkdp/bat) mediainfo |
 | preview-tabbed | `tabbed`/xembed based file previewer | bash | _see in-file docs_ |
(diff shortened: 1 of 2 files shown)


plugins/README.md
```diff
@@ -76,6 +77,8 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | wall | Set wallpaper or change colorscheme | sh | nitrogen/pywal |
 | x2sel | Copy `\n`-separated file list from system clipboard to sel | sh | _see in-file docs_ |
 
+¹: with kitty's `allow_remote_control` option on.
```
jarun, line 80: We can drop it from here. Keep this instruction in-file.

plugins/preview-kitty
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env sh
+
+# Description: File previewer for kitty term
```
jarun, line 3: File previewer for kitty term

->

File previewer for kitty term using NNN_FIFO

And drop the Note.

plugins/preview-kitty
```diff
@@ -0,0 +1,72 @@
...
+
+preview_file () {
+    clear
+    lines=$(($(tput lines)-1))
+    cols=$(tput cols)
+    mime="$(file -b --mime-type "$1")"
+
+    if [ -d "$1" ]; then
+        # Print directory tree
+        cd "$1" && \
```
jarun, line 39: shellcheck (CI) fails because of this:

```
In plugins/preview-kitty line 39:
 
        cd "$1" && \
 
                ^-- SC2015: Note that A && B || C is not if-then-else. C may run when A is true.
```

leovilok: Well this is the intended behaviour, adding an ignore.

plugins/preview-kitty
leovilok: Done.

plugins/README.md
leovilok: Done.

jarun:
Thank you!

Issue #582, opened 2020-05-20
kitty preview plugin needs to be stabilized


midexe6380:
kitty preview doesn't always work

for example
get last commit https://github.com/jarun/nnn/commit/0c916ae7b024c961df1ab1797035ee80501f66e3
`git clone https://github.com/jarun/nnn.git`
`make`
`make install`
get test repository with last commit https://github.com/wireapp/wire-desktop/commit/13e69df98f2c8c79ad61cd8326f4a1afc2effd04
`git clone https://github.com/wireapp/wire-desktop.git`
`cd wire-desktop`
`cd resources`
`cd macos`
`cd entitlements`

didn't display folder `entitlements` and file `child.plist`
although bat and exa separately perfectly display the contents of the file and folder
for clarity, I attach a video
https://youtu.be/2f5m68YhTg8

here I show that bat can normally view the file, but for some reason does not do this
https://youtu.be/2f5m68YhTg8?t=17

for comparison:
ranger displays this
https://youtu.be/2f5m68YhTg8?t=26

jarun:
@leovilok please take a look.

leovilok:
hum, I get the same issue with `preview-tui`, I suspect that our "smart" pointer comparison to check if we actually moved the cursor has a few false positive on string reuse. I'll look into this tomorrow. Thanks @midexe6380 for reporting!

jarun:
From the video it seems the problem is the timestamp of the directory and file matches. An easy fix could be checking if the mode matches. But if there's a directory within a directory with the same timestamp we will still see this issue.

But I would say that's still better than the timestamp. I will think some more on this.

jarun:
`struct entry` is 40 bytes (much smaller than the 4K max path length). I would suggest we store the whole thing and do a `memcmp` which is pretty fast.

jarun:
Note that there's still the worst case where everything matches but I believe that's a long shot. Someone has to prep it with prior knowledge.

jarun:
@midexe6380 please confirm if the latest patch works as you expect.

midexe6380:
@jarun  now is working
but there were other problems

example
get test repository
`git clone https://github.com/timvisee/ffsend.git`
cd `ffsend/pkg/scoop/`
`ffsend.json` file shows as binary without preview

midexe6380:
also if I run two or more nnn instances kitty preview doesn't work at all

midexe6380:
kitty preview plugin is very unstable 🙃

[...]

PR #587, opened 2020-05-21
Change binary file detection in preview-tui/kitty

leovilok:
Fixes `.json` (and probably other) text files with no `text/*` mime type, as reported in #582 (discussion) by @midexe6380.


jarun:
Thank you!

PR #588, opened 2020-05-22
Add -a option to generate a temporary NNN_FIFO

leovilok:
This feature was discussed both in #582 ("on the fly FIFO") and in dev discussion. The suggested flag, `-f`, is already in use for readline history, so I've put it to `-a` for "autofifo" ...

Currently `-a` overrides `NNN_FIFO`, I don't know which one should have precedence.

Labeled as WIP as there is no doc yet in this PR.


leovilok:
I added some docs.

src/nnn.c
```diff
@@ -7292,6 +7309,8 @@ int main(int argc, char *argv[])
 	notify_fifo();
 	if (fifofd != -1)
 		close(fifofd);
+	if (autofifo)
+		unlink(fifopath);
```
jarun, line 7313: If there's a failure and we exit before this, we will not unlink it.

Can we have a g_state and unlink in `cleanup()`?

src/nnn.c
```diff
@@ -6726,6 +6726,9 @@ static void usage(void)
 		"positional args:\n"
 		"  PATH   start dir [default: .]\n\n"
 		"optional args:\n"
+#ifndef NOFIFO
+		" -a      auto-setup temporary NNN_FIFO\n"
```
jarun, line 6730: Just

    auto NNN_FIFO

plugins/README.md
```diff
@@ -198,9 +198,9 @@ Usage examples can be found in the Examples section below.
 
 If `NNN_FIFO` is set, `nnn` will open it and write every hovered files. This can be used in plugins and external scripts, e.g. to implement file previews.
 
-If a `NNN_FIFO` is set globally, each `nnn` instance will write to it, and a process reading from the pipe will get hovered path from every instance, interleaved.
+The easiest way to set `NNN_FIFO` is to start `nnn` with the `-a` option, to automatically setup a temporary FIFO file for this `nnn` instance.
 
-If you want to prevent this and be sure to have a private pipe to one `nnn` instance, you can unlink (remove) the FIFO file. If you had opened the FIFO before and you have read from it (so that `nnn` have it opened too), you can still read from it while you don't close it. But new `nnn` instances will recreate a new FIFO not linked to the previous one.
+If a `NNN_FIFO` environment variable is set globally (and `-a` is not passed to `nnn`), each `nnn` instance will write to the same FIFO, and a process reading from the pipe will get hovered path from every instance, interleaved.
```
jarun, line 203: Ensure the wiki doc is in sync.

src/nnn.c
```diff
@@ -7154,6 +7165,12 @@ int main(int argc, char *argv[])
 
 #ifndef NOFIFO
 	/* Create fifo */
+	if (autofifo) {
+		g_tmpfpath[tmpfplen - 1] = '\0';
+		snprintf(g_buf, CMD_LEN_MAX, "%s/nnn-fifo.%d", g_tmpfpath, getpid());
+		setenv("NNN_FIFO", g_buf, TRUE);
+	}
+
 	fifopath = xgetenv("NNN_FIFO", NULL);
```
jarun, line 7174: This should be in else

If -a is passed, that gets the priority. Document that as well in our wiki configuration section.

leovilok: I still need to get fifopath from the environment, to avoid string allocation (sincewe need to put the FIFO path in the environment anyways, we don't want to have it duplicated).

src/nnn.c
leovilok: done.

src/nnn.c
jarun: No, please use g_states in line 404, nnn.c. Do not add globals if it's not absolutely required. In this case a but suffices.

nnn.1
```diff
@@ -54,6 +54,9 @@ to see the list of keybinds.
 .Nm
 supports the following options:
 .Pp
+.Fl a
```
jarun, line 57: Please add `a` in line 9 as well.

jarun:
Other than using g_states, please add the new option `a` to the completion scripts.

jarun:
@leovilok I have taken care of the above. Please ensure the wiki is in sync with the plugin docs.

src/nnn.c
leovilok: Oh I misunderstood that. Done now.

nnn.1
leovilok: Done.

leovilok:
Oh I didn't see you did it, thank you.

jarun:
No problem! ;)

leovilok:
I see that you've already updated the Live Preview wiki page, I updated [Usage -> Program options](https://github.com/jarun/nnn/wiki/Usage#program-options), I thinks that we're good.

jarun:
Yes, this looks good! All docs up to date.

0xACE:
I'm yet again way behind master and not familiar with all changes.

But does `-a` remove the `NNN_FIFO` file it has created upon exit? I couldn't find a reference to removal in this PR...

jarun:
Yes, it does. See `cleanup()`.

leovilok:
@0xACE there is a note on this in the man:
> 2. If the FIFO file doesn't exist it will be created, but not removed (unless it is generated by -a option).

PR #591, opened 2020-05-22
preview-kitty polish (async version + text render for SVGs)

leovilok:
The first commit kills file viewers when a new file comes down the FIFO (we use fancier viewers in this plugin after all).

The second commit preview SVG files in text mode. This is a matter of taste at this point, as slow rendering SVGs won't block anything, but it's definitively faster. What do you think of this?

Fixes issues discussed in #582


jarun:
Thank you!

PR #597, opened 2020-05-27
preview-tui: async version using $PAGER to scroll preview

leovilok:
Edit: actually it seems that I just don't reliably kill $PAGER

~It's been a hell to make this work (as always when job control is involved) because I think that:~
~1. `kill %%` kills only the last job is a pipeline and~
~2. the shell reuse previously used pipes, even pipes created in subshells (so previously supposed dead commands were writing to the new $PAGER)~

~I fixed it by using my own pipe (named FIFO), I could probably also alternatively make another call to $0...~

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -4,7 +4,7 @@
 #
 # Note: This plugin needs a "NNN_FIFO" to work.
 #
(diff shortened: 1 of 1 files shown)


jarun:
Works like a charm! Deal closed!

PR #598, opened 2020-05-27
Add LUP Mention

gpen365:
Added the Linux Unplugged blog post that includes NNN.


jarun:
Sorry, we link only nnn-exclusive articles. As I can see your article covers a much broader area. Would you be interested in writing a review of `nnn`? An area that would be of great help would be exploring the plugins.

gpen365:
Okay, 👍

PR #599, opened 2020-05-27
preview-tui: directly call $PAGER on text files

leovilok:
This allow colored preview for people using less configured with a lesspipe filter, and use a FIFO only when it's strictly needed (for tree view, which must use a pipe).

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -25,35 +25,36 @@ TERMINAL="${TERMINAL:-xterm}"
 PAGER="${PAGER:-less}"
 
 preview_file () {
-    kill %- %+ 2>/dev/null
+    kill "$(jobs -p)" 2>/dev/null
     clear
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

toddyamakawa:
@leovilok, would it have slightly better performance if instead of doing this:

```
        (
            exec > "$tmpfifopath"
            tree &
        )
```

You did this?
```
        (
            exec > "$tmpfifopath"
            tree
        ) &
```

leovilok:
@toddyamakawa I want to be sure that `exec > "$tmpfifopath"` returned (so that the FIFO is opened) before launching `rm  "$tmpfifopath"`. If this synchronisation was not needed I could replace those 4 lines with a single:
~~~sh
tree > "$tmpfifopath" &
~~~

PR #606, opened 2020-05-29
add 'wraps' attribute to allow autocompletion

basbebe:


basbebe:
> What does it do?

It tells fish shell that `n` is a wrapper for `nnn`.
That way tab-completion works also for `n`.
Bildschirmfoto 2020-05-29 um 16 56 30

basbebe:
see https://fishshell.com/docs/current/cmds/function.html#cmd-function
and https://fishshell.com/docs/current/cmds/complete.html#cmd-complete

jarun:
Cool!

PR #607, opened 2020-05-29
Cleanly handle SIGTERM and SIGHUP

leovilok:
Closing the host terminal should trigger SIGHUP, but a clean exit it expected on SIGTERM too, both are handled in this patch.

I see that SIGINT is handled whith `sigaction()`, and SIGQUIT with `signal()`. I chose to use `signal()`, it's an older API, but it's simpler (and avoids an extra struct on the stack), and we don't need the `sigaction` extra features. Should we use `signal()` for everything for consitency? or switch everything to the newer `sigaction`, which is recomended the standards?

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -372,14 +372,8 @@ static regex_t archive_re;
 #endif
 
 /* Retain old signal handlers */
-#ifdef __linux__
-static sighandler_t oldsighup; /* old value of hangup signal */
-static sighandler_t oldsigtstp; /* old value of SIGTSTP */
-#else
-/* note: no sig_t on Solaris-derivs */
-static void (*oldsighup)(int);
-static void (*oldsigtstp)(int);
-#endif
+static struct sigaction oldsighup;
+static struct sigaction oldsigtstp;
 
 /* For use in functions which are isolated and don't return the buffer */
 static char g_buf[CMD_LEN_MAX] __attribute__ ((aligned));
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -7230,6 +7236,8 @@ int main(int argc, char *argv[])
 		xerror();
 		return EXIT_FAILURE;
 	}
+	signal(SIGTERM, clean_exit_sighandler);
```
jarun, line 7239: I think we should move to sigaction() everywhere, now that we are doing this.

leovilok: I added a commit to replace `signal()` with `sigaction()` everywhere in `nnn.c`.
We might want to keep it a different commit in history, as it is not entirely related...

leovilok:
We might want to do some extra cleanup on SIGTERM, as for now on `pkill nnn` it leaves the terminal in a messy state...

leovilok:
I added a `exitcurses()` call to the SIGTERM/SIGHUP handler to cleanup terminal state on e.g. `pkill nnn`.

src/nnn.c
```diff
@@ -372,14 +372,8 @@ static regex_t archive_re;
 #endif
 
 /* Retain old signal handlers */
-#ifdef __linux__
-static sighandler_t oldsighup; /* old value of hangup signal */
-static sighandler_t oldsigtstp; /* old value of SIGTSTP */
-#else
-/* note: no sig_t on Solaris-derivs */
-static void (*oldsighup)(int);
```
jarun, line 380: Would Solaris still work?

@sjmulder ?

leovilok: There is an instance of `struct sigaction` in `main()` already, so if it works, this commit works too.

leovilok: Also, this [man page from 2005](https://www.unix.com/man-page/opensolaris/2/sigaction/) documents sigaction in Solaris.

The issue was the use of `sighandler_t` which is not needed in this PR.

jarun: OK. Thanks!

src/nnn.c
jarun: OK. I will merge this. thanks for the note.

jarun:
Thank you! This is a very nice improvement.

src/nnn.c
sjmulder: The SmartOS build has actually been [failing](https://us-east.manta.joyent.com/pkgsrc/public/reports/upstream-trunk/20200528.2251/nnn-3.1/build.log) for a while:

```
src/nnn.c: In function 'show_stats':
src/nnn.c:3642:43: error: 'FILE_MIME_OPTS' undeclared (first use in this function)
 3642 |    get_output(g_buf, CMD_LEN_MAX, "file", FILE_MIME_OPTS, fpath, FALSE);
      | 
```

so I wouldn't worry. If it doesn't work I'll do a PR to fix it again.

PR #610, opened 2020-05-29
Fix detecting end of evironment options

leovilok:
This allow using both NNN_OPTS and options with arguments from argv.


jarun:
Thank you!

PR #613, opened 2020-05-30
Allow setting arbitrary (1..8) CTX_MAX

leovilok:
This builds up on @jarun's recent commits and @0xACE's #570, plus a bit of macro magic (to generate SEL enum/key list/cases).

The way it handles non power of two CTX_MAX was found elsewhere in nnn code, there is no more `%` only int cmp/inc/dec/affectation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,8 @@ ifeq ($(O_NOFIFO),1)
 	CPPFLAGS += -DNOFIFO
 endif
 
-ifeq ($(O_CTX8),1)
-	CPPFLAGS += -DCTX8
+ifdef O_CTXMAX
+	CPPFLAGS += -DCTX_MAX=$(O_CTXMAX)
 endif
 
 ifeq ($(shell $(PKG_CONFIG) ncursesw && echo 1),1)
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -157,14 +157,8 @@
 #define EXEC_ARGS_MAX 8
 #define LIST_FILES_MAX (1 << 16)
 #define SCROLLOFF 3
-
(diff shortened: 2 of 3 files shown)


jarun:
Please let's not get unrealistic with this.

jarun:
If you need more than 8 contexts, open 2 instances in tmux. 16 contexts won't even show the contexts properly in a smaller terminals.

leovilok:
This doesn't support more than 8 ctx (and I don't personnaly use more than 2 or 3 by instance).

jarun:
Got it. Anyway the #ifdef is clearer than the `REPEAT(CTX_MAX, CTX_KEYDEF)`. That's where I thought we are adding more.

jarun:
I noticed one issue if you want to investigate. When compiling with 8 contexts and opening `nnn` it takes some time to load _only_ the first time.

I think it may have something to do with the option `-S` I use.

My `n` is `nnn -cEFnQrSx -t 900 $@`.

leovilok:
> I noticed one issue if you want to investigate. When compiling with 8 contexts and opening `nnn` it takes some time to load _only_ the first time.
> 
> I think it may have something to do with the option `-S` I use.
> 
> My `n` is `nnn -cEFnQrSx -t 900 $@`.

I get the same for 4 contexts: first invocation with `-S` with empty `~/.config/nnn/sessions` in slow, but then it's instant again.

It doesn't seems to be related to ctx number, may some fs cache effect?

jarun:
> It doesn't seems to be related to ctx number, may some fs cache effect?

That's what I wanted to confirm. I think we are good if it's not related to session.

PR #616, opened 2020-05-30
Enhanced preview and splitting

Recidiviste:
Hello there !

Add more options to preview, like syntax highlighting, archives content preview, pdf preview....
All of that only if the packages necessary are present (so same dependencies).
Plus, printing more informations using `stat` about the file if the type is unknown.
Finally, ask for the type of splitting when in Tmux

PS : I'm not a big shell writer, so if it's better to have case instead of if-elif-then block, I can change that. You have the editings right on this after all, so change it the way you want. I love your project.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -18,12 +18,14 @@
 #   single common preview window. I you provide different FIFO path, they
 #   will be independent.
 #
+#   
+#
 # Shell: POSIX compliant
 # Authors: Todd Yamakawa, Léo Villeveygoux
 
 TERMINAL="${TERMINAL:-xterm}"
 PAGER="${PAGER:-less -r}"
-SPLIT="h"
+SPLIT=
 
 lines=$(($(tput lines)-1))
 cols=$(tput cols)
@@ -34,6 +36,7 @@ preview_file () {
 
     encoding="$(file -b --mime-encoding "$1")"
     typ="$(file -b "$1" | sed 's/ .*//')"
+    LESSOPEN="${LESSOPEN:'|highlight --out-format=ansi --force %s'}"
 
     if [ -d "$1" ]; then
         # Print directory tree
@@ -52,7 +55,7 @@ preview_file () {
         )
 
         rm "$tmpfifopath"
-     elif [ "$typ" = "Zip" ] && [ "$(command -v zip)" ]; then
+    elif [ "$typ" = "Zip" ] && [ "$(command -v zip)" ]; then
      	# Display content of Zip
      	zip -sf "$1" | $PAGER &
      elif [ "$typ" = "gzip" ] && [ "$(command -v tar)" ]; then
@@ -68,7 +71,7 @@ preview_file () {
         # Binary file: just print filetype info
         echo '-------- Binary file --------'
         file -b "$1"
-	echo '------- Informations -------'
+	echo '------- Details -------'
 	stat "$1"
     else
         # Text file:
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -48,18 +48,18 @@ preview_file () {
         )
 
         rm "$tmpfifopath"
-     elif [ "$typ" = "Zip" ] && [ $(command -v zip) ]; then
+     elif [ "$typ" = "Zip" ] && [ "$(command -v zip)" ]; then
      	# Display content of Zip
      	zip -sf "$1" | $PAGER &
-     elif [ "$typ" = "gzip" ] && [ $(command -v tar) ]; then
+     elif [ "$typ" = "gzip" ] && [ "$(command -v tar)" ]; then
      	# Display content of archive
      	tar -tvf "$1" | $PAGER &
-     elif [ "$typ" = "PDF" ] && [ $(command -v pdftotext) ]; then
+     elif [ "$typ" = "PDF" ] && [ "$(command -v pdftotext)" ]; then
      	# Display content of PDF
      	pdftotext "$1" - | $PAGER &
-     elif ( [ "$typ" = "PNG" ] || [ "$typ" = "JPEG" ] ) && [ $(command -v viu) ]; then
+     elif ( [ "$typ" = "PNG" ] || [ "$typ" = "JPEG" ] ) && [ "$(command -v viu)" ]; then
      	# Display image in terminal (more or less lowered resolution)
-     	viu "$1" | head -n $lines
+     	viu "$1" | head -n "$lines"
     elif [ "$encoding" = "binary" ] ; then
         # Binary file: just print filetype info
         echo "-------- Binary file --------"
@@ -68,7 +68,7 @@ preview_file () {
 	stat "$1"
     else
         # Text file:
-	if [ $(command -v highlight) ]; then
+	if [ "$(command -v highlight)" ]; then
 		highlight --out-format=ansi --force "$1" | $PAGER &
 	else
 		$PAGER "$1" &
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -18,7 +18,8 @@
 #   single common preview window. I you provide different FIFO path, they
 #   will be independent.
 #
-#   
+#   Configure SPLIT to either "h" or "v" to set a 'h'orizontal split or a
+#   'v'ertical split
 #
 # Shell: POSIX compliant
 # Authors: Todd Yamakawa, Léo Villeveygoux
@@ -35,7 +36,7 @@ preview_file () {
     clear
 
     encoding="$(file -b --mime-encoding "$1")"
-    typ="$(file -b "$1" | sed 's/ .*//')"
+    typ="$(file -b --mime-type)"
     LESSOPEN="${LESSOPEN:'|highlight --out-format=ansi --force %s'}"
 
     if [ -d "$1" ]; then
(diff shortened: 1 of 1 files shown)


jarun:
Hi, thanks for the PR.

@leovilok can you please take a look? I have a small concern about extracting the archives and  listing them. Do you think it will slow the previewer down?

jarun:
Also, @Recidiviste CircleCI fails... probably due to `shellcheck`. Please take a look.

Recidiviste:
Done, I forgetted the `"` around some tests and also that `$lines` did disappeared between the version I was working on and the last version (sorry for being an amateur)

plugins/preview-tui
```diff
@@ -74,7 +94,16 @@ if [ "$PREVIEW_MODE" ] ; then
 fi
 
 if [ -e "${TMUX%%,*}" ] && [ "$(tmux -V | cut -c6)" -eq 3 ] ; then
-    tmux split-window -e "NNN_FIFO=$NNN_FIFO" -e "PREVIEW_MODE=1" -dh "$0" "$1"
+   printf "split 'h'orizontal/'v'ertical :"
```
jarun, line 97: @leovilok I think we discussed this at some point. We either have a variable or do this by determine the rows*cols. Prompting everything is not very user-friendly.

Recidiviste: Maybe if there's more cols than (lines * 2), split vertically (n³ | preview), else, split horizontally (n³ above ― preview below). The only problem is if cols and (lines * 2) are really close (for example 116 cols and 59 lines => easier with horizontal split, but vertical split using our condition).
Maybe add a variable ASK at the start, by default on false. If false, split using condition (or maybe set a default vertical split). If true, ask the type of split.
Writing that, I think it's getting more and more complex for a simple plugin. I suggest doing a first merge request for the different previews and a second for the type of split. Your choice.

jarun: I think we should do the following (tmux-only):

have a variable: SPLIT=h
this overrides all. if user sets to v than do vertical split

now, if the above is not set:
if lines > cols: do horizontal split
if cols > lines: do vertical split
if cols=lines: do vertical split

Recidiviste: Ok with that at the exception of the conditions lines > cols/lines < cols
On a terminal, there's way more cols than lines usually. Maybe add a multiplier (2 for example, even if it's not quite right) for the number of lines. Than we have

if (lines*2) > cols : do horizontal split
if (lines*2) < cols : do vertical split
if (lines*2) = cols : do vertical split

As said, it's your choice, it's not a big change in the code

jarun: OK. Let's try this.

Recidiviste:
Now it works. Don't look at the feed up there, I didn't struggled at all doing this simple thing...

leovilok:
Hi @Recidiviste and thanks for the PR, it's good to know that you could hack on `preview-tui`, plugins are made to be customisable after all.

I do have a few issues/comments on this PR:

1. You actually don't need the highlight call to have syntax highlighted text in less: you can configure less with the LESSOPEN env var to set an input filter (using the `highlight` command as LESSOPEN is actually a popular option)
2. (maybe the biggest issue) You cant really use lines like
    ~~~sh
    cmd | $PAGER &
    ~~~
   because the whole pipeline is one unique job in the shell, with `cmd`'s PID as the job's PID, so when you kill the job (because there is a new input) it actually only kills `cmd` and not `less`. With your version of the plugin, try highlighting several files with various filetypes, than on one file try scrolling in less: other less instance are still running so it will scroll a random less instance, showing any of the files you were previewing before. That's why I'm using a FIFO for `tree` view, we could work on a way to make it generic for any command if you want. (I hope that this explanation makes sense, j'aurais peut-être du te la faire en français :smile:)
3. I think that we should frefer `file -b --mime-type` over just `file -b`: the output should be more consistent while retaining the same info.
4. For archive preview, maybe `atool`'s `als` is a more universal tool? that's jut a matter of taste. @jarun with the async architecture it can't slow down anything, if it's not fast enough to list archive content before another file is highlighted... it just gets killed.
5. About viu/pdftotext: it's good to be able to have these options, but I'm feeling like we're reimplementing `nuke` plugin again... which is why I'm calling `nuke` directly in `preview-tabbed`. Maybe we could come up with a way to call `nuke` and still reliably kill what it spawns... there might be something to try there.
6. On tmux vertical/horizontal split: this code could use a bit of factorization, but I see that you autodetect width/height ratio which was a request from @jarun that I totally forgot about :smile:. I'm not sure about the SPLIT env var though, isn't there a Tmux configuration controlling the default split orientation?

I'm not sure that we'll get everything you added in `preview-tui` (it's supposed to be a barebone preview plugin after all), but there is definitively some good stuff there. And I hope that you had fun hacking it!

plugins/preview-tui
```diff
@@ -23,14 +23,17 @@
 
 TERMINAL="${TERMINAL:-xterm}"
 PAGER="${PAGER:-less -r}"
+SPLIT="h"
```
jarun, line 26: Please keep it empty in the beginning with comment.

Also, please document this behaviour in the Usage section.

plugins/preview-tui
```diff
@@ -47,13 +52,31 @@ preview_file () {
...
+     	pdftotext "$1" - | $PAGER &
+     elif ( [ "$typ" = "PNG" ] || [ "$typ" = "JPEG" ] ) && [ "$(command -v viu)" ]; then
+     	# Display image in terminal (more or less lowered resolution)
+     	viu "$1" | head -n "$lines"
     elif [ "$encoding" = "binary" ] ; then
         # Binary file: just print filetype info
-        echo "-------- Binary file --------"
+        echo '-------- Binary file --------'
         file -b "$1"
+	echo '------- Informations -------'
```
jarun, line 71: Informations -> Details

jarun:
> I'm not sure about the SPLIT env var though, isn't there a Tmux configuration controlling the default split orientation?

@Recidiviste in that case just drop split.

plugins/preview-tui
jarun: Please drop split. Probably add a note on the tmux variable.

leovilok: I was wrong, default layout can't be easily configured in tmux...

jarun: OK, then:



Please keep it empty in the beginning with comment.

Also, please document this behaviour in the Usage section.

plugins/preview-tui
Recidiviste: Done

jarun: I meant:

    	echo '------- Details -------'

plugins/preview-tui
```diff
@@ -47,13 +52,31 @@ preview_file () {
         )
 
         rm "$tmpfifopath"
+     elif [ "$typ" = "Zip" ] && [ "$(command -v zip)" ]; then
```
jarun, line 55: The alignment looks broken to me. Can you please check?

Recidiviste: Don't seems to me, 3 spaces like the rest of the file.

plugins/preview-tui
```diff
@@ -47,13 +52,31 @@ preview_file () {
...
+     	# Display content of Zip
```
jarun, line 56: This line is not aligned correctly. Basically all statements after each condition check.

jarun: There's a TAB.

plugins/preview-tui
jarun: It has 1 space extra wrt. line 38. Fix it please.

plugins/preview-tui
jarun: We do not use TABs in the scripts.

plugins/preview-tui
```diff
@@ -47,13 +56,31 @@ preview_file () {
...
-        echo "-------- Binary file --------"
+        echo '-------- Binary file --------'
         file -b "$1"
+	echo '------- Details -------'
+	stat "$1"
     else
         # Text file:
-        $PAGER "$1" &
+	if [ "$(command -v highlight)" ]; then
+		highlight --out-format=ansi --force "$1" | $PAGER &
```
jarun, line 80: And what's this alignment? I can see the differences in the GitHub diff. Please take a look yourself.

jarun:
Also, CI fails.

@KlzXS and @leovilok is there a way to enforce 4 space alignment using CI? I cannot spend time on this trivia.

plugins/preview-tui
```diff
@@ -47,13 +56,31 @@ preview_file () {
...
+    	pdftotext "$1" - | $PAGER &
+    elif ( [ "$typ" = "PNG" ] || [ "$typ" = "JPEG" ] ) && [ "$(command -v viu)" ]; then
+     	# Display image in terminal (more or less lowered resolution)
+     	viu "$1" | head -n "$lines"
     elif [ "$encoding" = "binary" ] ; then
         # Binary file: just print filetype info
-        echo "-------- Binary file --------"
+        echo '-------- Binary file --------'
         file -b "$1"
+	echo '------- Details -------'
```
jarun, line 75: You can see the difference between line 74 and 75.

plugins/preview-tui
```diff
@@ -47,13 +56,31 @@ preview_file () {
         )
 
         rm "$tmpfifopath"
+    elif [ "$typ" = "Zip" ] && [ "$(command -v zip)" ]; then
+    	# Display content of Zip
```
jarun, line 60: This line should be aligned to 4 spaces.

jarun: So 4 + 4 = 8 spaces from the left.

plugins/preview-tui
```diff
@@ -47,13 +56,31 @@ preview_file () {
...
         # Binary file: just print filetype info
-        echo "-------- Binary file --------"
+        echo '-------- Binary file --------'
         file -b "$1"
+	echo '------- Details -------'
+	stat "$1"
     else
         # Text file:
-        $PAGER "$1" &
+	if [ "$(command -v highlight)" ]; then
```
jarun, line 79: See line 78 and 79. I think your editor settings is the problem.

plugins/preview-tui
```diff
@@ -18,17 +18,26 @@
...
+
+lines=$(($(tput lines)-1))
+cols=$(tput cols)
 
 preview_file () {
     kill "$(jobs -p)" 2>/dev/null
     clear
 
     encoding="$(file -b --mime-encoding "$1")"
+    typ="$(file -b --mime-type)"
```
jarun, line 39: For me:

file -b --mime-type abc.pdf 
application/pdf

Will it work? Also, where are you passing the file name?

leovilok: The output from `file -b --mime-type` is not the same as `file -b`, so further check need to be adapted. But as MIME types are standards, they should be the same across different systems with different implementations of `file`.

jarun:
I just tested it in the current status and I noticed that the newly created pane just flickers and exits. No preview is seen. However, I can see that the splitting logic is in play (e.g. in my case with tmux it now tries to split horizontally).

@leovilok can you give it a try and see what all needs to be fixed? Probably we can pull it in and you can sanitize it in another iteration.

jarun:
@Recidiviste I have pushed partial changes form your PR at commit 5df231947392475ced257a6ba5e4cec3c3fbb5a9. This covers the terminal splitting logic.

I think you can re-raise the PR with only the mime related changes and the alignment issues fixed.

jarun:
@Recidiviste In subsequent commits ca8bc8bf3059be0402c16032823cb061281141cd and 4ef6ab1082090ae8d56da8f9487aa4c1357bb028 I have added the basic framework to detect and match mime type and file extension so users can add new types based on their preferences.

I think you should be able to extend you local `preview-tui` based on these. I am closing the PR as I would like to leave the implementation generic.

PR #622, opened 2020-05-31
Wrap FILE_MIME_OPTS use in check

sjmulder:
We leave it undefined on Solaris/Illumos (at the top of `nnn.c`)
because no such options exists, so only use it if we have it. This
fixes the build on OpenIndiana Hipster, an Illumos distribution.


jarun:
Thank you!

PR #625, opened 2020-06-02
preview-tui: simplify, add generic fifo_pager()

leovilok:
I've commented new filetype checks for now, as we need to discuss which ones
should be included by default, keeping in mind that this is supposed to be a
minimal, adaptable plugin.

Maybe the `tar` listing is a good candidate? as it's probably a dependency people already have installed.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -5,7 +5,8 @@
 # Note: This plugin needs a "NNN_FIFO" to work.
 #
 # Dependencies: tmux (>=3.0) or xterm or $TERMINAL, less or $PAGER,
-#               file, stat, tree, add you own! (see examples in code)
+#               file, stat, tree, man, tar, unzip, ...
+#                      ... add you own! (see examples in code)
 #
 # Usage:
 #   You need to set a NNN_FIFO path and set a key for the plugin,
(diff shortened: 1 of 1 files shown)


jarun:
> Maybe the tar listing is a good candidate?

Why not zip as well?

jarun:
Also, I am unable to scroll in pager (mine is `less`) with tmux. Please see if that can be fixed.

leovilok:
> Also, I am unable to scroll in pager (mine is `less`) with tmux. Please see if that can be fixed.

I don't think that it's related with `nnn`, try opening `less` in `tmux`: it have the same behaviour as in the preview plugin.

leovilok:
I uncommented man, tar, and unzip previewers, but I feel like there should be better extension checking for `tar` (maybe /*.tar.*/ intead of /gz|bz2/)

jarun:
tar can have many many combos. I have added only 2 of them because plugins are indicative. Users can add more.

jarun:
We have the same stand for `nuke` as well. There is no way we can (and no reason we have to) cover every app, every mime and every extension.

jarun:
Thank you!

PR #630, opened 2020-06-03
Added GPG encrypt and decrypt plugins

KlzXS:
The encrypt plugin can either encrypt using a key or a passphrase (only a single file with a passphrase, gpg limitation). If asymmetric is chosen it lists the available public keys that support encryption along with the user id of that key in `fzf`.

The decrypt plugin attempts to decrypt a file. If there is no secret key gpg silently prints a message in the background. Decrypted files are saved in `original_name.dec`, I couldn't think of anything better and writing out to the shell seems like the lesser option.

### DIFF
--- a/plugins/gpgd
+++ b/plugins/gpgd
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-# Description: gpg decryption
+# Description: Decrypts selected files using gpg. The contents of the decrypted file are stored in a file with extension .dec
+#
+# Note: If an appropriate private key cannot be found gpg silently prints a message in the background and no files are written.
 #
 # Shell: POSIX compliant
 # Author: KlzXS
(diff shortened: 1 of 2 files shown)


jarun:
The documentation needs to go to the in-file docs as well. Please add to the encryption plugin.

jarun:
Thank you!

jarun:
@KlzXS I've already merged it. But do you see any reason we can combine both encrypt and decrypt in one file and ask the user at the beginning which action to perform?

jarun:
In addition:

- why does it need to be bash script?
- we need to confirm with user whether to use selection or hovered file when both are present
- the entry needs to go to the readme

KlzXS:
> But do you see any reason we can't combine both encrypt and decrypt in one file and ask the user at the beginning which action to perform?

At the time it seemed like too many levels of prompts when selection is included.

It doesn't need to be bash, that's a typo. I'll fix that, add to the readme, and update the prompt.

jarun:
OK. We can keep them separate. But at least add support for current and selection.

KlzXS:
Submited #633. Suggest any further improvements there.

PR #631, opened 2020-06-03
bookmarks: Pipe error to PAGER and clean up code

toddyamakawa:
column -t is slow because it waits for the for loop to finish before formatting. I found `printf` to be an acceptable alternative.

Also when NNN_PIPE isn't set, I print the error message in `$PAGER` so users can figure out what's gone wrong.


jarun:
Thank you!

Issue #617, opened 2020-05-31
Can't get preview to work


kuntau:
#### Environment details (Put `x` in the checkbox along with the information)

[x] Operating System: MacOS 10.14.6
[ ] Desktop Environment:
[x] Terminal Emulator: iTerm2
[x] Shell: zsh
[ ] Custom desktop opener (if applicable):
[ ] Program options used:
[ ] Configuration options set:
[x] Issue exists on `nnn` master

```sh
NNN_PLUG=f:fzcd;o:fzopen;z:fzz;d:diffs;t:treeview;v:preview-tui
NNN_FIFO=/tmp/nnn.fifo
TMUX v3.1b
nnn v3.2
```
Maybe I'm just dumb, but I can't get the preview to work.

1. I hover on file
2. Press `;`
3. Press `v`
4. `nnn` just flickered but nothing show.. what did I miss? 🤔

Please help. Thank you

jarun:
Did you run tmux first and then start nnn from within?

kuntau:
Run tmux first

jarun:
What is the output of `tmux -V`?

kuntau:
tmux 3.1b

jarun:
Also, please try the following patch:

```
diff --git a/plugins/preview-tui b/plugins/preview-tui
index 7872b07..e467236 100755
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -73,7 +73,7 @@ if [ "$PREVIEW_MODE" ] ; then
     exit 0
 fi
 
-if [ -e "${TMUX%%,*}" ] && [ "$(tmux -V | cut -c6)" -eq 3 ] ; then
+if [ -e "${TMUX%%,*}" ] ; then
     tmux split-window -e "NNN_FIFO=$NNN_FIFO" -e "PREVIEW_MODE=1" -dh "$0" "$1"
 else
     PREVIEW_MODE=1 $TERMINAL -e "$0" "$1" &
```

jarun:
Forget the patch in that case. The issue is somewhere else.

jarun:
Try the following:

1. tmux
2. nnn -a
3. `;` followed by `v`

What do you see?

jarun:
I need fast follow-up here. Don't have a lot of time in hand. Plugins are indicative. In most cases we request users to modify and suit themselves. I am offering some help here because it's a very new plugin.

kuntau:
[...]

PR #632, opened 2020-06-03
preview-tui: wait killed previewers

leovilok:
This prevent annoying job control messages, an possibly more issues with long-to-exit previewer precesses. Discussed in #617

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -52,7 +52,7 @@ fifo_pager() {
 }
 
 preview_file () {
-    kill %- %+ 2>/dev/null
+    kill %- %+ 2>/dev/null && wait %- %+ 2>/dev/null
     clear
 
     encoding="$(file -Lb --mime-encoding -- "$1")"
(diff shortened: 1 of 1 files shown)


Kabouik:
Thanks for this. It greatly improved directory preview with Node.js' `tree-cli` (which I think is significantly slower than `tree`, but `tree` is not available on all platforms), however I still get the job control messages if I hover a new directory while the previous one is still being calculated:

Screenshot_20200604_001

Kabouik:
I replaced `tree-cli` with [`tree-node-cli`](https://github.com/yangshun/tree-node-cli), problem fixed. Sorry for polluting the comments here, but just wanted to confirm this was related to that one module.

jarun:
Thanks @leovilok!

@Kabouik would it be possible to prepare the asciicast without any error? I thought about it and an error in a showcased media would look bad. We can't explain the problem to users and they may think that it's our issue.

jarun:
@Kabouik please ignore the comment above. I saw your new cast later.

Kabouik:
Totally, that's what I wanted to do too when you said you could re-use it in the wiki. I think the latest I uploaded should be ok.

PR #633, opened 2020-06-04
Added selection prompt and README entry for gpg plugin

KlzXS:


plugins/README.md
```diff
@@ -40,6 +40,8 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 | fzz | Change to any directory in the z database with fzf | sh | fzf, z |
 | getplugs | Update plugins to installed `nnn` version | sh | curl |
 | gutenread | Browse, download, read from Project Gutenberg | sh | curl, unzip, w3m<br>[epr](https://github.com/wustho/epr) (optional) |
+| gpgd | Decrypt files using gpg | sh | gpg |
+| gpge | Encrypt files using gpg | sh | gpg, fzf (see in-file docs) |
```
jarun, line 44: We can drop the `(see in-file docs)`. Generally we add it when the deps list is too big.

KlzXS: So should it just be optional? I mean, it is optional it works just fine without it, you just can't use asymmetric encryption. Which is kind of the main feature of the plugin, is it not?

jarun: I meant mention gpg, fzf.

Issue #627, opened 2020-06-02
Unified feature-full preview-* script


marioortizmanero:
I was setting up nnn the other day (really good job, I love it!) and I found the preview scripts a bit too minimal (I know they've been released recently and they're being worked on). While I really liked how straightforward it was to understand and configure, I think a feature-full plugin that unifies `preview-kitty`, `preview-tui` and `preview-tabbed` could be quite interesting and easier to set-up. One of my issues was that some features were available on the tui script, but not on the kitty one, so I had to copy-paste code between them.

How about following a more ranger-like approach in this one, and having a complete and configurable preview script for *everything*? [Here's theirs, scope.sh](https://github.com/ranger/ranger/blob/master/ranger/data/scope.sh). It's quite similar to what one would expect on nnn previews (text previews, images...). It's supposed to be a standalone script; I tried to build a wrapper around it so that it can be used in nnn, too, and I'd say it's possible.

I replied to some post by @jarun a few days back on Reddit and he suggested me to open an issue here. I also mentioned in that message an issue I was having with specifically the preview-kitty script, but I'm not sure if this should go in a different issue, please let me know. Basically, when you use preview-kitty and open a file, the preview window remains open. It'd be great to close it once you open a file, and re-open it once you go back, but I'm not sure how this could be done, and if it happens with other preview scripts. Here's an example:

![image](https://user-images.githubusercontent.com/25647296/83578797-f0cd6b80-a537-11ea-9c3e-9964149d1148.png)

*I press return or `l`*

![image](https://user-images.githubusercontent.com/25647296/83578812-f7f47980-a537-11ea-8694-6dcf1fb47d39.png)

I could definitely help with the scripts. Let's just discuss what can be done first.

jarun:
> I think a feature-full plugin that unifies preview-kitty, preview-tui and preview-tabbed could be quite interesting and easier to set-up.

Please feel free to write one and contribute it back an a unified plugin. During development we figured that keeping them separate is a simpler way to maintain.

> Basically, when you use preview-kitty and open a file, the preview window remains open.

Now we are seeing a use case where people want this to happen so they can edit a file (say markdown, man page, html) and view the update live in the preview window. That contradicts your use case and we can't have them both. I would suggest you have the option to close the preview window in your unified plugin.

@leovilok is there a possibility of closing the preview window by pressing a key in the preview window?

leovilok:
[...]

PR #634, opened 2020-06-04
[WIP] Unified preview-tui and preview-kitty scripts

marioortizmanero:
This is a continuation of #627. It unifies `preview-tui` and `preview-kitty` into a single script. This makes maintaining them much easier, as most of their contents were the same, and for instance `preview-kitty` was missing some features included in the latest modifications of `preview-tui`, like the pager.

Here's the order this script follows to know where to show the previews:

1. tmux
2. kitty terminal (if `$KITTY_WINDOW_ID` is set, see https://github.com/kovidgoyal/kitty/issues/957)
3. other terminals (according to `$TERMINAL`)

There are two options available to handle the file previews:

* Using the `preview-tui` defaults as a standalone script only for nnn. It's quite limited, but more minimal.
* Using `preview-tui` as a wrapper for [`scope.sh` from ranger](https://github.com/ranger/ranger/blob/master/ranger/data/scope.sh), which already handles lots of different types of files.

For now it will only use the `scope.sh` script if it's found in `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/scope.sh`.

TODO:
* [ ] Bring in more features from `scope.sh`. Also, support `w3mimgdisplay` and test `catimg`.
* [ ] Documentation at the top of the script and in the wiki.
* [ ] Try to find `scope.sh` in the `$HOME/.config/ranger/scope.sh` directory, too? This would allow users to use a single script for both file managers.


plugins/preview-tui
```diff
@@ -55,38 +60,64 @@ preview_file () {
...
     if [ -n "$ext" ]; then
         ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
     fi
+    lines=$(($(tput lines)-1))
+    cols=$(tput cols)
+
+    # Trying to use scope.sh if it's available.
+    if [ -f "$SCOPE_SCRIPT" ] && [ -x "$SCOPE_SCRIPT" ]; then
+        fifo_pager "$SCOPE_SCRIPT" "$1" "$cols" "$lines" "$(mktemp -d)" \
+            "True" 2>/dev/null
```
leovilok, line 77: Why redirect stderr? Are you expecting error messages that we don't want to show to users ?

plugins/preview-tui
```diff
@@ -55,38 +60,64 @@ preview_file () {
...
+            fifo_pager unzip -l "$1"
+        elif [ "$ext" = "gz" ] || [ "$ext" = "bz2" ] ; then
+            fifo_pager tar -tvf "$1"
+        else
+            # Binary file: show file info
+            printf -- "-------- \033[1;31mBinary file\033[0m --------\n"
+            if exists mediainfo; then
+                fifo_pager mediainfo "$1" 2>/dev/null
+            else
+                fifo_pager file -b "$1"
```
leovilok, line 109: We're missing the `stat` command that someone added before, but as `fifo_pager` only support running 1 command ... I guess this is OK. (I didn't wrap it in `fifo_pager` because I thought that file + stat could fit in a window unless it's ridiculousely small)

plugins/preview-tui
```diff
@@ -55,38 +60,64 @@ preview_file () {
...
+            elif exists catimg; then
+                catimg "$1"
+            fi
+        elif [ "$mimetype" = "application/zip" ] ; then
+            fifo_pager unzip -l "$1"
+        elif [ "$ext" = "gz" ] || [ "$ext" = "bz2" ] ; then
+            fifo_pager tar -tvf "$1"
+        else
+            # Binary file: show file info
+            printf -- "-------- \033[1;31mBinary file\033[0m --------\n"
```
leovilok, line 105: This line is useless if you start a pager afterwards.

jarun:
Please rename to a new script, do not modify the existing ones.

plugins/preview-tui
```diff
@@ -55,38 +60,64 @@ preview_file () {
...
-        # Binary file: just print filetype info
-        echo "-------- binary file --------"
-        file -b "$1"
-        echo
-        stat "$1"
     else
-        # Text file:
-        $PAGER "$1" &
+        if exists bat; then
+            bat --paging=never --decorations=always --color=always \
```
leovilok, line 116: Don't use `--paging=never` ... or use `fifo_pager`, but as it is it will only show the end of the file.

jarun:
Why is the name scope.sh? We do not use .sh extension in the plugins. Also I would like to have a different name to what is used by another FM. It would confuse users.

@leovilok once the above is done, please review this.

jarun:
@leovilok what do you say about a single unified plugin instead of 3 separate ones?

plugins/preview-tui
```diff
@@ -114,6 +145,11 @@ if [ -e "${TMUX%%,*}" ] && tmux -V | grep -q '[ -][3456789]\.' ; then
     fi
 
     tmux split-window -e "NNN_FIFO=$NNN_FIFO" -e "PREVIEW_MODE=1" -d"$SPLIT" "$0" "$1"
+elif [ -n "$KITTY_WINDOW_ID" ]; then
```
leovilok, line 148: I'm not sure about prioritizing tmux over kitty, especially since it would probably mess up image preview if the previewer pane is inside a tmux pane in a kitty window. Are there people using both with an opinion on this?

Also checking $KITTY_WINDOW_ID is not enough, as it doesn't check for `allow_remote_control` availability. I think that we had a check for that in `preview-kitty`.

plugins/preview-tui
```diff
@@ -55,38 +60,64 @@ preview_file () {
...
+        if exists tree; then
+            fifo_pager tree
+        elif exists exa; then
+            fifo_pager exa -G --colour=always 2>/dev/null
+        else
+            fifo_pager ls --color=always
+        fi
+    elif [ "$encoding" = "binary" ]; then
+        if [ "${mimetype%%/*}" = "image" ] ; then
+            if [ -n "$KITTY_WINDOW_ID" ]; then
```
leovilok, line 92: Check if this works in a tmux pane in a kitty window.

marioortizmanero:
>Please rename to a new script, do not modify the existing ones.

I understand that you may not want support for scope.sh inside this same script, but combining both preview-kitty and preview-tui is the ultimate goal; and it would be impossible to do it without modifying the existing ones. Or do you mean just for some time for the sake of backward compatibility?

>Why is the name scope.sh? We do not use .sh extension in the plugins. Also I would like to have a different name to what is used by another FM. It would confuse users.

scope.sh is the name of the original file made by the ranger developers. I didn't change it because it's not tied to nnn at all. You'd just copy/symlink the script into the nnn plugins directory and done.

leovilok:
> @leovilok what do you say about a single unified plugin instead of 3 separate ones?

I'm not against merging what can be, If we keep it clear what dependencies are really required (not much) and what are totally optionnal (most stuff). But I bielive that it'll be hard to merge `preview-tabbed` with the other ones, as management of X windows is quite different from reusing the same terminal.

jarun:
@leovilok please proceed as you think would be the right way.

marioortizmanero:
> I'm not against merging what can be, If we keep it clear what dependencies are really required (not much) and what are totally optionnal (most stuff). But I bielive that it'll be hard to merge preview-tabbed with the other ones, as management of X windows is quite different from reusing the same terminal.

Yes, I haven't touched the documentation at the top of the script yet because I wanted to discuss how it should be organized. And I wouldn't add preview-tabbed into the mix, as it doesn't have much to do with preview-kitty and preview-tui and it doesn't make much sense to me. But it could definitely be done just to have less scripts.

plugins/preview-tui
marioortizmanero: scope.sh is designed to show error messages like `bat command not found`, as all it does is launch the commands sequentially and hope one of them works. Without redirecting stderr, the output is filled with useless messages. I think the guys at ranger also do this, but I would have to check it myself.

plugins/preview-tui
marioortizmanero: The thing is that preview-kitty and preview-tui had different commands so I wasn't sure what to include. `fifo_pager` is definitely unnecessary here, I just added it everywhere I could.

jarun:
BTW, if we can retire some plugins after merging please do so.

plugins/preview-tui
marioortizmanero: True. How should I do this with fifo_pager, then? Something like this?

```sh
print_bin_header() {
    printf -- "-------- \033[1;31mBinary file\033[0m --------\n"
}

if exists mediainfo; then
     fifo_pager print_bin_header && mediainfo "$1" 2>/dev/null
else
     print_bin_header
     file -b "$1"
fi
```

plugins/preview-tui
leovilok: `fifo_pager` just pipe the output of a command into $PAGER, while using a FIFO to keep access to the $PAGER's PID, to be able to kill it afterwards.
So if you need to pipe to a $PAGER, use `fifo_pager`, but if your output fits in the screen, you don't need it.

plugins/preview-tui
marioortizmanero: I didn't know image previews don't work inside tmux. Kitty should go first, then.

AFAIK the original preview-kitty doesn't check allow_remote_control. Is there a command to check that, or do you mean running a grep into the kitty config (quite ugly)?

leovilok:
I'm not sure how to handle the `scope.sh` situation ... do we:
- include it and maintain a fork
- tell people to get it from ranger
- (convince them to extract it to another repo :p )

plugins/preview-tui
marioortizmanero: `KITTY_WINDOW_ID` is apparently set even in tmux sessions:

image

Should this also check that tmux isn't being used, then?

marioortizmanero:
Oh I like the last option. We could talk with the ranger devs to join forces and make scope.sh less more generic and tied to ranger, moving it into a separate repository for other file managers to use/collaborate. But I doubt if it'll be possible for them, as for example kitty previews support is outside this file, and integrated into their code.

Otherwise I would make a "fork", because as I said, it might have to be modified.

plugins/preview-tui
leovilok: > Should this also check that tmux isn't being used, then?
yes, and maybe give back tmux the priority over kitty ...

plugins/preview-tui
leovilok: > AFAIK the original preview-kitty doesn't check allow_remote_control. Is there a command to check that, or do you mean running a grep into the kitty config (quite ugly)?

We had a check earlier, based on kitty authors' suggestion, wich was to run a basic remote control command and check for success (like `kitty @ ls`). It might sitll be in the file's history?

plugins/preview-tui
leovilok: No, because this will only page the header and than print file content unpaged.

`fifo_pager` executes its first argument and pass it the remaining arguments, so you can't use more than one command. You can use a shell function though:
~~~sh
print_bin_info () {
    printf -- "-------- \033[1;31mBinary file\033[0m --------\n"
    if exists mediainfo; then
        mediainfo "$1" 2>/dev/null
    else
        file -b "$1"
}

...

fifo_pager print_bin_info "$1"
~~~

or
~~~sh
print_bin_info () {
    printf -- "-------- \033[1;31mBinary file\033[0m --------\n"
    mediainfo "$1" 2>/dev/null
}

if exists mediainfo; then
     fifo_pager print_bin_info
else
     printf -- "-------- \033[1;31mBinary file\033[0m --------\n"
     file -b "$1"
fi
~~~

plugins/preview-tui
```diff
@@ -67,6 +68,15 @@ preview_file () {
     if [ -n "$ext" ]; then
         ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
     fi
+    lines=$(($(tput lines)-1))
+    cols=$(tput cols)
+
+    # Trying to use scope.sh if it's available.
+    if [ -f "$SCOPE_SCRIPT" ] && [ -x "$SCOPE_SCRIPT" ]; then
```
toddyamakawa, line 75: You don't need to check if the file exists and is executable right? Can you just check if it's executable?

marioortizmanero: I wasn't sure so I checked both. I'll only use -x then.

plugins/preview-tui
marioortizmanero: Okay, let's use the first one, as it seems more flexible.

plugins/preview-tui
leovilok: I saw you version checking `$TERMINAL`, it should work for the image preview, but I think that it's not enough to decide if you can open a kitty pane.

The check I was talking about was actually discarded before plugin inclusion, it was:
~~~sh
if [ -n "$KITTY_WINDOW_ID" ] && kitty @ ls 1>/dev/null 2>&1 ; then
~~~
(Note that it's a bit slow, so it should not be used in the preview loop, unless it's cached)

marioortizmanero: Yeah, I was checking the script history and I couldn't find what you were telling me. I made sure it works myself, so I'll add it.

jarun:
Guys, do let me know when this is good to go. I have marked WIP so I don't merge it accidentally.

leovilok:
@jarun do you have an opinion on the `scope.sh` case?

Appart from that we only lack documentation before merging this.
(@marioortizmanero other image preview methods can come later)

jarun:
scope.sh: go as you like. :+1:

marioortizmanero:
Some considerations after working on it for a bit more:

* Should we even list basic things in the documentation like unzip, tar, or man?
* Why is [`--allow-remote-control`](https://sw.kovidgoyal.net/kitty/launch.html#cmdoption-launch-allow-remote-control) not being used? Is it enabled by default? Can it be replaced by `allow_remote_control`? I don't understand that option.

Kitty can now split vertically or horizontally depending on `$SPLIT`, too! What do you think of how I implemented it? Apparently you can't set the layout only for a launch, so this first sets it to `splits` and restores it afterwards. I don't know if I missed that, but I should open a feature request on kitty, too. It seems quite useful. The issue is that for some reason when I try to restore the layout to the previous one, kitty breaks. So I might be doing it wrong??:

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/bin/../lib/kitty/__main__.py", line 121, in <module>
    main()
  File "/usr/bin/../lib/kitty/__main__.py", line 117, in main
    func(sys.argv[1:])
  File "/usr/bin/../lib/kitty/__main__.py", line 23, in remote_control
    rc_main(args)
  File "/usr/bin/../lib/kitty/kitty/remote_control.py", line 187, in main
    response = do_io(global_opts.to, send, no_response)
  File "/usr/bin/../lib/kitty/kitty/remote_control.py", line 136, in do_io
    received = io.simple_recv(timeout=10)
  File "/usr/bin/../lib/kitty/kitty/remote_control.py", line 115, in simple_recv
    read_command_response(self.tty_fd, timeout, ans)
OSError: [Errno 5] Input/output error
```

Also, tmux splits for some reason are inverted... A vertical split in tmux is this:

image

while both in kitty and vim it means the exact opposite:

image
image

So I just used what kitty and vim do, and inverted it for tmux... Quite confusing.

I've also added a fallback for images to show the binary information when no preview methods are available.

leovilok:
> I've worked on it a bit more. Should we even list basic things in the documentation like unzip, tar, or man?

We should list every **required** dependency in the file header. If there are few required stuff, add if in the plugins' `README.md` page too; but if there are too much dependencies just mention "*see in-file doc*" (see e.g. `nuke` plugin).

For additional, optional dependencies (where a fallback is available), maybe just add something like "*more optional previewers/windows available, look at the source*" (but the wording is probably bad, I suck at this).

And maybe instructions specifically for `scope.sh` too.

> Kitty can now split vertically or horizontally depending on `$SPLIT`, too! What do you think of how I implemented it? Apparently you can't set the layout only for a launch, so this first sets it to `splits` and restores the layout afterwards. I don't know if I missed that, but I should open a feature request on kitty, too. It seems quite useful. The issue is that for some reason [when I try to restore the layout to the previous one](https://github.com/jarun/nnn/pull/634/files#diff-2f4f01bd0a72ac577e4ee5c4c7f71562R195), kitty breaks. So I might be doing it wrong??:

Souds like a bug in kitty, you might want to discuss it in a kitty issue. Or maybe it's just some kind of tty io concurrency issue. @kovidgoyal warned us about that in the past, and maybe interleaving a `kitty @` call with `--no-response` and one without is not supported.

> So I just used what kitty and vim do, and inverted it for tmux... Quite confusing.

It just another way to see it (split as an action, not as an object) see https://github.com/tmux/tmux/issues/213

> I've also added a fallback for images to show the binary information when no preview methods are available.

Good, this keeps the requirements minimal.

kovidgoyal:
> Souds like a bug in kitty, you might want to discuss it in a kitty issue. Or maybe it's just some kind of tty io concurrency issue. @kovidgoyal warned us about that in the past, and maybe interleaving a `kitty @` call with `--no-response` and one without is not supported.
> 

If you think its a bug in kitty come up with a minimal script containing
remote control commands to reproduce it and I will be most happy to take a look.

kovidgoyal:
Oh and from the traceback above, it is breaking waiting for a response which is indeed because of tty sharing issues. I can easily add a --no-response to last-used-layout as well, to workaround that, or just use a dedicated socket for remote control.

leovilok:
@kovidgoyal thank you!

leovilok:
@kovidgoyal in `nnn` we don't start kitty, so we're dependent on how users configured it. We could require that they setup a socket, but we try to have minimal requirements. A `--no-response` flag for `last-used-layout` would help, indeed, if it fixes the issue.

kovidgoyal:
It's already added in master.

marioortizmanero:
> We should list every required dependency in the file header. If there are few required stuff, add if in the plugins' README.md page too; but if there are too much dependencies just mention "see in-file doc" (see e.g. nuke plugin).

Let's just leave it like that, then. It seems like a better idea to me.

> Oh and from the traceback above, it is breaking waiting for a response which is indeed because of tty sharing issues. I can easily add a --no-response to last-used-layout as well, to workaround that, or just use a dedicated socket for remote control.

Okay, thanks for the help! We'll have to wait until every distro has this option available. For now, I'd just use the `--no-response` flag, and in case it fails, just do nothing and keep the `splits` layout. @kovidgoyal, now that you're here, is it possible to launch a command with a specific layout only for that launch? I think it'd be better than first setting the layout, then launching the window/whatever, and finally restoring the original layout. Some flag like `--layout` for `kitty launch`. That way we could just do `kitty launch --layout splits --location "${SPLIT}split"` If you want I can just submit this feature request on the kitty repository instead.

kovidgoyal:
> > Oh and from the traceback above, it is breaking waiting for a response which is indeed because of tty sharing issues. I can easily add a --no-response to last-used-layout as well, to workaround that, or just use a dedicated socket for remote control.
> 
> Okay, thanks for the help! We'll have to wait until every distro has this option available. For now, I'd just use the `--no-response` flag, and in case it fails, just do nothing and keep the `splits` layout. @kovidgoyal, now that you're here, is it possible to launch a command with a specific layout only for that launch? I think it'd be better than first setting the layout, then launching the window/whatever, and finally restoring the original layout. Some flag like `--layout` for `kitty launch`. That way we could just do `kitty launch --layout splits --location "${SPLIT}split"` If you want I can just submit this feature request on the kitty repository instead.

If you launch in a new tab or os window, sure, but launching should not
change layouts as a side effect, since layout is a property of the tab
and not an individual window.

marioortizmanero:
Can confirm that using `--no-response` worked! I've redirected stderr to `/dev/null` with `2>&1` for now to avoid a very likely error message when launching the plugin.

What about `scope.sh`, then? Should we still support it? How can that be done? I'm leaving that to you guys because I'm really new to the project and your decisions will surely be more appropiate :)

jarun:
@leovilok let's _not_ incorporate scope.sh as a plugin. We can document the inegration.

I want to be very clear from support perspective. Tomorrow I do not want any user coming back and saying xyz feature that works with ranger and scope.sh is not working in case of `nnn` and `nnn` devs have to spend time debugging it.

I think the best approach would be to document how to integrate scope.sh with xyz plugin on an as-is basis in the Live Previews Wiki page.

jarun:
@marioortizmanero 

> Should we still support it? How can that be done?

Can you explain this? Are you saying scope.sh is not necessary for your plugin? If that's the case, what is the relevance wrt. this discussion?

marioortizmanero:
> Can you explain this? Are you saying scope.sh is not necessary for your plugin? If that's the case, what is the relevance wrt. this discussion?

I initially suggested `scope.sh` to integrate it into nnn to handle the previews, rather than what's currently being used, as it already supports lots of different methods and it's used outside of nnn, too. The problems are:

* Should this be optional, or the only and "official" way to manage TUI previews? Supporting more preview methods inside `preview-tui` is basically copying and pasting from `scope.sh`.
* How can this be integrated into nnn? Should nnn use a modified version to support kitty previews and other methods not included from ranger?

Sorry for not being clear enough.

jarun:
> Supporting more preview methods inside preview-tui is basically copying and pasting from scope.sh.

We'll let the users add whatever they need. I am pretty sure if 100 users modify it, there will be 100 different versions, so 1 big script doesn't fit the bill. When it comes to plugins we prefer to guide the usres to things themselves. For people who want ready-made stuff, there are many alternative file managers already.

> How can this be integrated into nnn?

We already support kitty previews in plugin `kitty-preview`. What did you unified plugin do differently? I don't understand the integration problem.

leovilok:
I'll be offline for a few days, so this might be merged before I'm back.
I reviewed the in file docs and last code changes: it seems good to me. Great job @marioortizmanero :+1: 

On the `scope.sh` issue: we can either leave it commented in the source with a note, or maybe just leave a note to let people know that they can take inspiration for ranger's `scope.sh` to add more file viewers.

`plugins/REAMDE.md` also needs to be updated (remove `preview kitty` line, redirect to in-file docs for dependencies).

jarun:
@leovilok thanks for the notes! I'll take care of these.

@marioortizmanero once I'm done, I'll request for a review.

jarun:
Thank you!

jarun:
@Kabouik and @basbebe can you please confirm the preview-tui plugin works fine? I would like to ensure it's flawless before the next release. We have invested a lot of time on this.

Kabouik:
Sure, I'll have a look after work later today.

jarun:
Thanks a lot!

Kabouik:
Seems to work pretty well for me. I like that text preview now shows syntax highlighting. That was not the case with the pager in the previous `preview-tui` version I was using.

I made custom changes locally, namely to set `tree` to `tree -L 3 -F`, and to replace `catimg "$1"` with `viu "$1" &`. On my SailfishOS phone, `viu` is the only solution since we have no terminal with SIXEL support, but even on PC I use Tilix which can't preview real images either, so `viu` is still convenient on PC. I think there are still many terminals that don't support SIXEL, so maybe it would be useful to include the `viu` fallback (or maybe just mention it in comments).

I have mixed feelings about the red colouring of "Binary file". On the one hand I like it, it's fancy and easy to spot. On the other hand it makes `nnn` looks less minimal for no obvious reason. Maybe synchronizing the "Binary file" colour with colour in use for the current context would allow a better integration. But that is very minor, I don't have any strong opinion against it anyway, I might just leave it as is.

jarun:
Thanks @Kabouik!

regd. the red color, I think we are good. It's easily visible and the user can change them anytime as it's in the plugin.

jarun:
As I can see, it's WIP. I prefer [epr](https://github.com/wustho/epr). But for now, let's leave this alone. I think most people who use previews would go for scope.sh or pistol.

marioortizmanero:
Should we change `tree` to `tree -L 3 - F` by default though? It makes much more sense to me as a default option, `tree` is almost unusable in directories like `~`. We could also add `viu` but that might be unnecessary.

jarun:
ok.

Issue #614, opened 2020-05-30
Problem using kitty preview


basbebe:
I'm having a similar problem like https://github.com/jarun/nnn/issues/578 (preview pane closes immediatley after opening).
`kitty @ launch man man` works fine.

I'm calling `NNN_PLUG='p:kitty-preview' n -a`.
When I open the plugin I see a short flicker of the preview, then it's gone.

This is my env:

```
COLORTERM=truecolor
DISPLAY=/private/tmp/com.apple.launchd.BGywV6KAIn/org.macosforge.xquartz:0
EDITOR=kak
HOME=/Users/bastian
HOMEBREW_GITHUB_API_TOKEN=XXXXXXXX
KITTY_WINDOW_ID=8
LANG=de_DE.UTF-8
LESS=-g -i -M -R -S -w -z-4
LESSOPEN=|/usr/local/bin/lesspipe.sh %s
LESS_ADVANCED_PREPROCESSOR=1
LOGNAME=bastian
MANPAGER=sh -c 'col -bx | bat -l man -p'
NNN_ARCHIVE=\.(7z|bz2|gz|tar|tgz|zip)
NNN_BMS=d:~/Documents;u:~/;D:~/Downloads/
NNN_COLORS=1234
NNN_OPENER=/Users/bastian/.config/nnn/plugins/nuke
NNN_PLUG=f:finder;o:fzopen;p:preview-kitty;d:diffs;t:nmount;v:imgview;g:-_git diff;l:-_git log;b:-_bat $nnn*;s:-pdfview
PAGER=less
PATH=/usr/local/bin:/usr/local/sbin:/Users/bastian/.local/bin:/Users/bastian/.cache/dein/repos/github.com/junegunn/fzf/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Library/Apple/usr/bin:/usr/local/MacGPG2/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Library/TeX/texbin:/usr/local/sbin:/Applications/kitty.app/Contents/MacOS
PWD=/Users/bastian
SHELL=/usr/local/bin/fish
SHLVL=1
SSH_AUTH_SOCK=/private/tmp/com.apple.launchd.dXY9WGEm9I/Listeners
STARSHIP_SHELL=fish
TERM=xterm-kitty
TERMINFO=/Applications/kitty.app/Contents/Frameworks/kitty/terminfo
[...]

PR #636, opened 2020-06-05
preview-kitty: fix crash on first file on dash

leovilok:
In the dash shell, when `exec < fifo` is interrupted by SIGCHLD, it exits.
So we replace it with `cat fifo |`.

Issue discussed in #614

@marioortizmanero I believe that you'll have to backport this to #634

### DIFF
--- a/plugins/preview-kitty
+++ b/plugins/preview-kitty
@@ -63,7 +63,9 @@ if [ "$PREVIEW_MODE" ] ; then
 
     preview_file "$1"
 
-    exec < "$NNN_FIFO"
+    # use cat instead of 'exec <' to avoid issues with dash shell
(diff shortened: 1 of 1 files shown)


jarun:
Travis fails with the following:

```
In plugins/preview-kitty line 67:
 
    cat "$NNN_FIFO" |\
 
        ^-- SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
```

Please add a note and shellcheck exception.

leovilok:
Oh sorry, I forgot to shellcheck it (even though I suspected there would be an issue).

It's fixed now.

jarun:
Thank you!

PR #642, opened 2020-06-09
preview-tui splits are inverted

marioortizmanero:
I forgot to invert them properly. It now works like this:

image

image

Which is how it used to work, and how you'd expect it.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -64,10 +64,10 @@ else
     TERMINAL="${TERMINAL:-xterm}"
 fi
 
-if [ -z "$SPLIT" ] && [ $(($(tput lines) * 2)) -gt "$(tput cols)" ] ; then
-    SPLIT='v'
-elif [ "$SPLIT" != 'v' ] ; then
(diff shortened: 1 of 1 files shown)


jarun:
Can you also add a dummy SPLIT= at the top so users know where to change?

marioortizmanero:
Shouldn't `SPLIT` be indicated as an environmental variable, like `SPLIT=v nnn` or with `export`? Otherwise I'd have to add `TERMINAL` to the top, too.

jarun:
Can we have the following:

    SPLIT="${SPLIT}"
    TERMINAL="{TERMINAL}"

jarun:
This is for users who do not want to export in the environment and use a fixed terminal with fixed size.

marioortizmanero:
What about

```sh
SPLIT=
TERMINAL=
```

or maybe

```sh
SPLIT="$SPLIT"  # you can set a permanent split here
TERMINAL="$TERMINAL"  # same goes for the terminal
```

jarun:
Second one is fine.

marioortizmanero:
Done. I also added a clarification regarding `SPLIT`, as it might be interpreted in two different ways (see tmux).

jarun:
Thank you!

PR #643, opened 2020-06-09
Kitty splits

marioortizmanero:
Continuation of https://github.com/jarun/nnn/pull/634

So turns out that after managing to get `--no-response` for `kitty @ last-used-layout`, it doesn't exactly work as I wanted to. `kitty @ launch` is non-blocking, so `kitty @ last-used-layout` is executed immediately afterwards, and the layout goes back to the original one, which is horizontal by default. The layout should only be restored after the script is done.

I also wanted to add a note indicating that this will only work for versions older than 0.17.4, but I'm not sure what number the next release is going to have. Can you give us a hint, @kovidgoyal?

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -180,7 +180,8 @@ if [ "$PREVIEW_MODE" ] ; then
         preview_file "$selection"
     done
 
-    # Restoring the previous preview for kitty users
+    # Restoring the previous layout for kitty users. This will only work for
+    # kitty >= 0.18.0.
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm sorry what hint are you looking for? The next kitty release is going to be 0.18.0

marioortizmanero:
I was looking for the next release that includes the new `--no-response` flag for `kitty @ last-used-layout`. Thanks.

jarun:
If this is also a separate branch, please re-raise on master.

PR #644, opened 2020-06-09
Fixes environmental variables not being passed on preview-tui

marioortizmanero:
As I explained at https://github.com/jarun/nnn/commit/a47d190654204b8a46c7be9cc641382e597c25ad, we're missing a few environmental variables when launching a new kitty window.

I think I haven't missed any, but please make sure before merging.


jarun:
Thank you!

jarun:
Please do not create new branches. We keep developing on master and stabilizing it. And we definitely definitely don't need them for plugin updates.

jarun:
Please re-raise on master.

PR #645, opened 2020-06-09
Pistol support

marioortizmanero:
As https://github.com/jarun/nnn/commit/a47d190654204b8a46c7be9cc641382e597c25ad suggests, this provides support for [pistol](https://github.com/doronbehar/pistol).

Should I open an issue in their repo to let them know and possibly mention nnn in their docs? They seem really ranger-oriented.


jarun:
> Should I open an issue in their repo to let them know and possibly mention nnn in their docs? They seem really ranger-oriented.

Sure thing! I don't see a problem. ranger, lf, nnn are just different choices for users.

jarun:
Please re-raise on master.

PR #647, opened 2020-06-10
preview-tui: make scope & pistol run before `file`

doronbehar:
Both Pistol and `scope.sh` run `file --mimetype` or an equivalent, internally. Thus, for speed, it's better to run them before almost anything else.

Specifically for Pistol, the motivation for his creation was to avoid as much as possible shell code when previewing files, so for this plugin, the least can be done is this.

I don't use nnn and I didnt test this patch. I just thought it's trivial.


jarun:
Thank you!

PR #649, opened 2020-06-11
Sane tree defaults and viu support in preview-tui

marioortizmanero:
Applying sane defaults in `preview-tui` for `tree`, and `viu` support, as suggested in here https://github.com/jarun/nnn/pull/634#issuecomment-641469277


jarun:
Thank you!

PR #659, opened 2020-06-18
Avoid SIGWINCH interruptions of NNN_PIPE IO

### LINKED ISSUE #656: Resizing a window for a TUI app called through NNN_PLUG results in immediate termination of application
Release and 4b4bf6c. Not sure if this is intended behavior, or whether the issue is Wayland or sway specific as I do not have a box with X on hand at the moment.

Arch + sway + alacritty
Debian 10 Testing + sway + alacritty

Repro:
1. Set arbitrary `e:-_nano $nnn*;`
2. Launch nano from plugin
3. nano is launched as expected; no issues with use
4. Move and/or resize window containing nnn/nano
5. nano exists and window either corrupts and/or returns `[server exited unexpectedly]`

At this point, the screen is unable to be redrawn with ^L. Issue occurs with tmux/vim/kakoune/kibi/et al. The expecation is that moving/resizing the window will also redraw and resize its contents as seen when the same applications listed above are launched through  _o ^O  Open with..._

leovilok:
Fixes #656

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4300,11 +4300,20 @@ static void rmlistpath()
 	}
 }
 
+static ssize_t read_nointr(int fd, void *buf, size_t count)
+{
+	ssize_t len;
+	do{
+		len = read(fd, buf, count);
+	} while (len == -1 && errno == EINTR);
+	return len;
+}
+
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #660, opened 2020-06-18
Fix image previews with pistol/scope.sh and add note

marioortizmanero:
When using pistol or scope.sh for image previews, one has to execute the `pistol` or `scope.sh` commands with `&`, or otherwise a black block from the image will remain after changing to a different file:

image

I've also added a small disclaimer for using kitty preview with those methods that's exclusive to nnn.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -53,7 +53,9 @@
 #   Kitty users need `allow_remote_control` set to `yes`. To customize the
 #   window split, `enabled_layouts` has to be set to `all` or `splits` (the
 #   former is the default value). This terminal is also able to show images
-#   without extra dependencies.
+#   without extra dependencies. For image previews with kitty using scope.sh
+#   or pistol, you'll need to configure it with --transfer-mode=stream, as
+#   it's done in this script.
 #
 # Shell: POSIX compliant
 # Authors: Todd Yamakawa, Léo Villeveygoux, @Recidiviste, Mario Ortiz Manero
@@ -116,14 +118,14 @@ preview_file () {
 
     # Trying to use pistol if it's available.
     if [ "$USE_PISTOL" -ne 0 ] && exists pistol; then
-        fifo_pager pistol "$1"
+        fifo_pager pistol "$1" &
         return
     fi
 
     # Trying to use scope.sh if it's available.
     if [ "$USE_SCOPE" -ne 0 ] && exists scope.sh; then
         fifo_pager scope.sh "$1" "$cols" "$lines" "$(mktemp -d)" \
-            "True" 2>/dev/null
+            "True" 2>/dev/null &
         return
     fi
 


leovilok:
`fifo_pager` already starts the command in background, so that should not be an issue...

marioortizmanero:
That's... weird. Can you reproduce my issue?

leovilok:
I'll have to install pistol or scope.sh, does this happen only in kitty?

marioortizmanero:
I can't get other image preview methods like `catimg` to work right now, so I'm using kitty. Here's my pistol config file:

```
image/* kitty +kitten icat --transfer-mode=stream %pistol-filename%
```

It can be reproduced with scope.sh or pistol when using kitty.

leovilok:
OK, so this issue you are fixing only happens in modified/configured scope.sh/pistol, maybe it can be fix in those places too. I'll try to reproduce anyways.

marioortizmanero:
> OK, so this issue you are fixing only happens in modified/configured scope.sh/pistol, maybe it can be fixed in those places too. I'll try to reproduce anyways.

Yes, because neither scope.sh nor pistol offer image previews by default.

I've just tried with `viu`, which compared to `catimg`, does work. But it's not the same as the kitty images, as it's just text, so I can't reproduce it with that one.

marioortizmanero:
Here's a scope.sh file with kitty previews enabled, too: http://ix.io/2py2/sh

leovilok:
I can reproduce with pistol now, with your config ... it looks like issues I fought agains when writing `preview-kitty`, kitty's icat is just not killed when it should be.

marioortizmanero:
Exactly. It's weird because if you resize the terminal a bit, the black block will go away. Maybe `--transfer-mode=stream` is causing the issue here? Or perhaps a combination of other flags could fix this, too. I've also tried `image/* kitty +kitten icat --silent --transfer-mode=stream --stdin=no %pistol-filename%` to no avail.

marioortizmanero:
To be fair I can reproduce this without involving nnn at all. Here's a simple script, similar to what nnn does:

```sh
#!/bin/sh

launcher() {
    # sleep 1000 &
    kitty +kitten icat ~/Pictures/Screenshots/2019-10-13_21-34.png | less &
}

main() {
    launcher

    echo "There should be a job for a sleep in here:"
    jobs

    echo "It should have been killed now:"
    kill %- %+ 2>/dev/null && wait %- %+ 2>/dev/null
    jobs
}

main
```

The script itself works, because the second echo isn't accompanied by any messages from `jobs`. It seems normal, but once you launch `less` or `nvim` for something else, you get the black box:

image

image

leovilok:
Well when we spawn kitty's icat from the plugin itself, its PID is registred in the shell's job list. So it's directly killed when we kill slow previewers. Whereas when it's spawned by pistol or scope.sh we actually kill pistol or scope.sh, which might kill child processes, but indirectly.

This is an issue I had multiple times in the past when working on this plugins: how to recursively kill background jobs. The issue is that job control is quite poorly defined in POSIX sh, and I'm not even sure about how to consistently get all child process' PID, or kill them recursively. For example, I think that `zsh` register a job item PID for each component of a pipeline, while `bash` doesn't.

In the current `preview-tui` plugin design, we only spawn single level background commands (no pipelines) just to avoid this issue, and reliably kill previewer processes. It's the only reason why we have `fifo_pager` btw.

If we could reliably recursively kill children (but not the `cat` that feeds us the FIFO) in a POSIX sh compatible way, we would avoid a lot of issues.

Kabouik:
Would it make sense to extend this PR to `preview-tui` leaving an empty/unused pane when used with `tmux` and when restoring a previously saved `tmux` layout (e.g., with `tmux-resurrect`)?

Example:

```
Open nnn in tmux and load preview-tui:

   +---------------------------+
   |             |             |
   |             |             |
   |             |             |
   |     nnn     |   preview   |
   |             |             |
   |             |             |
   |             |             |
   +---------------------------+


Save layout with tmux-resurrect, then reload it in a new terminal:

   +---------------------------+
   |             |             |
   |     nnn     |             |
   |             |             |
   +-------------+     bash    |
   |             |             |
   |   preview   |             |
   |             |             |
   +---------------------------+

```


> I've just tried with `viu`, which compared to `catimg`, does work. But it's not the same as the kitty images, as it's just text, so I can't reproduce it with that one.

Side question: do I understand correctly that `preview-tui` and `viu` are the only ways to preview images inside a tmux session? As far as I understand it's not possible to use SIXEL inside tmux, even with terminals that support images like Kitty.

jarun:
@leovilok please squash and merge when this is ready.

leovilok:
@marioortizmanero I managed to explicitely kill the whole subprocess tree (sparing only a `cat` and a `sh`, responsible for reading from the FIFO): https://github.com/leovilok/nnn/blob/preview-tui-monsterkill/plugins/preview-tui

But funnily enough this doesn't fixes the original black box issue: now I kill sub-sub processes, but `wait` can only wait for direct subprocesses.

@Kabouik I'm not familiar with `tmux`, but I think that tmux resurect tries to understand which commands are running in which pane, and respawn them. This can't work with preview plugins, as tmux can't setup NNN_FIFO out of nowhere.

On the side question: in unpatched tmux, I think that only text approximation of images is possible, but there are a ton of programms doing the same think as `viu`. Sixel is a "standard" technology to display images in the terminal, but it's not supported by `tmux`, neither by a lot of terminal emulators. Kitty uses it's own image viewing capability, not linked to Sixels (like terminology, iterm2).

toddyamakawa:
> Would it make sense to extend this PR to `preview-tui` leaving an empty/unused pane when used with `tmux` and when restoring a previously saved `tmux` layout (e.g., with `tmux-resurrect`)?

I'm not familiar with `tmux-resurrect`. Is it possible to to tell `tmux-resurrect` to ignore certain panes to restore? I don't see a way for `preview-tui` to fix this issue, unless I'm missing something.

marioortizmanero:
@Kabouik I think that should go in a separate issue to avoid mixing it up with this issue.

@leovilok I have to investigate this a bit more. I'll try to come up with something.

marioortizmanero:
Sorry about that, I'm on finals so I don't have much time right now. My last exam is this Thursday, so I'll hopefully get back at this then.

jarun:
All the best for your exams! :+1:

marioortizmanero:
I've been trying to come up with something but honestly it's just a mess. I was trying to use `setsid`, as this [SO answer suggests](https://unix.stackexchange.com/a/124148), or something that marks the future child processes so that they can be found easily with `ps -s $$`, but I can't get it to work. Also, `setsid` doesn't work for functions declared inside the script itself, nor I can get it to work with a pipe to `$PAGER` either. I think this would only be viable with something like `setsid bash -c "..."`. That way at least we would avoid not killing other background processes launched in the script.

I also tried to use pipelines instead of `fifo_pager` to simplify it a bit, and this seems to work, too:

```
pistol "$1" | $PAGER
pager_pid=$!
pistol_pid=$(jobs -p)
```

As long as there's a single pipeline, `jobs -p` works well. Otherwise `jobs -l` can do the trick, too. Is there a reason you decided to opt for the `fifo_pager` solution instead?

leovilok:
> I've been trying to come up with something but honestly it's just a mess. I was trying to use `setsid`, as this [SO answer suggests](https://unix.stackexchange.com/a/124148), or something that marks the future child processes so that they can be found easily with `ps -s $$`, but I can't get it to work. Also, `setsid` doesn't work for functions declared inside the script itself, nor I can get it to work with a pipe to `$PAGER` either. I think this would only be viable with something like `setsid sh -c "..."`. `setsid` would avoid not killing other background processes launched in the script.

Finding the whole sub-processes tree is possible with `ps`, but I think that the really hard part is waiting for them to exit after killing them. As `wait` can only wait for direct subprocesses.

> I also tried to use pipelines instead of `fifo_pager` to simplify it a bit, and this seems to work, too:
> 
> ```
> pistol "$1" | $PAGER
> pager_pid=$!
> pistol_pid=$(jobs -p)
> ```
> 
> As long as there's a single pipeline, `jobs -p` works well. Otherwise `jobs -l` can do the trick, too. Is there a reason you decided to opt for the `fifo_pager` solution instead, @leovilok?

Well i'm not sure that every shell handles this the same way, I initially though that `jobs -p` output was not well defined, because `zsh` don't print the standard defined output there (but maybe it's just in interactive mode?).

jarun:
Guys, if it's just a matter or an `&` document and move on.

leovilok:
@jarun it's not a matter or `&`, the code currently in this PR is wrong, it can generate zombi processes and stuff. The bug that it's trying to fix only happens when you customise other programs that are not even activated by default.

Actually, this PR (except maybe the comment at the top) could be close without merging.

But the work on relialy killing and waiting any subtree of processes is interesting, it could potentially make the code simpler (e.g. remove the `fifo_pager` shell function) and easier to hack on for beginners.

This PR is currently more about this discussion, but you can close it if you feel like it's cleaner, we can reopen another one later if we have something working.

jarun:
OK. I am closing this till we find a cleaner approach.

PR #662, opened 2020-06-19
Add plugin bulknew

KlzXS:
Keeps behavior of `nnn` when creating new files.

### DIFF
--- a/plugins/bulknew
+++ b/plugins/bulknew
@@ -4,6 +4,7 @@
 #              Plugin opens a temp file where each entry is to be written on a separate line
 #
 # Note: Only relative paths are supported. Absolute paths are ignored
+#       Leading and trailing whitespace in path names is also ignored
 #
 # Shell: POSIX compliant
 # Author: KlzXS
--- a/plugins/bulknew
+++ b/plugins/bulknew
@@ -1,0 +1,29 @@
+#!/usr/bin/env sh
+
+# Description: Allows for creation of multiple files/directories at the same time.
+#              Plugin opens a temp file where each entry is to be written on a separate line
+#
+# Shell: bash
+# Author: KlzXS
+
+EDITOR="${EDITOR:-vi}"
+TMPDIR="${TMPDIR:-/tmp}"
+
+printf "'f'ile / 'd'ir? "
+read -r resp
+
+if [ "$resp" = "f" ]; then
+	#shellcheck disable=SC2016
+	cmd='mkdir -p "$(dirname "{}")" && touch "{}"'
+elif [ "$resp" = "d" ]; then
+	cmd='mkdir -p {}'
+else
+	exit 1
+fi
+
+tmpfile=$(mktemp "$TMPDIR/.nnnXXXXXX")
+$EDITOR "$tmpfile"
+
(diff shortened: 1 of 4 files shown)


plugins/bulknew
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
+
+# Description: Allows for creation of multiple files/directories at the same time.
+#              Plugin opens a temp file where each entry is to be written on a separate line
+#
+# Shell: bash
```
leovilok, line 6: I don't see anything preventing this plugin to be POSIX shell compatible.

plugins/bulknew
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
+
+# Description: Allows for creation of multiple files/directories at the same time.
+#              Plugin opens a temp file where each entry is to be written on a separate line
```
leovilok, line 4: As you strip prefix '/' it could be worth mentionning that this only supports relative path, not absolute ones.

leovilok:
You also need to add an entry to `plugins/README.md`, you know the drill...

KlzXS:
> You also need to add an entry to plugins/README.md, you know the drill...

Funny thing. I remembered that that needed to be done, but then did something else and forgot.

plugins/bulknew
KlzXS: As far as I know `mktemp` is not part of POSIX. But we could maybe switch to something else that is.

leovilok: Or just declare it as a dependency (well it's available in GNU coreutils, busybox, osx, freeBSD and openbsd at least, so chances are that if you have an UNIX, you have `mktemp`). It's not related to bash anyways, it's an external command, and if you wanted to make it a bash script you'll have to use a bash shebang:
```sh
#!/usr/bin/env bash
```

plugins/bulknew
```diff
@@ -0,0 +1,31 @@
...
+#
+# Note: Only relative paths are supported. Absolute paths are ignored
```
leovilok, line 6: More like "Every path is treated as a relative path", currently absolute path aren't ignored, they are just pasted inside the current dir. (Well I'm not a native english speaker, so it might just be my interpretation...)

KlzXS: They are ignored just like when creating a new file in `nnn` with and absolute path.

Or am I misinterpreting "relative lath" and "no traversal" messages?

plugins/bulknew
KlzXS: That makes sense. I'll update all the scripts that use it. For some reason I thought it was a shell built-in.

KlzXS:
There is a slight difference between `nnn` and this plugin. `[space]/newfile` in `nnn` would create a dir called `[space]` and put `newfile` in there while the plugin would try to touch `/newfile`. If the all-whitespace directory is not the first one in the list then it creates it successfully.

But do we really need to support all-whitespace directories and files?

jarun:
Please document the difference in-file anyway.

jarun:
Guys, please squash and merge when you are ready.

leovilok:
I think we're good now. Merging.

gandhiShepard:
hell yeah! you all rock!

PR #665, opened 2020-06-23
Fix check when char is unsigned

sjmulder:
If char is unsigned (as on ARM) subtracting a larger number would
result in a wrap around, not a negative value.

    src/nnn.c: In function 'readpipe':
    src/nnn.c:4325:11: warning: comparison is always false due to limited
    range of data type [-Wtype-limits]

       if (ctx < 0 || ctx > CTX_MAX)
               ^


jarun:
Thank you!

PR #670, opened 2020-06-30
xconfirm prompts should indicate the expected action with [y/N]

marioortizmanero:
Currently, prompts followd by a `xconfirm` are just questions. It can be confusing because the user doesn't know what it's supposed to do. Should they type "yes", "y", "Y"? What happens if they Ctrl+C? What's the default action? Not enough information, IMO.

I think that's quite easy to solve by adding the typical `[y/N]` inside the question. The way I've added it is quite manual, but the prompt system is quite basic (I found using `#define`s to act as a map quite funny, maybe I'm not used to C :P), so I don't think it matters much. By "manual" I mean that knowing that all `xconfirm` prompts have the `[y/N]` block, so it could be integrated into that function. Didn't want to touch it anyway because it might overcomplicate it.


jarun:
Do you see the last line here? https://github.com/jarun/nnn/wiki#prompts

The first thing to do when one plans to use a powerful file manager like `nnn` is read the documentation. We don't need to add 30 bytes for that. And someone will not be happy with your solution either. They will also ask for `Y`.

It would also save your effort if you leave a note in the ToDo list first when you propose to change stable behaviour.

marioortizmanero:
Are *30* bytes more that important? I get that nnn is minimal, but something as basic as a prompt needs feedback in one way or another, and 30 bytes IMO is well worth it... I shouldn't have to read the wiki to know that.

This change was quite small, so I didn't bother to leave a note in the ToDo list, but I was just writing a comment regarding something similar... You can read it if you like, but I'm guessing you won't support these changes either.

jarun:
> I shouldn't have to read the wiki to know that.

Not required either.

1. What are your options for a binary question like `rm -rf %s file%s?`? Common sense says `y` or `n`.
2. If you are not unsure you'll press `n`. And the first time you are past it, you'll know that the answer is either `y` or `n` for these questions.

Yes, 30 bytes is that important when you want to keep the binary size minimal. Because if the binary is padded and you cross a sector size, you'll take another sector to store it, not 1 or 2 bytes. Yes, that's a long go for `nnn` but I do not want to add redundant text to the binary.

jarun:
And why does it feel so alien anyway? It just follows the most basic utilities cp, mv and rm:

    $ rm nnn       
    rm: remove write-protected regular file 'nnn'?

I would understand if there are more than 2 choices but binary is the simplest case.

marioortizmanero:
> And why does it feel so alien anyway? It just follows the most basic utilities cp, mv and rm:

Yeah, you're right about that. But using `y/n` is also a very common practice in other utilities like pacman, vim, ffmpeg... Even if you think it's already intuitive, and I agree with you to a certain point, it might not be so for others. When someone first sees a prompt like that they might just type `yes` or `no` because they're used to programs that use that answer instead, or maybe they'll just press the return key because they think it's the default action. It's a small change, but I consider it a good practice to follow.

PR #682, opened 2020-07-20
Allow absolute paths returned by fzf in fzcd plugin

jmeischner:
I wanted to use fzcd to always start searching for paths from the HOME directory

```sh
 export FZF_DEFAULT_COMMAND="fd --type d . $HOME"
```

this returns absolute paths and I thought supporting this in the fzcd plugin could be helpful for other people as well.


jarun:
@leovilok please review.

jarun:
@jmeischner please fix the following lint issue CI is complaining about:

```
In plugins/fzcd line 23:
 
    if [[ "$sel" = /* ]]; then
 
       ^-- SC2039: In POSIX sh, [[ ]] is undefined.
```

plugins/fzcd
```diff
@@ -20,11 +20,16 @@ if [ -n "$sel" ]; then
...
 
-    # Remove "./" prefix if it exists
-    sel="${sel#./}"
-    if [ "$PWD" = "/" ]; then
-	    nnn_cd "/$sel"
-    else
-	    nnn_cd "$PWD/$sel"
-    fi
-fi
+    case $sel in
```
jarun, line 23: Please add notes to document this new behaviour. Please refer to other plugins for syntax.

jmeischner: in such a way, or should I document it somewhere else additionally?

jarun: Something like this:

https://github.com/jarun/nnn/blob/dacf7b08113afd6940bd70681ecb9b41737e3946/plugins/bulknew#L6

@leovilok can you explain what this change is trying to do and why? I am not very clear still.

leovilok:
@jarun `fzcd` get a path from `fzf`, and uses `nnn_cd` to go to that path. Currently, it only supports relative path, and with default configuration it's not an issue because `fzf` only returns relative path. But `fzf` can be configured with environment variables to work on a different set of path, which can lead to returning absolute path.

This patch solves this issue by supporting absolute path returned by `fzf`.
The behaviour or the plugin is the same when a relative path is given, so I don't think that documenting it is usefull, it just fixes a bug on an edge case.

Another aproach would be to support both relative and absolute path in either the `nnn_cd` function in `.nnn-plugin-helper` or directly in the `nnn` plugin control API. Currently `printf "0crelative/path" > "$NNN_PIPE"` doesn't work and can even crash `nnn`, but that's not well documented.

jarun:
OK. In that case please see if you need any further changes and approve when ready.

jarun:
@jmeischner I guess we can get rid of the notes and comments in that case.

plugins/fzcd
```diff
@@ -20,11 +26,20 @@ if [ -n "$sel" ]; then
...
+        # Remove "./" prefix if it exists
+        sel="${sel#./}"
+        
+        if [ "$PWD" = "/" ]; then
+            nnn_cd "/$sel"
+        else
+            nnn_cd "$PWD/$sel"
+        fi;;
+    esac    
+fi
```
leovilok, line 45: There should always be a newline at the end of the file, see [the reason](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

plugins/fzcd
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env sh
 
 # Description: Run fzf and go to the directory of the file selected
+# Note: by adding an environment variable called FZF_DEFAULT_COMMAND
+# it's possible to manipulate the default collection of items
+# piped to fzf. For Example:
+#   export FZF_DEFAULT_COMMAND="fd -H --type d . $HOME"
+# Then all paths piped to fzf are absolute and therefore
+# they need to be handled in such a way by this plugin.
```
leovilok, line 9: The note is not really needed, as your PR doesn't affect normal behaviour and only fixes a uncommon issue.

jmeischner:
Hej @leovilok , thanks for the additional detailed description. I also thought about changing the behavior either in the `nnn_cd` function or in `.nnn-plugin-helper` but I was not completely sure which else this would effect and wanted to start with small steps :-)

I tested it on _zsh/alacritty_ and with relative and absolute paths and it worked. But not with every other shell out there.

jarun:
All good, thank you!

PR #686, opened 2020-07-22
Remote name can contain the path

microo8:
If there is a ':' in the remote name, let it be. So there's a option to
input just the `remotename` (that will be transformed to `remotename:`), or
`remotename:/with/path`.


jarun:
@microo8 thanks for the PR. However, there are more things to be taken care of so I fixed it myself. Please test master.

PR #694, opened 2020-08-10
Fix inability to enter directories on an XFS filesystem on top of an md device

ucs1:
nnn was refusing to enter any directories located on an XFS filesystem sitting on top of an md device, and showing me the error "unknown" in the status line.

This fix consists in adding a check for DT_UNKNOWN. Apparently, even on filesystems supporting d_type, readdir(3) can sometimes return a d_type of DT_UNKNOWN for performance reasons. Programs looking at d_type should always handle this case.

Issue encountered and fixed on Debian buster (XFS v4.20, md v4.1) using Linux kernels 5.7.6 and 4.19.60.


jarun:
Thanks a lot for the PR!

jarun:
Thank you!

jarun:
@ucs1 would it be possible for you to refresh the Arch Linux and Wikipedia wiki pages?

PR #698, opened 2020-08-12
Adding support for `most` pager

Recidiviste:
`most` pager, by default, does not support control characters. So, the idea is to add a test after the set of PAGER to change it to `less -R` so control chars can be interpreted

First, test for PAGER, than bat, than change.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -64,7 +64,7 @@ TERMINAL="$TERMINAL"  # same goes for the terminal
 USE_SCOPE="${USE_SCOPE:-0}"
 USE_PISTOL="${USE_PISTOL:-0}"
 PAGER="${PAGER:-less -R}"
-exists bat && [ "$PAGER" = "most" ] && PAGER="less -R"
+[ "$PAGER" = "most" ] && exists bat && PAGER="less -R"
 
 if [ -e "${TMUX%%,*}" ] && tmux -V | grep -q '[ -][3456789]\.'; then
     TERMINAL=tmux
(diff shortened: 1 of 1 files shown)


jarun:

plugins/preview-tui
```diff
@@ -64,6 +64,7 @@ TERMINAL="$TERMINAL"  # same goes for the terminal
 USE_SCOPE="${USE_SCOPE:-0}"
 USE_PISTOL="${USE_PISTOL:-0}"
 PAGER="${PAGER:-less -R}"
+[ "$PAGER" = "most" ] && exists bat && PAGER="less -R"
```
leovilok, line 67: OK, so `most` doesn't accept ANSI colors, and should be replaced by `less -R`, this souds correct.

But you can't just checkt for `bat`, there are other previewers that output ANSI colors to a pager (exa, pistol and/or scope.sh output when available, ...).

I think that we should always replace `most` when it's set as PAGER, and add a comment saying that we only support ANSI color capable pagers. Actually we could even force `less -R` as PAGER instead of getting it from the system, and then let the user edit the script if he/she wants something else, with a warning about colors.

Recidiviste: Well, the test for `bat` was here only to change the pager when colored output is printed. So the user could use his PAGER with the other stuff.

Anyway, what you said sounds good, so I'll go with that. Forcing wouldn't be a great idea tho :thinking:, but, with a warning (tell me what kind of warning tho), that should do the job

jarun:
Thank you!

PR #705, opened 2020-08-18
nuke: use macOS open when GUI enabled

gpanders:
For non-macOS users, this has no effect (`[ "$uname" = Darwin ]` is false). For macOS users, this has the effect of using `open` as the primary command for images, videos, and PDFs. `open` uses the default opener for the given filetype at the OS level, meaning the program that is launched is still user-configurable through the OS.

### DIFF
--- a/plugins/fzopen
+++ b/plugins/fzopen
@@ -29,5 +29,10 @@ case "$(file -biL "$entry")" in
     *text*)
         "${VISUAL:-$EDITOR}" "$entry" ;;
     *)
-        xdg-open "$entry" >/dev/null 2>&1 ;;
+        if uname | grep -q "Darwin"; then
+            open "$entry" >/dev/null 2>&1
+        else
+            xdg-open "$entry" >/dev/null 2>&1
+        fi
+        ;;
 esac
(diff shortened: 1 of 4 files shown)


jarun:
Thank you! Would it be possible to audit the other plugins for macOS compatibility? It would be of great help!

@leovilok please review the changes.

gpanders:
I updated `fzopen` and `imgview` as well. Just FYI, the `imgview` plugin is currently not compatible with macOS's version of `find` and `grep`. Specifically, the `-iregex` used in `find` doesn't work with BSD `find` without the `-E` option (which isn't supported by GNU `find`), and there is no `-z` flag to BSD `grep`. We can get 90% of the way there by replacing the `find` command with:

```sh
    find -L "$(dirname "$target")" -maxdepth 1 -type f \( \
        -iname '*.jpeg' -o \
        -iname '*.jpg' -o \
        -iname '*.bmp' -o \
        -iname '*.webp' -o \
        -iname '*.png' -o \
        -iname '*.gif' \
        \) -print0 | sort -z
```

And replacing `grep -ZznF` with `grep --null -nF`, but without a way to ingest null-separated values in `grep`, I don't see how to proceed further.

Fortunately, this isn't really a problem because the `view_dir` function in the `imgview` plugin doesn't work with macOS `open`, so that code path shouldn't be traveled on macOS.

jarun:
Thank you so much!

PR #707, opened 2020-08-19
Haiku: Some small fixes to the recipe and resource file

CodeforEvolution:
Forgot to throw these changes in, should be all good still with updating over at Haiku Ports.


misc/haiku/nnn.rdef
```diff
@@ -1,15 +1,15 @@
...
+// Next, run "rc nnn.rdef -o nnn.rsrc" to compile this resource file.
+// Then, run "xres -o nnn nnn.rsrc" to apply the resource file to nnn.
+// Finally, run "mimeset -f nnn" to refresh Haiku's MIME database.
 
 resource app_signature "application/x-vnd.Jarun-nnn";
 
 resource app_version {
 	major  = 3,
-	middle = 4,
+	middle = 5,
```
jarun, line 12: What does this signify? We are nowhere near 3.5 release yet.

CodeforEvolution: I sort of thought that since we're past version 3.4, we are on to the "development version" of the next one (in this case, I assume the next version would be 3.5). I can revert this if needed however.

jarun: > I can revert this if needed however.

I think that's would be good. We have the `B_APPV_DEVELOPMENT` in place already. How many ways are we going to convey that it's not a released version? An user checking out code and compiling is probably aware of that already. And it's also a burden on maintenance. Already I have added a point in release checklist to update the rdef file as dev after each update.

Also, when `B_APPV_DEVELOPMENT` is set, I would like to think of the version as the base on top of which new development has happened (but not a new release).

jarun: Took care of it.

CodeforEvolution:
Sorry I didn’t get back sooner about that, thank you!

jarun:
No problem! We just made a release. Chill! ;)

PR #708, opened 2020-08-19
Add support for Alexey Tourbin's QSORT code

leovilok:
See https://github.com/svpv/qsort

No benchmark yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,10 @@ ifeq ($(O_NO_TOURBIN_QSORT),1)
 	CPPFLAGS += -DNO_TOURBIN_QSORT
 endif
 
+ifeq ($(O_BENCHMARK),1)
+	CPPFLAGS += -DBENCH
+endif
+
 ifeq ($(shell $(PKG_CONFIG) ncursesw && echo 1),1)
 	CFLAGS_CURSES ?= $(shell $(PKG_CONFIG) --cflags ncursesw)
 	LDLIBS_CURSES ?= $(shell $(PKG_CONFIG) --libs   ncursesw)
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,8 @@ O_NOBATCH := 0  # no built-in batch renamer
 O_NOFIFO := 0  # no FIFO previewer support
 O_CTX8 := 0  # enable 8 contexts
 O_ICONS := 0  # support icons
+O_QSORT := 0  # use Alexey Tourbin's QSORT implementation
+O_BENCH := 0  # benchmark mode (stops at first user input)
 
 # convert targets to flags for backwards compatibility
 ifneq ($(filter debug,$(MAKECMDGOALS)),)
@@ -75,11 +77,11 @@ ifeq ($(O_ICONS),1)
 	CPPFLAGS += -DICONS
 endif
 
-ifeq ($(O_NO_TOURBIN_QSORT),1)
-	CPPFLAGS += -DNO_TOURBIN_QSORT
+ifeq ($(O_QSORT),1)
+	CPPFLAGS += -DTOURBIN_QSORT
 endif
 
-ifeq ($(O_BENCHMARK),1)
+ifeq ($(O_BENCH),1)
 	CPPFLAGS += -DBENCH
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,10 @@ ifeq ($(O_ICONS),1)
 	CPPFLAGS += -DICONS
 endif
 
+ifeq ($(O_NO_TOURBIN_QSORT),1)
+	CPPFLAGS += -DNO_TOURBIN_QSORT
+endif
+
 ifeq ($(shell $(PKG_CONFIG) ncursesw && echo 1),1)
 	CFLAGS_CURSES ?= $(shell $(PKG_CONFIG) --cflags ncursesw)
 	LDLIBS_CURSES ?= $(shell $(PKG_CONFIG) --libs   ncursesw)
--- a/misc/haiku/Makefile
+++ b/misc/haiku/Makefile
@@ -18,6 +18,8 @@ O_NOBATCH := 0  # no built-in batch renamer
 O_NOFIFO := 0  # no FIFO previewer support
 O_CTX8 := 0  # enable 8 contexts
 O_ICONS := 0  # support icons
+O_QSORT := 0  # use Alexey Tourbin's QSORT implementation
+O_BENCH := 0  # benchmark mode (stops at first user input)
 
 # convert targets to flags for backwards compatibility
 ifneq ($(filter debug,$(MAKECMDGOALS)),)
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -75,6 +75,10 @@ ifeq ($(O_ICONS),1)
 	CPPFLAGS += -DICONS
 endif
 
+ifeq ($(O_NO_TOURBIN_QSORT),1)
+	CPPFLAGS += -DNO_TOURBIN_QSORT
```
jarun, line 79: Can we have something smaller? Also this should go in:

- initialize at line 22
- Haiku Makefile

jarun (changes requested):

jarun:
Travis complains _mostly_ about macro parenthesis. Please fix those.

Makefile
jarun: Also, this increases the binary size as @KlzXS suggested. Let's have it as compile-in.

jarun:
Benchmark details from @leovilok for the record. It's faster than stdlib at the cost of increasing the binary size by 4KB.

It turns out that `populate()` was already instrumented in DEBUG mode, including both `dentfill()` and `qsort()` calls. So as an early result, we can already tell that Tourbin's qsort **is** faster:
(here I'm manually starting nnn in a test dir filled whith 100000 empty files with random names)
~~~
#### TOURBIN QSORT ####
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=3656519433
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=354925821
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=3652997582
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=3657911883
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=347739563
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=347168855
#### STDLIB QSORT ####
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=400908271
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=3676649821
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=378532354
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=382453839
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=397564860
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=379860967
[leo@franky tmp]$ grep nsec nnndbg 
ln 5141: ts2.tv_nsec - ts1.tv_nsec=3697438251
~~~

leovilok:
I've been trying to get more in-depth benchmarks, so I ran everything in valgrind (callgrind mode).
Here is basically what I got:

## Tourbin's qsort version

Total callgrind "cost": 685 487 119
Calls to `entrycmp`: 1 859 344
Screenshot (excuse the french UI):
tourbin-qsort

## stdlib's qsort

Total callgrind "cost": 582 280 595
Calls to `entrycmp`: 1 536 571
Screenshot:
std-qsort

## Discussion

So in this directory with 100000 empty files generated by `misc/test/genfiles.sh`, when quitting immediately, sorting the input file entries is really the big performance bottleneck in a `nnn` execution (more than 90% of the "cost" in both cases).
Of course valgrind is a slow VM, and doesn't allways reflect the real performmances, especially when sys calls are in the mix, but we can still count loops and function calls, and compare both qsort implementations.

Tourbin's QSORT, being implemented in macros, is not visible in the call graph. But we can count calls to `entrycmp`, much higher than in stdlib's qsort implementation, as expected considering that's what happens in Tourbin's benchmarks too.

Now this makes the overall callgrind "cost" of the Tourbin version higher, but the bare metal benchmarks are more important than what happens in Valgrind VM.

leovilok:
I forgot the callgrind traces, if you want to explore them by yourself with kcachegrind.
[callgrind-traces.zip](https://github.com/jarun/nnn/files/5098321/callgrind-traces.zip)

jarun:
Thanks a lot for the in-depth details!

Makefile
jarun: Maybe just use `O_QSORT` in Makefile and `TOURBIN_QSORT` in code.

leovilok:
So here is a more "real world" benchmark, with the unix sh builtin `time` (and with `nextsel()` always returning `SEL_QUIT`, so it's the "time to first user input"):
qsort-bench

 So there is no clear winner for `/`, or even `/usr/lib/` (which I think is one of the most crowded real directory on my system), but the new `QSORT` wins on (unrealistically?) entry heavy directory.

But the difference is barely noticeable on my system, it might be worth benchmarking on lower end hardware, where it could make a real difference. I might publish the benchmark code for that.

So do you think this QSORT is worth including?

jarun:
> it might be worth benchmarking on lower end hardware

I concur. Please run the benchmarks on a Pi or mid-range Android.

> So do you think this QSORT is worth including?

Of course! Please update the PR.

jarun:
> I might publish the benchmark code for that.

Please do so and I can run on the Pi 4 4GB version I have.

0xACE:
@leovilok that's a sexy diagram, how did you make it?

leovilok:
I just pushed some benchmark scripts & compilation mode, I'll really update the PR tomorrow.

@0xACE I wanted to learn some matplotlib, but in the end I only used it to plot the violin graph and the vertical axis, and I did colors, spacing, and adding labels in Inkscape :stuck_out_tongue:

jarun:
No problem! Please take your time.

Makefile
```diff
@@ -75,6 +75,14 @@ ifeq ($(O_ICONS),1)
...
+endif
+
+ifeq ($(O_BENCHMARK),1)
```
jarun, line 82: `O_BENCH`?

leovilok: OK

Makefile
leovilok: All done.

leovilok:
I don't own a Pi or a similar small computer, and I don't have a very old box running currently, so I can't really benchmark on ressource constrained devices. But if someone wants to try my benchmark scripts, I'm available for any help.

jarun:
I have one. I'll run it sometime.

jarun:
Thank you!

PR #714, opened 2020-08-28
Small plugin for jump. Also adds check for NNN_PIPE to autojump plugin

snide:
Hello! Thank you for this fantastic little program. I like using [jump](https://github.com/gsamokovarov/jump), which is very similar to auto-jump. Figured I'd add a small plugin for it.

Here is a small mp4 showing it working (exactly the same as autojump). I'm primarily a designer, so yell if I missed something!

https://snid.es/2020AUG/GlygMMHKZTW9hBrU.mp4

### DIFF
--- a/plugins/autojump
+++ b/plugins/autojump
@@ -9,6 +9,11 @@
 # Shell: POSIX compliant
 # Author: Marty Buchaus
 
+if [ -z "$NNN_PIPE" ]; then
+  echo 'ERROR: NNN_PIPE is not set' | ${PAGER:-less}
+  exit 2
+fi
+
 if which autojump >/dev/null 2>&1; then
     printf "jump to: "
     read -r dir
--- a/plugins/autojump
+++ b/plugins/autojump
@@ -10,8 +10,8 @@
 # Author: Marty Buchaus
 
 if [ -z "$NNN_PIPE" ]; then
-  echo 'ERROR: NNN_PIPE is not set' | ${PAGER:-less}
-  exit 2
+    echo 'ERROR: NNN_PIPE is not set' | ${PAGER:-less}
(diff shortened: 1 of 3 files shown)


KlzXS (approved):
@jarun we should add the `NNN_PIPE` check in `autojump` plugin as well.

snide:
@KlzXS Added it for you.

plugins/autojump
```diff
@@ -9,6 +9,11 @@
 # Shell: POSIX compliant
 # Author: Marty Buchaus
 
+if [ -z "$NNN_PIPE" ]; then
+  echo 'ERROR: NNN_PIPE is not set' | ${PAGER:-less}
```
jarun, line 13: I think the indent is 2 chars, we use 4. Can you please update it in both the plugin changes?

jarun (changes requested):

plugins/autojump
snide: Absolutely! My bad.

jarun: No problem! You got most of the things right! Even I miss sometimes. ;)

snide: Done.

jarun:

jarun:
I missed something... why not just merge these 2 plugins into one?

Keep jump as default and autojump as fallback.

jarun:
We do combine near-similar-functionality utilities in a single plugin.

jarun:
@snide you can follow the `fzopen` plugin to combine these 2 into a single one.

jarun:
Took care of it.

snide:
@jarun Ah. Thanks. That makes sense. Good to know for next time!

Issue #375, opened 2019-11-04
"which" vs "command -v"


0xACE:
@eugene55 [wrote](https://github.com/jarun/nnn/pull/374#issuecomment-549212666_)
>why not use command -v?
>Look at description here: https://github.com/koalaman/shellcheck/wiki/SC2230

Throughout the source we use `which`, and he proposes we use `command -v`.

Rather than just changing that specific plugin I figured I'd ask what we do with the entire codebase as it's probably better to be consistent.

I have previously run into this discussion and from what I remember no one had a universal solution but the text he is citing makes it look like the proper solution...

jarun:
I agree :+1: .

jarun:
A few points to add:

- `command` looks like a shell built-in, which means we need to invoke the shell in `nnn.c` which will do this. I think `which` will be much faster. @eugene55 can you please do a profiling with 10 random files in `/usr/local/bin` and share with us?
- I can't think of any OS without `which`. It's in *BSD, [Solaris](https://docs.oracle.com/cd/E19253-01/816-5165/6mbb0ma0m/index.html) and *nix. Probably the next version of POSIX should include it.

For now, I would suggest we update our scripts with `command -v` and leave the source code with `which`.

annagrram:
May I ask what was the problem with `which`?

jarun:
POSIX compliance: https://docs.oracle.com/cd/E19253-01/816-5165/6mbb0ma0m/index.html

eugene55:
Speed of shell spawning depends on actual shell implementation. in my case: 
```
$ time for run in {1..1000}; do which vi > /dev/null; done

real	0m3.360s
user	0m1.322s
sys	0m2.136s
$ time for run in {1..1000}; do bash -c 'command -v $0 > /dev/null' vi; done

real	0m11.182s
user	0m7.055s
sys	0m3.963s
$ time for run in {1..1000}; do dash -c 'command -v $0 > /dev/null' vi; done

real	0m3.301s
user	0m1.095s
sys	0m2.330s
```
I agree with suggestion to change just scripts, leaving source code as is

jarun:
The larger problem I see is the link says `which` is _non-standard_.

Why??? @koalaman did you audit the code and notice any non-compliance?

If not, then by your logic every utility that doesn't appear in the POSIX docs is _non-standard_. Utilities are added in open source everyday. They can be POSIX-compliant or not. But I don't think it's correct to term them _non-standard_.

[...]

PR #715, opened 2020-08-29
Add comment for imv command in imgview plugin. Check with which instead of version

snide:
Adds a comment to mention that `imvr` is often `imv` in certain linux distros. Also makes all the checks use `which` instead of `command -v` which can also vary from system to system.

### DIFF
--- a/plugins/imgview
+++ b/plugins/imgview
@@ -43,22 +43,24 @@ if uname | grep -q "Darwin"; then
     if [ -f "$1" ]; then
         open "$1" >/dev/null 2>&1 &
     fi
-elif command -v imvr >/dev/null 2>&1; then
+# `imvr` is often callable as `imv` on Linux distros
+# You might need to change the reference below
+elif which imvr >/dev/null 2>&1; then
     if [ -f "$1" ]; then
         view_dir imvr "$1" >/dev/null 2>&1 &
     elif [ -d "$1" ] || [ -h "$1" ]; then
         imvr "$1" >/dev/null 2>&1 &
     fi
-elif command -v sxiv >/dev/null 2>&1; then
+elif which sxiv >/dev/null 2>&1; then
     if [ -f "$1" ]; then
         view_dir sxiv "$1" >/dev/null 2>&1 &
     elif [ -d "$1" ] || [ -h "$1" ]; then
         sxiv -qt "$1" >/dev/null 2>&1 &
     fi
-elif command -v viu >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


jarun:
On Ubuntu:

```
1 ~$ which imv           
imv not found
1 ~$ which imvr
/usr/bin/imvr
1 ~$ dpkg -S /usr/bin/imvr                               
imv: /usr/bin/imvr
```

But I can understand that's an issue with Ubuntu packaging. I agree the program should be installed as imv.

Can we check for both to get rid of this issue? Look for imv first and then imvr.

jarun:
Little unrelated to this. But can we replace instances of `command -v` with `which in all the plugins? (Probably the next PR).

We have a detailed discussion in #375. I would really appreciate if you make the change.

leovilok:
The issue with the `imv` binary name is that it conflicts with the tool `imv` (interactive move) from `renameutils`. I think that renameutils might be older, so some distros might have changed `imv`'s executable name to avoid conflicts. I think that in Arch both packages are simply conflicting with each other.

jarun:
Yes, I suspected that.

What do you propose? We just live with what we have right now?

leovilok:
There are a few option:
* keep what we have, maybe with a comment explaining the issue
* check the binary, the thing is that they don't even have common options to check versions:
  here is the first line of renameutils `imv --version` (you can't use `-v`):
```
imv (renameutils) 0.12.0
```
   but with `imv` you need to use `-v` (not `--version`), and you get:
```
Version: v4.1.0
```
There is a way to tell them appart using the return value:

| command | renameutils' `imv` | `imv` image viewer |
|-----|-----|-----|
| `imv -v` | 1 | 1 |
| `imv --version` | 0 | 1 |

So we can use this logic:

```sh
imv --version 2>/dev/null 1>&2 && echo renameutils || echo image viewer
```

But is it worth it?

jarun:
> But is it worth it?

No. We can just have what we have now and live with it.

@snide please update the note on this and we can conclude.

Also, please confirm if you are planning to work on the `command -v` removal.

snide:
Cool. I'll give it a shot today. Thanks for the detailed explanation.

snide:
Updated code and comment based upon the discussion in this thread. Again, I'm a designer, so hopefully I tracked it right. Take that as a compliment, the code is readible :D

snide:
Updating the other plugins now too.

jarun:
Thank you!

Issue #717, opened 2020-08-30
Better coloring of icons / add support for nerdfont


snide:
Any interest if I take a stab at coloring the various extension icons?

My method would be using the primary color associated with that lang. I personally find this helps me spot files a lot quicker and it's what I'm using to seeing in other vim explorers, so having some built in defaults would be nice.

Examples of things I know as a web dev

- TSX, JSX React files being their teal
- JS / Node files being green
- Sass / Ruby files being pink / red

Don't know if I can do something similar to check for dot files? Usually I've seen them somewhat muted out.

I'm sure this can be fairly subjective, but doing a local compile I noticed how many icons were defined in `icons-in-terminal.h` and not in `icons.h`

Totally understand if this isn't wanted, and might be considered out of scope. I'm happy to just fill in the gaps of missing extensions and not apply color if that's the case (for example... scss is defined in `icons-in-terminal.h` but not `icons.h`).

Lemme know, happy to do the gruntwork over a day or two if folks would find it useful.

![image](https://user-images.githubusercontent.com/324519/91649447-2bf11500-ea29-11ea-8a6f-5513b986d834.png)

snide:
Exploring this a bit more, I'm realizing I'd probably want to use my own patched Nerd font instead (which is more consistent in styling then the mixed-bag-on-purpose icons-in-terminal lib). Trying to think of a way to make this more friendly for customization without having to edit the source file directly.

I think to make this easier to theme, you might want to separate this into three files

- icon-colors
- icon-definitions
- icon (which imports the two above)

That way a user can edit the colors and definitions separately without worrying about touching the main compile file. I might also suggest deduplicating the `icons-in-terminal` definitions file. It has so many variants (three for React), which would make it hard to touch it separately from the `icons` file which right now calls a mix of variants. Using a single batch of their sources and normalizing the names might be helpful (then you could use the single definition and apply one of the three -- or your own patch -- you prefer).

These are just suggestions from a habitual themer. The way it's set up currently I can certainly patch my own font in easily. I mostly worry about upgrades. Setting up the icon defs and colors separately would help there.

![image](https://user-images.githubusercontent.com/324519/91650039-4c23d280-ea2f-11ea-80c6-ae27a198e7dc.png)

jarun:
> want to use my own patched Nerd font

So you'll have to set Nerd Font as your terminal/system Monospace font?

@KlzXS for his input.

[...]

PR #722, opened 2020-09-06
Add nerdfont defs file and normalize vars

snide:
Fixes https://github.com/jarun/nnn/issues/717

This is a draft PR that I'd appreciate some testing on. I made the following changes:

* I normalized the variables in `icons.h` to match `ICON_` and `COLOR_` prefixing that don't reference the icon sets directly. I similarly tried to clean up var naming where i could (ex: all extension vars are now `ICON_EXT_` ...etc)
* I added a `icons-nerdfont.h` file with definitions that match.
* I removed all the unused icons in `icons-in-terminal.h` now that there's some normalization. Is there a place similar to https://www.nerdfonts.com/cheat-sheet where you can visually search their icons? If not, it may be worth keeping the defs in the file if only for reference.
* I added colors for all the major language extensions based on the colors from their logos.
* I added a couple missing extensions (sass, scss...etc)

### What this means

* To add a new extension. You would add the var in the three icon files, referencing as needed.
* When compiled NNN, you simply comment out / comment in the icon pack you like and everything should just work.

### Other notes

* Because I assume they were made with intent, I did not change the icon-in-terminal icon references. This means they should remain largely unchanged from before.
* For the nerdfont icon choices I tried to stick to a single icon set when possible to normalize the spacing, style and size. This meant using the `mdi` set as often as possible.
* To the above, I could normalize the actual icons being used between the two sets if we want. Both "packs" tend to use a lot of the same icons. I personally find the choices in the nerdfont ones superior, but I didn't want to be pushy.
* The icon-in-terminal fonts look bigger. I think this is due to them not being matched against the font being used in the terminal. Don't know if there's anything to fix there, just calling it out. The sizing for nerdfont looks "correct" on my system.

@jarun or one of the maintainers, if you could give me a generaly :+1: on the methods I used, I'll go ahead and write up some docs while we do some testing.

### TODO

- [ ] Write documentation once the basics of this PR are thought to be OK

### Icons in terminal

image

image

### Nerdfont

image

image

---

# Wiki content / testing

## File icons

`nnn` supports mutliple ways to show icons. They are turned off by default and are not provided in the default distribution to avoid shipping third party resources. Currently we support [icons-in-terminal](https://github.com/sebastiencs/icons-in-terminal) for those wishing to use icons alongside a normal code font. Alternatively, you can utilize Nerdfont's patched fonts to add icons to the font itself. The later is more common for Vim users who need fonts for various powerline plugins. Regardless of the method, each method will produce similar experiences within nnn. You will need to recompile nnn rather than use the default release in both cases.

### To enable icons-in-terminal icons

- [Install](https://github.com/sebastiencs/icons-in-terminal#installation) icons-in-terminal
- Clone the nnn repo
- Edit `src/icons.h` and make sure to include the `icons-in-terminal.h` file at the top
- Compile `nnn` with `make O_ICONS=1` from the root

### To enable Nerdfont icons

- [Download and install](https://www.nerdfonts.com/) a patched Nerdfont.
- Apply that font as your terminal emulator's font. This will vary from emulator to emulator, but usually involves editing a config file or changing a setting within a GUI menu system.
- Edit `src/icons.h` and make sure to include the `icons-nerdfont.h` file at the top
- Compile `nnn` with `make O_ICONS=1` from the root

### Move the newly compiled nnn to where your programs normally go

Once compiled, you'll notice an `nnn` executable is now in your root. Copy this file to wherever such programs are normally stored. A good way to check is to do a `which nnn` on where the packaged release was installed. Simply replacing that file should be enough.


src/icons.h
```diff
@@ -1,11 +1,19 @@
-#include "icons-in-terminal.h"
+/*
+ * This file loads in variables either from icons-in-terminal.h or nerdfont.h.
+ * Only include one and not the other.
+ * If you are a Vim user you are likely already using a patched font and Nerdfont will be easier
+*/
+
+// #include "icons-in-terminal.h"
+#include "icons-nerdfont.h"
```
snide, line 8: Right now you just comment one out one of the other. It'd be nice if this was maybe done through some sort of build flag instead? I don't know enough about the build system to set that up.

KlzXS (changes requested):
I took a quick look at the changes, noted some spacing mismatches.
I'll give it a closer look and chack that everything works later today.

@jarun do you think we should have another file like `misc/test/genfiles.sh`, but for icons? Generate an empty file for each rule.

src/icons.h
```diff
@@ -60,201 +80,203 @@ static const struct icon_pair icons_name[] = {
...
-	{"scala",    MFIZZ_SCALA,          0},
-	{"sh",       MFIZZ_SCRIPT,         0},
-	{"slim",     FA_FILE_CODE_O,       0},
-	{"sln",      DEV_VISUALSTUDIO,     0},
-	{"sql",      MFIZZ_MYSQL,          0},
-	{"srt",      FA_COMMENTS_O,        0},
-	{"sub",      FA_COMMENTS_O,        0},
-	{"svg",      FA_FILE_IMAGE_O,      IMAGE_COLOR},
+  {"sass",       ICON_EXT_SASS,      COLOR_CSS},
+  {"scss",       ICON_EXT_SCSS,      COLOR_CSS},
```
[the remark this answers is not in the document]
snide: I don't write in C files... well... ever. Is there something missing in a spaces vs tabs thing here? It seems to align correctly for me in my vim, but it could be a local setting I have.

image

src/icons-nerdfont.h
```diff
@@ -0,0 +1,245 @@
+#ifndef ICONS_NERDFONT
```
jarun, line 1: Please use

    #ifndef NERD

and make option `O_NERD`.

jarun (changes requested):

jarun:
@KlzXS can you please help @snide with the guideline?

@snide please let me know how to test this. Please start from how to install the patched nerd etc. (same would go to docs as well). I am on Xubuntu 20.04.

src/icons.h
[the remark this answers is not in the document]
KlzXS: It seems so. You used 2 spaces while we use tabs.

snide:
@jarun Testing instructions are now in the description at the bottom. Scratching my head trying to figure out how to make a pull request against the `nnn.wiki` repository. Github seems to make this unnecessarily complex, but that's the text I would add.

0xACE:
@snide, the wiki page is edited by the button on the top right:

cap

Regarding projects and indentation, I have been using this in vim and never bothered with it again: https://github.com/tpope/vim-sleuth

src/icons-in-terminal.h
```diff
@@ -1,3736 +1,245 @@
...
-# define FIRACODE_NAMEME_1114119 "\ueea0"
+// Hex codes reference icons-in-terminal
+// https://github.com/sebastiencs/icons-in-terminal
 
-#endif // ICONS_IN_TERMINAL
+// Generics
+#define ICON_DIRECTORY     "\ue155"
+#define ICON_FILE          "\ue0f6"
+#define ICON_EXEC          "\ue0f3"
+#define ICON_MANUAL        "\uf5bd"
```
KlzXS, line 11: Did you maybe forget to change this codepoint? icons-in-terminal doesn't map anything to it. And to me it looks like a spray bottle?

jarun:
> Scratching my head trying to figure out how to make a pull request against the nnn.wiki repository

Please edit directly when the PR is merged.

KlzXS:
General notes:
- why do libraries (`a`, `so`, `dll`), ebooks (`epub`), object files (`o`) and manpages (`1`) all use the same icon?
- `txt` files used to have a separate icon from general file icon
- `doc` and `docx` documents no longer have word icon
- `cue` files are text metadata (might be misleading to use audio icon)

Nerdfonts ():
- Now it might be just me but for certain icons there is an extra space between icon and filename
2020-09-08-002618_323x138_escrotum
Notice how `nnn*` is shifted one space over. I got Sauce Code Pro, but family shouldn't matter, right?
- `iso` and `sln` icons are missing for me. Displayed as only space.
- choice for markdown is a bit weird for my taste, but if that's what's commonly used...

Aside from these few issues I'm having with nerdfonts everything is good.

jarun:
@snide can you please address the comments from @KlzXS? I am planning to include this in the new release.

snide:
Will do. I'll get to it tomorrow.

jarun:
Thank you!

snide:
Feedback addressed. @jarun I have no idea how the make files work. I tried making a `O_NERD` setting, but couldn't really figure it out. If you don't mind applying a commit on top, I'd appreciate it, then I can learn how it works. For now, it has nerdfont/icons-terminals as a comment/uncommon at the top of `icons.h`. Again, feel free to edit, I have no idea what's best here.

---

@KlzXS 

> * why do libraries (`a`, `so`, `dll`), ebooks (`epub`), object files (`o`) and manpages (`1`) all use the same icon?

This was the same as the `icons-in-terminal` settings. I changed the dll one. The rest likely need some consideration. Rather leave this for another PR since it isn't a regression and replicates what existed.

> * `txt` files used to have a separate icon from general file icon

✅

> * `doc` and `docx` documents no longer have word icon

✅

> * `cue` files are text metadata (might be misleading to use audio icon)

✅ Made this a playlist for both sets. FWIW, it was like this in the old set as well, but is an easy enough change.

>   Notice how `nnn*` is shifted one space over. I got Sauce Code Pro, but family shouldn't matter, right?

✅ This was due to an icon issue. I switched the icon.

> * `iso` and `sln` icons are missing for me. Displayed as only space.

✅ Fixed by changing the icon. Think this is an upstream issue (possibly a hex or docs error), but found it hard to diagnose. It could also be that we have older versions of the font installed.

> * choice for markdown is a bit weird for my taste, but if that's what's commonly used...

For small icons that try to deliver a lot, I go with readability / bigger glyphs.

snide:
Also I'll cut a video whenever you clean this up so that I can document how to compile locally with icons. Figure that's gonna be something a lot of people want to do.

jarun:
> I tried making a O_NERD setting, but couldn't really figure it out.

No problem! We'll take care of it.

> Also I'll cut a video whenever you clean this up so that I can document how to compile locally with icons.

Sure thing!

Can you please update the docs [here](https://github.com/jarun/nnn/wiki/Advanced-use-cases#file-icons) on how to install and use patched NERD fonts so this works? This is _a must_ for people like me who just want to follow the instructions and get the icons working.

jarun:
Thank you!

jarun:
@KlzXS there's a big problem. As I see now, this PR actually replaces all icons in terminal icons in icons.h with nerd icons. So once this is in, and in my case I have only icons in terminal, I only see unicode squares.

To resolve this, we need to refactor the code. I definitely wanted a single file that can understand both the definitions and pick up the right icons.

I will clean this up. What do you think is the right direction? Having both the icons separate or allow the usage together?

snide:
@jarun the instructions are updated.

I think the definition separation is fine. I don't see anyone ever wanting to use both of them together. If possible I'd make a separate compile command for each and it would just pull the correct file. Sorry, that was what I was mentioning with the make file. I didn't think you'd merge before you added that portion.

jarun:
> I think the definition separation is fine.

OK. I will proceed accordingly. Thanks!

jarun:
@snide support for both icons-in-terminal and icons-nerdfont is enabled at commit 1fc1b8be8b6cd5d43f7443486dc97cce133f9053.

Please confirm it works as expected and update the documentation in Wiki.

Note:

To compile in icons-nerdfont, run:

    sudo make O_NERD=1 strip install

jarun:
I see an issue:

```
2020-01-20 07:24  664   30.77M  ﳜ 50th Anniversary of the Moon Landing (English).mkv
2019-07-20 08:03  664    23.2K   50th Anniversary of the Moon Landing (English).vtt
```

For the video, the icon is after 50.

0xACE:
To me it looks like left to right got shifted in that message... That icon maybe some multibyte special case that your terminal is interpreting differently... Rendering text is a nasty business...

KlzXS:
I second the right to left thing. For me it displays some arabic, I don't have icons in the browser.

jarun:
Guys, if you have a patch, PR... if you think it's OK, I'll live with it till some user reports it.

jarun:
@snide did you get some time to work on the plugins video? An introductory video with some plugin usage examples would greatly benefit new `nnn` users.

PR #731, opened 2020-09-14
make sxiv automatically play animations

tonijarjour:
Not sure how to make this work on the file check, but for the directory just add the -a flag.


jarun:
Thank you!

jarun:
Plugins are indicative. If sxiv is your favourite utility, feel free to hack the plugin and enable it for directories.

PR #734, opened 2020-09-18
preview-tabbed will now exit with ^C

cdarkly:
preview-tabbed plugin script will now exit when ^C is pressed in nnn.

### DIFF
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -103,7 +103,6 @@ kill_viewer () {
 sigint_kill () {
 	kill_viewer
 	kill "$TABBEDPID"
-	disown
 	exit 0
 }
 
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -100,6 +100,13 @@ kill_viewer () {
         fi
 }
 
+sigint_kill () {
+	kill_viewer
+	kill "$TABBEDPID"
+	disown
+	exit 0
+}
+
 previewer_loop () {
     unset -v NNN_FIFO
     # mute from now
@@ -108,6 +115,7 @@ previewer_loop () {
     MAINWINDOW="$(xdotool getactivewindow)"
 
     start_tabbed
+    trap sigint_kill SIGINT
 
     xdotool windowactivate "$MAINWINDOW"
 


plugins/preview-tabbed
```diff
@@ -100,6 +100,13 @@ kill_viewer () {
         fi
 }
 
+sigint_kill () {
+	kill_viewer
+	kill "$TABBEDPID"
+	disown
```
leovilok, line 106: Why `disown` here?

cdarkly: I was simply trying to match what happens during a normal exit path, as that is the last command listed in the preview-tabbed script, before it exits during a normal nnn quit.

Maybe running "disown -a" would be a better choice, to stop **all** jobs incase something was missed by the kill commands?

I'm assuming there was a reason for including it at the end of the original script.

cdarkly: I played a bit with the disown command in bash, and I guess it does not stop or terminate any suspended or running jobs. It simply removes them from the table of active jobs for a given parent shell.

So at this point I'm not really sure why it's included in the original script. If all child processes are killed, is there a scenario that you've seen where something is lingering and must be disowned?

leovilok: In the original script, the process to be disowned is main loop:
```bash
previewer_loop < "$NNN_FIFO" &
```
It's spawned in background mode because we ant the script to exit quickly to give the hand back to `nnn` itself (`nnn` waits for plugin completion). I just added `disown` so that when the plugin exits, it doesn't kills the backgrounded main loop. I'm not entirely sure it's required though.

Anyways, `disown` is in theexit path of the plugin script, but not of the plugin's main loop, so it shouild not be required in the sigint handler.

jarun:
@cdarkly please update as discussed and I can close this one.

jarun:
Thank you!

plugins/preview-tabbed
cdarkly: I tried running preview-tabbed without disown, and it seemed to work fine.

After some reading, whether or not disown is needed depends on whether or not the bash shell will send a SIGHUP when it exits.

Bash has an option called "huponexit", and searching for that online leads to some discussions that indicate it is not turned on by default for interactive shells.

So maybe it's best to leave it there. I'm not really sure.

But I agree that there is no need for it when exiting with SIGINT. What you explained makes sense now. Thank you.

PR #742, opened 2020-09-28
Move selection marker in detail mode next to filename

### LINKED ISSUE #741: Change selection mark position with detail mode
Is there an option to move the selection mark to the front of the entry name with detail mode? I think it is clearer which entry has been selected.
nnn_ups

KlzXS:
Fixes #741 

It really is easier to see which file is selected when the marker is next to the filename (and icon). Also keeps the placement consistent next to filename.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3789,9 +3789,9 @@ static void printent_long(const struct entry *ent, uint namecols, bool sel)
 		if (pair && fcolors[pair])
 			attrs |= COLOR_PAIR(pair);
 
+		attroff(attrs);
 
 #ifdef ICONS_ENABLED
-		attroff(attrs);
 		addstr("  ");
 #endif
 
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3680,7 +3680,8 @@ static void printent_long(const struct entry *ent, uint namecols, bool sel)
 	/* Directories are always shown on top */
 	resetdircolor(ent->flags);
 
-	addch((ent->flags & FILE_SELECTED) ? '+' : ' ');
+	/* Padding from the edge, ex selection marker */
+	addch(' ');
 
 	if (attrs)
 		attron(attrs);
@@ -3787,9 +3788,16 @@ static void printent_long(const struct entry *ent, uint namecols, bool sel)
 
 		if (pair && fcolors[pair])
 			attrs |= COLOR_PAIR(pair);
+
+
 #ifdef ICONS_ENABLED
 		attroff(attrs);
 		addstr("  ");
+#endif
+
+		addch((ent->flags & FILE_SELECTED) ? '+' : ' ');
+
+#ifdef ICONS_ENABLED
 		if (sel)
 			attrs &= ~A_REVERSE;
 		print_icon(ent, attrs);


KlzXS:
We should probably consider merging (in some way) the `printent...()` functions to remove the need to apply the same patch in two places. After all doesn't `printent_long()` just print some info before the stuff that `printent()` prints?

src/nnn.c
```diff
@@ -3787,9 +3788,16 @@ static void printent_long(const struct entry *ent, uint namecols, bool sel)
 
 		if (pair && fcolors[pair])
 			attrs |= COLOR_PAIR(pair);
-#ifdef ICONS_ENABLED
+
 		attroff(attrs);
+
+#ifdef ICONS_ENABLED
 		addstr("  ");
```
jarun, line 3795: I don't get it. if icons are enabled, we show the `+` after 2 spaces? Wouldn't it be one space followed by `+`?

jarun (changes requested):

jarun:
> just print some info before the stuff that printent() prints?

The long version looks far too complicated and then there's an additional check required per file. Now we just adjust the pointer and we are done.

src/nnn.c
KlzXS: I don't know, I've been wondering myself what all those two space prints were.

KlzXS:
I don't see what's stopping us from gathering the info for the file, printing it, and then telling `printent()` to print selection marker, icon and filename after that. For each file.

But then again I difn't dive too deep into understanding everything in there.

0xACE:
> I don't see what's stopping us from gathering the info for the file, printing it, and then telling `printent()` to print selection marker, icon and filename after that. For each file.
> 
> But then again I difn't dive too deep into understanding everything in there.

The code could be reconstructed handle it like you want. I use this personally to print a column indicating relative line numbers:

cap

Though it would require reconstruction of `printent()` and `printent_long()` I could look whenever i find the time to solve my merge conflicts with `master`

jarun:
> I don't see what's stopping us from gathering the info for the file, printing it

This would still need checking detail mode flag per file.

jarun:
@KlzXS the correct patch would be:

```diff
diff --git a/src/nnn.c b/src/nnn.c
index 82a80eb..f67d992 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3676,11 +3676,15 @@ static void printent_long(const struct entry *ent, uint namecols, bool sel)
                        : (g_state.oldcolor ? A_DIM : COLOR_PAIR(C_MIS));
        uint len;
        char *size;
+       char selgap[]="  ";
+
+       if (ent->flags & FILE_SELECTED)
+               selgap[1] = '+';
 
        /* Directories are always shown on top */
        resetdircolor(ent->flags);
 
-       addch((ent->flags & FILE_SELECTED) ? '+' : ' ');
+       addch(' ');
 
        if (attrs)
                attron(attrs);
@@ -3769,14 +3773,14 @@ static void printent_long(const struct entry *ent, uint namecols, bool sel)
        }
 
        if (g_state.oldcolor) {
-               addstr("  ");
+               addstr(selgap);
                if (!ln) {
                        attroff(A_DIM);
                        attrs ^=  A_DIM;
                }
        } else {
 #ifndef ICONS_ENABLED
-               addstr("  ");
+               addstr(selgap);
 #endif
                if (ent->flags & FILE_MISSING)
                        pair = C_MIS;
@@ -3789,7 +3793,7 @@ static void printent_long(const struct entry *ent, uint namecols, bool sel)
                        attrs |= COLOR_PAIR(pair);
 #ifdef ICONS_ENABLED
                attroff(attrs);
-               addstr("  ");
+               addstr(selgap);
                if (sel)
                        attrs &= ~A_REVERSE;
                print_icon(ent, attrs);
```

But a major problem is, the `+` is now in info color (grey/dim) making ti very difficult to see:



To bypass it we would have to add more code to discontinue the dim/grep color (and enable it again for missing files).

I would leave it as it is if it's not pressing. When you are pressing `+` you know exactly on which file you are because of the reverse coloring.

0xACE:
Quick comment: I would prefer if we avoid merging letters next to file names as it blocks "word" selecting filenames because the word boundary may include the extra `+` infront of the file name.

jarun:
Yes, that too.

jarun:
Fixed this at 499127c69889cc29191236e1eb2510247d92ae05.

PR #747, opened 2020-09-29
Fix rollover bug

### LINKED ISSUE #743: The -R option doesn't work with -l option

[x] Operating System: Debian 10
[x] Desktop Environment: fvwm
[x] Terminal Emulator: xfce4-terminal
[x] Shell: bash
[ ] Program options used:
[ ] Configuration options set:
[ ] Issue exists on `nnn` master

I found that the -R option only works when scrolling exactly to the last entry. For example, there are 17 files in a folder. run nnn -R -l 5, then I enter that folder, and use the mouse wheel to scroll. It will go to 1/17, 6/17, 11/17, 16/17, then rollover to 4/17,... 14/17, 2/17... and go on until 17/17.

leovilok:
Fixes #743

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5973,7 +5973,7 @@ nochange:
 #if NCURSES_MOUSE_VERSION > 1
 			/* Scroll up */
 			if (event.bstate == BUTTON4_PRESSED && ndents && (cfg.rollover || cur)) {
-				if(!cfg.rollover && cur < scroll_lines)
+				if (!cfg.rollover && cur < scroll_lines)
 					move_cursor(0,0);
 				else
 					move_cursor((cur + ndents - scroll_lines) % ndents, 0);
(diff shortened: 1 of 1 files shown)


leovilok:
Sorry for the branch created in `nnn` repo, I'm on a horrible network that block git pull/push, so I had to use the github web UI.

leovilok:
And I realise tha maybe the expected behaviour is a little bit different:

- currently it stops scrolling if there is less space than what's provided in the `-l` option
- but one might expect to go to the top/end and get blocked here

What do you think ?

jarun:
> but one might expect to go to the top/end and get blocked here

This.

leovilok:
Done.

src/nnn.c
```diff
@@ -5972,15 +5972,21 @@ static bool browse(char *ipath, const char *session, int pkey)
 			}
 #if NCURSES_MOUSE_VERSION > 1
 			/* Scroll up */
-			if (event.bstate == BUTTON4_PRESSED && ndents && (cfg.rollover || cur >= scroll_lines)) {
-				move_cursor((cur + ndents - scroll_lines) % ndents, 0);
+			if (event.bstate == BUTTON4_PRESSED && ndents && (cfg.rollover || cur)) {
+				if(!cfg.rollover && cur < scroll_lines)
```
jarun, line 5976: Space after `if`. Same at line 5986.

Is there no way to do it without an additional condition check?

@KlzXS ?

jarun (changes requested):

src/nnn.c
leovilok: Well you can hide it in a ternary operator, but it's really a special case, so we need to check it at some point.

PR #756, opened 2020-10-13
fix fzf bookmarks plugin environment variable

fdplacido:
Setting a custom $BOOKMARKS_DIR env is overwritten by the plugin, so only the default directory can be used.
- Check if the variable is defined before setting the default.


jarun:
CI fails. Please fix.

jarun:
Thank you!

PR #762, opened 2020-10-15
fix bookmarks plugin on empty output

fdplacido:
If no bookmark was selected from `fzf` with bookmarks plugin (i.e. pressing Esc key), nnn would change to `$BOOKMARKS_DIR`.
- With this PR the output would be to keep nnn in the current directory, and only change directory if a bookmark is selected.

### DIFF
--- a/plugins/bookmarks
+++ b/plugins/bookmarks
@@ -44,7 +44,11 @@ get_links() {
         [ -d "$entry" ] || continue
 
         printf "%20s -> %s\n" "$(basename "$entry")" "$(readlink -f "$entry")"
-    done | fzf | awk 'END { print "'"$BOOKMARKS_DIR"'/"$1 }'
+    done | fzf |
+    awk 'END {
+        if(length($1) == 0){ print "'"$PWD"'" }
+        else { print "'"$BOOKMARKS_DIR"'/"$1 }
+        }'
 }
 
 # Choose symlink with fzf
--- a/plugins/bookmarks
+++ b/plugins/bookmarks
@@ -49,7 +49,6 @@ get_links() {
         if (length($1) == 0) { print "'"$PWD"'" }
         else { print "'"$BOOKMARKS_DIR"'/"$1 }
     }'
-
 }
 
 # Choose symlink with fzf
--- a/plugins/bookmarks
+++ b/plugins/bookmarks
@@ -46,9 +46,10 @@ get_links() {
         printf "%20s -> %s\n" "$(basename "$entry")" "$(readlink -f "$entry")"
     done | fzf |
     awk 'END {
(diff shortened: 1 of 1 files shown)


plugins/bookmarks
```diff
@@ -44,7 +44,11 @@ get_links() {
         [ -d "$entry" ] || continue
 
         printf "%20s -> %s\n" "$(basename "$entry")" "$(readlink -f "$entry")"
-    done | fzf | awk 'END { print "'"$BOOKMARKS_DIR"'/"$1 }'
+    done | fzf |
+    awk 'END {
+        if(length($1) == 0){ print "'"$PWD"'" }
```
jarun, line 49: Please fix all indentation issues. Your last PR also had an issue which I fixed in commit bfcf296334fdb8b4800044d6e6dece1176af749d.

jarun (changes requested):

plugins/bookmarks
jarun: Updated the commit hash.

TBH `nnn` is past the stage where we expect these type of novice/repeat issues in PRs. Please be more mindful of that when you raise the PRs. We really don't have the time to spend to figure and rework basic stuff.

fdplacido: Just to be sure, you mean everything to be in the same line?

Sorry about the previous contribution. In this one I used 4 spaces for indentation, as stated in CONTRIBUTING.md. For breaking the long line I assumed that lines longer than 80 columns should be breaked, as stated in the Linux kernel coding style.

Maybe a clang-format file is a good idea.

jarun:
```
    awk 'END {
        if (length($1) == 0) { print "'"$PWD"'" }
        else { print "'"$BOOKMARKS_DIR"'/"$1 }
    }'
```

plugins/bookmarks
```diff
@@ -46,9 +46,10 @@ get_links() {
         printf "%20s -> %s\n" "$(basename "$entry")" "$(readlink -f "$entry")"
     done | fzf |
     awk 'END {
-        if(length($1) == 0){ print "'"$PWD"'" }
+        if (length($1) == 0) { print "'"$PWD"'" }
         else { print "'"$BOOKMARKS_DIR"'/"$1 }
-        }'
+    }'
+
```
jarun, line 52: Please remove this redundant newline.

plugins/bookmarks
fdplacido: Thanks, fixed. I'll be more careful next time.

jarun: There's a redundant newline. Please fix that as well.

jarun (approved):

jarun:
Thank you!

PR #764, opened 2020-10-19
Support zoxide for jumping

toadle:
There is even more autojump-clones. I like to use https://github.com/ajeetdsouza/zoxide because it supports querying it's database and some other things.

I though I'd add another plugins to use it in nnn.


jarun:
Please add the functionality to the `autojump` plugin. It already supports autojump and jump.

toadle:
@jarun Oh, you are right. Didn't notice that. I'll change it and submit a fix.

jarun:
Thanks!

toadle:
@jarun New version!

BTW: I noticed that when calling

```
curl -Ls https://raw.githubusercontent.com/jarun/nnn/master/plugins/getplugs | sh
```

while having `nnn` 3.4 will not get me the plugins from `master´-branch, but rather some version-specific stuff. So I won't get the current `autojump`-plugin. That was the root cause for me not seeing that `autojump` supports several scripts already.

jarun:
> while having nnn 3.4 will not get me the plugins from master-branch

That's correct. We can't always support new plugins with older releases so the plugins are release-fixed.

jarun:
Thank you!

jarun:
@toadle I have added some minor modifications at 57163ffa5de7c2c8ab136e5f50ac56a7ce0dac7c.

PR #765, opened 2020-10-21
Modified dups to allow selecting files for removal

KlzXS:
The script displays all the duplicate files in a directory grouped by md5 hash and size displayed above each grouping. Each grouping is followed by a newline. To keep a file remove it from the temp file, the script will attempt to remove every line not beginning with "md5sum".

If anyone has suggestions to improve performance they are more than welcome.

### DIFF
--- a/plugins/dups
+++ b/plugins/dups
@@ -25,7 +25,7 @@ printf "\
 ## After editiing this file you will be prompted to remove some of them.
 ## You can choose between removing all the commented out files, all the uncommented ones or none at all.
 ## All the lines begining with '##','#md5sum' or 'md5sum' will be ignored either way.
-## If you choose to remove, you will be given a choice between removing with force or interactively for each file.
+## If you choose to remove, you will be given a choice between removing with force or interactively for each file.\n
 " > "$tmpfile"
 
 # shellcheck disable=SC2016
@@ -39,7 +39,7 @@ g
 :loop
 N
 /.*\n$/!b loop
-p' | sed -E 's/^.{32} (.*) d[0-9]*$/\1/' > "$tmpfile"
+p' | sed -E 's/^.{32}  (.*) d[0-9]*$/\1/' >> "$tmpfile"
 
 "$EDITOR" "$tmpfile"
 
@@ -47,9 +47,9 @@ printf "Remove commented files? (yes/no/abort) [default=a]: "
 read -r commented
 
 if [ "$commented" = "y" ]; then
-	sedcmd="/^(##|#?md5sum|[^#]).*/d"
+	sedcmd="/^(##|#?md5sum|[^#]).*/d; /^$/d; s/^# *(.*)$/\1/"
 elif [ "$commented" = "n" ]; then
-	sedcmd="/^(#|#?md5sum).*/d"
+	sedcmd="/^(#|#?md5sum).*/d; /^$/d; s/^ *(.*)$/\1/"
 else
 	printf "Press any key to exit"
 	read -r _
@@ -59,10 +59,13 @@ fi
 printf "Remove with force or interactive? (f/i) [default=i]: "
 read -r force
 
-rmcmd="'rm -$force \"\$0\" \"\$@\" < /dev/tty'"
-
-# shellcheck disable=SC2016
-sed -e "$sedcmd" "$tmpfile" | tr '\n' '\0' | xargs -0 sh -c "$rmcmd"
+if [ "$force" = "f" ]; then
+	#shellcheck disable=SC2016
+	sed -E "$sedcmd" "$tmpfile" | tr '\n' '\0' | xargs -0 sh -c 'rm -f "$0" "$@" </dev/tty'
+else
+	#shellcheck disable=SC2016
+	sed -E "$sedcmd" "$tmpfile" | tr '\n' '\0' | xargs -0 sh -c 'rm -i "$0" "$@" </dev/tty'
+fi
 
 rm "$tmpfile"
 
--- a/plugins/dups
+++ b/plugins/dups
@@ -6,10 +6,39 @@
 #
 # Dependencies: find md5sum sort uniq xargs
 #
-# Shell: POSIX compliant
+# Note: bash compatible required for mktemp
+#
+# Shell: bash
 # Authors: syssyphus, KlzXS
 
-find . -size +0 -type f -printf "%s %p\n" | sort -rn | sed -n 'N; /^\([0-9]*\) .*\n\1.*$/p;$d;D' | awk '{printf("%s\0", substr($0, index($0, $2)))}' | xargs -0 md5sum | sort | uniq -w32 --all-repeated=separate
+# If the size of a file has more that $size_digits digits the file will be misplaced
+# 12 digits fit files up to 931GiB
+
+EDITOR="${EDITOR:-vi}"
+TMPDIR="${TMPDIR:-/tmp}"
+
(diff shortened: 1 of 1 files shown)


KlzXS:
Don't merge until I benchmark another approach I want to try.

jarun:
Some thoughts:

1. We need to ask at some point in the workflow whether the user wants to delete the lines left in the tmp file or not? This should be a cumulative question based on which the whole deletion process can be aborted.

2. It would also be a good idea to add a commented (hashed note) in the top of the tmp file that the lines left will be deleted.

3. I think we can take an approach of omitting hashed lines in the tmp file. So the user has another choice to hash out the files he wants to keep. To make things simple, we can show the hash + size line commented as well.

KlzXS:
> cumulative question

What do you mean by cumulative?

Also please don't use hash as in # and has as in hashsum in the same context. It took me a second. :)

jarun:
I meant a overall/general prompt that allows the user to abort the whole process or removing files one by one. Or you can provide an option in the per 

> file prompt to abort the remaining process.

Hehe, I meant skip the lines beginning with a `#`

jarun:
@KlzXS any update on this?

KlzXS:
I've been struggling to find much time to work on this lately but I'll try my best.

jarun:
Sure thing!

jarun:
@KlzXS any update here? I will plan for the next release once this is in.

jarun:
shellcheck fails:

```
sed -e $sedcmd "$tmpfile" | tr '\n' '\0' | xargs -0 sh -c "$rmcmd"
       ^-- SC2086: Double quote to prevent globbing and word splitting.
```

Also, please remove the _work in progress_ label once you are done.

jarun:
I see the following issue:

Created a file abc with character i in it. Created a duplicate name def.

Ran the plugin and commented both abc and def as follows:

```
# ./abc
#./def
```

Then I did the following:

```
Remove commented files? (yes/no/abort) [default=a]: y
Remove with force or interactive? (f/i) [default=i]: i
#md5sum: 372e25f23b5a8ae33c7ba203412ace30 size: 2 bytes: 1: rm -i "$0" "$@" < /dev/tty: not found
```

Finally, none of the 2 files are removed. Please take a look.

jarun:
Note that there is a space before `./abc` but not before `./def`. That's intentional.

KlzXS:
That should no longer be a problem, I tried some test similar to yours.

plugins/dups
```diff
@@ -4,12 +4,70 @@
...
+EDITOR="${EDITOR:-vi}"
+TMPDIR="${TMPDIR:-/tmp}"
+
+size_digits=12
+tmpfile=$(mktemp "$TMPDIR/.nnnXXXXXX")
+
+printf "\
+## This is an overview of all duplicate files found.
+## After editiing this file you will be prompted to remove some of them.
+## You can choose between removing all the commented out files, all the uncommented ones or none at all.
```
jarun, line 26: We need better documentation. Please explain more. Are the commented files removed or uncommented ones?

> commented out files

Single comment of double comment?

The next line indicates commented lines are ignored. From what I see, files in commented lines are deleted.

jarun:
I just want to ensure we are doing this right as it has potential to remove unwanted user data.

@leovilok @0xACE @snide can you guys spend some time on some random tests on this plugin?

jarun (changes requested):

plugins/dups
```diff
@@ -4,12 +4,70 @@
...
+# Dependencies: find md5sum sort uniq xargs gsed
 #
-# Shell: POSIX compliant
+# Note: bash compatible required for mktemp
+#
+# Shell: bash
 # Authors: syssyphus, KlzXS
 
-find . -size +0 -type f -printf "%s %p\n" | sort -rn | sed -n 'N; /^\([0-9]*\) .*\n\1.*$/p;$d;D' | awk '{printf("%s\0", substr($0, index($0, $2)))}' | xargs -0 md5sum | sort | uniq -w32 --all-repeated=separate
+# If the size of a file has more that $size_digits digits the file will be misplaced
```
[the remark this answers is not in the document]
jarun: Also does it indicate files lesser than or greater than 931 GiB will be correctly detected?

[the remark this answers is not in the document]
KlzXS: We want to filter all the files with duplicate filesizes, `uniq` doesn't have an option to filter by a specific space separated field, it only has an option to compare by the first `n` characters. So I need `find` to print the filesize with a constant number of characters. I chose 12, no real reason, we could go with less as you probably aren't going to be using this plugin on such huge files.

> Why can't we ignore such file (instead of misplacing)?

We could, but do you really expect someone to use this on a TB file? I haven't seen such a large file. Ever.

plugins/dups
KlzXS: I've simplified this a bit.

All the double comment lines (`##`) are always ignored, now the hash and size are also double commented.
Of the remaining non-empty lines the user has a choice to remove all the commented ones (`#`) or all the uncommented ones. (`Remove commented files? (yes/no/abort) [default=a]:`)

plugins/dups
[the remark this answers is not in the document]
jarun: Ahh OK. I thought it had to do with file size limit (rather than file size print length limit). ;)

plugins/dups
jarun: So `no` to the question means remove uncommented files?

If that's the case, I would say we stick to remove commented files and remove the `no` option. In any batch delete case, a man should be forced to be extra careful.

By default all the results are uncommented so I would say we don't delete uncommented as a fixed strategy. If the user wants to delete something he has to explicitly comment it.

jarun: Trust me, being a little less smart here is far better than a user coming back complaining he lost his files even before he understood clearly how the plugin works.

And _Comment the files you want to delete_ is a simple and clear instruction.

plugins/dups
```diff
@@ -4,12 +4,67 @@
...
+g
+
+:loop
+N
+/.*\n$/!b loop
+p' | sed -E 's/^.{32}  (.*) d[0-9]*$/\1/' >> "$tmpfile"
+
+"$EDITOR" "$tmpfile"
+
+printf "Remove commented files? (yes/no) [default=n]: "
```
jarun, line 45: Would it be a good idea to provide an option `export` to export and exit? In case the user wants to inspect and act manually?

KlzXS: I mean when in the editor the user can save the file wherever they like. And then they can just abort. I don't really see much of a reason to add an export.

jarun: I agree. Missed that. We are good!

jarun:
Thank you so much @KlzXS! This is a significant value addition.

jarun:
I see an issue if I do not comment any file for removal. Editor is vim. Without changing the tmp file I did `:q`.

Then:

```
Remove commented files? (yes/no) [default=n]: y
Remove with force or interactive? (f/i) [default=i]: i
rm: cannot remove 'sh': No such file or directory
```

I think when there is no file to remove rm is still invoked.

KlzXS:
Since we are already using GNU sed, let's use the `-r` aka `--no-run-if-empty` flag from GNU xargs.

jarun:
Sure thing!

jarun:
Took care of it at commit 1418ecde0dc28dc7d32a9ffdcf0721db7bd2d624.

PR #796, opened 2020-11-20
deactivate INOTIFY for CYGWIN

magomi:
inotify is a Linux kernel feature. It is not available within CYGWIN.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -35,8 +35,10 @@
 #if defined(__arm__) || defined(__i386__)
 #define _FILE_OFFSET_BITS 64 /* Support large files on 32-bit */
 #endif
+#if !defined(__CYGWIN__)
 #include <sys/inotify.h>
 #define LINUX_INOTIFY
+#endif
 #if !defined(__GLIBC__)
 #include <sys/types.h>
 #endif


src/nnn.c
```diff
@@ -35,8 +35,10 @@
 #if defined(__arm__) || defined(__i386__)
 #define _FILE_OFFSET_BITS 64 /* Support large files on 32-bit */
 #endif
+#if !defined(__CYGWIN__)
```
jarun, line 38: How about the MINGW, MINGW32 and MINGW64 families?

jarun (changes requested):

src/nnn.c
jarun: I'll add these. Just try to keep an eye out when you raise a PR. `nnn` is a very responsive project.

magomi: > 
> 
> How about the MINGW, MINGW32 and MINGW64 families?

I think this should also be checked. I'm sure these platforms too have no inotify support. 

But I have no environments other than linux and cygwin to test this. Therefore I decided to take the save way for this change.

magomi: > 
> 
> I'll add these. Just try to keep an eye out when you raise a PR. `nnn` is a very responsive project.

Sorry for that. Didn't expected this PR to be merged this fast. 

Thanks for the work on nnn. It's an awesome tool.

jarun: No problem! Thank you for your contribution.

I did sanitize this at commit 6f7a4f723b44d8a6abe4ee9e037da2cc5ca910b6.

PR #799, opened 2020-11-22
nuke plugin: use $VISUAL environment variable

tchici:
falls back to previous behavior if $VISUAL is not set ($EDITOR then vi)

I think it's convention that `$VISUAL` should be checked first. `$EDITOR` is meant for non-visual text editor (like `ed`) on older terminals.


jarun:
Thank you!

jarun:
If possible, please run a quick audit on the other plugins as well.

PR #800, opened 2020-11-22
use $VISUAL environment variable for plugins

tchici:
falls back to previous behavior if $VISUAL is not set ($EDITOR then vi)

Here are the rest of them.

### DIFF
--- a/plugins/.nmv
+++ b/plugins/.nmv
@@ -16,7 +16,7 @@
 # Shell: bash
 # Author: KlzXS
 
-EDITOR="${EDITOR:-vi}"
+EDITOR="${VISUAL:-${EDITOR:-vi}}"
 TMPDIR="${TMPDIR:-/tmp}"
 INCLUDE_HIDDEN="${INCLUDE_HIDDEN:-0}"
 VERBOSE="${VERBOSE:-0}"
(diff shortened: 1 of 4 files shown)


jarun:
Thank you so much!

tchici:
No problem. Thanks for creating the best file manager!

jarun:
@tchici we need to revert this. The program is supposed to wait for these plugins. Using `$VISUAL` breaks option `-E`.
Also see: https://github.com/jarun/nnn/wiki/Basic-use-cases#detached-text

jarun:
Note: no issues with `nuke`.

tchici:
Would it be possible to check for the `-E` option in the plugin then use `$EDITOR` otherwise `$VISUAL` then `$EDITOR` then `vi`? I know in Linux its pretty easy to check the command line arguments of other processes (assuming no sandboxing).

For example, something like `[ -n "$(ps -p $PPID -o args | grep E)" ]`.

Not sure how well it will work on other operating systems, like FreeBSD, which I think might have an option to prevent other processes checking arguments of other processes (correct me if I am wrong).

jarun:
I think as we have a fallback (vi), even if EDITOR is not defined we are good. It's too trivial to change the communication with the plugin and probably too late as well.

jarun:
BTW, plugins themselves are indicative. Users are supposed to edit them as per their needs and use them.

tchici:
Makes sense :+1:

Issue #81, opened 2018-02-23
Open new programs not as child of nnn?


javeq1:
I really like how fast it is to browse directories with nnn and I would like to use it as my main file manager, but I have a small problem and I don't know if there is a configuration to solve it.
When using nnn if, for example, I open a pdf using enter then *evince* is started with the opened file. If then I close the terminal containing nnn evince is killed too. Is there a way to keep programs started by nnn opened when closing nnn?
Thank you for your work!

jarun:
Can you try this? https://github.com/jarun/nnn#nnn-blocks-on-opening-files

jarun:
I believe this would solve your problem. So closing the issue. If it's not working, feel free to reopen.

javeq1:
I am already using that because I had the problem of nnn being blocked when opening a file. Now that problem is solved but there is still this strange problem I described you.
I am using Arch Linux with i3wm. I installed nnn using the aur package. Don't know if it is important but I am also using zsh as my shell instead of bash

jarun:
I have tested on my system with evince, can't reproduce the issue. I ran with xfce4-terminal and terminator.

I'm afraid you'll have to try the following:

1. check out the latest code from master
2. check `ps` output to figure out which process is the parent of the evince process

javeq1:
I installed latest version using nnn-git from the aur and I still have the problem.
Using ps I found out that:
 1. evince parent is "xdg-open";
 2. "xdg-open" parent is nnn.

I am using terminator as a terminal, but I have the same problem with xfce4-terminal.
I also noticed that if I quit nnn without closing the terminal than evince is not killed, even if after that I close the terminal

jarun:
OK. I gues you have to set the PPID to init. Can you please raise a PR?


Regards,

Arun

jarun:
I am working on this tonight. :+1:

jarun:
Can you please confirm if the patch on master fixes your issue?

javeq1:
I confirm that it's solved! Thank you!

PR #816, opened 2020-12-15
Fix the same issue as #81 but with nuke

kyechou:
## Issue description

When I set `NNN_OPENER` to nuke, I had the same issue as #81, where the GUI child process exits when I close nnn.

My desktop environment is Arch Linux with bspwm and sxhkd. My terminal emulator is termite. `nnn` is started as `termite -e nnn`. The same issue can be reproduced with xfce4-terminal, and the latest commit of `nnn` on master.

## Solution

The PR pretty much does the same thing as [`70c39a`](https://github.com/jarun/nnn/commit/70c39a063d1097d33b8967b9a31618b5f826db72), but with nohup in nuke's script.

### DIFF
--- a/plugins/nuke
+++ b/plugins/nuke
@@ -89,10 +89,10 @@ is_mac() {
 
 handle_pdf() {
     if [ "$GUI" -ne 0 ] && is_mac; then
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #821, opened 2020-12-20
plugins: dvtm: add plugin for open file in dvtm window

vkochan:
Add plugin to open selected file by $EDITOR in separate
DVTM window.

Signed-off-by: Vadym Kochan <vadim4j@gmail.com>


jarun:
Why not document this instead? Also, CI fails. Please check.

PR #822, opened 2020-12-21
plugins: preview-tui: video thumbnails in preview-tui

j0hnnybash:
(no idea wether you or anyone is interested in this.)
its quite easy to create temporary thumbnails for videos with ffmpegthumbnailer and show them in a preview window.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -158,10 +158,11 @@ preview_file () {
     elif [ "$encoding" = "binary" ]; then
         if [ "${mimetype%%/*}" = "image" ] || [ "${mimetype%%/*}" = "video" ]; then
             if [ "${mimetype%%/*}" = "video" ] && [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
-		videothumb="/tmp/videothumb.$$.png"
-		ffmpegthumbnailer -s 0 -i "$1" -o "$videothumb"
-		set "$videothumb"
-	    fi
+                videothumb="/tmp/videothumb.$$.png"
+                ffmpegthumbnailer -s 512 -i "$1" -o "$videothumb" >/dev/null 2>&1
+                set "$videothumb"
+                trap 'rm "$videothumb"' EXIT
+            fi
             if [ "$TERMINAL" = "kitty" ]; then
                 # Kitty terminal users can use the native image preview method.
                 kitty +kitten icat --silent --transfer-mode=stream --stdin=no \
@@ -214,7 +215,6 @@ if [ "$PREVIEW_MODE" ] ; then
         kitty @ last-used-layout --no-response >/dev/null 2>&1
     fi
 
-    rm $videothumb
     exit 0
 fi
 
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -156,7 +156,12 @@ preview_file () {
             fifo_pager ls --color=always
         fi
     elif [ "$encoding" = "binary" ]; then
-        if [ "${mimetype%%/*}" = "image" ] ; then
+        if [ "${mimetype%%/*}" = "image" ] || [ "${mimetype%%/*}" = "video" ]; then
+            if [ "${mimetype%%/*}" = "video" ] && [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
+		videothumb="/tmp/videothumb.$$.png"
+		ffmpegthumbnailer -s 0 -i "$1" -o "$videothumb"
+		set "$videothumb"
+	    fi
             if [ "$TERMINAL" = "kitty" ]; then
                 # Kitty terminal users can use the native image preview method.
                 kitty +kitten icat --silent --transfer-mode=stream --stdin=no \
@@ -167,19 +172,6 @@ preview_file () {
                 viu -t "$1"
             else
                 fifo_pager print_bin_info "$1"
-            fi
-        elif [ "${mimetype%%/*}" = "video" ] && [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
-            ffmpegthumbnailer -s 0 -i "$1" -o "/tmp/videothumb.png"
-            if [ "$TERMINAL" = "kitty" ]; then
-                     # Kitty terminal users can use the native image preview method.
-                     kitty +kitten icat --silent --transfer-mode=stream --stdin=no \
-                          "/tmp/videothumb.png" &
-                elif exists catimg; then
-                     catimg "/tmp/videothumb.png"
-                elif exists viu; then
-                     viu -t "/tmp/videothumb.png"
-                else
-                     fifo_pager print_bin_info "$1"
             fi
         elif [ "$mimetype" = "application/zip" ] ; then
             fifo_pager unzip -l "$1"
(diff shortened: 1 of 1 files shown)


jarun:

plugins/preview-tui
```diff
@@ -163,6 +168,19 @@ preview_file () {
             else
                 fifo_pager print_bin_info "$1"
             fi
+        elif [ "${mimetype%%/*}" = "video" ] && [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
+            ffmpegthumbnailer -s 0 -i "$1" -o "/tmp/videothumb.png"
+            if [ "$TERMINAL" = "kitty" ]; then
```
leovilok, line 173: we might mant to do something to avoid repeating the image diplay code

plugins/preview-tui
```diff
@@ -163,6 +168,19 @@ preview_file () {
             else
                 fifo_pager print_bin_info "$1"
             fi
+        elif [ "${mimetype%%/*}" = "video" ] && [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
+            ffmpegthumbnailer -s 0 -i "$1" -o "/tmp/videothumb.png"
```
leovilok, line 172: if we create this file, can we delete it when we're done with it

leovilok, line 172: also we might want the filename to include something like the previewer's PID, so that multiple preview sessions on the same machine don't override each other's preview image.

leovilok (commented):
Just by reading the code (not actually testing anything) it looks OK (I commented on a few parts though).

I don't know if it's reasonably fast.

Also, it adds complexity to a script we wanted to keep relatively simple, but I think that it adds a feature that's important & usefull for a lot of people.

plugins/preview-tui
j0hnnybash: we could use the same loop for images and videos and use set to replace $1 with the thumbnail path:

```
if [ "${mimetype%%/*}" = "image" ] || [ "${mimetype%%/*}" = "video" ]; then
        if [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
                ffmpegthumbnailer -s 0 -i "$1" -o "/tmp/videothumb.png"
                set "/tmp/videothumb.png"
        fi
        if [ "$TERMINAL" = "kitty" ]; then
```

plugins/preview-tui
j0hnnybash: we could use something like /tmp/$$_videothumb.png, but thats only useful if we find a simple way to delete the files afterwards.

jarun:
@leovilok please let me know when the review is done.

leovilok:
@j0hnnybash the solutions you suggested sound good to me:

- reuse the same code for image and ffpmeg based video preview
- use something like `/tmp/$$_videothumb.png` for the video path (maybe put it in a shell variable)

To delete the image, you can try to `rm` it just after launching the image preview command, but you'll have to check if it still works with `kitty`'s `icat`, because it's launched in the background, so the image might be deleted too soon.

plugins/preview-tui
```diff
@@ -156,7 +156,12 @@ preview_file () {
             fifo_pager ls --color=always
         fi
     elif [ "$encoding" = "binary" ]; then
-        if [ "${mimetype%%/*}" = "image" ] ; then
+        if [ "${mimetype%%/*}" = "image" ] || [ "${mimetype%%/*}" = "video" ]; then
+            if [ "${mimetype%%/*}" = "video" ] && [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
+		videothumb="/tmp/videothumb.$$.png"
+		ffmpegthumbnailer -s 0 -i "$1" -o "$videothumb"
+		set "$videothumb"
+	    fi
```
j0hnnybash, line 164: use the if loop for image previews for videos.
include the PID in the thumbnails filename.

plugins/preview-tui
```diff
@@ -222,6 +214,7 @@ if [ "$PREVIEW_MODE" ] ; then
         kitty @ last-used-layout --no-response >/dev/null 2>&1
     fi
 
+    rm $videothumb
```
j0hnnybash, line 217: as you assumed, instantly deleting the temp file doesn't work in kitty.

plugins/preview-tui
```diff
@@ -151,7 +156,12 @@ preview_file () {
             fifo_pager ls --color=always
         fi
     elif [ "$encoding" = "binary" ]; then
-        if [ "${mimetype%%/*}" = "image" ] ; then
+        if [ "${mimetype%%/*}" = "image" ] || [ "${mimetype%%/*}" = "video" ]; then
+            if [ "${mimetype%%/*}" = "video" ] && [ "$USE_VIDEOTHUMB" -ne 0 ] && exists ffmpegthumbnailer; then
+		videothumb="/tmp/videothumb.$$.png"
+		ffmpegthumbnailer -s 0 -i "$1" -o "$videothumb"
+		set "$videothumb"
+	    fi
```
leovilok, line 164: Indentation is inconsistent.

plugins/preview-tui
leovilok: Good.

plugins/preview-tui
leovilok: Does it work the way you did it there?
Even if we close the preview window?

j0hnnybash: only works when I cleanly exit nnn, so not ideal.
do you know why there's a "&" after kitty +kitten icat...?
without "&" the script waits for kitty to open the preview, which means we could delete the thumbfile instantly.

leovilok:
Does it work without the `&`?

I don't remember why it was added (I don't use kitty myself),
maybe because it's slow, so you'll have to wait for every image to render
when you're scrolling the file list, while launching it in the background
allows to kill it before the image is rendered when you just want to see
the next one.

Anyways, we might have to write something specifically for kitty again:

`ffmpegthumbnailer` probably gives a small, quick to display image, and
the call to `ffmpegthumbnailer` probably takes more time than the call to
Kitty's `icat`, so I think that if the mime type is video, we can call
`icat` synchronously (without the `&`).

Then we won't have any issue removing the file just after the call.

Of course, it the previewer is interrupted while `ffmpegthumbnailer`
generates the preview, or when an image previewer is painting it, this
would leave the file around ...

j0hnnybash:
you are exactly right, without '&' the image previewer waits for all the pictures to get rendered.

I think I found a solution, which archives our goals without too much code meddling.
with `trap` the file gets deleted as soon as the previewer script is getting exited. 
in addition I decreases the thumbnail size to 512p in width, which results in .png files of about 150kb.

an alternative for the record:
```
(sleep 1
rm $videothumb) &
```
waits one second for the preview to open and deletes the file, which works but as you pointed out leaves the file around when the previewer exits at the wrong time.

leovilok:
@j0hnnybash I'm OK with this solution, if you can confirm that you tried it and it works.

@jarun If everything works as intended, it's ready to merge in my opinion.

jarun:
Thank you!

jarun:
@j0hnnybash please confirm.

j0hnnybash:
I've tested this with Kitty and Alacritty/tmux/catimg, both combinations seem to work.

jarun:
Thank you!

PR #825, opened 2020-12-26
add NNN_ONCHDIR option to handle chdir events

vkochan:
It might be useful (looks like in some special cases)
to have ability of handling dir changes from nnn to make
custom routines (indexing, project pre-init).

Signed-off-by: Vadym Kochan <vadim4j@gmail.com>

### DIFF
--- a/nnn.1
+++ b/nnn.1
@@ -374,6 +374,11 @@ used. A single combination of arguments is supported for SHELL and PAGER.
     NOTE: 'nuke' is a file opener available in the plugin repository.
 .Ed
 .Pp
+\fBNNN_ONCHDIR:\fR specify a handler when directory is changed.
+.Bd -literal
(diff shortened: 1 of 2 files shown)


0xACE:
I think this is too much... Also I think it's better that you use the FIFO pipe and detect directory changes that way, though I'm not sure if that gets triggered when you switch contexts (didn't even check if this PR even handles that itself)...

vkochan:
ok, closing.

0xACE:
Ah, you can relax with that. I don't make executive decisions. Just voicing my opinion.

My specific conflict with this proposal is running `spawn()` when you run into a new directory, where as the same information can be deduced from the existing `NNN_FIFO`, meaning this is a redundant feature. Though someone who remembers it better than me could probably correct me if I'm wrong.

vkochan:
you are right, fifo writing is better than process spawning, looks like
fifo is enough to implement it.

PR #828, opened 2021-01-06
Use HTTPS everywhere (mechanical edit using util from https-everywhere)

ypid:
```Shell
~/src/EFForg/https-everywhere/utils/rewriter/rewriter.js .
```

Ref: https://github.com/EFForg/https-everywhere/tree/master/utils/rewriter

PR #830, opened 2021-01-06
Revert logo.hvif file without HTTPS being replaced

ypid:
This was not intended and slipped my self-review.

Fixes: #828
Thanks to: @0xACE


jarun:
I have already removed this file.

PR #834, opened 2021-01-09
Updated the Preview-Tabbed , 

wimstockman:
Changed the mime detector to xdg-mime this offers support for symlinks which file doesn't
and added support for postscript files with zathura

### DIFF
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -50,16 +50,17 @@
 
 
 XDOTOOL_TIMEOUT=2
-PAGER=${PAGER:-"vim -R"}
+PAGER=${PAGER:-"less -r"}
 NUKE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/nuke"
+TREEVIEW="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/treeview"
 
 
-if which xterm >/dev/null 2>&1 ; then
+if which st >/dev/null 2>&1 ; then
+   TERMINAL="st -w"
+elif which xterm >/dev/null 2>&1 ; then
     TERMINAL="xterm -into"
 elif which urxvt >/dev/null 2>&1 ; then
     TERMINAL="urxvt -embed"
-elif which st >/dev/null 2>&1 ; then
-    TERMINAL="st -w"
 else
     echo "No xembed term found" >&2
 fi
@@ -138,7 +139,11 @@ previewer_loop () {
 
         kill_viewer
 
-        MIME="$(file -b --mime-type "$FILE")"
+       if which xdg-mime >/dev/null 2>&1 ; then
+          MIME="$(xdg-mime query filetype "$FILE")"
+       else
+          MIME="$(file -b --mime-type "$FILE")"
+       fi
 
         case "$MIME" in
             video/*)
@@ -169,8 +174,15 @@ previewer_loop () {
                     term_nuke "$XID" "$FILE"
                 fi
                 ;;
+            application/postscript)
+                if which zathura >/dev/null 2>&1 ; then
+                    zathura -e "$XID" "$FILE" &
+                else
+                    term_nuke "$XID" "$FILE"
+                fi
+                ;;
             inode/directory)
-                $TERMINAL "$XID" -e nnn "$FILE" &
+                $TERMINAL "$XID" -e $TREEVIEW "$FILE" &
                 ;;
             text/*)
                 if [ -x "$NUKE" ] ; then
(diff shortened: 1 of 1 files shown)


plugins/preview-tabbed
```diff
@@ -50,16 +50,17 @@
 
 
 XDOTOOL_TIMEOUT=2
-PAGER=${PAGER:-"vim -R"}
+PAGER=${PAGER:-"less -r"}
 NUKE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/nuke"
+TREEVIEW="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/treeview"
 
 
-if which xterm >/dev/null 2>&1 ; then
```
leovilok, line 57: In the in-file docs (top of the file), we say that we default to `xterm`, not `st`. We could change the priority, but do you have a reason for that?

plugins/preview-tabbed
```diff
@@ -169,8 +174,15 @@ previewer_loop () {
...
+            application/postscript)
+                if which zathura >/dev/null 2>&1 ; then
+                    zathura -e "$XID" "$FILE" &
+                else
+                    term_nuke "$XID" "$FILE"
+                fi
+                ;;
             inode/directory)
-                $TERMINAL "$XID" -e nnn "$FILE" &
+                $TERMINAL "$XID" -e $TREEVIEW "$FILE" &
```
leovilok, line 185: Well I touht that nnn itself is a good folder preview app. It's fast and its probably already what the user is used to.

Also it emulates a kind of 2 panel view.

I can understand why you'd want to have a tree view there, but I'm not sure I agree…

plugins/preview-tabbed
```diff
@@ -138,7 +139,11 @@ previewer_loop () {
 
         kill_viewer
 
-        MIME="$(file -b --mime-type "$FILE")"
+       if which xdg-mime >/dev/null 2>&1 ; then
```
leovilok, line 142: xdg-mime looks interesting, is it as wide spread as `file` though?

leovilok (commented):
The postscript part looks good.

plugins/preview-tabbed
leovilok: Oh you fallback on file?
It's fine in that case...

plugins/preview-tabbed
wimstockman: O the Treeview, is a personal setting.
Wasn't went for the commit.
But FYI , the Treeview is piped with bat to less gives it a nice touch and easy to search :-)

plugins/preview-tabbed
wimstockman: Xdg-mime is a tool from xdg-utils.
So if you have xdg-open you probably have xdg-mime.

plugins/preview-tabbed
wimstockman: Yes, I do have a reason. I think users of ST also set their colors scheme fonts etc hard coded in st. xterm then uses completely other colors. For me personal I use black background in st. Never use xterm and when using the preview-tabbed I first didn't no why the background was white. 
So it makes more sense to make the default the one used mostly by the user.

jarun:
CI fails. Please check.

plugins/preview-tabbed
jarun: Please leave treeview out of the mix.

jarun (changes requested):

jarun:
@wimstockman 

This is a 8-day old PR. If I do not get a response in the next 2 days, Ill be closing it. Thanks!

plugins/preview-tabbed
```diff
@@ -182,7 +182,7 @@ previewer_loop () {
                 fi
                 ;;
             inode/directory)
-                $TERMINAL "$XID" -e $TREEVIEW "$FILE" &
+	        $TERMINAL "$XID" -e nnn "$FILE" &
```
jarun, line 185: Please fix the alignment. This is elementary.

jarun (changes requested):

plugins/preview-tabbed
```diff
@@ -50,16 +50,17 @@
 
 
 XDOTOOL_TIMEOUT=2
-PAGER=${PAGER:-"vim -R"}
+PAGER=${PAGER:-"less -r"}
 NUKE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/nuke"
 
 
-if which xterm >/dev/null 2>&1 ; then
+
```
jarun, line 57: Please do not add unnecessary newlines.

jarun (changes requested):

jarun:
The new PR seems to be #842... has the same problems.

PR #839, opened 2021-01-18
Updated wall plugin to the standard utility file

bagnacauda:
Small edit to use the standard utility file to get the mimetype instead of the mimetype script, which probably is based on Perl, in order to reduce the dependencies.

### DIFF
--- a/plugins/wall
+++ b/plugins/wall
@@ -12,7 +12,7 @@ cmd_exists () {
 }
 
 if [ -n "$1" ]; then
-    if [ "$(mimetype --output-format %m "$1" | awk -F '/' '{print $1}')" = "image" ]; then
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

0xACE:
Bad PR...
`file --mime-type "filename with spaces.png"  | awk '{print $2}'` yields `with` as 'mimetype`

KlzXS:
You can use `awk '{print $NF}'` to print the last column.

jarun:
I'll revert.

jarun:
@0xACE and @KlzXS fixed this at commit 5d8738544b0acb3ddf4d827601c9df8d7f07d6b4.

PR #851, opened 2021-02-09
adding support for ueberzug into plugins/preview-tui

Lucas-Summers:
I was able to add support for ```ueberzug``` in the ```preview-tui``` plugin. However, I couldn't support it in the ```preview-tabbed``` plugin because for some reason ```ueberzug``` does not want to work inside a terminal embedded in a ```tabbed``` window.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -172,7 +172,7 @@ preview_file () {
             elif exists viu; then
                 viu -t "$1"
             elif exists ueberzug; then
-                preview_ueberzug "$1"
+                preview_ueberzug "$cols" "$lines" "$1"
             else
                 fifo_pager print_bin_info "$1"
             fi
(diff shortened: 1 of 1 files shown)


jarun:
CI fails, please check.

jarun:
Here's the error:

```
In plugins/preview-tui line 201:
    ueberzug layer --parser json 0< <(
                                    ^-- SC2039: In POSIX sh, process substitution is undefined.


In plugins/preview-tui line 203:
        read
        ^--^ SC2162: read without -r will mangle backslashes.

For more information:
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, process substitution...
  https://www.shellcheck.net/wiki/SC2162 -- read without -r will mangle backs...
```

leovilok:
The first error is because `<( )` is bash syntax, we try to keep this plugin in POSIX sh.

Don't worry about integrating ueberzug into preview-tabbed, we can use a graphical viewer there, so we don't need this hack that paste an image onto the terminal window.

jarun:
Thank you!

luukvbaal:
Isn't grouping commands a bashism as well? It doesn't work on my system with `/usr/bin/sh` symlinked to `bash`, but does work when changing `#!/usr/bin/env sh` to `#!/usr/bin/env bash` .
Using `dash` results in `ueberzug` displaying a black box which disappears instantly.

Using bash it does work but it only successfully displays every other picture on my system i.e. pressing j 4 times results in 2 pictures displayed and 2 empty preview windows.

PR #852, opened 2021-02-11
state.runctx now retains bitwidth as cfg.curctx

0xACE:
Didn't test this out, but judging by the name of the variable this
should be a bugfix


jarun:
Thank you!

PR #856, opened 2021-02-16
Add check for empty selection

### LINKED ISSUE #855: Clear selection (mm) does not delete .selection file causing copy/move to error
#### Environment details

[x] Operating System: Arch Linux
[ ] Desktop Environment: None (i3wm)
[x] Terminal Emulator: alacritty
[x] Shell: zsh
[ ] Program options used:
[ ] Configuration options set:
[x] Issue exists on `nnn` master

1. Select one or more files with `space` or `m`
2. Clear selection with `mm`
3. Try to copy (`p`) or move (`v`) with empty buffer

##### Expected result
The `~/.config/nnn/.selection` file is deleted and the copy/paste operation returns `0 selected` at the bottom of the window. No file is copied/moved.

##### Actual result
An error message is shown:
```
cp: cannot stat 'sh': No such file or directory
'Enter' to continue
```
The content of `.selection` file is deleted but the empty file is still in `~/.config/nnn/`.

This does not happen when clearing selection with `space`. The problem is that if the window is redrawn (after switching the context), the `+` symbol disappears and the file cannot be deselected with `space`.

KlzXS:
Fixes #855

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2084,6 +2084,11 @@ static bool cpmvrm_selection(enum action sel, char *path)
 {
 	int r;
 
+	if (!selbufpos) {
+		printmsg(messages[MSG_0_SELECTED]);
+		return FALSE;
+	}
+
 	if (!selsafe())
 		return FALSE;
 


jarun:
Thank you!

PR #857, opened 2021-02-16
Favor selection over current for nmv

KlzXS:
Inspired by 600286b6272 I updated the batch rename plugin as well.


jarun:
Thank you!

PR #858, opened 2021-02-16
Fix frozen terminal caused by opener

KlzXS:
`xdg-open` would launch a cli program that would contest the terminal for character input resulting in a unresponsive terminal that's hard to use. There are many references to this issue, but I'm bad at finding them.

If you send `xdg-opens`'s `std{in,out,err}` to the shadow realm it seems to prevent it from launching cli programs (probably because they don't want to work without any I/O, but I guess you can may be able to find some oddballs) so it defaults to some gui program (probably firefox) and that doesn't block the terminal.

This should be tested to make sure the other openers are okay with silencing their `std{in,out,err}`.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -215,6 +215,7 @@
 #define F_NORMAL  0x08  /* spawn child process in non-curses regular CLI mode */
 #define F_CONFIRM 0x10  /* run command - show results before exit (must have F_NORMAL) */
 #define F_CHKRTN  0x20  /* wait for user prompt if cmd returns failure status */
+#define F_ALLNULL 0x44  /* stdin, stdout and stderr mapped to /dev/null */
 #define F_CLI     (F_NORMAL | F_MULTI)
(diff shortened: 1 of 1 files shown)


jarun:
@KlzXS can you share some test cases?

KlzXS:
Right. I use `LICENSE` but any `text/plain` file should do. Just hover over it and press `l` or `Enter` to open it.
In my case it decides to open with `vim` for you it may be some other terminal editor or maybe even `less` depends on your configuration. Doesn't matter really, as long as it's a terminal program that accepts input from `stdin` the problem holds.

You won't see anything change but your keypresses will now be "ignored" at random, effectively making the terminal unusable.

KlzXS:
I forgot to mention. Of course if you have it setup so that `text/plain` files are opened in a gui editor like sublime you won't be experiencing this with text files. Right now I can't think of a non-text file that is commonly opened in terminal, but any such would also exhibit this behavior.

jarun:
Thank you!

Issue #781, opened 2020-11-15
ToDo list


jarun:
Rolled from #719.

### Cooking

- [x] REPL command prompt (<kbd>Esc</kbd> or <kbd>Enter</kbd> to exit)
- [x] invert selection with <kbd>A</kbd>
- [x] option `-u` removed (always prefer selection to hovered)
- [x] visit start dir on <kbd>@</kbd> when start path is a file
- [x] exit filter mode and redraw on <kbd>^L</kbd> if no last filter
- [x] plugin `fzcd` now selects the chosen file (#876)
- [x] `ueberzug` support in plugin `preview-tui`
- [x] new plugin `preview-tui-ext` with extra preview support
- [x] clear selection after successful plugin invocation
- [x] add method to sync subshell `$PWD` in WIki
- [x] clear selection on single file deletion (#812)
- [x] copy between instances not working (#864)
- [x] plugin `togglex` to toggle exe mode of a selection (#813)
- [x] fix `memccpy()` buffer overlap fault on macOS (#786)
- [x] show `0 selected` msg on cp/mv with empty selection (#855)
- [x] fix frozen terminal caused by opener (#858)
- [x] migrate macOS CI to GitHub workflows, retire Travis

### Up for grabs

- [ ] update only current context data when saving session (#837)
- [ ] generate static binary for arm, aarch with releases using CI (#623)
- [ ] support predefined filters like bookmarks
- [ ] open new terminal window in spawned shell directory (#759)

For anything else please discuss in this thread.

**Contribution [guideline](https://github.com/jarun/nnn/blob/master/misc/CONTRIBUTING.md).**

ghistes:
Not sure if this feature already exists but I would find it useful if you could color regular files based on their file-types (primitively given by a file's extension).

So maybe an environment variable that would allow me to associate file-extensions with colors.

If you use icons this functionality already exists to a certain degree (e.g. you get a specific icon for pdf-files), but I fould icons to be irritating, nevertheless would like to be able to visually differentiate file-types.

jarun:
You are right, this is done with icons. There's no plan to add colors to file name text otherwise.

KlzXS:
I mean, you could compile it with icons, then just tell it to print an empty string as the icon. Setting the padding to an empty string as well.

Use either `O_ICONS` or `O_NERD` and empty out the corresponding strings.

KlzXS:
If you don't want icons with your colors, that is.

rieje:
[...]

PR #869, opened 2021-03-08
preview-tui improvements

luukvbaal:
While waiting for a response in #851 I took it upon myself to improve the ueberzug integration more akin to [vifmimg](https://github.com/cirala/vifmimg). 

As far as I can tell, the drawing problem I mentioned in #851, as well as the bashism are fixed. Besides that I added moving gif, pdf, audio, video, epub, and font preview. 

One issue is that in using the FIFO to keep the image on the screen without a `read -r`(which the other PR used and which caused the next image to not be drawn), is that when resizing the window the image origin stays at the original location when the ueberzug FIFO queue was opened.
 
The only way I can see to fix this is to let nnn pass the SIGWINCH to its children, which I believe should be sent by ncurses on window resize. I think SIGWINCH could then be trapped in `preview-tui` to close/reopen the ueberzug FIFO causing the image to be drawn correctly.

I did my best to adhere by the current modality of the script although I am not sure about the factoring. I suppose the gif/pdf etc preview can be refactored to be included in the other image drawing methods as well although I guess only kitty would be useful for pdf/epub preview. Feel free to refactor however you like if you are interested in merging this.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -34,6 +34,7 @@
 #                To use:
 #                1. install ImageMagick
 #                2. set/export USE_GIFPREVIEW as 1
+#                3. set/export LOOP_GIF as 1 if you want to loop gif preview
 #    - optional: ffmpeg for audio thumbnails
 #                To use:
 #                1. install ffmpeg
@@ -97,8 +98,9 @@ USE_AUDIOPREVIEW="${USE_AUDIOPREVIEW:-0}"
 USE_FONTPREVIEW="${USE_FONTPREVIEW:-0}"
 USE_EPUBPREVIEW="${USE_EPUBPREVIEW:-0}"
 USE_PDFPREVIEW="${USE_PDFPREVIEW:-0}"
-USE_GIFPREVIEW="${USE_GIFPREVIEW:-0}"
 USE_OFFICEPREVIEW="${USE_OFFICEPREVIEW:-0}"
+USE_GIFPREVIEW="${USE_GIFPREVIEW:-0}"
+LOOP_GIFS="${LOOP_GIFS:-0}"
 PAGER="${PAGER:-less -R}"
 
 if [ "$USE_UEBERZUG" -ne 0 ]; then
@@ -258,9 +260,12 @@ preview_ueberzug() {
                         mkdir -p "/tmp/$3"
                         convert -coalesce "$3" "/tmp/$3/${3##*/}.png"
                     fi
-                        for frame in $(find /tmp/"$3"/*.png | sort -V); do
-                            ueberzug_layer "$1" "$2" "$frame"
-                            sleep 0.1
+                        while true; do
+                            for frame in $(find /tmp/"$3"/*.png | sort -V); do
+                                ueberzug_layer "$1" "$2" "$frame"
+                                sleep 0.1
+                            done
+                            [ "$LOOP_GIFS" -eq 0 ] && return
                         done &
                         gifloop=$!
                         return
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -189,19 +189,19 @@ preview_file () {
         fi
     elif [ "$encoding" = "binary" ]; then
         if [ "$USE_UEBERZUG" -ne 0 ] && exists ueberzug; then
-            if [ "$ext" = "gif" ] && [ "$USE_GIFPREVIEW" -ne 0 ]; then
+            if [ "$ext" = "gif" ]; then
                 preview_ueberzug "$cols" "$lines" "$1" "gif"
             elif [ "${mimetype%%/*}" = "image" ]; then
                 ueberzug_layer "$cols" "$lines" "$1"
-            elif [ "${mimetype%%/*}" = "audio" ]; then
+            elif [ "$USE_AUDIOPREVIEW" -ne 0 ] && [ "${mimetype%%/*}" = "audio" ]; then
                 preview_ueberzug "$cols" "$lines" "$1" "audio"
-            elif [ "${mimetype%%/*}" = "video" ]; then
+            elif [ "$USE_VIDEOPREVIEW" -ne 0 ] && [ "${mimetype%%/*}" = "video" ]; then
                 preview_ueberzug "$cols" "$lines" "$1" "video"
-            elif [ "$ext" = "pdf" ]; then
+            elif [ "$USE_PDFPREVIEW" -ne 0 ] && [ "$ext" = "pdf" ]; then
                 preview_ueberzug "$cols" "$lines" "$1" "pdf"
-            elif [ "$ext" = "epub" ]; then
+            elif [ "$USE_EPUBPREVIEW" -ne 0 ] && [ "$ext" = "epub" ]; then
                 preview_ueberzug "$cols" "$lines" "$1" "epub"
-            elif [ "${mimetype%%/*}" = "font" ]; then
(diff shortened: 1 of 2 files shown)


luukvbaal:
Sorry, forgot some things while adopting my personal `preview-tui` to the upstream version.

Also added support for previewing office documents.

luukvbaal:
Sorry, I didn't properly test the WINCH trap. I'm not sure why but when I send a SIGWINCH to `preview-tui` the trap does not seem to trigger. Thought it might have something to do with the fact that that ueberzug is running in the background according to [docs](https://www.gnu.org/software/bash/manual/html_node/Signals.html#Signals)(assuming its the same in POSIX, couldn't find equivalent docs). However when I detach the `ueberzug` process with `nohup` the signal still doesnt trigger the trap. 

In my experience this PR works flawlessly in static windows but I'm afraid I'm gonna need assistance in order to get window resizing to work. Not sure how much time I will have to take another look at that.

luukvbaal:
I refactored the preview generation to be used by `kitty`, `catimg` and `viu` as well. Not sure how useful document preview with `catimg` and `viu` is but it does work. Also tested it in `kitty` which works as well (albeit a lot slower than `ueberzug`).

luukvbaal:
As far as I can tell the only defect right now is the image origin for window resizing and using `ueberzug` in `kitty`. When using `ueberzug` in `kitty` the image is drawn with its origin at the center of the entire window as opposed to the center of the preview split.
image
Using `tmux` the image is drawn correctly in the preview split for a static window.
image
When resizing the window the image is drawn incorrectly with its origin as it was before resizing.
image

The preview [script](https://github.com/cirala/vifmimg) for `vifm` can prevent this by using the coordinates passed by `vifm` but I'm not sure how realistic that is to add to `nnn` since the preview split is created by external programs.

As I said before I thought sending the `SIGWINCH` signal from `nnn` to its child plugins could be used to trap and restart the `ueberzug` fifo process, setting the origin to the new preview split but I wasn't able to get `preview-tui` to trap any signals.

luukvbaal:
Added support for `atool` archive preview as already suggested in the [wiki](https://github.com/jarun/nnn/wiki/Usage#configuration). 

Also somewhat opinionated changes to `PAGER` env and `tree` command with an for a more satisfying (IMO) directory preview.

jarun:

luukvbaal:
I managed to get window resize working, albeit a bit hacky. Added a `pkill -f -n preview-tui` in the `WINCH` trap to prevent multiple `nnn_fifo()` loops from spawning. Can't really follow why that's necessary but so be it, if someone knows a more clean way to do it feel free to assist.

leovilok:
There is a lot of work there, this makes the plugin quite complete, thank you @luukvbaal !

I could not review every line thoughtfully, but it looks good overall.

An issue I see (but not really specific to this PR) is that this plugin was supposed to be simple and hackable, and this is less and less the case ... we're closer and closer to ranger's preview stuff.
We might want to add a new `preview-simple` plugin with as few code as possible to show the basic of FIFO usage, for people to base their plugin ideas on.

luukvbaal:
Yeah I can see why you would like to keep it simple, I already added some quite personal preview options and was about to add some more but I can refrain from adding those to this PR and keep it hackable. The main feature here is the `ueberzug` integration which I think works really well now and could benefit users of the script.

jarun:
The nature of the script is, the more you add the more personal choices creep in. I see several new categories added.

I suggest we put this script somewhere in the wiki and link it from the minimal script that we have today so people can refer to both.

jarun:
Please rename this to `preview-tui-ext` (ext for extended) and let's have ti as a separate plugin. I need this by next Monday as I am planning to make a release.

jarun:
Closed my mistake. Reopened.

luukvbaal:
I assume `preview-tui` should be reverted to how it was on previous release then, since it included non-functional `ueberzug` support on master as far as I could tell. Let me know if you disagree.

luukvbaal:
Updated the [Live-previews](https://github.com/jarun/nnn/wiki/Live-previews) page of the wiki.

jarun:
Yes, I think we should remove `ueberzug` from `preview-tui` and all extra perks in preview-tui-ext.

luukvbaal:
Alright, I already did so in https://github.com/jarun/nnn/pull/869/commits/d913aa214efffd4ff4a47618d2185f7540b901c2. I might add support for previewing other file types in the future but don't think I will before Monday so as far as I'm concerned you can merge this.

jarun:
Thank you!

luukvbaal:
Just noticed the plugin only works when starting `nnn` with the `-P` flag. When running the plugin manually the `SIGWINCH` trap seems to trigger in the first instance of the script, killing itself. I'll try and see if I can fix it but I don't fully understand why the two cases(`-P` flag vs manual) are different so idk...

jarun:
> When running the plugin manually the SIGWINCH trap seems to trigger in the first instance of the script, killing itself.

That is the common use case. I am not sure why you didn't test that earlier. Anyway, I am kind of blind about plugins, so please fix it yourself. @leovilok may be able to help you.

jarun:
This also means we have now thrown away ueberzug support from preview-tui which was added by @Lucas-Summers. I am sure if he has added at least it was working partially.

@luukvbaal do you see what kind of problems we encounter when we let in too many personal preferences into plugins? And the major problem with that is we do not have the time to play around and fix scripts for everyone's setups anymore.

@Lucas-Summers can you please check whatever issue @luukvbaal mentioned with ueberzug is a real thing and if we can fix it in preview-tui?

jarun:
@luukvbaal is it possible to add support for just `ueberzug` in `preview-tui` with all issues fixed? If yes, please raise the PR.

luukvbaal:
> > When running the plugin manually the SIGWINCH trap seems to trigger in the first instance of the script, killing itself.
> 
> That is the common use case. I am not sure why you didn't test that earlier. Anyway, I am kind of blind about plugins, so please fix it yourself. @leovilok may be able to help you.

Didn't imagine the logic would be different for the two cases. I always start `nnn` with the `-P` flag, hence why I didn't notice it before.

luukvbaal:
> This also means we have now thrown away ueberzug support from preview-tui which was added by @Lucas-Summers. I am sure if he has added at least it was working partially.
> @Lucas-Summers can you please check whatever issue @luukvbaal mentioned with ueberzug is a real thing and if we can fix it in preview-tui?

I'm also curious if @Lucas-Summers can confirm the issues I encountered, he at least acknowledged some problems in https://github.com/jarun/nnn/issues/781#issuecomment-796349743.

> @luukvbaal is it possible to add support for just `ueberzug` in `preview-tui` with all issues fixed? If yes, please raise the PR.

I'm sure it is, however earlier you said to remove `ueberzug` from `preview-tui`.
> Yes, I think we should remove `ueberzug` from `preview-tui` and all extra perks in preview-tui-ext.

 In any case I first have to figure out how to make it work when calling it manually and when started with the `-P` flag simultaneously. Only managed to make it work for either case...

> @luukvbaal do you see what kind of problems we encounter when we let in too many personal preferences into plugins? And the major problem with that is we do not have the time to play around and fix scripts for everyone's setups anymore.

All "personal preferences" I added remained optional so those shouldn't cause any problems, not testing running the plugin manually is my bad.

luukvbaal:
Figured it out, I'll raise a PR later

jarun:
Thanks!

PR #875, opened 2021-03-13
Fix preview-tui-ext and ueberzug support for preview-tui

luukvbaal:


luukvbaal:
Feel free to delete `preview-tui-ext` if u feel that its unnecessary and up to the user to define more preview options.

luukvbaal:
In any case I think it makes more sense to just use whats available so I got rid of the config variables  with precedence `kitty` > `ueberzug` > `catimg` > `viu`.

jarun:
Thank you so much!

jarun:
Thank you so much!

PR #877, opened 2021-03-14
cleanup fifo

luukvbaal:
I noticed leftover fifo files in `TMPDIR`. Not sure how important since its in `TMPDIR` anyways but this cleans up the fifo for `ueberzug` when closing the preview pane.


jarun:
Thank you!

PR #891, opened 2021-03-17
preview-tui(-ext) subshell job control

luukvbaal:
Fixes pager manipulation in `preview-tui(-ext)`. The job control added in https://github.com/jarun/nnn/commit/2acc7bd3246674c8f1cc039e36d1afb5b4287824 was non-functional for me in `dash`, only working in `bash`. I assume it was supposed to work in `dash` as well @leovilok, any idea why it doesn't? The `%- %+` seems to be POSIX compliant so I'm not sure but running that version of the script results in an extra pager spawning whenever the selection changes when ran in `dash`, not killing the previous pager. 
```
luuk       65256    3383  0 21:17 pts/6    00:00:00  \_ sh /home/luuk/.config/nnn/plugins/preview-tui
luuk       65265   65256  0 21:17 pts/6    00:00:00      \_ less -R
luuk       65276   65256  0 21:17 pts/6    00:00:00      \_ less -R
luuk       65291   65256  0 21:17 pts/6    00:00:00      \_ less -R
luuk       65302   65256  0 21:17 pts/6    00:00:00      \_ less -R
luuk       65313   65256  0 21:17 pts/6    00:00:00      \_ less -R
luuk       65324   65256  0 21:17 pts/6    00:00:00      \_ less -R
luuk       65335   65256  0 21:17 pts/6    00:00:00      \_ less -R
luuk       65348   65256  0 21:17 pts/6    00:00:00      \_ less -R
```
This leaves the preview window irresponsive to input as it has multiple pagers running inside. 

Anyways it becomes even more difficult to keep track of the PIDs with the subshells introduced by the `ueberzug` integration. 
This is the only way I could come up with to keep track of the previewer PIDs. If anyone knows of a way to do this without writing to disk(`$TMPDIR`), do tell.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -55,6 +55,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [picker](picker) | Pick files and list one per line (to pipe) | sh | nnn |
 | [preview-tabbed](preview-tabbed) | Tabbed/xembed based file previewer | bash | _see in-file docs_ |
 | [preview-tui](preview-tui) | Tmux/kitty/xterm/`$TERMINAL` based file previewer | sh | _see in-file docs_ |
+| [preview-tui-ext](preview-tui-ext) | Meant to be an exhaustive version of [preview-tui](preview-tui) | sh | _see in-file docs_ |
 | [pskill](pskill) | Fuzzy list by name and kill process or zombie | sh | fzf, ps, sudo/doas |
 | [renamer](renamer) | Batch rename selection or files in dir | sh | [qmv](https://www.nongnu.org/renameutils/)/[vidir](https://joeyh.name/code/moreutils/) |
 | [ringtone](ringtone) | Create a variable bitrate mp3 ringtone from file | sh | date, ffmpeg |
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -120,6 +120,7 @@ print_bin_info() {
 }
 
 preview_file () {
+    clear
     # Trying to use pistol if it's available.
     if [ "$USE_PISTOL" -ne 0 ] && exists pistol; then
         fifo_pager pistol "$1"
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -58,7 +58,7 @@
 #   without extra dependencies.
 #
 # Shell: POSIX compliant
-# Authors: Todd Yamakawa, Léo Villeveygoux, @Recidiviste, Mario Ortiz Manero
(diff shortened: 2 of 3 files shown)


jarun:
@leovilok can you please review?

luukvbaal:
Thoughts on adding to/replacing the current `preview-tui` demo on the [README](https://github.com/jarun/nnn/blob/master/plugins/README.md) with [this](https://s4.gifyu.com/images/nnn-preview-tui.gif) one I made to demo the `ueberzug` preview?@Kabouik @jarun

jarun:
Definitely a great improvement with new capabilities. Please go ahead.

jarun:
But please don't upload the gif in the repo. Link to it.

luukvbaal:
Should I add or replace it?

luukvbaal:
Replaced it, tell me if you want to keep both.

jarun:
No, one is enough.
But we need to reduce the size. It's 2.32MB!

try gifsicle...

luukvbaal:
2.32MB is large for a 33 second long 1080p gif? I can reduce the resolution I suppose since I didn't use the full size in the readme, I'll see how large it gets. What is an appropriate size you think?

jarun:
It took long enough on my system to show up. I have a reasonable network connection speed. If I were a user, I would just miss it and scroll down.

Keep it around 1 MB.

asciinema is the best because it doesn't load the whole file immediately. Can we do something like that? Take a single frame and put an arrow that links to the original gif.

luukvbaal:
`asciinema` won't record the `ueberzug` preview so that wouldn't work here. The gif for `preview-tabbed` on the [wiki](https://github.com/jarun/nnn/wiki/Live-previews) is 1.61MB as well and is only 7 seconds long so... I dont think 2.32MB is outlandishly large but I halved the resolution resulting in a 800kB gif(more blurry on the readme of course), should I use that?

luukvbaal:
Went ahead and merged the new preview demo (made it ~1MB) in #896, forgot about the other `README.md` change in this PR but whatever...

jarun:
Please do not self-merge. Wait for at least one other collaborator to approve. Only I merge directly because of my odd timings but get it reviewed by @0xACE or @KlzXS if I am not sure of the change.

jarun:
Thank you!

0xACE:
There has been a lot of activity on this topic the past few weeks and I have exhausted all my allotted time resources elsewhere.

I will not be able to set my eyes on this project for the next week at least. So don't hold your breath.

PR #896, opened 2021-03-19
preview-tui demo

luukvbaal:


jarun:
Is it playing correctly for you? I don;t see the image previews...

luukvbaal:
Uhm yeah playing fine here

jarun:
I think we can do this -> use the previous 2.32 MB gif, but use an image with an arrow on it to indicate that clicking on it would take you to the demo.

jarun:
That way, the image will render faster on the page and interested people can click on it to view the demo.

luukvbaal:
Fine I'll do that tommorow, It's late here hehe.

jarun:
I'm reverting the earlier commits. Please don't self merge.

PR #900, opened 2021-03-20
preview-tui fixes

luukvbaal:
Redraws preview on resize for `preview-tui` and `preview-tui-ext`.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -192,7 +192,7 @@ image_preview() {
         viu -t "$3" &
         echo "$!" > "$GIFPID"
     else
-        fifo_pager print_bin_info "$1"
+        fifo_pager print_bin_info "$3"
     fi
 }
 
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -232,6 +232,7 @@ preview_fifo() {
         echo "$selection" > "$CURSEL"
     done
     [ "$TERMINAL" != "kitty" ] && exists ueberzug && rm "$FIFO_UEBERZUG"
+    pkill -P "$$" >/dev/null 2>&1
     rm "$PAGERPID" "$GIFPID" "$CURSEL" >/dev/null 2>&1
 }
 
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -232,8 +232,8 @@ preview_fifo() {
         echo "$selection" > "$CURSEL"
     done
     [ "$TERMINAL" != "kitty" ] && exists ueberzug && rm "$FIFO_UEBERZUG"
-    pkill -P "$$" >/dev/null 2>&1
     rm "$PAGERPID" "$GIFPID" "$CURSEL" >/dev/null 2>&1
+    pkill -P "$$" >/dev/null 2>&1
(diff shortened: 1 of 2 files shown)


luukvbaal:
Just noticed a regression when opening `preview-tui` manually instead of with the `-P` flag. Not sure when this was introduced since I usually use the `-P` flag but the `tmux` split is not closed when quitting `nnn`.

Previously my changes to `preview-tui` didn't work at all for the manual case and this was fixed by adding 
the `[ "$PREVIEW_MODE" ] ` to the `SIGWINCH` trap. I suppose this happens because creating the window split triggers `SIGWINCH`, causing `preview-tui` to kill itself.

I'm still not sure why its necessary and how to distinguish the two cases in the plugin script so any suggestions would be appreciated. Can't tell why the "disowned" pager is left over when quitting `nnn` though and when it was introduced.

luukvbaal:
The latest commit that properly closes the `tmux` preview split when running `preview-tui` by plugin key is https://github.com/jarun/nnn/commit/1571ea2b35f4944f8515ae7647181bb206e10113.

It seems I only confirmed it was actually opening in the first place after https://github.com/jarun/nnn/commit/ac72d2c66ed6d1e76ea7fb23bf2832cd5ce62491 and somehow managed to miss the fact that it doesn't close when quitting `nnn` since adding `ueberzug` previews.

Had no one else noticed this?

luukvbaal:
The root problem is that somehow the program execution is different when opening `preview-tui` with its plugin key or with the `-P` flag. Leading to the following two problems:
1. The preview pane is left open when exiting `nnn` after running `preview-tui` with its plugin key.
2. When running `preview-tui` with its plugin key, an extra `preview_file "$1"` is necessary before `preview_fifo &` to actually preview the first selection. This causes `preview-tui` openened by the `-P` flag to flicker as well since it's running `preview_file` twice. I guess the FIFO is not populated with the first selection in the first case but is populated in the latter.

Not sure if this can/should be fixed in `nnn` such that the two cases can be treated the same in `preview-tui` or if it should be fixed in `preview-tui` and if so I'm not sure how yet.

luukvbaal:
> 1. The preview pane is left open when exiting `nnn` after running `preview-tui` with its plugin key.
> 2. When running `preview-tui` with its plugin key, an extra `preview_file "$1"` is necessary before `preview_fifo &` to actually preview the first selection. This causes `preview-tui` openened by the `-P` flag to flicker as well since it's running `preview_file` twice. I guess the FIFO is not populated with the first selection in the first case but is populated in the latter.

1. should be fixed in https://github.com/jarun/nnn/pull/900/commits/37092c1b09ce2ffb5a69f43774f28b3618207251 and actually concerned any `preview-tui` ran in an existing `tmux` session as opposed to started with `tmux` as discussed in #901. Difference in process tree I suppose.
2. remains but is just a minor annoyance, not sure if worth trying to fix in `nnn`.

jarun:
@leovilok please take a look at problem 2. I have a hunch we might have seen this before.

jarun:
Thank you!

PR #903, opened 2021-03-21
fix preview-tui-ext fontpreview

luukvbaal:
Never actually tested it, oops. This works for otf and ttf.


jarun:
Thank you!

jarun:
I think it might be a good idea to make generate preview return something when preview is not possible.

I see the problem with audio files (mp3, flac)... there si no embedded thumbnail so te preview pane is blanc. Can we print the regular binary info in this case?

luukvbaal:
@jarun havent actually encountered such a file myself but I suppose https://github.com/jarun/nnn/pull/904 should work.

PR #904, opened 2021-03-21
preview-tui-ext generate_preview fallback

luukvbaal:
Fallback for case described in https://github.com/jarun/nnn/pull/903.

### DIFF
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -255,7 +255,11 @@ generate_preview() {
             video) ffmpegthumbnailer -i "$3" -o "$TMPDIR/$3.png" -s 0 -q 10 >/dev/null 2>&1 ;;
         esac
     fi
-    image_preview "$1" "$2" "$TMPDIR/$3.png"
+    if [ -f "$TMPDIR/$3.png" ]; then
+        image_preview "$1" "$2" "$TMPDIR/$3.png"
+    else
+        fifo_pager print_bin_info "$3"
+    fi
 }
 
 image_preview() {


jarun:
Thank you!

luukvbaal:
Alternatively we could restore the `print_bin_info` removed in https://github.com/jarun/nnn/commit/4a45e45b3077f79b0b6f34ca6b2d1d29413f516e and just have it return when no image is generated.

jarun:
That's also fine. Now it works for most cases other than `m4a` files. Please check ones.

luukvbaal:
I will tomorrow

jarun:
Sure thing! GN!

jarun:
One more improvement would be interesting - to view the images centered. However, not required if it's too much work.

luukvbaal:
We can also fit the image better using one of the `scaler` [options] https://github.com/seebye/ueberzug#command-formats) from `ueberzug`(assuming you're talking about `ueberzug` previews).

 I think `fit_contain` works well in most cases, but even `distort` could be used if you really want to fill the preview pane. Centering the image could also be possible but do you still think it's necessary when using `scaler`?

luukvbaal:
> That's also fine. Now it works for most cases other than `m4a` files. Please check once.

Could not reproduce using a sample file downloaded `here`, would it be possible to share an affected file?

jarun:
Directly running mediainfo shows the file info.

Get the audio of [this](https://www.youtube.com/watch?v=4q9UafsiQ6k) with youtube-dl in m4a format and see.

luukvbaal:
> Get the audio of this with youtube-dl in m4a format and see.

That previews a black image on `preview-tui-ext` master for me, can you confirm that that's what you're seeing? Not exactly sure why since running `ffmpeg` on it manually errors with `Output file #0 does not contain any stream` without it producing any output file. Somehow `ffmpeg` ran from `preview-tui-ext` creates an empty image.

luukvbaal:
Oh this happens because the mime-type for the `m4a` file is `video/mp4` so `ffmpegthumbnailer is being run to generate the preview rather than `ffmpeg`.

jarun:
I think we should use `mimetype` if it is available. It reports correctly:

```
 ~/Music$ mimetype $nnn
Pearl Jam - Black.m4a: audio/mp4 
2 ~/Music$ file -Lb --mime-type $nnn          
video/mp4
```

luukvbaal:
> I think we should use `mimetype` if it is available. It reports correctly:

We are using `mimetype` for video/audio, the problem is that `m4a` does not always have the same `mimetpe`. Mine reports:
```
file --mime-type ~/test.m4a
/home/luuk/test.m4a: video/mp4
```
causing it to run `ffmpegthumbnailer` which produces an empty image.

PR #905, opened 2021-03-21
preview-tui-ext jpg previews

luukvbaal:
Switch to generating `jpg` previews over `png` since it seems to be a lot faster, especially noticeable on larger `pdf` files.

Also add image scaling for `ueberzug` using one of the `scaler` [options](https://github.com/seebye/ueberzug#command-formats) to better fit the preview pane.


jarun:
Thank you!

PR #906, opened 2021-03-21
preview-tui-ext ffmpegthumbnailer fallback

luukvbaal:
Should fix https://github.com/jarun/nnn/pull/904#issuecomment-803508048, please confirm.
If there are more cases where preview generation can fail it might indeed be better to refactor and check `generate_preview` return status.


jarun:
Along with this, we should also use `mimetype` if it is available:

```
 ~/Music$ mimetype -Lb $nnn
audio/mp4
2 ~/Music$ 
2 ~/Music$ file -Lb --mime-type $nnn          
video/mp4
```

jarun:
The mimetype as first option needs to go to `nuke`, `preview-tui`, `preview-tui-ext` and `preview-tabbed`.

jarun:
Thank you!

jarun:
Forget about `mimetype`: it beaks opening shell scripts - `application/x-shellscript`. We can't retest all basic stuff due to `m4a`.

PR #907, opened 2021-03-21
fill preview-tui preview

luukvbaal:
Was this here for any reason? Filling the entire preview pane looks better for images in my opinion.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -143,7 +143,7 @@ preview_file () {
     if [ -n "$ext" ]; then
         ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
     fi
-    lines=$(($(tput lines)-1))
+    lines=$(tput lines)
     cols=$(tput cols)
 
     # Otherwise, falling back to the defaults.
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -148,8 +148,8 @@ preview_file () {
 
     # Detecting the exact type of the file: the encoding, mime type, and
     # extension in lowercase.
-    encoding="$(file -Lb --mime-encoding -- "$1")"
-    mimetype="$(file -Lb --mime-type -- "$1")"
+    encoding="$(file -bL --mime-encoding -- "$1")"
+    mimetype="$(file -bL --mime-type -- "$1")"
     ext="${1##*.}"
     if [ -n "$ext" ]; then
         ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -232,7 +232,7 @@ generate_preview() {
             audio) ffmpeg -i "$3" "$TMPDIR/$3.jpg" -y >/dev/null 2>&1 ;;
             epub) gnome-epub-thumbnailer "$3" "$TMPDIR/$3.jpg" >/dev/null 2>&1 ;;
             font) fontpreview -i "$3" -o "$TMPDIR/$3.jpg" >/dev/null 2>&1 ;;
-            gif) if exists ueberzug || [ "$TERMINAL" = "kitty" ]; then
+            gif) if exists ueberzug && exists convert || [ "$TERMINAL" = "kitty" ]; then
                     if [ ! -d "$TMPDIR/$3" ]; then
                         mkdir -p "$TMPDIR/$3"
                         convert -coalesce "$3" "$TMPDIR/$3/${3##*/}.jpg"
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -154,7 +154,7 @@ preview_file () {
     if [ -n "$ext" ]; then
         ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
     fi
-    lines=$(($(tput lines)-1))
+    lines=$(tput lines)
     cols=$(tput cols)
 
     # Otherwise, falling back to the defaults.
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -148,8 +148,8 @@ preview_file () {
 
     # Detecting the exact type of the file: the encoding, mime type, and
     # extension in lowercase.
-    encoding="$(file -bL --mime-encoding -- "$1")"
-    mimetype="$(file -bL --mime-type -- "$1")"
+    encoding="$(file -Lb --mime-encoding -- "$1")"
+    mimetype="$(file -Lb --mime-type -- "$1")"
     ext="${1##*.}"
     if [ -n "$ext" ]; then
         ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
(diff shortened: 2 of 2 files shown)


plugins/preview-tui
```diff
@@ -143,7 +143,7 @@ preview_file () {
     if [ -n "$ext" ]; then
         ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
     fi
-    lines=$(($(tput lines)-1))
```
luukvbaal, line 146: Was this here for any reason?

jarun: Don't think so.

jarun:
Thank you!

jarun:
Can you please have a look if the non-text previews can be centered? That would really look great.

luukvbaal:
I assume you mean centered horizontally when an portrait style image is stretched over the height and is much smaller than the full width of the preview pane?

luukvbaal:
I think you can do it with something like this.
```
imagewidth="$(identify -format "%w" "$3")"
imageheight="$(identify -format "%h" "$3")"
if [ "$SPLIT" = "v" ]; then
    previewwidth="$(($(xwininfo -id $WINDOWID | sed -n 's/Width: \([0-9]*\)/\1/p')) / 2)"
    previewheight="$(xwininfo -id $WINDOWID | sed -n 's/Height: \([0-9]*\)/\1/p')"
else
    previewwidth="$(xwininfo -id $WINDOWID | sed -n 's/Width: \([0-9]*\)/\1/p')"
    previewheight="$(($(xwininfo -id $WINDOWID | sed -n 's/Height: \([0-9]*\)/\1/p')) / 2)"
fi

if (image does not fill preview width); then 
     xoffset=???
else
    xoffset=0
fi
```
it should then be possible to calculate the if condition and the xofsset, not sure how yet.

luukvbaal:
Which can then be passed to `ueberzug` as the `x`coordinate in `ueberzug_layer()`. Not sure if the same would be possible for `viu`/`catimg`/`kitt`.

luukvbaal:
Like this I think.
```
imageratio="$((imagewidth / imageheight))"
previewratio="$((previewwidth / previewheigth))"
if [ "$imageratio" -lt "$previewratio" ]; then 
    xoffset=??
else
    xoffset=0
fi
```

luukvbaal:
The `xoffset` should then be (`previewwidth` - `ueberzugwidth`) / 2.
`ueberzugwidth` is the width of the drawn image when the height of the preview pane is filled and the width of the image is potentially a lot smaller than the width of the preview pane. 
I think you might need the terminal cell dimensions for this? 
Is this what you had in mind to be centered @jarun?

luukvbaal:
```
imageheight="$(identify -format "%h" "$3")"
if [ "$SPLIT" = "h" ]; then
		panewidth="$(($(xwininfo -id $WINDOWID | sed -n 's/Width: \([0-9]*\)/\1/p') / 2))"
                paneheight="$(xwininfo -id $WINDOWID | sed -n 's/Height: \([0-9]*\)/\1/p')"
	else
		panewidth="$(xwininfo -id $WINDOWID | sed -n 's/Width: \([0-9]*\)/\1/p')"
		paneheight="$(($(xwininfo -id $WINDOWID | sed -n 's/Height: \([0-9]*\)/\1/p') / 2))"
	fi
		imageratio="$(echo "scale=5;$imagewidth / $imageheight" | bc)"
		paneratio="$(echo "scale=5;$panewidth / $paneheight" | bc)"
	if [ "$(echo "$imageratio < $paneratio" | bc)" -eq "1" ]; then
		previewwidth="$(echo "scale=5;$paneheight * $imageratio" | bc)"
		xoffset="$(echo "scale=0;($panewidth - $previewwidth) / 2" | bc)"
		echo $panewidth $previewwidth $xoffset
	else
		  xoffset=0
	fi
```
`echo $panewidth $previewwidth $xoffset` actually yields the expected result: 960 819.08320 70.
I thought that would shift the `ueberzug` preview by 70 pixels but it seems to shift it way out of view, maybe 70 terminal cells?

luukvbaal:
Dividing by 10 yields an almost centered image. 

If the coordinate arguments passed to `ueberzug` are indeed `cols` and `lines` it won't be possible to exactly center the image. 

Not sure if this semi-centering is better than the standard NW alignment.

luukvbaal:
https://github.com/luukvbaal/nnn/commit/d3add22689c2672e007808357888547531320a3f is the best I could come up with. See if it works for you and is what you meant by centering. Again doesn't seem possible to exactly center the image because the offset argument to `ueberzug` seems to be in terminal lines/cols, not pixels.

Not sure if pixel addressing for the offset is possible in the `ueberzug` `python` library and if a PR could be made to expose pixel offset to the `json` interpreter.

jarun:
Leave it. I see other side effects. We are great with what we have. Thanks for checking it out!

luukvbaal:
Chaning 
``` 
x = int((self.x + pane_offset.left) * term_info.font_width +
term_info.padding_horizontal)
```

to 

``` 
x = int(pane_offset.left) * term_info.font_width + int(self.x)
term_info.padding_horizontal)
```

in `ueberzug`'s `ui.py` produces the desired result. Exposing pixel addressing like this to the `json` parser should be fairly simple I suppose but afaik seebye does not accept PRs. Don't know if he would be willing to add this functionality.

jarun:
It's OK. Don't bother much.

luukvbaal:
Also noticed `ueberzug` has a signal handler for `SIGWINCH` while looking through the source; wondering why `ueberzug_refresh()`  is necessary in `preview-tui`.

luukvbaal:
Figured out how to pass the correct value from `preview-tui`, https://github.com/luukvbaal/nnn/commit/a822d8119c79a117ca3c0d8a075c455ca2404f20 works to center the image horizontally. Is this what you meant? What side effects are you seeing?

luukvbaal:
[This](https://github.com/luukvbaal/nnn/blob/df7987d34680ce3f592965237073e5b89b83c262/plugins/preview-tui-ext) works well to center the image horizontally with terminal cols precision. I don't see any side effects myself.

Centering on pixel level would require a change in `ueberzug` as far as I can tell. Is this what you had in mind @jarun, should we merge this?

luukvbaal:
I don't think really think the approximate centering looks better than the standard NorthWest alignment, so I suggest we leave this be unless pixel offset in `ueberzug` gets (re)added (removed in https://github.com/seebye/ueberzug/commit/aae4e557192868a8f63e3175f213fe632a39404a).

jarun:
Completely agree,

luukvbaal:
I suppose there's no way to let `nnn`'s statusbar text spill over into a horizontal `tmux` split is there?

jarun:
Nope.

luukvbaal:
Yeah I figured. 

What do you think about converting non `jpeg` images to `jpeg` for a somewhat slower first preview in exchange for a smoother experience afterwards?

jarun:
I think converting and showing huge PNGs can be faster than rendering them as PNGs.

luukvbaal:
Yep, I also thought it might be good to constraint the preview image dimensions to the preview pane dimensions. No point in generating images larger than the preview pane.

luukvbaal:
That approach can lead to blurry previews however when resizing the window after the preview has been generated. Perhaps just hardcoding the constraints to half the screen resolution makes sense?

jarun:
No, reduction in size is even more painful. Perhaps leave the mages as they are. People may actually want to view them exact.

luukvbaal:
More expensive to convert(once), less expensive to display(multiple times). I'd say the final the end goal of previews is browsing back and forth as smooth as possible no?

PR #910, opened 2021-03-23
cache and constraint preview dimensions

luukvbaal:
Sorry for the large amount of changes in a single commit, I'll do a review and comment.


plugins/preview-tui-ext
```diff
@@ -76,6 +76,9 @@ USE_PISTOL="${USE_PISTOL:-0}"
 PAGER="${PAGER:-less -P?n -R}"
 ARCHIVES="$(echo "$NNN_ARCHIVE" | sed 's/.*(\(.*\)).*/\1/;s/|/ /g')"
 TMPDIR="${TMPDIR:-/tmp}"
+NNN_PREVIEWDIR="${NNN_PREVIEWDIR:-$XDG_CACHE_HOME}/nnn"
```
luukvbaal, line 79: Allow choosing directory for generated image previews, `XDG_CACHE_HOME` default?

plugins/preview-tui-ext
```diff
@@ -76,6 +76,9 @@ USE_PISTOL="${USE_PISTOL:-0}"
...
+NNN_PREVIEWWIDTH="${NNN_PREVIEWWIDTH:-1920}"
+NNN_PREVIEWHEIGHT="${NNN_PREVIEWHEIGHT:-1080}"
```
luukvbaal, line 81: Use maximum dimensions for preview generation

plugins/preview-tui-ext
```diff
@@ -168,19 +171,21 @@ preview_file () {
...
-        elif [ "${mimetype%%/*}" = "image" ]; then
-            image_preview "$cols" "$lines" "$1"
+        if [ "${mimetype%%/*}" = "image" ]; then
+            if [ "${mimetype#*/}" = "gif" ] && exists convert; then
+                generate_preview "$cols" "$lines" "$1" "gif"
+            elif [ ! "${mimetype#*/}" = "jpeg" ]; then
+                generate_preview "$cols" "$lines" "$1" "image"
+            else
+                image_preview "$cols" "$lines" "$1"
+            fi
```
luukvbaal, line 181: Remove `xcf` case and `generate_preview` anything that is not `jpeg`

plugins/preview-tui-ext
```diff
@@ -168,19 +171,21 @@ preview_file () {
...
         elif [ "${mimetype%%/*}" = "audio" ] && exists ffmpeg; then
             generate_preview "$cols" "$lines" "$1" "audio"
         elif [ "${mimetype%%/*}" = "video" ] && exists ffmpegthumbnailer; then
             generate_preview "$cols" "$lines" "$1" "video"
-        elif [ "$ext" = "pdf" ] && exists pdftoppm; then
+        elif [ "${mimetype#*/}" = "pdf" ] && exists pdftoppm; then
             generate_preview "$cols" "$lines" "$1" "pdf"
-        elif [ "$ext" = "epub" ] && exists gnome-epub-thumbnailer; then
+        elif [ "${mimetype#*epub}" != "$mimetype" ] && exists gnome-epub-thumbnailer; then
             generate_preview "$cols" "$lines" "$1" "epub"
```
luukvbaal, line 189: Use `mimetype` where possible

plugins/preview-tui-ext
```diff
@@ -226,19 +231,19 @@ preview_file () {
 }
 
 generate_preview() {
-    if [ ! -f "$TMPDIR/$3.jpg" ]; then
-        mkdir -p "$TMPDIR/${3%/*}"
+  if [ ! -f "$NNN_PREVIEWDIR/$3.jpg" ] || [ -n "$(find -L "$3" -newer "$NNN_PREVIEWDIR/$3.jpg")" ]; then
+        mkdir -p "$NNN_PREVIEWDIR/${3%/*}"
```
luukvbaal, line 235: Only generate preview when it doesn't exist or the file is newer than the existing preview

plugins/preview-tui-ext
```diff
@@ -250,20 +255,16 @@ generate_preview() {
...
+                    mv "$NNN_PREVIEWDIR/${3%/*}/$filename.jpg" "$NNN_PREVIEWDIR/$3.jpg" ;;
+            pdf) pdftoppm -jpeg -f 1 -singlefile "$3" "$NNN_PREVIEWDIR/$3" >/dev/null 2>&1 ;;
+            video) ffmpegthumbnailer -m -s0 -i "$3" -o "$NNN_PREVIEWDIR/$3.jpg" >/dev/null 2>&1 || rm "$NNN_PREVIEWDIR/$3.jpg" ;;
         esac
     fi
-    if [ -f "$TMPDIR/$3.jpg" ]; then
-        image_preview "$1" "$2" "$TMPDIR/$3.jpg"
+    if [ -f "$NNN_PREVIEWDIR/$3.jpg" ]; then
+        image_preview "$1" "$2" "$NNN_PREVIEWDIR/$3.jpg"
     else
```
luukvbaal, line 268: Switch to `NNN_PREVIEWDIR` and constraint dimensions where possible

plugins/preview-tui-ext
```diff
@@ -338,6 +335,8 @@ if [ "$PREVIEW_MODE" ]; then
 
     if [ "$TERMINAL" != "kitty" ] && exists ueberzug; then
         mkfifo "$FIFO_UEBERZUG"
+        trap 'ueberzug_refresh' WINCH
+        trap 'rm "$FIFO_UEBERZUG" "$PAGERPID" "$GIFPID" "$CURSEL"' INT HUP EXIT
         tail --follow "$FIFO_UEBERZUG" | ueberzug layer --silent --parser json &
```
luukvbaal, line 340: Move to pre-existing if statement with same conditions

jarun:
@leovilok please review.

luukvbaal:
I think storing previews on disk and keeping them between reboots makes sense. Can result in a growing cache directory but I think it's worth it. 

Could also create a cleanup script to remove previews from the cache directory of which the files are no longer on disk(in the same location).

jarun:
Do not leave user data in cache please. The user may not even know.

luukvbaal:
What do you suggest then, keeping `NNN_PREVIEWDIR` default as `/tmp` and mentioning setting it to `XDG_CACHE_DIR` if you want to keep previews between reboots?

jarun:
Yes, that would be fine. But do NOT cache anything by default.

leovilok (approved):
Looks good to me.

plugins/preview-tui-ext
luukvbaal: Apparently not all `pdf`'s / `epub`'s have the same `mimetype`, found some with `application/octet-stream`.

jarun:
Thank you!

PR #912, opened 2021-03-23
Show path in terminal title

### LINKED ISSUE #911: Set xterm title
Is it possible to make `nnn` report the working directory (probably just `basename`) to the terminal emulator? This would be similar to what `set title` does in Vim.

TL;DR:

Screen Shot 2021-03-23 at 15 46 14

leovilok:
This is the straightforward way, there is no check on the compatibility of terminals (though I didn't see any issue in linux framebuffer terminal which obviously doesn't implement this feature).

I don't do any cleanup on exit, but it seems to get cleared anyways.

Fixes #911

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5994,7 +5994,7 @@ begin:
 	}
 
 	/* Set terminal window title */
-	printf("\033]2;%s\007", path);
+	printf("\033]2;%s (%s)\007", xbasename(path), path);
 	fflush(stdout);
 
 	if (g_state.selmode && lastdir[0])
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5993,6 +5993,10 @@ begin:
 		setdirwatch();
 	}
 
+	/* Set terminal window title */
+	printf("\033]2;%s\007", path);
+	fflush(stdout);
+
 	if (g_state.selmode && lastdir[0])
 		lastappendpos = selbufpos;
 


leovilok:
Oh I just realised that an explainatory comment in the source could be usefull there, I'm editing the PR.

salutis:
Amazing! That said, here are 2 issues I see with the feature as implemented now:

1. The terminal title is not reset when `nnn` exits. This is perhaps not that big of an issue, but other programs clean after themselves. Then, `mutt` does not.
2. The current implementation, while better than nothing, is not very practical. I would prefer to see `Bibliography` instead of `/Users/salutis/Documents/Bibliography`. The title must be usable when truncated; in other words, the most important information goes first. Vim gets this right, for instance; it sets the title by default to e.g. `Scratchpad.tex (~/Documents) - VIM`. The most important part of the path is `dirname`. Ideally, we would also shorten the home directory to `~` like others do.

leovilok:
I changed the title to "basename (full/path)".

> The terminal title is not reset when nnn exits. This is perhaps not that big of an issue, but other programs clean after themselves. Then, mutt does not.

The title seems to get reset in xterm, but not in other terminals, I'm not sure how to reset it though, set en empty string?

> As a bonus, we would also shorten the home directory to ~ like other programs do.

There is actually already a way to display a shorted version of a path, used to show the current path in the UI when the window is too narrow, might be usefull here ...

salutis:
> I changed the title to "basename (full/path)".

I tried it. Amazing work, @leovilok!

> The title seems to get reset in xterm, but not in other terminals, I'm not sure how to reset it though, set en empty string?

I just played with Vim, and it seems to me that Vim restores the title to back the original one, i.e. to whatever it had been before Vim changed it.

> There is actually already a way to display a shorted version of a path, used to show the current path in the UI when the window is too narrow, might be usefull here ...

A cherry on top!

leovilok:
> Looks like Vim restores the title to back the original one, i.e. to what it had been before Vim changed it to something else.

I'm not sur about what to do there :

- xterm cleans its title by itself, but other terminals don't,
- setting an empty title with `printf("\033]2;\007")` or `printf("\033]0;\007")` doesn't work anywhere I tried and,
- xterm specific escape sequences to store/load titles doesn't seem to work outside of xterm (i tried on urxvt and st).

salutis:
@leovilok That would explain why this is still not fixed in `mutt`. 🤣 Cannot we store it in a variable and restore to that upon exit? Or perhaps we simply leave it be like `mutt`...

leovilok:
I tried in `kitty` and both clearing method actually worked there... so we might try to do something.

Here are 2 branches with each title clearing method, @salutis  can you try them in the teminal emulatrs you care about? (if you want it, of course...)

- https://github.com/leovilok/nnn/tree/term-title-clear-empty
- https://github.com/leovilok/nnn/tree/term-title-clear-stack

leovilok:
> > There is actually already a way to display a shorted version of a path, used to show the current path in the UI when the window is too narrow, might be usefull here ...
> 
> A cherry on top?

Well we don't know how much space we have for the, so we don't know how much to compress the path.

Replacing the $HOME prefix with in a path with '~' looks easy enough to implement, and might be used in other places too, but I won't go into that personnaly.

salutis:
> I tried in `kitty` and both clearing method actually worked there... @salutis can you try them in the teminal emulatrs you care about?

I tried both Kitty and Apple’s Terminal on macOS.

Both branches work, but only one of them works correctly.

Example:

1. I start the terminal emulator.
2. The emulator uses a builtin title, call it _B_.
3. I set the title with `echo -e "\e]0;HELLO\a"`.
4. I run `nnn`. 
5. `nnn` updates the title.
6. I quit `nnn`.

With `term-title-clear-empty`, the title reverts (incorrectly) to _B_. With `term-title-clear-stack`, the title reverts (correctly) to `HELLO`.

TL;DR `term-title-clear-stack` is better.

salutis:
> Replacing the $HOME prefix with in a path with '~' looks easy enough to implement, and might be used in other places too, but I won't go into that personnaly.

Perhaps I could do it. (First time C contributor. Ha!) Found this:

https://stackoverflow.com/questions/15991241/how-do-i-achieve-tilde-expansion-in-c

jarun:
Thank you!

PR #913, opened 2021-03-24
plugin: cleanfilename: clean filename to be more shell-friendly

wustho:
Basically renaming file/dir to be more shell friendly by replacing spaces and any other non-A-Za-z0-9.- char with underscore ().

eg. '[qwer] sfasdf.txt' --> '_qwer__sfasdf.txt'

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -20,6 +20,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [bookmarks](bookmarks) | Use named bookmarks managed with symlinks | sh | fzf |
 | [boom](boom) | Play random music from dir | sh | [moc](http://moc.daper.net/) |
 | [bulknew](bulknew) | Create multiple files/dirs at once | bash | sed, xargs, mktemp |
+| [cleanfilename](cleanfilename) | Clean filename to be more shell-friendly | sh | sed |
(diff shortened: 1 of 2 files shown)


wustho:
Forgot description. Here is one:

Basically renaming file/dir to be more shell friendly by replacing spaces and any other non-A-Za-z0-9._- char with underscore (-).

eg. '[qwer] sfasdf.txt' --> '_qwer__sfasdf.txt'

jarun:
Thank you!

jarun:
Does it handle escape sequences a well? How about single/double quotes?

jarun:
@wustho can you please respond to my questions above?

In addition, does it remove leading and trailing spaces?

wustho:
@jarun hey there, sorry just read my email...

It should handle escape sequence and and single/double quotes as well...

About leading, or trimming, it wont remove them but instead replace them with _. I change IFS to newline and split selection by nullbytes. So `'  qwe  '` will be `'__qwe__'`...

In the cost that it can't handle new line...

jarun:
Please add these finer details to the plugin. Users may assume something else and misuse.

wustho:
Sorry, @jarun , any idea how can I put those details into words? English is not my first lang...

wustho:
Ok, I'll try later... I think we could make it to handle newline also, but it's inefficient to do it with posix sh... Array will totally helps.

jarun:
Feel free to move to bash and use array. Please ensure the documentation is accurate.

PR #915, opened 2021-03-24
Save/load the terminal window title using xterm esc sequences

leovilok:
Restores the previous terminal window title on some xterm compatible terminal emulators (e.g. kitty and Apple's Terminal).

Following report from @salutis in #912.


jarun:
Thank you!

PR #916, opened 2021-03-25
plugins: cleanfilename: fixed to not overwrite existing file

wustho:
When trying to clean filename `qwer[ty` while there is already file named `qwer_ty`, then the new cleaned file will be prepended by '_' become `_qwer_ty` so not to overwrite existing `qwer_ty`.

### DIFF
--- a/plugins/cleanfilename
+++ b/plugins/cleanfilename
@@ -51,7 +51,7 @@ fi
 for i in $targets; do
     if [ "$i" != "$(cleanup "$i")" ]; then
         tmp=''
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

jarun:
@wustho can you update `cleanfilename` to confirm before overwriting any existing file? In the lines of https://github.com/jarun/nnn/pull/979.

It's much safer than overwriting without any confirmation, whatever may be the argument.

jarun:
In case you can't respond here, please respond in the ToDo list.

Issue #267, opened 2019-05-22
[Feature request] Changing file entry order


magpie514:
#### Environment details (Put `x` in the checkbox along with the information)

[x] Operating System: Ubuntu Linux 19.04 amd64/Android 4.4.4
[x] Desktop Environment: KDE Plasma 5
[x] Terminal Emulator: Konsole
[x] Shell: Bash
[ ] Custom desktop opener (if applicable): N/A
[x] Issue exists on `nnn` master

nnn is pretty excellent at doing management tasks, I've been making a habit of using it to do basic quick file management and directory navigation in both my desktop and phone and I'm pretty happy about it so far.

But I have found a little inconvenience, not really a dealbreaker, but using long file display is a bit distracting to me, and I find short file display to be too barebones. I'd be pretty good if it could be customized or reversed so instead of:
` <date> <perms> <indicators> <file/dir name> `
if can start with the file name first like:
`<file/dir name> <indicators> <perms> <date>'`
which is a more familiar layout, personally speaking.

Of course while technically a small change it's just a "would be nice to have" thing, so I understand if you find it not worth the effort, feel free to close the issue if so.

jarun:
The first 3 fields are fixed-length while the file name field varies. The combined length of all the fields is limited by the number of columns your terminal has. So even if we show the names int he first column, we will have to truncate it when it's too long.

It's a huge amount of work to stabilize. we don't have that bandwidth.

You have the following options:

- use a wider terminal
- use `nnn` in the light mode where you see only the filename

luukvbaal:
For anyone still interested in this I'm currently maintaining a [fork](https://github.com/luukvbaal/nnn) that implements this.

Kabouik:
Thanks for it by the way @luukvbaal. I'm not using it currently because I don't use icons in `nnn` so having details before filenames like in `ls -la` is not a particular inconvenience for me. Also I was not sure the fork would be easy to maintain given how actively `nnn` is being developed. But I do like that your fork gives flexibility regarding the order I starred it because I may well try it some day.

This was already discussed recently and it was decided that the compilation option for custom column orders is out of scope for upstream, so I'm not commenting that, but what about the better (in my opinion) right-alignment of file sizes? I don't remember if it was decided to keep this particular feature out.

luukvbaal:
> Also I was not sure the fork would be easy to maintain given how actively nnn is being developed.

[...]

PR #919, opened 2021-03-25
new config: detail column order

luukvbaal:
This is my attempt to implement a new config variable to allow setting the order of the columns in detail mode.

Introduces a new environment variable `NNN_DCOLUMNS` to be set by user preferred order of the following characters:
- "n": entry name
- "o": owner:group
- "p": permission bits
- "P": `ls -l` style permissions
- "t": time

I'm not sure about the quality of the code as I had never worked with `ncurses` before. Was hard to wrap my head around the `attr` bitmasks in `printent_long` but I seem to have made it work properly. Although there might be some redundant `attroff`/`attron` calls, not sure.

I did come across #267 (after implementing this) so I'm curious to know if you think this implementation is acceptable. As you said for smaller terminal sizes the name needs to be cropped (could also consider falling back to "p" first when using "P") which is not working properly in this PR.

If that is required before merging I'll take another look at it and any other remarks.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3894,7 +3894,7 @@ static void printent_long(const struct entry *ent, uint_t namecols, bool sel)
 				len = sizes.maxsizeln + 2 - (sizeind ? 1 : xstrlen(size));
 				break;
 			case 'n':
-				len = sizes.maxnameln + (!nameind ? 1 : 0)  - xstrlen(ent->name);
+				len = MIN(namecols,sizes.maxnameln) + (!nameind ? 1 : 0)  - MIN(namecols,xstrlen(ent->name));
 				break;
 #ifndef NOUG
 			case 'o':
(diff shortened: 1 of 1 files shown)


luukvbaal:
Not sure what kind of rewrite is required to properly crop and restore non detailed view when the terminal gets too small.

I'll look into it myself further but for now this seems to be working fine in large terminals.

jarun:
Please do not make cosmetic changes without a discussion in the ToDo list first.

Did you see the amount of change in this PR? Who is going to maintain this change in the name of flexibility after a year?

luukvbaal:
I thought discussion could be done here, hence the draft PR. Not sure what maintenance is required if implemented properly.

jarun:
We fixed several issues with column length, broken lines etc. So we made a length fixed for fixed columns and a variable field at the end for lengths.

We like to discuss bigger changes first because I personally don;t like contributors to waste their valuable time on things we may not want.

luukvbaal:
Understandable, I hope it's not time wasted given I can properly finish the feature such that it might be merged.

Even if it won't be merged it's already usable for me so it won't totally wasted...

PR #920, opened 2021-03-25
change ^ v to unicode arrows

luukvbaal:
Wondering if this can be done harmlessly. I tested ~20 fonts on my system which all included this unicode character, including linux tty font.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5865,7 +5865,7 @@ static void redraw(char *path)
 	/* Go to first entry */
 	if (curscroll > 0) {
 		move(1, 0);
-		addch('^');
+		addstr("↑");
 	}
 
 	move(2, 0);
@@ -5888,7 +5888,7 @@ static void redraw(char *path)
 	/* Go to first entry */
 	if (i < ndents) {
 		move(xlines - 2, 0);
-		addch('v');
+		addstr("↓");
 	}
 
 	statusbar(path);


KlzXS:
CI says that makes the function too complex. I wouldn't argue against that assertion.

We use ncursesw so it should be able to handle unicode string just fine. But do see what you can do about the failing CI to "simplify" `browse()`.

luukvbaal:
> CI says that makes the function too complex. I wouldn't argue against that assertion.
> 
> We use ncursesw so it should be able to handle unicode string just fine. But do see what you can do about the failing CI to "simplify" `browse()`.

Not sure what to do about it, this change doesn't affect the `browse()` function does it? It's in the `redraw()` function.

KlzXS:
Huh, you are right. I guess it calculates complexity in depth. Still, it should be resolved. And fixing up `browse()` is definetly welcome.

From the first moment I saw the 1500 line switch I didn't like it, but I don't have anything that would simplify it without impacting the binary size.

luukvbaal:
> Huh, you are right. I guess it calculates complexity in depth. Still, it should be resolved. And fixing up `browse()` is definetly welcome.
> 
> From the first moment I saw the 1500 line switch I didn't like it, but I don't have anything that would simplify it without impacting the binary size.

Has nothing to do with depth complexity or this commit, the statements threshold was exceeded in https://github.com/jarun/nnn/commit/9c218d41e8d75c43516b51379c1dd4564378b045. I suppose a rewrite of `browse()` is indeed in order if we want to pass the complexity test.

luukvbaal:
I did find some fonts which don't contain this unicode arrow so I guess this is a no go. Unless we can somehow determine the active font and whether it supports the unicode character.

luukvbaal:
Checking whether a font supports the glyph is quite trivial with `fontconfig`, determining the active font terminal I'm not so sure about.

luukvbaal:
Doesn't seem to be possible without additional dependencies. It's possible with `lsof` and `sed` but I doubt we will want to require `lsof` as a dependency.

Closing unless we want to accept that users with an (uncommon) font not supporting the arrow will see nothing.

jarun:
I remember we used these in help in earlier versions and had issues on *bsd where the binary was detected as rogue (?). @annagrram found it.

jarun:
Commit 0144f440609b4fb7dcdf9e3325e6d1de3e6085bf.

PR #921, opened 2021-03-26
Show non-matching guid in statusbar

luukvbaal:
Display `xxx:xxx` instead of `- -` in statusbar for guid's without matching user/group on record.


luukvbaal:
Oops branched from my `detail-columns` branch instead of `master`, fixing...

jarun:
Please share an example for visual understanding.

luukvbaal:
```
sudo touch test
sudo chown 1234:1234 test
./nnn -U
```
image

jarun:
Thank you!

PR #922, opened 2021-03-27
Replace $HOME value with '~' in terminal title

salutis:
_Note: I am not a C programmer by trade. Please guide me._ ❤️

Spawned by #912 and @leovilok.

patch

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -998,7 +998,7 @@ static char *xextension(const char *fname, size_t len)
 	return xmemrchr((uchar_t *)fname, '.', len);
 }
 
-static char *xtildaname(char *path)
+static char *xtildapath(char *path)
 {
 	int pathlen = xstrlen(path);
 	int homelen = xstrlen(home);
@@ -6012,10 +6012,10 @@ begin:
 
 	if (!g_state.picker) {
 		/* Set terminal window title */
-		char *nicepath = xtildaname(path);
-		printf("\033]2;%s (%s)\007", xbasename(path), nicepath);
+		char *tildapath = xtildapath(path);
+		printf("\033]2;%s (%s)\007", xbasename(path), tildapath);
 		fflush(stdout);
-		free(nicepath);
+		free(tildapath);
 	}
 
 	if (g_state.selmode && lastdir[0])
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -998,7 +998,7 @@ static char *xextension(const char *fname, size_t len)
 	return xmemrchr((uchar_t *)fname, '.', len);
 }
 
-static char *xtildapath(char *path)
+static char *xtildepath(char *path)
 {
 	int pathlen = xstrlen(path);
 	int homelen = xstrlen(home);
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -998,6 +998,19 @@ static char *xextension(const char *fname, size_t len)
 	return xmemrchr((uchar_t *)fname, '.', len);
 }
 
+static char *xtildepath(char *path)
```
jarun, line 1001: So after a printf() and a fflush(stdout) now we have to execute this poor-performing logic every time we load a directory in `nnn`?

@leovilok ?

salutis: @jarun Would it be better to use `strcat` and `strcpy`? I would like to get this right.

salutis:
@jarun 

> closed this 21 minutes ago

Why is this PR closed? I would like to get it right.

jarun:
Ideally I would want the following:

```
index 0202040..9abf9f7 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5925,6 +5925,7 @@ static bool browse(char *ipath, const char *session, int pkey)
        int r = -1, presel, selstartid = 0, selendid = 0;
        const uchar_t opener_flags = (cfg.cliopener ? F_CLI : (F_NOTRACE | F_NOSTDIN | F_NOWAIT));
        bool watch = FALSE;
+       const uchar_t homelen = (uchar_t)xstrlen(home);
 
 #ifndef NOMOUSE
        MEVENT event;
@@ -5998,8 +5999,22 @@ begin:
 
        if (!g_state.picker) {
                /* Set terminal window title */
-               printf("\033]2;%s (%s)\007", xbasename(path), path);
+               if (is_prefix(path, home, homelen)) {
+                       home[homelen] = path[homelen - 1];
+                       path[homelen - 1] = '~';
+                       r = 1;
+               } else
+                       r = 0;
+
+
+               printf("\033]2;%s\007", r ? &path[homelen - 1] : path);
                fflush(stdout);
+
+               if (r) {
+                       path[homelen - 1] = home[homelen];
+                       home[homelen] = '\0';
+               }
+
        }
 
        if (g_state.selmode && lastdir[0])
```

How long do you think it would take for us to get to this?

Anyway. now that we have this, I really hate cosmetic new code in a stable product. So I am thinking of having a hybrid solution where we can combine the addressbar code in redraw() with this to show a more concise location and have a single function to do this.

jarun:
I have something like `~/G/p/nnn` in mind. What do you say?

jarun:
Please see commit e5d44ce223489656700081ebb1156414eb003e22.

salutis:
Dear @jarun,

I studied your code carefully (with a pencil and paper), and I reread all your words. I now clearly understand what you meant by "now we have to execute this poor-performing [...] logic". Instead of that, we can swap a single character and then subscript the buffer. That is such an elegant and fast solution; thank you so much for sharing! To answer the question, how long would take for us to get to your solution? A couple of rounds, that is for sure. To address your other points, I think using a tilde in the address bar is a great idea; `~` is easier on eyes and standard in Unix.

P.S. As for the nonstandard `~/G/p/nnn` pseudo-path, I am not a fan.

-- Rudy

jarun:
OK. I have pushed it in.

Also, thanks for your understanding. Not everyone takes criticism positively. My notes were not about _your coding capabilities_ but the performance of the code itself.

jarun:
Leaving out `~/G/p/nnn`. In the address bar we use it when the path length exceeds columns.

PR #926, opened 2021-03-28
Exchange nerd-font shell icon

luukvbaal:
Fixes incorrect spacing for shell scripts with nerd font icons enabled.

Old:
image

New:
image


jarun:
Thank you!

Issue #807, opened 2020-12-05
How to on/off preview-tui without leaving nnn?


akimdi:
I want to disable the preview-tui when I don't need it.

0xACE:
Don't think it is supported, you could implement it yourself.

I'm heading to bed so I can't be of help right now.

An idea I'll leave before I go to bed: imagine if you could toggle the `preview-tui` with 1 keystroke from `nnn` 🥲

akimdi:
> An idea I'll leave before I go to bed: imagine if you could toggle the `preview-tui` with 1 keystroke from `nnn` 🥲

it's my dream 😇

jarun:
I am using preview-tui with xterm. I press `^C` on the preview pane and it will close. I press the plugin shortcut again if I need it.

leovilok:
> it's my dream

@akimdi You have the power to make your dreams come true :

1. make a simple plugin that closes the preview window, e.g. run this:
```sh
echo 'kill $(lsof -tac cat "$NNN_FIFO")' > ~/.config/nnn/plugins/close-preview-tui
chmod +x ~/.config/nnn/plugins/close-preview-tui
```
2. add this plugin to a key
```sh
# open preview window with Alt+p , close it with Alt+c
NNN_PLUG='p:preview-tui;c:close-preview-tui' nnn -a
```

(a prettier version is left as an exercise to the reader :stuck_out_tongue: )

akimdi:
> ```shell
> NNN_PLUG='p:preview-tui;c:close-preview-tui' nnn -a
> ```

works fine but not always. 
Thanks :)

toddyamakawa:
I guess the solution would be to create a `toggle-tui-preview` command. All it would have to do is first check whether or not a preview pane is open. If it is open, close it. If it isn't open, open it.

luukvbaal:
You can now toggle `preview-tui(-ext)` with the same key due to
https://github.com/jarun/nnn/blob/b97e2d02ea92c390977a94810b3f0afc076b81a3/plugins/preview-tui-ext#L150

PR #927, opened 2021-03-28
preview-tui-ext: match nnn directory preview

luukvbaal:
Tries to match the `tree` directory preview to `nnn` by removing pager prompt, `tree` info (`--noreport`), `tree` indentation (`-i` flag) and cropping to `lines - 3` lines (-3 from `nnn` statusbar + context + spacing). 

~Also adds a new optional dependency [devicon-lookup](https://github.com/coreyja/devicon-lookup) to display icons in front of the `tree` output.~

Old:
image

New:
image

New without `iconlookup`:
image

Potential caveats are that users with `devicon-lookup` in their `PATH`, but not using icons in `nnn`, will still get icons in the preview pane. If that's an issue a variable toggle could be considered as well. 
The icons wont match the ones used in `nnn` perfectly either since `devicon-lookup` strictly handles source files, not e.g. `pdf` or `tar`.
Also the color has to be matched to the directory color used in `nnn` manually.

P.S. My `tmux` split seperator somehow dissapeared recently, that's not intended I suppose.

### DIFF
--- a/plugins/.iconlookup
+++ b/plugins/.iconlookup
@@ -1,0 +1,415 @@
+#!/usr/bin/sh
+
+# Description: Print icons in front of list of directories/files
+
+# Dependencies: awk
+
+# Usage
+#    1. Set colors and/or icons to your liking
+#    2. preview-tui-ext uses the script to prepend icon to directory listings
+#    3. Aditionally you can consider adding it to your PATH and/or FZF_DEFAULT_COMMAND to
+#    make it work with various fzf plugins (make sure you also add --ansi to your FZF_DEFAULT_OPTS)
+
+# Shell: POSIX compliant
+
+# Author: Luuk van Baal
+
+icon_lookup() {
+awk 'BEGIN { FS="." }
+{
+    if ($0 ~/^\.$/)
+        next
+    ent[NR] = $0
+    ext[NR] = $NF
+    limit = ENVIRON["limit"]
+}
+END {
+# Set your ANSI colorscheme below (https://en.wikipedia.org/wiki/ANSI_escape_code#Colors).
+# Default uses standard nnn icon colors, 8 and 24-bit nord themes are commented out.
+colordepth=8      #colordepth=8        #colordepth=24
+color_dirtxt=39   #color_dirtxt=111    #color_dirtxt="129;161;193"
+color_filetxt=15  #color_filetxt=111   #color_filetxt="129;161;193"
+color_default=39  #color_default=111   #color_default="129;161;193"
+color_video=93    #color_video=110     #color_video="136;192;208"
+color_audio=220   #color_audio=150     #color_audio="163;190;140"
+color_image=82    #color_image=150     #color_image="163;190;140"
+color_docs=202    #color_docs=173      #color_docs="208;135;112"
+color_archive=209 #color_archive=179   #color_archive="235;203;139"
+color_c=81        #color_c=150         #color_c="163;190;140"
+color_java=32     #color_java=139      #color_java="180;142;173"
+color_js=47       #color_js=109        #color_js="143;188;187"
+color_react=39    #color_react=111     #color_react="129;161;193"
+color_css=199     #color_css=110       #color_css="136;192;208"
+color_python=227  #color_python=68     #color_python="94;129;172"
+color_lua=19      #color_lua=167       #color_lua="191;97;106"
+color_document=15 #color_document=173  #color_document="208;135;112"
+color_fsharp=31   #color_fsharp=179    #color_fsharp="180;142;173"
+color_ruby=160    #color_ruby=150      #color_ruby="163;190;140"
+color_scala=196   #color_scala=139     #color_scala="143;188;187"
+color_vim=28      #color_vim=109       #color_vim="143;188;187"
+
+    switch (colordepth) {
+        case "4":
+            escape="\033["
+            break;
+        case "8":
+            escape="\033[38;5;"
+            break;
+        case "24":
+            escape="\033[38;2;"
+            break;
+    }
+
+# icons[][1] contains icon and icons[][2] contains color
+    icons["directory"][1] = ""; icons["directory"][2] = color_default
+    icons["file"][1] = ""; icons["file"][2] = color_default
+    icons["exec"][1] = ""; icons["exec"][2] = color_default
+    icons["manual"][1] = ""; icons["manual"][2] = color_docs
+    icons["pipe"][1] = "ﳣ"; icons["pipe"][2] = color_default
+    icons["socket"][1] = "ﳧ"; icons["socket"][2] = color_default
+    icons["door"][1] = "➡"; icons["door"][2] = color_default
+
+# top level and common icons
+    icons[".git"][1] = ""; icons[".git"][2] = color_default
+    icons["desktop"][1] = "ﲾ"; icons["desktop"][2] = color_default
+    icons["briefcase"][1] = ""; icons["briefcase"][2] = color_default
+    icons["document"][1] = ""; icons["document"][2] = color_default
+    icons["downloads"][1] = ""; icons["downloads"][2] = color_default
+    icons["music"][1] = ""; icons["music"][2] = color_default
+    icons["musicfile"][1] = ""; icons["musicfile"][2] = color_audio
+    icons["pictures"][1] = ""; icons["pictures"][2] = color_default
+    icons["picturefile"][1] = ""; icons["picturefile"][2] = color_image
+    icons["public"][1] = ""; icons["public"][2] = color_default
+    icons["templates"][1] = "陼"; icons["templates"][2] = color_default
+    icons["videos"][1] = ""; icons["videos"][2] = color_default
+    icons["videofile"][1] = "ﳜ"; icons["videofile"][2] = color_video
+    icons["changelog"][1] = ""; icons["changelog"][2] = color_docs
+    icons["configure"][1] = ""; icons["configure"][2] = color_default
+    icons["license"][1] = ""; icons["license"][2] = color_docs
+    icons["makefile"][1] = ""; icons["makefile"][2] = color_default
+    icons["archive"][1] = "遲"; icons["archive"][2] = color_archive
+    icons["script"][1] = ""; icons["script"][2] = color_default
+    icons["cplusplus"][1] = ""; icons["cplusplus"][2] = color_c
+    icons["java"][1] = ""; icons["java"][2] = color_java
+    icons["clojure"][1] = ""; icons["clojure"][2] = color_default
+    icons["js"][1] = ""; icons["js"][2] = color_js
+    icons["linux"][1] = ""; icons["linux"][2] = color_default
+    icons["fsharp"][1] = ""; icons["fsharp"][2] = color_fsharp
+    icons["ruby"][1] = ""; icons["ruby"][2] = color_ruby
+    icons["c"][1] = ""; icons["c"][2] = color_c
+    icons["html"][1] = ""; icons["html"][2] = color_default
+    icons["react"][1] = ""; icons["react"][2] = color_react
+    icons["python"][1] = ""; icons["python"][2] = color_python
+    icons["database"][1] = ""; icons["database"][2] = color_default
+    icons["worddoc"][1] = ""; icons["worddoc"][2] = color_document
+    icons["playlist"][1] = "蘿"; icons["playlist"][2] = color_audio
+    icons["opticaldisk"][1] = "ﴞ"; icons["opticaldisk"][2] = color_archive
+
+# numbers
+    icons["1"][1] = icons["manual"][1]; icons["1"][2] = icons["manual"][2]
+    icons["7z"][1] = icons["archive"][1]; icons["7z"][2] = icons["archive"][2]
+
+# a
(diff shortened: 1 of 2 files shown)


luukvbaal:
Theoretically the same could be achieved without `devicon-lookup` with just a very long `awk` or `sed` command/script. Would also allow exactly matching the icons used in `nnn`.

luukvbaal:
Wrote a lookup script in `awk` using the icons from `icons-nerdfont.h`, I suppose  anyone using `icons-in-terminal` would have to adapt.

Can that shellcheck error be suppressed? `# shellcheck disable=1071` doesn't seem to work.

jarun:
CI fails. Please check.

@leovilok please review this.

luukvbaal:
Had to remove the color output of `tree` since it was messing up the `awk` script, color is now set in `iconlookup` with ANSI codes.

@jarun I know, `shellcheck` is run on it since it's in the plugin folder I suppose but `shellcheck` errors because it doesn't support `awk`.
Can it be suppressed somehow or should I wrap it in a shell script?

luukvbaal:
`iconlookup` now uses the same color macros as defined in `colors.h`. As a result the `tree` output in the preview pane exactly matches `nnn`'s own directory listing, provided the same colors are used.

FollieHiyuki:
Can you use hex code instead of the icons themselves (if possible)?
This will make it easier to look up the icon names for the users, and match the style of the icon part of the source code.

luukvbaal:
As far as I could tell it's not possible with pure `awk` no. If you know otherwise I could reconsider.

Btw the `awk` script itself could also be omitted from the `nnn` repo since its not really an `nnn` plugin, perhaps that would be more appropriate? Already hosted it [here](https://github.com/luukvbaal/iconlookup).

luukvbaal:
Would also avoid the CI fail.

luukvbaal:
Wen't ahead and pushed the change

jarun:
You can ignore the `awk` thing: https://github.com/koalaman/shellcheck/wiki/Ignore

But this is a huge lookup for each file considering it's a script. Is this necessary? Can you make it conditional on a flag if not done already?

No external downloads please. Have it in the repo.

@KlzXS can you please share your thoughts on this? This is a script and I am not really opposed to this change. But everytime I look at the text rendering speed on WSL via putty, or the performance when loading big files, I feel we are moving away from the original philosophy of `nnn` - a snappy thin utility for power users.

I am asking this because I liked this plugin. But now I am seeing it's no longer remaining a simple script... the size is increasing, there in an option for icon lookup (it's a big list) just for displaying a directory contents (@luukvbaal is a tree view still possible with these changes)? I don't understand why I would rather not invoke ranger from the plugin and do the same thing?

jarun:
@luukvbaal I am not opposed to this change... just looking for a second opinion from someone else contributing for years on `nnn`.

luukvbaal:
It's an optional dependency for the `preview-tui-ext` plugin, just like the others. Makes more sense to have it outside the `nnn` repo imo. 

The lookup is done on the output of `tree` not on a per file basis if that's what you thought. Not very expensive: 
```
cd /
tree -L 3 -f | iconlookup
3830 directories, 22265 files
tree -L 3 -F  0.05s user 0.05s system 98% cpu 0.104 total
iconlookup  0.13s user 0.05s system 75% cpu 0.241 total
```

 It is optional and thus conditional, `preview-tui-ext` only runs `iconlookup` when found in `PATH`.

The size of `preview-tui-ext` does not necessarily increase just like it doesn't with the other optional dependencies.

A tree view should be possible with `iconlookup` but would require additional formatting, currently the icon is printed at the beginning of a line. It was written with `tree -L 1` in mind to serve as a Miller column in the `preview-tui-ext` allowing identical formatting as `nnn`.

Invoking `ranger` is obviously not equivalent. 

To me the only question is what the `tree` default output should be for `preview-tui-ext` without `iconlookup` in path. It used to be a tree view with depth 3, perhaps it should be restored to that but even without icons the view presented in this PR seems aesthetically pleasing to me.

jarun:
> It's an optional dependency for the preview-tui-ext plugin, just like the others.

It's not a standard utility you install using `apt-get`. It's a random script. So we should have it added to the `preview-tui-ext` plugin.

jarun:
After making it in-file please add a flag to ensure users can enable if they want. Document these and the flag should be off by default.

I think we should keep the default tree view as it is right now.

jarun:
> even without icons the view presented in this PR seems aesthetically pleasing to me.

In that case, leave it with the new view. People like me should be fine with `preview-tui`.

luukvbaal:
I have no interest in bringing that 300 line `awk` script inside `preview-tui-ext`, doesn't make sense to me. If that's what you feel is required I'll just close.

jarun:
We can't tell users that one script will download an external non-standard script and they have to allow that. why can't you add it as a separate dot script like `.nmv`, get it installed with other plugins and use it?

luukvbaal:
Users would not have to allow a script downloading another because that would not happen. User should install `iconlookup` to their path themselves just like all the other dependencies. 

The initial implementation was with [devicon-lookup](https://github.com/coreyja/devicon-lookup), would you have wanted that binary inside the plugin folder as well? 

I wrote `iconlookup` because `devicon-lookup` doesn't do folder icons and to allow matching the rest of the icons and colors to `nnn` but it is a separate script that can be used to prepend icons to e.g. standalone `fzf` as well. Again, having it in the `nnn` repo doesn't make sense to me.

jarun:
> User should install iconlookup to their path

Installation generally happens from a trusted repository. In this case it's downloading and copying in the path. Tomorrow if the script changes there is no way for users to know that because there is no release vehicle.

> would you have wanted that binary inside the plugin folder as well?

No I wouldn't probably allow linking to a third-party repo as well sue to security reasons.

I suggested adding this as `.iconlookup` inside the plugin dir.

I have more clarity from your last answer. It's better not to have this redundant dependency. `tree` is available everywhere and it should work.

luukvbaal:
Again it would be an optional dependency, whether it's redundant is up to the user to decide.

I'm fine with having it as `.iconlookup` in plugin dir. Let's wait for @KlzXS their thoughts then.

jarun:
It's an optional _non-standard_ dep. We prefer not to guide users to the path where they download scripts from third-party.

Yes, that's what I was suggesting earlier also, didn't understand why it's equivalent to having `devicon-lookup` binary in `nnn` repo. Having `.iconlookup` makes most sense to me.

I am not sure when @KlzXS can respond back. The question was more futuristic. I am OK with having this one in as I indicated earlier. So re-opening the PR. Please make the necessary changes for `.iconlookup`. And please keep it disabled by a flag by default.

KlzXS:
@luukvbaal 
I agree with @jarun, `iconlookup` should be present in the repo. I think it's more convenient that way for the user.

> 2. Make sure the script is in your PATH

> Potential caveats are that users with devicon-lookup in their PATH, but not using icons in nnn, will still get icons in the preview pane

You say this in the script. Why? I don't really see the need to put this script in PATH. Bundling the script allows you to keep it out of PATH since you know it's bundled with the plugin. No cross-contamination with icons for users who don't want it.

@jarun 
> But everytime I look at the text rendering speed on WSL via putty, or the performance when loading big files, I feel we are moving away from the original philosophy of nnn

I don't think WSL is the tool you would measure such performance with. People who use it swear by performance of WSL2, but are openly critical of WSL for bad performance. Try that yourself if it's bad for you.

I haven't properly read the script, but in theory the performance should be somewhere along the lines of `nnn`s lookup performance. `awk` is made for such manipulations and is optimized for them. But the best way to measure performance is to actually go ahead, use the script and measure it.

KlzXS:
@luukvbaal what exactly is the point of the outer while loop?

```awk
j = 0;
toprocess = $NF

while (toprocess > 0 && j < length(icons)) {
    for (i in ent) {
        if (ext[i] in icons) {
            matched[i] = icons[ext[i]]
            toprocess--
        }
        j++
    }
}
```

luukvbaal:
> didn't understand why it's equivalent to having `devicon-lookup` binary in `nnn` repo. Having `.iconlookup` makes most sense to me.

The point was that both scripts can be used for general use outside of `nnn`. I was saying that either the script or the `devicon-lookup` don't really belong in `nnn`'s plugin directory in my opinion.

>  And please keep it disabled by a flag by default.
> I don't really see the need to put this script in PATH.

I was saying there are people who have `devicon-lookup`(the rust binary) in their path for general use with `fzf`. `exists devicon-lookup` in `preview-tui` would then show icons in the preview pane and they might not have icons enabled in `nnn`, making it look out of place. Anyways since we now have a separate script this point is irrelevant. But since we have a separate script than can be used outside of `nnn` also I suggest a comment about symlinking `.iconlookup` to `PATH` if you want to have `tree` output with icons, rather than a flag to enable it. Agreed?

> @luukvbaal what exactly is the point of the outer while loop?

It's not necessary, good point.

leovilok:
If you don't preview the directory recursively with `tree`, why not use `nnn` itself in the preview window:

-  `nnn` is in C, it's small and fast
- it's installed
- it looks exactly like `nnn`
- you can use it as a dual pane file manager if you feel like it

I think that it's what I do in `preview-tabbed`.

luukvbaal:
The point is to allow looking inside the hovered directory, not run a dual pane file manager.
Not sure starting a second instance of `nnn` for every hovered directory is appropriate.

leovilok:
In this PR you're removing the pager and trying to cut the `tree` output to the size of the preview window:

1. this doesn't allow scrolling the list in the pager
2. this doesn't work well with long lines (filenames that are longer than the width of the preview window

> The point is to allow looking inside the hovered directory, not run a dual pane file manager.
> Not sure starting a second instance of nnn for every hovered directory is appropriate.

Well the thing is `nnn`is supposed to be really fast, it should be quicker than any ad-hoc shell pipeline, even if we don't use the dual pane accidental feature.

KlzXS:
@luukvbaal I'm not really keen on adding individual scripts into `PATH`. You could mention it somewhere, but also mention configuring it without adding it into `PATH`. The default should really be no icons I think that's the obivous choice.

@leovilok I guess the argument is that it's less visual clutter. No status bars, no extra info, just the entires inside. As for the scrolling I don't think a preview needs scrolling for a dir listing.

luukvbaal:
> 1. this doesn't allow scrolling the list in the pager

You're right but this is not a feature I often use and would never use for directory previews. If I want to look further into the preview directory I'll just press `l` in nnn.

> 2. this doesn't work well with long lines (filenames that are longer than the width of the preview window

I guess that's true I think I'll crop the length as well for my use.

> Well the thing is `nnn`is supposed to be really fast, it should be quicker than any ad-hoc shell pipeline, even if we don't use the dual pane accidental feature.

But it results in a very different view than the one proposed in PR i.e. double statusbar, contexts etc. Anyways if you're all so opposed to the change go ahead and close the PR again. I don't mind either way.

leovilok:
I don't have an opinion on the devicon stuff (I don't use icon fonts), but using a pager for the dir preview was added upon user request, so I think that it should be kept if possible. For example, you can put the whole pipeline (without the `head` part) in a shell function (something like `preview_dir`) and call `fifo_pager preview_dir`.

I don't mind not using `nnn` to preview dirs, I just find it funny that it worked so well when I tried it.

luukvbaal:
Can be merged as far as I'm concerned, feel free to change the default `tree` options when `ICONLOOKUP` is not set. Not sure what you guys want to be the default directory preview.

jarun:
> Not sure what you guys want to be the default directory preview.

The current change is OK. We are leaving it with `tree`. Users can modify and use if they need.

jarun:
Thank you!

jarun:
I copied `.iconlookup` into the plugin dir but I get the following:

    /home/vaio/.config/nnn/plugins/preview-tui-ext: 173: /home/vaio/.config/nnn/plugins/.iconlookup: not found

jarun:
```
$ /bin/ls -l /home/vaio/.config/nnn/plugins/.iconlookup
-rwxrwxr-x 1 vaio vaio 19912 Mar 30 20:56 /home/vaio/.config/nnn/plugins/.iconlookup
```

luukvbaal:
Uhm, weird. Not sure why that would happen, it works fine on my end.

luukvbaal:
What happens when you just enter `"${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins/.iconlookup"` into your shell?

jarun:
zsh: /home/vaio/.config/nnn/plugins/.iconlookup: bad interpreter: /usr/bin/sh: no such file or directory

jarun:
```
$ which sh 
/bin/sh
```

luukvbaal:
Ah should change the interpreter to env sh, my bad.

jarun:
Also in the call please use `$(dirname "$0")"/.iconlookup`. We use the same for plugin helper as well. Short and sweet.

jarun:
@KlzXS thanks for your insights!

luukvbaal:
Sure, made the [PR](https://github.com/jarun/nnn/pull/931).

akimdi:
@luukvbaal
Plugin preview-tui-ext doesn't work after update.
https://github.com/jarun/nnn/issues/807#issuecomment-827140764

PR #928, opened 2021-03-28
Add new plugin

exoton:


plugins/cdpath
```diff
@@ -0,0 +1,54 @@
...
+. "$(dirname "$0")"/.nnn-plugin-helper
+
+get_dirs() {
+    # Run a list of directories and symbolic links to directories through fzf
+    IFS=':'
+    for path in $CDPATH; do
+        for entry in "$path"/*; do
+            if [ -d "$entry" ]; then
+                name="$(basename "$entry")"
+                printf "%-24s :" "${name:0:24}"
```
KlzXS: CI fails
```
In plugins/cdpath line 26:
printf "%-24s :" "${name:0:24}"
^----------^ SC2039: In POSIX sh, string indexing is undefined.
```

plugins/cdpath
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env sh
+
+# Description: 'cd' to the directory from CDPATH
+#
+# Details: If the CDPATH environmet variable is not set, the default value of
+#          ${XDG_CONFIG_HOME:-$HOME/.config}/bookmarks will be used.
```
KlzXS, line 6: Make it `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/bookmarks` to keep it compativle with the other plugin and also to keep all things `nnn` in one place.

KlzXS (changes requested):
Address the comments. I see you've submited a commit while I was looking the changes over, I'll take a look at that now.

luukvbaal:
Uhm isn't this exactly the same as the [bookmarks](https://github.com/jarun/nnn/blob/master/plugins/bookmarks) plugin?

luukvbaal:
Forgive me, you already mention it in the description. What does it add exactly?

KlzXS:
@luukvbaal as far as I cam see it widens the search looking through everything in `CDPATH` not just in `nnn/bookmarks`. Seems useful enough for people who use that env.

luukvbaal:
I see, might make more sense to add that to the existing `bookmarks` plugin?

exoton:
It also includes regular directories, not just the symbolic links. I don't want to add anything to the existing plugin. My version is very short and clean and it has differences that are important to me: display format, field separator, etc.

exoton:
It also allows for spaces in the basename / bookmark name.

plugins/cdpath
```diff
@@ -0,0 +1,50 @@
...
+. "$(dirname "$0")"/.nnn-plugin-helper
+
+get_dirs() {
+    # Run a list of directories and symbolic links to directories through fzf
+    IFS=':'
+    for path in $CDPATH; do
+        for entry in "$path"/*; do
+            if [ -d "$entry" ]; then
+                name="$(basename "$entry")"
+                [ -h "$entry" ] && entry="$(readlink -f "$entry")"
```
KlzXS, line 26: I missed this originally. `readlink` is not POSIX. Either mention that or remove it. I think it might be better to remove it, I personally like it more when a symlink doesn't get resolved.

KlzXS:
One more comment from me. If nobody objects after that I think we can merge.

plugins/cdpath
exoton: It's used in the `bookmarks` plugin too.

KlzXS: Well leave it then if you will. When I get the time to work on these things I'll go through all the plugins and document that stuff. Not that it would matter to basically anyone, but still.

KlzXS (approved):

exoton:
I've removed the `readlink()` call.

plugins/cdpath
exoton: I don't know how important it is to stay strictly POSIX compliant nowadays. Anyways, I've found a way around it. I personally like the resolved symbolic link - it's an extra piece of information that can be useful.

Thanks for the quick review and approval on a Sunday night.

jarun:
Thank you! I would also have wanted this functionality in the bookmarks plugin but I see why it's different.

exoton:
If you set your `CDPATH` to `$HOME/.cache/nnn/bookmarks` you can use it as a bookmarks replacement.

jarun:
Can you please mention that in the plugin explicitly?

exoton:
Do you want this comment as an additional commit or as a new PR?

jarun:
In this PR.

exoton:
It's already merged and closed. I'll create a new one.

PR #929, opened 2021-03-29
Improve mdoc markup

sjmulder:
- Replace in-line styling with semantic macros (Ev, Pa, etc)
 - Mark up tables and lists with Bl
 - Add cross references and links (Xr, Sx)
 - ..lots more
 - Make mandoc -Tlint (mostly) happy.

### DIFF
--- a/nnn.1
+++ b/nnn.1
@@ -3,7 +3,7 @@
 .Os
 .Sh NAME
 .Nm nnn
-.Nd The unorthodox terminal file manager
+.Nd the unorthodox terminal file manager
 .Sh SYNOPSIS
 .Nm
 .Op Ar -aAcCdDeEfFgHJKlnQrRSuUVwxh
--- a/nnn.1
+++ b/nnn.1
@@ -3,548 +3,712 @@
 .Os
 .Sh NAME
 .Nm nnn
-.Nd The unorthodox terminal file manager.
+.Nd The unorthodox terminal file manager
 .Sh SYNOPSIS
 .Nm
-.Op Ar -a
-.Op Ar -A
+.Op Ar -aAcCdDeEfFgHJKlnQrRSuUVwxh
 .Op Ar -b key
-.Op Ar -c
-.Op Ar -C
-.Op Ar -d
-.Op Ar -D
-.Op Ar -e
-.Op Ar -E
-.Op Ar -f
-.Op Ar -F
-.Op Ar -g
-.Op Ar -H
-.Op Ar -J
-.Op Ar -K
-.Op Ar -l
-.Op Ar -n
 .Op Ar -p file
 .Op Ar -P key
-.Op Ar -Q
-.Op Ar -r
-.Op Ar -R
 .Op Ar -s name
-.Op Ar -S
 .Op Ar -t secs
 .Op Ar -T key
-.Op Ar -u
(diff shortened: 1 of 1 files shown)


nnn.1
```diff
@@ -3,548 +3,712 @@
 .Os
 .Sh NAME
 .Nm nnn
-.Nd The unorthodox terminal file manager.
+.Nd The unorthodox terminal file manager
```
jarun, line 6: Oh, we need that end `.`. I says - _enough said_. ;)

jarun (changes requested):

nnn.1
```diff
@@ -3,548 +3,712 @@
...
+.Pc .
+Set
+.Ql BTERM=xterm-256color
+to address it.
 .Sh AUTHORS
 .An Arun Prakash Jana Aq Mt engineerarun@gmail.com ,
 .An Lazaros Koromilas Aq Mt lostd@2f30.org ,
-.An Dimitris Papastamos Aq Mt sin@2f30.org .
+.An Dimitris Papastamos Aq Mt sin@2f30.org ,
+.An Sijmen J. Mulder Aq Mt ik@sjmulder.nl
```
jarun, line 712: After the existing 3 entries, I prefer to add new contributors names in the project main page. Can you please add your entry there instead of here? Please add it in the current PR.

jarun (changes requested):

nnn.1
sjmulder: Ah, I thought that was just about the manual. Never mind then, I'll remove it.

nnn.1
sjmulder: Actually `mandoc -Tlint` flagged that:

    mandoc: nnn.1:6:41: STYLE: trailing delimiter: Nd The unorthodox terminal file manager.

In fact looking at other manuals it's not just without trailing full stop but also without initial upper case:

    $ apropos print
    arch (1)             - print machine hardware name (same as uname -m)
    asprintf (3)         - print to allocated string
    base32 (1)           - base32 encode/decode data and print to standard output
    base64 (1)           - base64 encode/decode data and print to standard output
    blkid (8)            - locate/print block device attributes
    bsd-from (1)         - print names of those who have sent mail
    cat (1)              - concatenate files and print on the standard output
    date (1)             - print or set the system date and time
    ...

So I've made it align with that.

nnn.1
```diff
@@ -3,548 +3,711 @@
 .Os
 .Sh NAME
 .Nm nnn
-.Nd The unorthodox terminal file manager.
+.Nd the unorthodox terminal file manager
```
jarun, line 6: Please keep the capital `T` in the and the full stop and the end. We do not want to change the tagline.

jarun (changes requested):

nnn.1
jarun: Basically do not change anything in this line.

jarun:
We have increased around 200 lines and I see that now to add a single word I have to look up what is the syntax. We are not doing this,

jarun:
@sjmulder I have picked up some changes in commit 34f20ea67895c8954d10011f0bcb195e8b5586cb. I am not comfortable with the other changes. Don't really have the time to understand a new man page format and maintain it.

PR #932, opened 2021-04-01
.iconlookup: refactor

luukvbaal:
Refactored to print entries one by one rather than printing only after lookup for all the entries is done.
Not sure why I had the `END` rule originally but it's redundant and getting rid of it makes processing ~2 times faster:
```
tree -L 4 -F --noreport -i  0.24s user 0.24s system 93% cpu 0.517 total
~/.config/nnn/plugins/.iconlookup  0.28s user 0.29s system 56% cpu 1.000 total
```
vs
```
tree -L 4 -F --noreport -i  0.28s user 0.25s system 89% cpu 0.596 total
iconlookup  0.25s user 0.29s system 88% cpu 0.604 total
```

### DIFF
--- a/plugins/.iconlookup
+++ b/plugins/.iconlookup
@@ -6,24 +6,17 @@
 
 # Usage
 #    1. Set colors and/or icons to your liking
-#    2. preview-tui-ext uses the script to prepend icon to directory listings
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #934, opened 2021-04-04
Refactor printent/printent_long (DRY)

exoton:


jarun:
@KlzXS please review this.

src/nnn.c
```diff
@@ -3598,83 +3598,107 @@ static void print_time(const time_t *timep)
...
-			ind = '/';
-		} else
-			ind = '@';
+		return (ent->flags & DIR_OR_LINK_TO_DIR) ? '/' : '@';
+	case S_IFSOCK:
+		return '=';
+	case S_IFIFO:
+		return '|';
+	}
+	return '?';
```
jarun, line 3635: I see a `?` at the end of device names:

```
 2021-04-02 21:56  660        c  loop-control?
 2021-04-02 21:56  660        b  loop0?
```

jarun, line 3635: I see a `?` at the end of device names:

```
 2021-04-02 21:56  660        c  loop-control?
 2021-04-02 21:56  660        b  loop0?
```

Maybe you want to return `\0` for devices?

src/nnn.c
```diff
@@ -3598,83 +3598,107 @@ static void print_time(const time_t *timep)
...
-	addch((ent->flags & FILE_SELECTED) ? '+' : ' ');
+	switch (ent->mode & S_IFMT) {
+	case S_IFDIR:
+		if (!g_state.oldcolor)
+			attrs |= A_BOLD;
+		break;
+	case S_IFLNK:
+		if (ent->flags & DIR_OR_LINK_TO_DIR && !g_state.oldcolor)
+			attrs |= A_BOLD;
+		if (!detailed && g_state.oldcolor)
```
jarun, line 3691: I think we should drop the `!detailed` check here.

jarun:
> And what about the difference in the color pair? It's 0 in the 8-bit light mode and not 0 in the detail mode.

More details on this please. Are you referring to symlinks? Probably share the line numbers as well.

src/nnn.c
```diff
@@ -3694,163 +3721,56 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
...
 {
-	bool ln = FALSE;
-	char ind1 = '\0', ind2 = '\0';
-	uchar_t pair = 0;
-	int attrs = sel ? (A_REVERSE | (g_state.oldcolor ? A_DIM : COLOR_PAIR(C_MIS)))
-			: (g_state.oldcolor ? A_DIM : COLOR_PAIR(C_MIS));
-	uint_t len;
+	int entry_type = ent->mode & S_IFMT;
 	char *size;
-	char selgap[] = "  ";
```
jarun, line 3712: I think `selgap` is gone now leading to reverse effect touching the icons for the selected item:

Original:



Now:

src/nnn.c
```diff
@@ -3598,83 +3598,107 @@ static void print_time(const time_t *timep)
 	       t->tm_year + 1900, t->tm_mon + 1, t->tm_mday, t->tm_hour, t->tm_min);
 }
 
-static void printent(const struct entry *ent, uint_t namecols, bool sel)
+static char get_details_indicator(const struct entry *ent)
```
jarun, line 3601: Please keep the function name short: maybe `get_detail_ind`.

src/nnn.c
```diff
@@ -3598,83 +3598,107 @@ static void print_time(const time_t *timep)
...
+}
 
-		if (!ent->size)
-			pair = C_UND;
-		else if (ent->flags & HARD_LINK)
-			pair = C_HRD;
-		else if (!pair)
-			pair = C_FIL;
-		break;
+static char get_name_indicator(const struct entry *ent)
```
jarun, line 3621: Please keep the function name short: maybe `get_name_ind`.

jarun (changes requested):

src/nnn.c
```diff
@@ -3694,163 +3721,56 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
...
 	if (attrs)
 		attroff(attrs);
-
 	if (ind)
 		addch(ind);
 	addch('\n');
 }
 
-static void printent_long(const struct entry *ent, uint_t namecols, bool sel)
+static void printent_details(const struct entry *ent)
```
jarun, line 3729: Just `print_details`.

src/nnn.c
exoton: Yes. I was looking at that and I wasn't sure what is better. The hole in the reversed background looks weird (to me) in both cases.

exoton:
> > And what about the difference in the color pair? It's 0 in the 8-bit light mode and not 0 in the detail mode.
> 
> More details on this please. Are you referring to symlinks? Probably share the line numbers as well.

Yes, for symlinks (it's actually a return and not an assignment) on the line 3651 after the refactoring.
```
if (detailed || !g_state.oldcolor)
	pair = (ent->flags & SYM_ORPHAN) ? C_ORP : C_LNK;

```

src/nnn.c
exoton: Thanks, I've missed that one.

jarun:
> Yes, for symlinks (it's actually a return and not an assignment) on the line 3651 after the refactoring.

If `g_state.oldcolor` always dim symlinks... both in light and detailed modes.

src/nnn.c
jarun: Yes, the older one looks cleaner due to additional space.

jarun:
Mostly looks OK. I see one problem though.
We do not show icons in 8-color mode, so the whole entry for the selected entry should be in reverse in detail mode (no gap).

Earlier:



Now:



You can easily test this with `nnn -C`.

jarun:
BTW, I am very happy with this PR. We have reduced around 80 lines of code.

@KlzXS nice idea!

KlzXS:
I will take a look at this. Perhaps tonight. Just need to get a bit of time for it all.

src/nnn.c
```diff
@@ -3694,163 +3724,59 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
...
+	addch(' ');
+	attron(attrs1 | attrs2);
+	print_details(ent);
 
-#ifndef NOLOCALE
-	addwstr(unescape(ent->name, namecols));
-#else
-	addstr(unescape(ent->name, MIN(namecols, ent->nlen) + 1));
+#ifdef ICONS_ENABLED
+	attroff(attrs2);
```
jarun, line 3771: I think the following would make the reverse continuous if oldcolor is enabled:

```
3770 #ifdef ICONS_ENABLED
3771         if (!g_state.oldcolor)
3772                 attroff(attrs2);                                                                      
3773 #endif
```

jarun (changes requested):

jarun:
Thank you for your contribution! I'll take care of the minor discontinuity.

@KlzXS please review anyway. I am merging this for testing. Planning to make a release soon.

exoton:
I don't know if this is an acceptable style, but in this particular case it's easier to read and it further reduces the line count :)
```
	switch (ent->mode & S_IFMT) {
	case S_IFDIR:
	case S_IFREG:  return ' ';
	case S_IFLNK:  return '@';
	case S_IFSOCK: return '=';
	case S_IFIFO:  return '|';
	case S_IFBLK:  return 'b';
	case S_IFCHR:  return 'c';
	}

```

jarun:
Sure thing! I will make the change.

exoton:
The same goes for `get_name_ind`

exoton:
The same goes for `get_name_ind`

jarun:
Done!

exoton:
Thank you!

exoton:
> BTW, I am very happy with this PR. We have reduced around 80 lines of code.
> 
> @KlzXS nice idea!

Sorry, but I didn't know about this idea until my PR was ready. I did it in my own interests :)

jarun:
@KlzXS proposed this refactoring almost a year back. Doesn't take the credit away from you, don't worry. We could never get our hands to it.

luukvbaal:
> Yes. I was looking at that and I wasn't sure what is better. The hole in the reversed background looks weird (to me) in both cases. I should have kept the old behavior.

@exeton This was my main gripe with the current detail view as well and my reason for writing the `NNN_DCOLUMNS` PR. I'm rewriting my fork after the `printent()` refactor and in view of @jarun their concerns and trying to stay closer to upstream I came up with a more sensible implementation in https://github.com/jarun/nnn/compare/master...luukvbaal:namefirst. 

I got rid of the configuration variable and just used the order name > size > permissions > time as this seems the most sensible column order to me. 

The result is a much smaller patch (also due to the refactor from this PR) and the following view:
demo

Looks and feels much better than the current detail view in my opinion(no gap in the selection, no jarring jump of the name position when toggling detail view, better use of space by calculating max entry length(tradeoff)) but if you still think it's not worth merging @jarun, nevermind.

jarun:
> if you still think it's not worth merging @jarun, nevermind

@luukvbaal I hope you don't have a grudge because the earlier PR was not merged. We work as a team and this project existed long before you pitched in. There can be decisions you may not agree with, please bear with that.

Coming to the latest iteration of your change, yes it does look good with the existing columns. We do NOT want to provide unnecessary information in the name of new/customizable columns. The problem I see is the dancing width depending on the entry with the maximum name length. All columns move together and that isn't a good user experience. The current arrangement in `nnn` leaves the columns in the same position because the time, size and permission fields are fixed-width. `name` is towards the right because it's variable-width. I guess `ls` default also does the same because of the same reasons. I don't want everything shifting left/right together. Do you have a solution for that?

jarun:
Also, you mentioned your issues with the reverse-gap when icons are used. Many users complained when we introduced icons. That's why it's under conditional compilation. Icons may look good but for many users `nnn` is not just about being a _fancy_ tool. There are too many of those available already.

jarun:
@0xACE @KlzXS @leovilok do you guys have any suggestions for a static UX with the name towards the left?
I don't want a large gap between the name and the size fields either. I thought of taking into account the longest filename in the directory but that won't work with event updates when a new file is added, removed etc.

exoton:
I agree that changing width is bad, but I like the look of `mc` and `lf` with `drawbox`.

jarun:
@exoton an image please... I don't understand `drawbox`.

exoton:
Now that we are in the open, the real reason for my refactoring was to experiment with different print options.

jarun:
> the real reason for my refactoring was to experiment with different print options

we wanted a simpler and unified approach towards printing.

exoton:
Untitled

exoton:
Untitled

exoton:
Untitled

exoton:
I like simplicity but I also know that look and feel preferences are very different and very subjective. If it doesn't cost much to provide configurable l&f then why not. The current l&f will be the default option.

jarun:
Sorry, no such disruptive changes for look and feel.
We are not going to maintain all that code.

I made some small tweaks in the current view. How does this look?

- reverse only the filename in detail view
- don't reverse the details for hovered but show an arrow in the left
- always show selection (`+` symbol) in bright

Without icons:



With icons:

exoton:
What if there is no gap? How will icon look on the reversed background?

jarun:
> What if there is no gap?

What do you mean?

We are not showing icons in reversed background (not even currently).

exoton:
What if we do?
BTW, I wasn't proposing the drastic changes in the display. Just a couple of examples of how it looks when the details are on the right side. Almost all file managers that I know show details on the right side.

jarun:
> What if we do?

I think we tried that once (by mistake) and it didn't look good. The icons should show in natural color. In any case, I want to get rid of the details in reverse.

> Almost all file managers that I know show details on the right side.

We are trying to maximize the utility of the screen real estate to show file names. It's very useful when we get the advantage. Even `ls` does the same by default.

exoton:
But when you show details, you have the same space for the file name no matter if the details are on the left side or on the right side. Right?

jarun:
That's correct. The problem is file name length varies.

exoton:
Personally, I think that details should always be on the far right side. If your terminal window is too wide - enable the preview :smile:

luukvbaal:
> I don't want everything shifting left/right together. Do you have a solution for that?

The solution would be to align the details to the right, something like:
test

I'm not sure I would prefer that over keeping it left aligned though. Yes it dances around but this is only prevalent when scrolling very fast through a directory with a large number of files(and looks funny to me). When just navigating through folders the dancing around is not that jarring to me and side having the column length equal to the largest largest entry provides the best use of space.

exoton:
Three different people - three different opinions.

I like it. May be because I'm used to it. You should judge from the usability point of view. In most cases the focus is on the file name and it really helps when the details are not very close.

KlzXS:
I check out your changes @exoton. Great work! That's basically what I wanted done, but in a much better way than I had imagined.

Concerning the left-aligned filenames. I don't like putting details all the way on the right. I use i3 so in that case I would have some 30cm (call it 11.8 inches) of empty space on screen and that's not pleasant to look at. If you don't use a tiling wm it might be better for you. 

I prefer the way `mc` does it, have a column of fixed width (or set by some variable) and put as much of the filename as can fit, making sure the extension is shown if there is one. But that would require a lot of changes and would complicate the display logic. I assume that's what you implemented in the PR you mentioned, I haven't taken a look at that.

luukvbaal:
Rather than a fixed column width I made it dynamic according to the largest entry that has to be printed. As a tiling WM user myself this gives me the most conveniently arranged(file name on the left) and aesthetically pleasing(no icons in between details/name) view but I suppose the issue is that it's subjective.

exoton:
@KlzXS Thank you.
I hear you guys, but we should also think about the minority of users who are not using tiling WM :smile:

jarun:
Even if we try to make the column width fixed, on the wide screen a short name and the details are so far apart that there is no way to link them easily. The details are already greyed to they are unobtrusive. However, if a user is using the detail mode that means she may need to refer to the details any time. We can't make it even harder for her.

We have discussed this enough. I do not see any real issue with the current order and will retain the same.

Archangel-Live:
Let me add my two cents as a normal user: I'm relatively new to nnn and I use it because there are no bells and whistles. When I discovered the detail mode I instantly thought: "Hey, that's well done. Better than the view offered by other filemanagers". Please keep this as it is and don't try to be fancy.

Exeton:
Oh hey exoton. Nice to meet you!

exoton:
Oh, hello! I thought luukvbaal just made a typo.

luukvbaal:
It was a typo haha.

For anyone interested I worked out all the kinks and finished a clean patch to print the names first in detail mode:
https://github.com/luukvbaal/nnn/commit/22082edc31c33b5bb3e28cfaa030208280057afc.

image

PR #939, opened 2021-04-07
print_details: printw align

luukvbaal:
Seems more elegant to me than the `blanks` solution unless there is some other downside to using `printw()` I'm not aware of?

Was looking for a solution to avoid the print loop in my name first patch https://github.com/luukvbaal/nnn/commit/4deec222abf9aebaeb09778218d12404802c79f1 where the `blanks` solution can't be applied and thought I would suggest it for upstream as well.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3747,7 +3747,7 @@ static void print_details(const struct entry *ent)
 		char *size = coolsize(cfg.blkorder ? ent->blocks << blk_shift : ent->size);
 		printw("%*c%s", 9 - (uint_t)xstrlen(size), ' ', size);
 	} else
-		printw("%*c%s", 8, ' ', get_detail_ind(ent->mode));
+		printw("%*c%c", 8, ' ', get_detail_ind(ent->mode));
 }
 
 static void printent_long(const struct entry *ent, uint_t namecols, bool sel)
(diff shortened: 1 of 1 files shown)


jarun:
printw() parses the format string and passes the result to addstr(). I wanted to get rid of it (as much as possible) so we can avoid the formatting checks.

jarun:
Accidentally closed. Reopened and merged. Thanks!

luukvbaal:
> printw() parses the format string and passes the result to addstr(). I wanted to get rid of it (as much as possible) so we can avoid the formatting checks.

I see, can you explain why you merged it in this case? Do you expect it to be faster than the multiple `addstr()/addch()` calls?

luukvbaal:
If I had seen this commit https://github.com/jarun/nnn/commit/ed23e494a9064eb29879cfd9c24fa4965f1ea4d4 before I wouldn't have made this PR in the first place. I do wonder why you think it's better in this case(assuming you do since you merged it).

jarun:
Yes, I think the performance difference is negligible given we do not have any numeric to string conversion.

luukvbaal:
I see, thanks for the insight!

jarun:
Also see commit 903a4b3a66bb37a029906ef20b97ca68619a5310. Reduces many lines of code and the numeric conversions are only for values < 99 so I guess the performance would be similar even if the naive approach is taken.

PR #946, opened 2021-04-13
Update README demo

luukvbaal:
It's now 113.1K, FYI the image right below is 682.4K...


jarun:
Can you reduce the size of the below image also?
Please keep the dimensions same.

luukvbaal:
Exported as `jpg` without EXIF data etc in GIMP resulting in a 218.1K image.

jarun:
Thank you so much!

Issue #950, opened 2021-04-17
ICONLOOKUP not work in preview-tui-ext


pooh2pooh:
Hi!

I can't turn on the icons in the plugin preview-tui-ext, I made export ICONLOOKUP=1 and export FZF_DEFAULT-OPTS="--ansi" and install fzf

Terminal: kitty

luukvbaal:
The `iconlookup` script uses nerd font icons by default, are you using a patched nerd font in your terminal?

pooh2pooh:
Yes. If you run: ls | iconlookup | fzf

luukvbaal:
Is `.iconlookup` present in your `nnn` plugin directory? That's where `preview-tui-ext` looks for the iconlookup script, not in your `PATH`.

luukvbaal:
Not sure how you are exporting `ICONLOOKUP` but please check if #951 fixes it for you.

pooh2pooh:
.iconslookup file is in the plugins folder,
there are no errors,
export ICONLOOKUP in the .zshenv and terminal

updated preview-tui-ext also does not show icons :(

Screenshots:
https://d.poohprod.ru/2021-04-20_285x101.png
https://d.poohprod.ru/2021-04-20_607x166.png

luukvbaal:
Hmm I don't know. Can you confirm whether its entering the `ICONLOOKUP` if statement by changing
`tree -L 1 --dirsfirst -F --noreport -i | head -n "$((lines - 3))" | "$(dirname "$0")"/.iconlookup -l "$cols" -B "$BSTR" -b " "` to e.g. `echo test`.

pooh2pooh:
after installing tree everything worked! thanks :)

jarun:
@pooh2pooh is `tree` mentioned in the list of dependencies in `preview-tui-ext`?

luukvbaal:
It's listed as one of three possible dependencies(`tree`, `exa` or `ls`), I'll make a PR to prepend icons to `exa` and `ls` output as well since both support the `-F` flag for indicators.

jarun:
OK.

PR #951, opened 2021-04-18
preview-tui-ext: pass env

luukvbaal:
Fix environment variables not exported in profile(potential fix for #950)


jarun:
Thank you!

PR #954, opened 2021-04-21
preview-tui: directory preview/iconlookup

luukvbaal:
* Switch to `ls` for `ICONLOOKUP` case
* Remove `fifo_pager` from `exa` command to actually show grid (`-G` flag)
* Restore `tree` depth 3
* Fix `.iconlookup` typos
* Add note regard `NNN_ARCHIVE` (#955)

### DIFF
--- a/plugins/.iconlookup
+++ b/plugins/.iconlookup
@@ -165,9 +165,9 @@ awk 'BEGIN {
 
 # i
     icons["ico"][1] = icons["picturefile"][1]; icons["ico"][2] = icons["picturefile"][2]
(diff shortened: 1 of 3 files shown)


luukvbaal:
Even though `exa` and `tree` support the `-F` flag to show indicators, it makes no sense to use anything other than `ls` for `.iconlookup` when the goal is to mimic the `nnn` directory preview.

Might add support to `.iconlookup` to maintain the `tree` indentation in which case there would be merit to piping tree into `.iconlookup`.

jarun:
Thank you!

jarun:
On a side note: if it happens that a PR is in progress use the work in progress label so I know I should not merge it.

PR #955, opened 2021-04-21
preview-tui-ext: use atool for rpm/deb

rck:
atool does good job in previewing the content of deb/rpm files, use it.

### DIFF
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -210,6 +210,8 @@ preview_file() {
             fifo_pager unzip -l "$1"
         elif [ "$ext" = "gz" ] || [ "$ext" = "bz2" ]; then
             fifo_pager tar -tvf "$1"
+        elif [ "$mimetype" = "application/x-rpm" ] || [ "$mimetype" = "application/vnd.debian.binary-package" ] && exists atool; then
+            fifo_pager atool -l "$1"
         else
             fifo_pager print_bin_info "$1"
         fi


luukvbaal:
If you export `NNN_ARCHIVE` in your profile as suggested in the [wiki](https://github.com/jarun/nnn/wiki/Usage#configuration), `preview-tui-ext` will already preview `deb/rpm` files using `bsdtar` or `atool`:
```
elif [ "${ARCHIVES#*$ext}" != "$ARCHIVES" ]; then
        if exists bsdtar; then
            fifo_pager bsdtar -tvf "$1"
        elif exists atool; then
            fifo_pager atool -l "$1"
        fi
```
 I would suggest exporting `NNN_ARCHIVE` making this addition unnecessary, perhaps an additional comment in `preview-tui-ext`  regarding this is warranted.

luukvbaal:
I think this can be closed with https://github.com/jarun/nnn/pull/954/commits/558fa528a5f4ec1b3a6206a0e75155b418389846.

rck:
if you do that and one has bsdtar intalled (next to atool), then you basically get this:

```
bsdtar -tvf some.deb
-rw-r--r--  0 0      0           4 Apr 20 13:42 debian-binary
-rw-r--r--  0 0      0        1448 Apr 20 13:42 control.tar.xz
-rw-r--r--  0 0      0      848532 Apr 20 13:42 data.tar.xz
```

which is pretty worthless. It works better for rpms

luukvbaal:
`bsdtar` seems to output only the toplevel of the archive yeah, not sure how that's worthless but it might not be preferred. In any case if you prefer `atool` output you can just move those around but preference in that sense should probably be left to the user.

jarun:
Yes, the tool preference is up to the user.

jarun:
Thank you!

luukvbaal:
@jarun merging this like this is kind of pointless as it comes after the `elif [ "${ARCHIVES#*$ext}" != "$ARCHIVES" ]` statement so this will never be executed when `NNN_ARCHIVE` is exported as suggested in the documentation.

I don't mind but this seems unnecessary to add as `deb` and `rpm` previews were already supported.

jarun:
Ahh, I thought it was some improvement from you. I will revert this back if it never executes.

luukvbaal:
Assuming `NNN_ARCHIVE` is exported it would never execute. Perhaps replacing the `NNN_ARCHIVE` check with the same list of extensions in `preview-tui-ext` makes more sense though.

jarun:
I am not sure why we need to tie `NNN_ARCHIVE` with previewers. Plugins are free to have their own deps.
Can you decouple this?

rck:
I think packages can deserve some special handling, so I would have adapted the documentation for `NNN_ARCHIVE` to not include deb and rpm. And keep my suggested code as it handles debs better if bsdtar is installed

rck:
> `bsdtar` seems to output only the toplevel of the archive yeah, not sure how that's worthless 

it shows the exact same content for every deb. That IMO does not help the user

jarun:
> it shows the exact same content for every deb. That IMO does not help the user

Does atool do any better?

rck:
> > it shows the exact same content for every deb. That IMO does not help the user
> 
> Does atool do any better?

yes, sure. it prints the files in the deb. atool works for both, rpm and deb, listing the *content* (== the files)

luukvbaal:
> it shows the exact same content for every deb. That IMO does not help the user

Fair point, I suppose `atool` deservers higher precedence for previewing `deb` archives in that sense.

jarun:
Yes, I agree. bsdtar anyway doesn't work well with some rar versions.

@luukvbaal can we decouple `NNN_ARCHIVE` from plugins?

luukvbaal:
> @luukvbaal can we decouple `NNN_ARCHIVE` from plugins?

Yeah sure. Substituting the `if/else` block for `ext` comparisons with a `switch` case to allow easier grouping of extensions as opposed to very convoluted if statements would make the most sense to achieve that in my opinion. If no one is against that I can make a PR for that.

luukvbaal:
On second thought I'm not so sure that's easily possible since we are checking for `mimetype` as well, I'll leave it as is.

PR #957, opened 2021-04-21
preview-tui-ext: reformat archive preview

luukvbaal:
* Change archive preview precedence `atool` > `bsdtar`
* Add `case` block for easier changing of precedence for archive preview on a per-extension basis

I know I prefer `bsdtar`'s more concise  preview for `7z` archives for example where `atool` leads with
```
7-Zip [64] 17.04 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.04 (locale=en_US.UTF-8,Utf16=on,HugeFiles=on,64 bits,8 CPUs x64)

Scanning the drive for archives:
1 file, 101502970 bytes (97 MiB)

Listing archive: Measurement_Data_Lab2.7z

--
Path = Measurement_Data_Lab2.7z
Type = 7z
Physical Size = 101502970
Headers Size = 398
Method = LZMA:24
Solid = +
Blocks = 1
```

Should such cases be included by default or be left up to the user?

### DIFF
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -16,8 +16,7 @@
 #    - unzip
 #    - tar
 #    - man
-#    - optional: bsdtar or atool for additional archive preview (make sure to export NNN_ARCHIVE
-#                as suggested in the wiki: https://github.com/jarun/nnn/wiki/Usage#configuration)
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

Issue #959, opened 2021-04-22
Get the selected file for live preview to work with a MS Window previewer


jtourlamain:
**Is your feature request related to a problem? Please describe.**

I can't get the live preview to work on WSL2 running on MS Windows.
On windows I use quicklook from https://github.com/QL-Win/QuickLook to preview my files. I would like to do the same from nnn running in WSL2. 

**Describe the solution you'd like**

If a hook is available that provides the file name and path every time the selection changes a bridge script can be executed to preview the file in the quicklook app. A bridge with quicklook exists so you can call a cli command to change the file in the preview window (https://github.com/QL-Win/QuickLook/wiki/Integration-and-Development)
The selected path in nnn can be found in the .lastd file, but I can't find the filename that is currently selected in nnn. If I cat the fifo file. nothing can be seen (I guess I don't understand how it works). 

Opening files in MS Windows apps works via a command plugin ```-_cmd.exe /C start $nnn*``` but that principle doesn't fit for a live preview.

Alternative:
The live view solutions with tmux works on windows terminal for text files.

luukvbaal:
See if #960 works for you. I'm not familiar with `QuickLook` and I'm not quite sure what's the point of the `Bridge.exe` but this seems to work fine.

jarun:
Closed at #960.

PR #960, opened 2021-04-22
preview-tui: add QuickLook support for WSL

luukvbaal:
Add [QuickLook](https://github.com/QL-Win/QuickLook) support for WSL as suggested in https://github.com/jarun/nnn/issues/959.

Perhaps this should be a separate plugin since essentially bypasses all logic provided by `preview-tui` but on the other hand having all previewers in a single script has merit.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -6,9 +6,10 @@
 # For a more extended version of this script with additional optional dependencies, see preview-tui-ext.
 #
 # Dependencies:
-#    - Supports 3 independent methods to preview with:
+#    - Supports 4 independent methods to preview with:
 #        - tmux (>=3.0), or
 #        - kitty with allow_remote_control on, or
+#        - QuickLook on WSL (https://github.com/QL-Win/QuickLook)
 #        - $TERMINAL set to a terminal (it's xterm by default).
 #    - less or $PAGER
 #    - tree or exa or ls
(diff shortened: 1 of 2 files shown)


jtourlamain:
@luukvbaal wow that was quick! When launching the preview, nnn returns to the command prompt but the preview windows opens. When hitting CTRL-C I return to nnn and the preview window functions as desired. Is there a way to disable the preview again, because if I close the preview window and select another file, the preview reopens.

luukvbaal:
When not running in `tmux` `QuickLook` will be used to preview everything. I also added `QuickLook` as an additional option for `preview_image` so `preview-tui(-ext)` will now use `QuickLook` to preview images when running in `tmux`. 

@jtourlamain not sure how to address the fact that you can't disable previews for the pure `QuickLook` preview mode. When running inside `tmux`, closing the preview pane will also stop `QuickLook` previews.

luukvbaal:
As far as I'm concerned this can be merged. 

The `QuickLook` and `$TERMINAL` preview modes could use a solution to kill the preview process when pressing `^C` in `nnn` but I couldn't find a reliable way to do that in the script. We need a way to kill all child processes from a trap in the main plugin shell while not blocking `nnn` input but I couldn't figure it out.

jtourlamain:
ok great thx

jarun:
Thank you!

luukvbaal:
> Is there a way to disable the preview again, because if I close the preview window and select another file, the preview reopens.

@jtourlamain #961 allows you to do this by pressing the plugin key a second time.

jtourlamain:
your implementation works like magic! I can start the previewer and stop it with the same command 👍

PR #961, opened 2021-04-22
preview-tui: close previewer

luukvbaal:
Closes the previewer (all 4 options) on second invocation of the plugin key by killing the previewer fifo. Also assumes the `QuickLook` previewer is open and closes it by running it a second time with the same argument(https://github.com/QL-Win/QuickLook/issues/557#issuecomment-540130609).


jarun:
Thank you!

PR #963, opened 2021-04-23
preview-tui: reformat QuickLook

luukvbaal:
Get rid of redundant `pgrep` and reduce tests inside main preview loop.


jarun:
Thank you!

PR #966, opened 2021-04-25
preview-tui: reformat toggle

luukvbaal:
Reduce flickering on toggle by wrapping in function and sending to background.

Edit: just noticed I messed up the environment variables passed to `kitty`/`tmux` will fix, probably tomorrow though.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -67,27 +67,75 @@ USE_SCOPE="${USE_SCOPE:-0}"
 USE_PISTOL="${USE_PISTOL:-0}"
 PAGER="${PAGER:-less -P?n -R}"
 TMPDIR="${TMPDIR:-/tmp}"
-NUMPREVIEWTUI="$(($(find "$TMPDIR" -maxdepth 1 -name 'nnn-preview-tui-pagerpid*' 2>/dev/null | wc -l) + 1))"
-PAGERPID="$TMPDIR/nnn-preview-tui-pagerpid.$NUMPREVIEWTUI"
-GIFPID="$TMPDIR/nnn-preview-tui-gifpid.$NUMPREVIEWTUI"
-CURSEL="$TMPDIR/nnn-preview-tui-selection.$NUMPREVIEWTUI"
-FIFO_UEBERZUG="$TMPDIR/nnn-preview-tui-ueberzug-fifo.$NUMPREVIEWTUI"
 
-[ "$PAGER" = "most" ] && PAGER="less -R"
+startpreview() {
+    NUMPREVIEWTUI="$(($(find "$TMPDIR" -maxdepth 1 -name 'nnn-preview-tui-pagerpid*' 2>/dev/null | wc -l) + 1))"
+    PAGERPID="$TMPDIR/nnn-preview-tui-pagerpid.$NUMPREVIEWTUI"
+    GIFPID="$TMPDIR/nnn-preview-tui-gifpid.$NUMPREVIEWTUI"
+    CURSEL="$TMPDIR/nnn-preview-tui-selection.$NUMPREVIEWTUI"
+    FIFO_UEBERZUG="$TMPDIR/nnn-preview-tui-ueberzug-fifo.$NUMPREVIEWTUI"
 
-if [ -e "${TMUX%%,*}" ] && tmux -V | grep -q '[ -][3456789]\.'; then
-    TERMINAL=tmux
-elif [ -n "$KITTY_WINDOW_ID" ] && kitty @ ls >/dev/null 2>&1; then
-    TERMINAL=kitty
-else
-    TERMINAL="${TERMINAL:-xterm}"
-fi
+    [ "$PAGER" = "most" ] && PAGER="less -R"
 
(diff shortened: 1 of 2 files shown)


luukvbaal:
This works great with `tmux` and `$TERMINAL` options but for some reason doesn't play well with `kitty`. Seems like running 
https://github.com/jarun/nnn/blob/1fc7fde819d6a7e082055ff10407f4744d862090/plugins/preview-tui#L105 and https://github.com/jarun/nnn/blob/1fc7fde819d6a7e082055ff10407f4744d862090/plugins/preview-tui#L110 from a background process sends garble to stdin for some reason, causing `nnn` to hang and the split is never opened.

luukvbaal:
I'll see if I can figure it out but I'm wondering how do we feel about dropping `kitty` split support from `preview-tui` and instead suggest running `tmux` @jarun @leovilok?. `tmux` with `ueberzug` inside with `kitty` is way more responsive than the `kitty` splits and previews and provides a better experience in my opinion.

luukvbaal:
> Seems like running
> https://github.com/jarun/nnn/blob/1fc7fde819d6a7e082055ff10407f4744d862090/plugins/preview-tui#L105
> 
> and
> https://github.com/jarun/nnn/blob/1fc7fde819d6a7e082055ff10407f4744d862090/plugins/preview-tui#L110
> 
> from a background process sends garble to stdin for some reason, causing `nnn` to hang and the split is never opened.

This can be avoided (tested and confirmed) using the `@ --to` but would require users to start `kitty` with the `--listen-on` flag. Any opinions on this or dropping `kitty` support?

jarun:
> This can be avoided (tested and confirmed) using the @ --to but would require users to start kitty with the --listen-on flag. Any opinions on this or dropping kitty support?

Is the issue intermittent? I didn't face any issues with kitty when I tried it.

Also, can this be a one time setting in the kitty config file?

luukvbaal:
It happens every time I start `preview-tui`. Interesting that it's not an issue for you, the cause is that escape codes are being printed to the terminal(see https://github.com/kovidgoyal/kitty/issues/1501) which seem to be caught as keypresses by `nnn` on my system. Unfortunately it cannot be set in the configuration file(https://github.com/kovidgoyal/kitty/issues/2557).

luukvbaal:
Actually the man page states "For UNIX sockets, this can also be specified in kitty.conf." It is possible https://github.com/kovidgoyal/kitty/commit/85b55b31b6d4c0e22345a2593c6c04721e63af48.

luukvbaal:
`eval KITTYSOCKET="$(awk '/^listen_on/ {print $2}' "${XDG_CONFIG_HOME}"/kitty/kitty.conf)-$(pstree -sap $$ | awk -F',' '/kitty/ { print $2; exit}')"` and  `kitty @ --to "$KITTYSOCKET"` works to extract the socket set in `kitty.conf`. 
However, I'm still curious if anyone is against dropping `kitty` support from `preview-tui` in favor of `tmux` since it works in `kitty` and is way faster.

jarun:
Someone already using kitty may not be interested in learning tmux.

luukvbaal:
> Someone already using kitty may not be interested in learning tmux.

Yeah I suppose, not sure what to do with this PR then. 
I really like the improvement for `tmux` and `$TERMINAL` case but doesn't seem applicable to the `kitty` previews. It works for tui previews but the `kitten icat` image preview doesn't seem to play well the `@ --to` flag. Focus goes to the preview pane and the image disappears immediately without the `--hold` flag.

luukvbaal:
It might well be possible, I never actually used `kitty` so perhaps someone who has can assist.

jarun:
Just install it and check it out.

luukvbaal:
I have it installed of course to test the plugin I was just saying someone who actually uses it might have an easier time figuring it out.

jarun:
Ahh. OK. No I hardly use previews myself.

luukvbaal:
> Seems like running
> https://github.com/jarun/nnn/blob/1fc7fde819d6a7e082055ff10407f4744d862090/plugins/preview-tui#L105
> 
> and
> https://github.com/jarun/nnn/blob/1fc7fde819d6a7e082055ff10407f4744d862090/plugins/preview-tui#L110
> 
> from a background process sends garble to stdin for some reason, causing `nnn` to hang and the split is never opened.

I made some further changes and this issue is not present anymore, not sure what caused it unfortunately. Can be merged now as far as I'm concerned.

jarun:
Thank you! I am sure users will get back if they see any problems. ;)

luukvbaal:
To reiterate: this PR immediately sends `preview-tui` to the background to avoid flickering(shell prompt coming through) in the terminal window. The flickering is mostly gone on all terminals I've tested except `alacritty` which seems to draw at a higher fps and the prompt sometimes comes through, still an improvement to the old situation though.

If we want to get rid of the flicker entirely for the preview plugins I suppose we have to detach the process in `nnn` already. Probably not worth the extra logic though as we would have to make a distinction between normal plugins and plugins that only do something in the background.

PR #970, opened 2021-04-26
preview-tui: error to foreground

luukvbaal:
Bring `NNN_FIFO` error to foreground and `printf` instead of `echo`

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -8,7 +8,7 @@
 # Dependencies:
 #    - Supports 4 independent methods to preview with:
 #        - tmux (>=3.0), or
-#        - kitty with allow_remote_control on, or
+#        - kitty with allow_remote_control and listen_on set in kitty.conf, or
 #        - QuickLook on WSL (https://github.com/QL-Win/QuickLook)
 #        - $TERMINAL set to a terminal (it's xterm by default).
 #    - less or $PAGER
@@ -240,10 +240,10 @@ image_preview() {
         ueberzug_layer "$1" "$2" "$3"
     elif exists catimg; then
         catimg "$3" &
-        echo "$!" > "$GIFPID"
+        printf "%s" "$!" > "$GIFPID"
     elif exists viu; then
         viu -t "$3" &
-        echo "$!" > "$GIFPID"
+        printf "%s" "$!" > "$GIFPID"
     else
         fifo_pager print_bin_info "$3"
     fi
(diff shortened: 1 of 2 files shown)


luukvbaal:
The changes from #966 require the `listen_on` config option or `--listen-on` flag for `kitty` previews.

This option relegates the kitty remote control to a socket instead of terminal escape sequences which don't get caught properly when sent from a background process for some reason.

If anyone has any insight as to why this might be the case and if it can be fixed, I would be interested to know. For now however, suggesting the use of the `listen_on` config seems fine to me.

luukvbaal:
> If anyone has any insight as to why this might be the case and if it can be fixed, I would be interested to know. For now however, suggesting the use of the `listen_on` config seems fine to me.

I would say it's caused by the race condition for `/dev/tty` @kovidgoyal mentions in  in https://github.com/kovidgoyal/kitty/issues/2658#issuecomment-629600217.
Suggesting the use of `listen_on` seems the best way forward in my opinion.

jarun:
Thank you!

jarun:
@luukvbaal can you please add the kitty specific notes in the ToDo list in brief? We need to put it in the release notes for users who are on kitty.

jarun:
I added it. Also, in my experience previews are pretty smooth now with tmux.

luukvbaal:
> @luukvbaal can you please add the kitty specific notes in the ToDo list in brief? We need to put it in the release notes for users who are on kitty.

Kitty users will also be greeted with an error about the `listen_on` config but having it in the changelog is good. Thanks for adding it.

> I added it. Also, in my experience previews are pretty smooth now with tmux.

Yeah I myself use tmux previews as well and it really feels as if its part of nnn with the recent changes in my opinion!

jarun:
I see an issue with preview-tui-ext... when I hover on a directory, the previewer shows all entries in that dir recursively. All defaults. If the dir has a subdir with too many entries, I don't get to see the other entries in the subdir.

luukvbaal:
Yeah I don't see the point of it either but this is the `tree` output that has always been default in `preview-tui` as far as I can tell. Except the `-i` flag is left there by mistake from when I used `tree` for `iconlookup` preview. #972 restores the old default.

We can discuss a new default I guess, a non-recursive preview of the hovered directory makes the most sense to me(the current iconlookup view) but I think @leovilok suggested the paged `tree` view be kept default since it was a requested feature.

PR #975, opened 2021-04-27
Fix #974: picker mode broken

### LINKED ISSUE #974: picker mode broken
#918 is open again on master when using `NNN_OPTS=x` in picker mode. 
It would be nice to make picker mode work with this option set

luukvbaal:


luukvbaal:
On second thought just add back the `g_state.picker` test, retaining the other X functionality in picker mode.

jarun:
> On second thought just add back the g_state.picker test, retaining the other X functionality in picker mode.

Yes, just block the print.

jarun:
I will take care of the check. Thanks!

jarun:
Ahh OK. Ignore my comment. We are good as the X11 tweaks are not in the same place.

Issue #854, opened 2021-02-12
Unmount remote: change default from "hovered" to "hovered or parent"


Kabouik:
Currently when pressing `u` to unmount a remote folder, the user can write the name of the remote or unmount the hovered folder (which is default). Given that nnn is often built around optimizing the number of key presses for many actions, would it make sense to default to unmounting not only hovered folder, but any remote parent to the currently hovered folder? 

For instance, say I mount `pro` into `~/.config/nnn/mounts/pro`, then use this remote to browse some files in `~/.config/nnn/mounts/pro/home/pro/Pictures/Screenshots`, I would want to unmount it once I'm done, and I see only situations were it would be more straightforward to directly unmount with `u` then `return` while I'm in this child folder instead of browsing back to `~/.config/nnn/mounts/pro` or typing "pro". Am I missing some other situations where this would be an inconvenience?

0xACE:
I agree... but i do this instead (unmount all my mounts, trigger it when im finished with everything)
```
err=0
for d in ~/.config/nnn/mounts/*/; do
    mountpoint -q -- "$d"
    if [ "$?" -eq "0" ]; then
        fusermount3 -u "$d" || err=1
    fi
done

[ "$err" -eq "1" ] && read -n 1 -s -r -p "Failed to unmount. Press any key to continue"

exit 0
```

Feel free to fix the script, it may be incredibly broken...

Kabouik:
Interesting, thanks 0xACE. There are cases where unmounting all remotes at the same time could be inconvenient forme. Typically I would mount my phone temporarily to copy some pictures or documents, which is completely unrelated to the other remotes I may mount.

Surely I could leave my phone mounted, but as was discussed some months ago in another issue, I'm not comfortable with keeping wireless devices mounted in in a `~/.config/nnn` subfolder when not in use, I would likely forget about them eventually, and possibly break things badly when trying to stash or git pull in my `nnn` folder at a later point.

0xACE:
Yeah I agree, thats why I wish `u` would also unmount if you are in a sub-directory of a mounted directory... Alas I'm lazy and went with the script that unmounts them all (and honestly i forget to unmount them more often than you think). you could very easily re-adjust the script to do what you like... but really I'm hoping for that it your feature request gets accepted, as it makes sense to me that it should unmount the parent if it's a `nnn` mount...

[...]

PR #977, opened 2021-04-29
[Plugins] Fuzzy find plugins and run them; unmount mountpoints from their subfolders

Kabouik:
### fzfplug
I like having plugins bound to specific keys in my `~/.bashrc`, but there are some plugins that I would like to use more occaionnally, or simply others that I overlooked for way too long just because I know I have to edit and source my `~/.bashrc` or enter a prompt to just try them. This is an attempt to solve this issue.

The script tries to run a plugin with `$1` first for plugins that are made to work with hovered target, and if that fails, it tries again with no specific target (which apparently still takes any nnn selection into account). This is very hacky but seemed to work in the plugins that I could test. WIthout this, plugins like `upload` would fail as if they were run with no target file. 

### unmount-parent
See #854. After many tests, I found that the command to change nnn active directory (`printf "%s" "0c$dir" > "$NNN_PIPE"`) was breaking the nnn UI every time. I have checked other scripts using this command or `nnn_cd` from the plugin helper, I couldn't find any difference with the way I used the command, but there's indeed no UI break in their case. I finally decided to remove that part from the script to at least have a first working solution that I think would still be useful to those often mounting/unmounting remotes, even though it will obviously require lazy unmount when invoked from a subfolder since nnn being in a subfolder keeps the resource busy.

[asciicast](https://asciinema.org/a/0iuGLrjgjN2lsnwdziHSQI5rI)


jarun:
CI fails. Please fix.

@0xACE please review this.

jarun:
There are some shellcheck errors left.

Kabouik:
> There are some shellcheck errors left.

Sorry, I didn't know all those `read` options were undefined in POSIX. Thanks for the shellcheck tip! 

I see that the plugin `imgur` uses `read -rp`, but if I use the same, shellcheck still reports a warning in mine. I don't know much about POSIX, maybe I am missing something obvious elsewhere in the scripts?

luukvbaal:
The `imgur` plugin uses the `#!/usr/bin/env bash` shebang, and thus does not need to be POSIX compliant. 
Change the shebang to `bash` if you want to use the `read` options, or use a POSIX compliant [solution](https://stackoverflow.com/a/27875395) for the prompt.

plugins/README.md
```diff
@@ -87,8 +89,6 @@ Plugins are installed to `${XDG_CONFIG_HOME:-$HOME/.config}/nnn/plugins`.
 
 ## Configuration
 
-Set environment variable `NNN_PLUG` to assign keybinds and invoke plugins directly using the plugin shortcut (<kbd>;</kbd>) followed by the assigned key character. E.g., with the below config:
```
luukvbaal, line 90: Why remove this documentation?

Kabouik: That is totally a mistake when trying to fix the merge conflict yesterday in the middle of the night. I'll get back to it and try to fix that later today, thanks for noticing.

Kabouik:
> The `imgur` plugin uses the `#!/usr/bin/env bash` shebang, and thus does not need to be POSIX compliant.
> Change the shebang to `bash` if you want to use the `read` options, or use a POSIX compliant [solution](https://stackoverflow.com/a/27875395) for the prompt.

Excellent, thanks. I'll update that accordingly soon.

jarun:
@Kabouik I would prefer `sh` if it's not absolutely necessary to use `bash`.

Kabouik:
Cool, thanks. And excellent @luukvbaal, I was asking exactly that when you posted. I fixed `fzfplug` with this, will commit later with the rest. 

However I did use `$REPLY` in `unmount-parent` (in only one of the `read -p` lines). Is that OK if I make `fzfplug` POSIX-compliant and switch `unmount-parent` to bash? Note that the latter script will need some improvements in the future to better avoid lazy unmount when possible (_i.e._, make `nnn` context leave the subfolder before `fusermount`, which I tried to do in #854 but hit some UI bug).

luukvbaal:
> > where one would have to specifically hit Return to exit instead of any key
> 
> as with any other plugin in `nnn` now and I don't see a problem there.

It's not a problem per se, it's just that the `read -n1 -s` bashisms are more user friendly and `cfg=$(stty -g); stty raw -echo; head -c 1; stty "$cfg"` is a way to achieve the same functionality in POSIX shell.

I was considering updating the prompts in all the plugins but if you don't want to introduce this into the plugins I'll leave it be.

Kabouik:
It did happen to me quite a few times to expect that any key would exit, and then to be tempted to use backspace when I saw my random characters being printed to avoid sending unintentional instructions to a script in fear that it would interpret them. That is why I find the `read -n1 -s` behaviour less ambiguous. However I totally understand that you want to keep scripts straightforward and human-readable, that makes sense too. I think I took into account your comments and made both scripts POSIX-compliant again.

While I'm here, the reason I use so `clear` before several `printf` is because I observed all `nnn` scripts and builtin functions print text on top of each other if invoked successively, or if they print several things in a row. See for instance `cpg` on a copy/paste of two files: 

ss-2021-04-30_205416

Is that a known issue? I work around this in those scripts with the `clear`.

jarun:
The scripts are independent and we want users to find what they did if anythnig goes wrong after running a plugin.

Kabouik:
I'm thinkiing it might be best to clear before printing a new one instead, so that one going back to the shell could still see the log/warning/error.

jarun:
I'll think about it. Not in this PR.

@0xACE and @KlzXS please share your opinion on clearing the screen after running a plugin.

Kabouik:
> The scripts are independent and we want users to find what they did if anythnig goes wrong after running a plugin.

Was that a reply to the `clear` comment? In any case in those scripts, `clear` only happens when new text is printed, so the last error should stay readable. When `fzfplug` fails to execute a script, it keeps the error output from the other script. For instance, this is the result with `treeview` with my `$EDITOR` that doesn't understand the `-` at the end of this script:

ss-2021-04-30_210851

luukvbaal:
I would say catching and printing errors, followed by a prompt to return to nnn, provides a more user friendly experience.

jarun:
> I would say catching and printing errors,

Errors can be silent... like a file removed by a script that also prints the removed files. The user may want to go back and see what files were removed by the script.

jarun:
Thank you!

plugins/README.md
```diff
@@ -27,6 +27,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [fzcd](fzcd) | Change to the directory of a fuzzy-selected file/dir | sh | fzf |
 | [fzhist](fzhist) | Fuzzy-select a cmd from history, edit in `$EDITOR` and run | sh | fzf, mktemp |
 | [fzopen](fzopen) | Fuzzy find a file in dir subtree and edit or open | sh | fzf, xdg-open |
+| [fzfplug](fzfplug) | Fuzzy find, preview and run nnn plugins (or any plugin in an extra custom directory) | sh | fzf |
```
jarun: Is this alphabetically sorted? And why start with fzf and not follow the other plugin name patters that use fzf? This is very basic. Please be more careful for alter PRs.

Kabouik: You're absolutely right, it should be `fzplug` and this is actually how I sorted it without realizing. Do you want another PR to fix that? Sorry for missing that.

jarun: I took care of it.

plugins/fzfplug
```diff
@@ -1,32 +1,37 @@
...
 # Shell: POSIX compliant
 # Author: Kabouik
 
-# nnn PLUGINS SOURCE
-nnnplugins="$HOME/.config/nnn/plugins"
+# OPTIONAL SCRIPTS SOURCES
+# Leave blank or fill with the absolute path of a folder containing executable scripts other than nnn plugins
+# (e.g., "$HOME/.local/share/nautilus/scripts", since there are numerous Nautilus script git repositories).
+# Add extra variables if need be, but be sure to call them in the find command below at lines 28:49 and 30:49.
+CUSTOMDIR1="$HOME/.local/share/nautilus/scripts"
```
Kabouik, line 20: Better replace that with ` CUSTOMDIR1="" ` for default value in my opinion. This issue was introduced in the last commit before you merged unfortunately.

Kabouik:
You may want to change line 20 from `CUSTOMDIR1="$HOME/.local/share/nautilus/scripts"` to `CUSTOMDIR1=""` in a later revision.

The pre-filled value shouldn't break the plugin even if the directory does not exist, but pre-filling a custom option may not be desired as default. I commented above but I am not sure you received notifications for inline comments after merge.

plugins/fzfplug
jarun: Fixed in commit 95bbe5b3a7ec54d282a793344b172610df64f1e9.

PR #979, opened 2021-04-30
Make batchrename interactive

KlzXS:
Added the requested improvement. We no longer ignore `stderr` output from `mv` as that is what is used to display the overwrite prompts.

Please confirm that it works as expected.


KlzXS:
Should we do `.nmv` plugin as well?

jarun:
Yes, we should have this for `.nmv` too.

jarun:
Please update this PR.

jarun:
Thank you!

PR #981, opened 2021-05-01
preview-tui: kitty gif/pager fallback

luukvbaal:
* Avoid gif loop for `kitty` terminal since `icat` displays them natively
* Add pager fallback for matched text extensions


jarun:
Thank you!

PR #983, opened 2021-05-01
preview-tui: fix cleanup/resize

luukvbaal:
Fixes some issues I noticed that occurred when opening/closing `preview-tui` without changing selection and/or with first selection on a gif.


jarun:
Thank you!

PR #984, opened 2021-05-01
preview-tui: fix preview dir

luukvbaal:
Generated previews were being placed in `$NNN_PREVIEWDIR/nnn/previews/nnn/previews` with `NNN_PREVIEWDIR` set due to `preview-tui` being called recursively.

Fix and suggest setting to `$XDG_CACHE_HOME/nnn/previews` instead.


jarun:
Thank you!

PR #987, opened 2021-05-06
Use cfg.showdetail

luukvbaal:
Passing `(printptr == &printent_long)` seems redundant to me when we have `cfg.showdetail`.


jarun:
Thank you!

PR #989, opened 2021-05-08
Fix adjust_cols for icons

luukvbaal:
Account for the icon length when determining when to fall back to light mode to avoid
image


jarun:
Thank you!

Issue #935, opened 2021-04-05
ToDo list


jarun:
Rolled from #881.

### Cooking

- [x] <kbd>Bksp</kbd>/<kbd>Del</kbd> at empty filter prompt to refresh dir
- [x] fix zombies left behind after running plugins (#999)
- [x] consider nanosecond resolution when sorting by time (#978)
- [x] check external selection in `'c'urrent / 's'el` prompt (#976)
- [x] support picker plugins e.g. `fzopen` (https://github.com/mcchrish/nnn.vim/issues/82)
- [x] make batch rename interactive (#971)
- [x] cached uid/gid for performance improvement
- [x] support [QuickLook](https://github.com/QL-Win/QuickLook) on WSL in `preview-tui*` (#959)
- [x] toggle `preview-tui*` with the same _custom_ plugin key
- [x] smoother preview toggling in `preview-tui*` (#966)
- [x] config `listen_on` should be set in kitty.conf for `preview-tui*` (#970)
- [x] minimal `bat` style in `preview-tui*`, honors [BAT_STYLE](https://github.com/sharkdp/bat#output-style)
- [x] plugin `unmount-parent`: unmount a remote mountpoint from within
- [x] plugin `xdgdefault`: set the default application for hovered file type
- [x] plugin `fzplug`: fuzzy find, preview and run other plugins
- [x] plugin `cmusq`: queue/play music in `cmus`
- [x] plugin `mocplay` renamed to `mocq`
- [x] go to last dir on <kbd>~</kbd> (HOME) or <kbd>`</kbd> (ROOT) key repeat
- [x] key <kbd>^Space</kbd> for range selection/clear dropped, use <kbd>m</kbd> (#998)
- [x] a [Themes](https://github.com/jarun/nnn/wiki/Themes) wiki page for users to add and refer to
- [x] show selection mark (`+`) in reverse bold for improved visibility
- [x] reversed block instead of `>` to mark hovered entry in detail mode
- [x] make option `O_CKBOARD` retired in favour of reversed block cursor
- [x] retain filter in _nav-to-type_ mode after file open
- [x] script to statically compile with `netbsd-curses` + `musl` on Ubuntu
- [x] fixes for `netbsd-curses` + `musl` (https://github.com/sabotage-linux/netbsd-curses/issues/46)
- [x] fix broken picker mode with `NNN_TMPFILE` exported and <kbd>q</kbd> to quit
- [x] disable xterm title setting in picker mode (#974)
[...]

PR #992, opened 2021-05-08
Merge and simplify unescape

luukvbaal:
While considering to implement the suggestion in https://github.com/jarun/nnn/issues/935#issuecomment-835291032 for my fork and looking at `unescape()` I thought the while loop to reduce the number of wide chars to max columns seemed unnecessary.

This merges `unescape()` for the default and `NOLOCALE` case and adjusts the number wide chars to `maxcols` in the same call to `mbstowcs()` by providing `maxcols` instead of `NAME_MAX` as the character limit, avoiding the extra while loop.

Also came across the issue mentioned in #989 while testing this. Without #989 applied `wbuf[maxcols] = '\0';` results in a segfault with icons enabled and `maxcols < 2`.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3415,58 +3415,33 @@ static void resetdircolor(int flags)
  * Adjust string length to maxcols if > 0;
  * Max supported str length: NAME_MAX;
  */
-#ifndef NOLOCALE
+#ifdef NOLOCALE
+static char *unescape(const char *str, uint_t maxcols)
+{
+	char * const wbuf = g_buf;
+	char *buf = wbuf;
+
+	xstrsncpy(wbuf, str, maxcols);
+#else
 static wchar_t *unescape(const char *str, uint_t maxcols)
 {
 	wchar_t * const wbuf = (wchar_t *)g_buf;
 	wchar_t *buf = wbuf;
-	size_t lencount = 0;
 
 	/* Convert multi-byte to wide char */
-	size_t len = mbstowcs(wbuf, str, NAME_MAX);
+	mbstowcs(wbuf, str, maxcols);
+	wbuf[maxcols] = '\0';
+#endif
 
-	len = wcswidth(wbuf, len);
+	while (*buf) {
+		if (*buf <= '\x1f' || *buf == '\x7f')
+			*buf = '\?';
 
-	/* Reduce number of wide chars to max columns */
-	if (len > maxcols) {
-		while (*buf && lencount <= maxcols) {
-			if (*buf <= '\x1f' || *buf == '\x7f')
-				*buf = '\?';
-
-			++buf;
-			++lencount;
-		}
-
-		lencount = maxcols + 1;
-
-		/* Reduce wide chars one by one till it fits */
-		do
-			len = wcswidth(wbuf, --lencount);
-		while (len > maxcols);
-
-		wbuf[lencount] = L'\0';
-	} else {
-		do { /* We do not expect a NULL string */
-			if (*buf <= '\x1f' || *buf == '\x7f')
-				*buf = '\?';
-		} while (*++buf);
+		++buf;
 	}
 
 	return wbuf;
 }
-#else
-static char *unescape(const char *str, uint_t maxcols)
-{
-	ssize_t len = (ssize_t)xstrsncpy(g_buf, str, maxcols);
-
-	--len;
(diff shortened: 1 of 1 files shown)


jarun:
I remember there was some issue with broken wide characters because of which we had to consider [`wcswidth()`](https://man7.org/linux/man-pages/man3/wcswidth.3.html).

The third param in [`mbstowcs()`](https://linux.die.net/man/3/mbstowcs) denotes max wide characters that are written to dest and not bytes. Some wide characters take more than one column (say, in Indian languages). I am not sure about this change.

luukvbaal:
Ah I see. Yeah I hadn't considered that, my bad. Feel free to close as you see fit.

jarun:
Yes, let's drop it.

luukvbaal:
> Yes, let's drop it.

@jarun Were you able to verify that this implementation doesn't work with any characters btw? I tested it with a wide glyph I have on my system and it seems to work fine:
image
image

Seems like a good improvement to me to merge the two `unescape()` functions and to drop the while loop for determining the widechar column width so if it works I'd say why not. If you have any other problems with the implementation of course nevermind.

jarun:
It breaks. And please give me a break. Run `mktest.sh` and test with the korean chars.

I really keep busy with different stuff and I need people to stop harping on the same string.

jarun:
OK. I understand you didn't know about the korean string we had for testing.

luukvbaal:
I see, yeah I didn't know about the test string sorry.

jarun:
No problem!

luukvbaal:
BTW I found an implementation that works for the Korean strings while still merging and simplifying the two `unescape()` functions and still getting rid of the "Reduce number of wide chars to max columns"  while loop: https://github.com/luukvbaal/nnn/commit/f0b665a62c383d65540692a6b534c41cf101b57c Let me know if I should reopen the PR or if you have no interest in merging the two nevermind.

jarun:
I ma thinking of the following detection logic:

```diff
diff --git a/src/nnn.c b/src/nnn.c
index dde27b6..9ba8028 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3417,7 +3417,7 @@ static void resetdircolor(int flags)
  * Max supported str length: NAME_MAX;
  */
 #ifndef NOLOCALE
-static wchar_t *unescape(const char *str, uint_t maxcols)
+static wchar_t *unescape_w(const char *str, uint_t maxcols)
 {
        wchar_t * const wbuf = (wchar_t *)g_buf;
        wchar_t *buf = wbuf;
@@ -3457,19 +3457,23 @@ static wchar_t *unescape(const char *str, uint_t maxcols)
 
        return wbuf;
 }
-#else
+#endif
+
 static char *unescape(const char *str, uint_t maxcols)
 {
        ssize_t len = (ssize_t)xstrsncpy(g_buf, str, maxcols);
 
        --len;
-       while (--len >= 0)
+       while (--len >= 0) {
+#ifndef NOLOCALE
+               if (g_buf[len] & 0x8)
+                       return NULL;
+#endif
                if (g_buf[len] <= '\x1f' || g_buf[len] == '\x7f')
                        g_buf[len] = '\?';
-
+       }
        return g_buf;
 }
-#endif
 
 static off_t get_size(off_t size, off_t *pval, uint_t comp)
 {
@@ -3754,7 +3758,10 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
                ++namecols;
 
 #ifndef NOLOCALE
-       addwstr(unescape(ent->name, namecols));
+       if (unescape(ent->name, MIN(namecols, ent->nlen) + 1))
+               addstr(g_buf);
+       else
+               addwstr(unescape_w(ent->name, namecols));
 #else
        addstr(unescape(ent->name, MIN(namecols, ent->nlen) + 1));
 #endif
```

luukvbaal:
Hmm okay. https://github.com/luukvbaal/nnn/commit/92e01d4594edabb8028c4c9a06c0ae977fc2294c is my final implementation of how I would merge the two and I see no problems with it.

Seems more elegant to me and should be more performant due to dropping a while loop but do whatever you feel is best.

jarun:
The detection (plus on-the-fly conversion) logic performs better on English-native systems but not on non-English-native systems.

luukvbaal:
Sure, made the PR #993.

PR #993, opened 2021-05-09
Merge and simplify unescape

luukvbaal:
Following discussion from #992


luukvbaal:
Moved `lencount` to inside the if statement since we only need it when `len > maxcols`.

PR #994, opened 2021-05-10
Drop printptr for if-statement

luukvbaal:
Was wondering if we should drop `printptr` now that we can replace it with a single if-statement resulting in the same logic.


jarun:
Yes, I think it's time to do that.

jarun:
Thank you!

jarun:
With the following make option I see a 4 KiB reduction (111 to 107) in binary size on my system. Good job!

    make CFLAGS+=-march=corei7 O_NORL=1 O_NOMOUSE=1 O_NOBATCH=1 O_NOSSN=1 O_QSORT=1 O_NOUG=1 O_NOX11=1 O_NOLOC=1 O_NOFIFO=1 strip

PR #995, opened 2021-05-10
Unescape to void

luukvbaal:
Not sure if it's an improvement over returning the local buffer but since we have the global buffer anyways I thought we might as well use it in the `addstr` call instead of returning the local buffer. What do you think?

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3414,28 +3414,24 @@ static void resetdircolor(int flags)
  * Adjust string length to maxcols if > 0;
  * Max supported str length: NAME_MAX;
  */
+static void unescape(const char *str, uint_t maxcols)
+{
 #ifdef NOLOCALE
-static char *unescape(const char *str, uint_t maxcols)
-{
-	char * const wbuf = g_buf;
-	char *buf = wbuf;
+	char *buf = g_buf;
 
-	xstrsncpy(wbuf, str, maxcols);
+	xstrsncpy(buf, str, maxcols);
 #else
-static wchar_t *unescape(const char *str, uint_t maxcols)
-{
-	wchar_t * const wbuf = (wchar_t *)g_buf;
-	wchar_t *buf = wbuf;
-	size_t len = mbstowcs(wbuf, str, maxcols); /* Convert multi-byte to wide char */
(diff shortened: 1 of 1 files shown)


jarun:
No, I think the current implementation is fine.

jarun:
In the 2 places I used the comma operator, I did it to execute the first function call/assignment within the same condition check.

jarun:
unescape returning a value is much cleaner.

luukvbaal:
Yeah agreed.

luukvbaal:
However I'm not sure why we have `wbuf` and `buf` in `unescape()`, can we not drop `wbuf` i.e.
```diff
#ifdef NOLOCALE
static char *unescape(const char *str, uint_t maxcols)
{
-	char * const wbuf = g_buf;
-	char *buf = wbuf;
+	char *buf = g_buf;

-	xstrsncpy(wbuf, str, maxcols);
+	xstrsncpy(buf, str, maxcols);
#else
static wchar_t *unescape(const char *str, uint_t maxcols)
{
-	wchar_t * const wbuf = (wchar_t *)g_buf;
-	wchar_t *buf = wbuf;
-	size_t len = mbstowcs(wbuf, str, maxcols); /* Convert multi-byte to wide char */
+	wchar_t *buf = (wchar_t *)g_buf;
+	size_t len = mbstowcs(buf, str, maxcols); /* Convert multi-byte to wide char */

-	len = wcswidth(wbuf, len);
+	len = wcswidth(buf, len);

	if (len >= maxcols) {
		size_t lencount = maxcols;
		while (len > maxcols) /* Reduce wide chars one by one till it fits */
-			len = wcswidth(wbuf, --lencount);
+			len = wcswidth(buf, --lencount);

-		wbuf[lencount] = L'\0';
+		buf[lencount] = L'\0';
	}
#endif

@@ -3446,7 +3444,7 @@ static wchar_t *unescape(const char *str, uint_t maxcols)
		++buf;
	}

-	return wbuf;
+	return buf;
}
```

luukvbaal:
Ah right, I was avoiding that by passing `g_buf` to `printw()` in my fork, sorry.

PR #997, opened 2021-05-11
print .0 for file sizes

luukvbaal:


Kabouik:
Glad to see this merged, thanks!

luukvbaal:
You're right, the current change only helps for Kilobyte files. We would need a more elaborate change to `coolsize()` to align larger files.

luukvbaal:
I could look into making it look like:
```
 2020-10-23 23:24  664    5.29M  coreutils-8.32.tar.xz
 2020-09-20 22:10  775    4.00K  xtrlock
 2020-01-18 21:40  664    5.61M  Bash-Snippets.7z
```
Original:

```
 2020-10-23 23:24  664    5.29M  coreutils-8.32.tar.xz
 2020-09-20 22:10  775       4K  xtrlock
 2020-01-18 21:40  664    5.61M  Bash-Snippets.7z
```

jarun:
4.0 and 4.0000 all are 4. Let's leave it as it is.

luukvbaal:
Sure they are, the point is having the column aligned properly though:yum:

Kabouik:
For what it's worth (not much, I'm just a user), I agree with @luukvbaal. The file size is the only right-aligned column currently, which shows some intent to ease visualization of aligned numbers, but it partly fails due to decimals being removed when they are `.00`.

On the other hand, I agree that adding `.00` everywhere might clutter the view when in a directory containing files with round sizes only. Not sure how common that is, but this is the (togglable) Details view anyway, so perhaps one nnn view where users may not prioritize a minimal display.

PR #1000, opened 2021-05-12
Wait for child on plugin invocation

### LINKED ISSUE #999: Question: nnn leaves zombies while running plugins
I noticed that each time i run plugin in nnn it creates zombie nnn child processes.
It happens both with plugins which run in background, such as
``` sh
#!/bin/sh
nohup zathura "$1" >/dev/null 2>&1 &
```
and foreground plugins, such as
``` sh
#!/bin/sh
less -S "$1"
```
After each plugin call nnn gets new child zombie, which will remain till death of the main nnn process.
exmpl
This also happens with `imgview` with `NNN_PLUG='i:imgview'`.

**Thus question**: Is that intented?

I hope i'm not being too stupid, but i didn't manage to find an answer or a similar question myself.

KlzXS:
Fixes #999

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4785,6 +4785,8 @@ static void readpipe(int fd, char **path, char **lastname, char **lastdir)
 
 static bool run_selected_plugin(char **path, const char *file, char *runfile, char **lastname, char **lastdir)
 {
+	int status;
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1001, opened 2021-05-12
nerdfont support for chess files(pgn,fen) haskell(hs,lhs) files and hpp

CantoroMC:
If this pr is accepted, do you like if a provide some other changes to nerdfont icons?
Also, i would like to ask .....
I really like more the previous current item and selection indicators, if I provide a pr to reintroduce them with a compiler flag, will it be accepted, or have you decided to stick with the current method?

### DIFF
--- a/src/icons-nerdfont.h
+++ b/src/icons-nerdfont.h
@@ -118,6 +118,7 @@
 /* H */
 #define ICON_EXT_H         ICON_C
 #define ICON_EXT_HH        ICON_CPLUSPLUS
+#define ICON_EXT_HS        "\ue777"
 #define ICON_EXT_HTACCESS  ICON_CONFIGURE
 #define ICON_EXT_HTPASSWD  ICON_CONFIGURE
 #define ICON_EXT_HTM       ICON_HTML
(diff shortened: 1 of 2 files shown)


jarun:
Thanks for the PR!

> do you like if a provide some other changes to nerdfont icons?

> I really like more the previous current item and selection indicators

Please remind me what was the previous current indicator.

By previous selection indicator do you mean just the `+` without reverse for selection?

Please share the details.

@luukvbaal please review this.

CantoroMC:
The previous indicator were > for the current item and + for the selection.
The selection has now also reversed colors which I don't like (I asked cause I understand that my taste may not be shared).

jarun:
I would like to leave the current indicator as it is because it's much easier to find the hovered line.

For the selection marker I initially chose bold `+`.
However, switched to the current style because it's only temporary and it's much easier to find visually.

CantoroMC:
Yes of course. Thanks again.

jarun:
Thank you!

luukvbaal:
Thanks for extending the icon list, looks good to me.

I must say I'm also not a big fan of the inverse current/selection indicators. I did like the change to bold which I guess I will be using on my personal builds.

jarun:
Bold doesn't work for people who disable it in the terminal.

PR #1002, opened 2021-05-12
Add chess/haskell icons to .iconlookup

luukvbaal:
Define grouping macros for chess/haskell icons added in #1001 as we have done for other icons used for multiple file extensions. Also add them to `.iconlookup`.


jarun:
Thank you!

PR #1006, opened 2021-05-14
Kill kitty icat in preview-tui

luukvbaal:
Should fix https://github.com/jarun/nnn/issues/935#issuecomment-840850150.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -71,7 +71,7 @@ TMPDIR="${TMPDIR:-/tmp}"
 startpreview() {
     NUMPREVIEWTUI="$(($(find "$TMPDIR" -maxdepth 1 -name 'nnn-preview-tui-pagerpid*' 2>/dev/null | wc -l) + 1))"
     PAGERPID="$TMPDIR/nnn-preview-tui-pagerpid.$NUMPREVIEWTUI"
-    GIFPID="$TMPDIR/nnn-preview-tui-gifpid.$NUMPREVIEWTUI"
+    IMGPID="$TMPDIR/nnn-preview-tui-gifpid.$NUMPREVIEWTUI"
     CURSEL="$TMPDIR/nnn-preview-tui-selection.$NUMPREVIEWTUI"
     FIFO_UEBERZUG="$TMPDIR/nnn-preview-tui-ueberzug-fifo.$NUMPREVIEWTUI"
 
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

luukvbaal:
@Kabouik did you see https://github.com/jarun/nnn/issues/935#issuecomment-840937419? Can you confirm that this commit fixed the queuing issue?

Kabouik:
Thanks a lot @luukvbaal. I still observe some queuing when holding down <kbd>Up<kbd> or <kbd>Down<kbd> with kitty's `icat` even on my PC. I haven't pulled master on the phone yet but that should be the same.

I'm happy to try `ueberzug` if it performs better (damn, I just spent a few hours during the last two days to configure my `kitty.conf` and finally have image previews in my terminal thanks to `icat`, I didn't know performance was notoriously bad compared to alternatives!). However I think right now `preview-tui*` prioritizes kitty's `icat` if available. We might want to change that if `ueberzug` is also installed and performs better, no?

luukvbaal:
How does the queuing manifest? For me when I hold down <kbd>Down</kbd> in `kitty`, nothing is drawn by `icat` until I release <kbd>Down</kbd> i.e. the preview window stays empty. I don't experience any queuing of previews as before this commit.

Not sure, people using `kitty` will probably expect image previews to be handled by `icat`. Your best bet would probably to switch to something like `alacritty` if you want to use `ueberzug` previews. 

On the other hand, if the user indeed installed `ueberzug` I suppose we might also reasonably expect them to want `ueberzug` previews in `kitty`. Would need feedback from `kitty` for such a change. Otherwise, just move the `ueberzug` if clause above `kitty` in `image_preview` youself.

jarun:
> Otherwise, just move the ueberzug if clause above kitty in image_preview youself.

This. Once again, plugins are indicative.

Kabouik:
That's totally fine by me, no worries. I was just wondering if kitty users who also have installed `ueberzug` may be intersted in using it by default since it performs better. On the other hand, they may also have installed it to have image support in other terminal emulators but still prefer to use `icat` when in kitty, I don't know.

This is what the queuing looks like on my end (in `icat`) even after the changes of this PR:

https://user-images.githubusercontent.com/7107523/118283663-21938000-b4bf-11eb-86e8-8226a439d490.mp4

This is on a PC with hardware acceleration here (the queuing is obviously more severe on the slow machine with no hardware acceleration).

Kabouik:
Nevermind, there are conflicts on my machine between my ongoing PR and yours here, which I hadn't noticed. This is the cause.

PR #1008, opened 2021-05-14
Fix conflict between #1004 #1006

Kabouik:
#1004 was merged seconds before I fixed a copy/paste error that had come unnoticed, this is the fix. Sorry for the extra work.


Kabouik:
Closing since it can easily be included in a single #1009 PR.

PR #1009, opened 2021-05-14
If/else to case in preview-tui

luukvbaal:
Replace `if/else` clauses with `case` block in view of optimizations mentioned in https://github.com/jarun/nnn/issues/935#issuecomment-841088727. Should maintain exactly the same logic as previously.

I'm not sure how I feel about maintaining `preview-tui` in the future though. If it were up to me I would just rename `preview-tui-ext` to `preview-tui` and let users edit that how they see fit instead of extending the more minimal `preview-tui`. The extended script works fine without the optional dependencies installed.

### DIFF
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -131,7 +131,7 @@ startpreview() {
             --env "PREVIEW_MODE=1" --env "PAGER=$PAGER" --env "TMPDIR=$TMPDIR" \
             --env "USE_SCOPE=$USE_SCOPE" --env "SPLIT=$SPLIT" --env "TERMINAL=$TERMINAL"\
             --env "USE_PISTOL=$USE_PISTOL" --env "BAT_STYLE=${BAT_STYLE:-numbers}" \
-            --env "PAGERPID=$PAGERPID" --env "GIFPID=$GIFPID" --env "FIFO_UEBERZUG=$FIFO_UEBERZUG" \
+            --env "PAGERPID=$PAGERPID" --env "IMGPID=$IMGPID" --env "FIFO_UEBERZUG=$FIFO_UEBERZUG" \
             --env "ICONLOOKUP=$ICONLOOKUP" --env "NNN_PREVIEWDIR=$NNN_PREVIEWDIR" \
             --env "NNN_PREVIEWWIDTH=$NNN_PREVIEWWIDTH" --env "NNN_PREVIEWHEIGHT=$NNN_PREVIEWHEIGHT" \
             --env "CURSEL=$CURSEL" --location "${SPLIT}split" "$0" "$1" >/dev/null
--- a/plugins/preview-tui-ext
+++ b/plugins/preview-tui-ext
@@ -200,6 +200,63 @@ print_bin_info() {
     fi
 }
 
+handle_mime() {
+    case "$2" in
+        image/jpg) image_preview "$cols" "$lines" "$1" ;;
+        image/gif) generate_preview "$cols" "$lines" "$1" "gif" ;;
+        image/*) generate_preview "$cols" "$lines" "$1" "image" ;;
+        video/*) generate_preview "$cols" "$lines" "$1" "video" ;;
+        audio/*) generate_preview "$cols" "$lines" "$1" "audio" ;;
+        application/font*|application/*opentype) generate_preview "$cols" "$lines" "$1" "font" ;;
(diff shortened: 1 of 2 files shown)


plugins/preview-tui-ext
```diff
@@ -355,7 +351,8 @@ generate_preview() {
 image_preview() {
     clear
     if [ "$TERMINAL" = "kitty" ]; then
-        # Kitty terminal users can use the native image preview method.
+        # Kitty terminal users can use the native image preview method although one might consider
+        # installing ueberzug and moving the ueberzug clause to the top since it performs alot better.
```
luukvbaal, line 355: Add note about `kitty` vs `ueberzug` performance consideration

plugins/preview-tui-ext
```diff
@@ -246,68 +301,9 @@ preview_file() {
...
-        else
-            fifo_pager pager "$1"
-        fi
-    elif [ "$mimetype" = "text/troff" ]; then
-        if exists man; then
-            fifo_pager man -Pcat -l "$1"
-        else
-            fifo_pager pager "$1"
-        fi
+        handle_mime "$1" "$mimetype" "$ext" "bin"
```
luukvbaal, line 304: Pass "bin" as argument to `handle_mime()` to maintain the `if [ "${encoding#*)}" = "binary" ]` condition.

plugins/preview-tui-ext
```diff
@@ -198,6 +198,63 @@ print_bin_info() {
...
+        application/font*|application/*opentype) generate_preview "$cols" "$lines" "$1" "font" ;;
+        */*office*|*/*document*) generate_preview "$cols" "$lines" "$1" "office" ;;
+        application/zip) fifo_pager unzip -l "$1" ;;
+        text/troff)
+            if exists man; then
+                fifo_pager man -Pcat -l "$1"
+            else
+                fifo_pager pager "$1"
+            fi ;;
+        *) handle_ext "$1" "$3" "$4" ;;
```
luukvbaal, line 217: If no `mimetype` matches, check extensions

plugins/preview-tui-ext
```diff
@@ -198,6 +198,63 @@ print_bin_info() {
...
+            fi ;;
+        md) if exists glow; then
+                fifo_pager glow -s dark "$1"
+            elif exists lowdown; then
+                fifo_pager lowdown -Tterm "$1"
+            else
+                fifo_pager pager "$1"
+            fi ;;
+        pdf) generate_preview "$cols" "$lines" "$1" "pdf" ;;
+        *) if [ "$3" = "bin" ]; then
```
luukvbaal, line 250: `print_bin_info()` or `pager` depending on whether or not we came from `if [ "${encoding#*)}" = "binary" ]`

luukvbaal (commented):
@leovilok please review if you can

Kabouik:
Thanks for picking that up @luukvbaal. Seeing how a simple change like 1004 can be hectic if I'm the one doing it, I'm glad you stepped forward!

I feel the same about `preview-tui`. I know its merits are it's simpler than `preview-tui-ext`, but we're talking about 11.3 and 17.2 KB files, respectively, and a single 17.2 KB file is still lighter than that. Also, of course users should only customize the one they intend to use, but I feel uneasy editing `preview-tui-ext` to my liking and leaving `preview-tui` out of sync, so I end up patching both for no reason even when there's no PR planned.

Kabouik:
Unless I am missing some magic Github feature that would automatically merge, those two `preview-tui*` plugins are pre-#1004 (_i.e._, `bat` changes), are they not?

luukvbaal:
Correct, I was working pre-#1004. It will merge fine since there are no merge conflicts, I'll rebase though.

Kabouik:
> Correct, I was working pre-#1004. It will merge fine since there are no merge conflicts, I'll rebase though.

Since #1008 is rather small, maybe take that into account at the same time and scrap #1008 before it's merged?

luukvbaal:
As you want, if you close it I'll add it here

Kabouik:
Thanks!

jarun:
Thank you!

luukvbaal:
> I'm not sure how I feel about maintaining preview-tui in the future though. If it were up to me I would just rename preview-tui-ext to preview-tui and let users edit that how they see fit instead of extending the more minimal preview-tui. The extended script works fine without the optional dependencies installed.

@jarun thoughs on this?

jarun:
After the next release.

PR #1010, opened 2021-05-15
[Plugin] Queue/play in cmus player

Kabouik:
A cmus alternative to existing audio player plugins. I hope you'll find that a relevant addition to `boom` and `mocplay`.


Kabouik:
https://user-images.githubusercontent.com/7107523/118360357-78fd2300-b587-11eb-8259-02297d364e50.mp4

The powerful nnn selection system (individual files through multiple directories and/or later edition of selection in $EDITOR) makes nnn a really great tool to manipulate an audio play queue for those who like directory-based music libraries.

luukvbaal:
You should wrap the code for starting `cmus` in a function and use that instead of duplicating the code e.g.
```
start_cmus() {
    printf "cmus is not running, starting it in a new window."
    case "$TERMINAL" in
        kitty | gnome-terminal | st)
            nohup "$TERMINAL" -- cmus > /dev/null 2>&1 &
            cmuspid=$!
            ;;
        havoc)
            nohup "$TERMINAL" cmus > /dev/null 2>&1 &
            cmuspid=$!
            ;;
        "")
            nohup x-terminal-emulator -e cmus > /dev/null 2>&1 &
            cmuspid=$!
            ;;
        *)
            nohup "$TERMINAL" -e cmus > /dev/null 2>&1 &
            cmuspid=$!
            ;;
    esac
    until [ -n "$cmuspid" ]; do # Give the new terminal some time to open
        sleep 0.25
    done
}
```
Also use `if ! pgrep cmus >dev/null; then` instead of `if [ "$(pgrep cmus)" = "" ]; then`

jarun:
@Kabouik great to find another Leonard Cohen fan. ;)

Kabouik:
Thanks for the comments! It should be much better now.

Also, the way I was waiting for the `cmus` process to start before using `cmus-remote` was very ugly and partly broken (I was monitoring the new terminal window PID instead of the `cmus` PID). It should now wait for `cmus` to start in the new terminal window in a cleaner way; tested on both fast and slow computers. I also added an optional `xdotool` clause to keep focus on `nnn` when `xdotool` is installed.

plugins/cmusqueue
```diff
@@ -0,0 +1,84 @@
...
+    read -r REPLY
+fi
+
+# If [s]election, then start cmus in new window (if necessary) and append selection to queue
+if [ "$REPLY" = "s" ]; then
+    if ! pgrep cmus >/dev/null; then
+        start_cmus
+        printf "Appending files to cmus queue\n"
+        # Queue, change view and play
+        cmus-remote -C "view 4"; xargs < "$selection" -0 cmus-remote -q; cmus-remote -p; unset cmuspid
```
luukvbaal, line 61: Can't you just end the if statement here and use the same `cmus-remote` command for both cases?

plugins/cmusqueue
```diff
@@ -0,0 +1,82 @@
...
+            nohup "$TERMINAL" cmus >/dev/null 2>&1 &
+            ;;
+        "")
+            nohup x-terminal-emulator -e cmus >/dev/null 2>&1 &
+            ;;
+        *)
+            nohup "$TERMINAL" -e cmus >/dev/null 2>&1 &
+            ;;
+    esac)
+    until [ -n "$cmuspid" ]; do # Give the new terminal some time to open
```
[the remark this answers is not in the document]
Kabouik: Thanks!

plugins/cmusqueue
```diff
@@ -0,0 +1,82 @@
...
+        cmuspid="$(pgrep cmus)"; sleep 0.20
+    done
+    if [ -x "$(command -v xdotool)" ]; then
```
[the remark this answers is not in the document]
Kabouik: Thanks!

plugins/cmusqueue
Kabouik: I prefer not to force playback with `cmus-remote -p` when the user already has a cmus instance running but paused, as this would mean he/she may not necessarily intend to play files right at the moment. By contrast, if one has no cmus running but decides to open it in a new window, it's safe to assume starting playback should be the default behaviour.

plugins/README.md
```diff
@@ -20,6 +20,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [cdpath](cdpath) | `cd` to the directory from `CDPATH` | sh | fzf |
 | [chksum](chksum) | Create and verify checksums [✓] | sh | md5sum,<br>sha256sum |
 | [cleanfilename](cleanfilename) | Clean filename to be more shell-friendly [✓] | bash | sed |
+| [cmusqueue](cmusqueue) | Queue files/dirs in cmus player | sh | cmus, pgrep |
```
jarun, line 23: Please rename to `cmusq`. The shorter the sweeter as long as it's meaningful.

plugins/cmusqueue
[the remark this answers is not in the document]
Kabouik: This has started to print the PID in the terminal though, I'm currently looking for ways to hide that.

plugins/cmusqueue
luukvbaal: I see

plugins/cmusqueue
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env sh
+
+# Description: Add selection or hovered file/directory to cmus queue.
+# If adding selection, files/dirs are added in the same order they were selected in nnn.
```
jarun, line 4: Please add as `Notes`.

Follow the format in `autojump` plugin. It's clean and well-readable.

jarun (changes requested):

plugins/cmusqueue
```diff
@@ -0,0 +1,84 @@
...
+# cmus will be started in a new window if not already running and playback will start immediately.
+# If cmus is already running, files will just be appended to the queue (no forced playback).
+#
+# Todo: Add cava as optional dependency
```
jarun, line 8: TODO

jarun: See plugin `bookmarks`.

plugins/cmusqueue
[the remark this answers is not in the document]
luukvbaal: `pidof cmus >/dev/null` sorry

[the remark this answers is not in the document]
Kabouik: Hum, apparently there's no `-q` flag in my version of `pidof`.

```
$ pidof -V
pidof from procps-ng UNKNOWN
```

Helpful. :<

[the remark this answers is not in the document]
luukvbaal: Then use `pidof cmus >/dev/null`

plugins/cmusqueue
```diff
@@ -0,0 +1,84 @@
...
+#
+# Dependencies: cmus, pgrep, xdotool (optional)
```
jarun, line 10: Please add deps below description.

Kabouik:
Thanks for the multiple helpful comments. I think I have taken care of them. Also harmonized my other plugins' descriptions according to @jarun comments.

jarun:
Thanks!

@luukvbaal please do a final review and merge if all is well.

luukvbaal:
Made sure to exit if `cmus` is missing and some other style changes. @Kabouik please confirm it still works

Kabouik:
Thanks for the improvements!

I will further test later when I am back at my computer, but from what I can see on my phone, I seem to have lost some functionality with the latest changes @luukvbaal. Namely, when a new cmus window is opened, the files are not enqueued and the view is not changed. I assume the `fill_queue()` part is running too early, before the cmus pid is found maybe?

Also with the change in `bat`  detection in `fzplug`, `bat` is no longer used and `fzplug` defaults to `cat`. I had tried that change you made (after you suggested it for `xdotool` detection in `cmusq`), but observed that and postponed. Can you reproduce?

luukvbaal:
Can't reproduce either issue with `fzplug` and `cmusq` from master😅

Kabouik:
Hum, that's good news I guess! I still have to figure out why I'm getting this though. `cmusq` was working on both machines (slow and fast) when I pushed the last commit, and the issue with `bat` detection in `fzplug` also occurred on my other computer running another distro (and confirmed now on my phone).

luukvbaal:
The first issue might indeed be that the `cmus` server is not ready to accept `cmus-remote` commands on slow machines, although I can't reproduce it. Should be fixed by
```diff
-    until pidof cmus >/dev/null; do sleep 0.1; done
+    until cmus-remote -C; do sleep 0.1; done
```

Your second issue is because the if/else clauses should be switched around lol, sorry for not testing. Making the PR.

Kabouik:
Yeah I expect the issue only shows up on machine that need some time to open a new terminal window. 

I can confirm the diff works, thanks. I don't exactly understand how that works since I see no difference when executing `cmus-remote -C` manually whether cmus is running or not, but I assume it fails silently until cmus is detected. Great idea!

luukvbaal:
Correct, see #1011 for the `fzlplug` fix.

PR #1011, opened 2021-05-15
Wait until cmus is ready and fix fzplug preview command

luukvbaal:
Fix issues mentioned in https://github.com/jarun/nnn/pull/1010#issuecomment-841713987

### DIFF
--- a/plugins/cmusq
+++ b/plugins/cmusq
@@ -40,7 +40,7 @@ start_cmus() {
             nohup "$TERMINAL" -e cmus & ;;
     esac
     # Give the new terminal some time to open
-    until pidof cmus >/dev/null; do sleep 0.1; done
+    until cmus-remote -C; do sleep 0.1; done
     [ -n "$nnnwindow" ] && xdotool windowactivate "$nnnwindow"
 } >/dev/null 2>&1
 
--- a/plugins/fzplug
+++ b/plugins/fzplug
@@ -31,19 +31,13 @@ nnnpluginsdir="$HOME/.config/nnn/plugins"
 
 # Preview with bat if installed
 if type bat >/dev/null; then
-    plugin=$(find "$nnnpluginsdir" "$CUSTOMDIR1" "$CUSTOMDIR2" \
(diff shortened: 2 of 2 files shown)


jarun:
Thank you!

Kabouik:
Thanks!

Kabouik:
Although this improved the code a lot, there is one relatively minor drawback to using a single `$plugin` command. The `fzf` `:wrap` option is now enforced whether `bat` is detected or not, even though `bat` does a better job when it knows the terminal width. The prior version used `:wrap` only for the `cat` variant.

See comparison (master on the left, previous commit on the right):

Screenshot_20210515_001

Kabouik:
For this particular plugin, removing `bat` line numbers would help:

Screenshot_20210515_002

No real need to show line numbers in plugins a user is about to execute (and not to edit) anyway.

luukvbaal:
Agreed, I would also just get rid of the wrap option for `fzf`

Kabouik:
With no `:wrap`, some plugin descriptions may be unreadable depending on the window size. I found myself using new plugins much more frequently since `fzfplug` due to the `fzf` preview pane.

jarun:
@luukvbaal and @Kabouik are you guys familiar with vimscript?

luukvbaal:
I guess, to me a preview window is not meant for actual reading but rather just for a quick glance at the context. Line wrapping only makes it harder to read to my eyes.

luukvbaal:
Not particularly well versed in it @jarun. I know about the open issues for `nnn.vim` though, I'll take a look if I ever I feel like it.

jarun:
OK thanks!

Kabouik:
> I guess, to me a preview window is not meant for actual reading but rather just for a quick glance at the context. Line wrapping only makes it harder to read to my eyes.

I agree to some extent but I see `fzplug` as a script launcher/explorer and I am not necessarily familiar with those scripts since I haven't authored them. `fzplug` doesn't offer a shortcut to edit/view the scripts, as opposed to files previewed with `preview-tui*` that can be opened with just a keystroke, therefore exhaustive preview means discoverability for me in that case.

jarun:
No problem!

PR #1013, opened 2021-05-16
Preview-tui cleanup

luukvbaal:
Further cleanup of `preview-tui*`
- `start_preview()` `if/else` to `case`
- per-function redirection instead of per-command
- remove some code duplication in `preview_image()`
- `< "$NNN_FIFO"` instead of `cat` in `preview_fifo()` works fine in `dash` for me
- job control through `FIFOPID` tmpfile

Let me know if you want me to do a more detailed self review but If you're fine with merging for wider testing I wont bother.


jarun:
Just awesome!

PR #1015, opened 2021-05-16
Fix preview-tui zombie pager

luukvbaal:
I can't explain the behavior I was seeing but ONLY when starting `nnn` with a command such as `st -c file tmux new-session -A -s nnn nnn -Pp` would result in a zombie `tmux` pane with an empty pager when closing `nnn` (Introduced by #1013).

Didn't happen when just manually starting `nnn` and toggling `preview-tui*` or even when starting with `preview-tui*` open with `nnn -Pp` in an existing `tmux` session. 

Even when the pager wasn't used for the previewed entry at all(e.g. `ICONLOOKUP` dir entry), it would open after exiting `nnn` and leave an empty `tmux` pane.

Anyhow swapping around the tmpfiles `rm` and `pkill` seems to fix it although I can't really explain why.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -307,8 +307,8 @@ if [ "$PREVIEW_MODE" ]; then
     preview_file "$PWD/$1"
     preview_fifo &
     wait "$!"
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

@leovilok can you please look into it?

PR #1018, opened 2021-05-17
Preview-tui fixes

luukvbaal:
- Restore `viu/catimg` previews after regression in #1013 (was redirected to `/dev/null`)
- Handle `WINCH` for all image previewers, not just `ueberzug` 
- Reduce tests for `ueberzug` case by testing if named pipe `FIFO_UEBERZUG` exists
- Restore cleanup trap to remove tmpfiles when exiting `nnn` with <kbd>q</kbd>


jarun:
Thank you!

PR #1019, opened 2021-05-18
Preview-tui fix winch_handler

luukvbaal:
`pkill` in `winch_handler()` should be exclusive to `ueberzug` case. Resizing for the non-`ueberzug` case was broken, this fixes that. Hard to test all these cases 😅


jarun:
Thank you!

luukvbaal:
Lol didn't know this was syntactically correct but the `then` is in the incorrect place.

PR #1030, opened 2021-05-30
plugin:nmount Update UX and change dependencies

geekodour:
The plugin `nmount` is nice but has some issues like:
- It uses pmount which is not directly available in some package
managers, this change intends to replace it with udevil which does the
same thing.
- The instructions were not very clear on how should the device name be
specified. Improved that UX.

Signed-off-by: Hrishikesh Barman <hrishikeshbman@gmail.com>

### DIFF
--- a/plugins/nmount
+++ b/plugins/nmount
@@ -1,12 +1,10 @@
 #!/usr/bin/env sh
 
-# Description: Toggle mount status of a device using pmount
+# Description: Toggle mount status of a device using udevil
 #              If the device is not mounted, it will be mounted.
 #              If the device is mounted, it will be unmounted and powered down.
 #
-# Dependencies: lsblk
-#
-# Usage: Runs `lsblk` on 'l', exits on 'Return`.
+# Dependencies: lsblk, udevil, udisks2
 #
 # Notes:
 #   - The script uses Linux-specific lsblk to list block devices. Alternatives:
@@ -18,13 +16,17 @@
 # Shell: POSIX compliant
 # Author: Arun Prakash Jana
 
-prompt="device name ['l' lists]: "
+prompt="device name [eg. sdXn]: "
+inst_prompt () {
+  printf "\nEnsure you aren't still in the mounted device.\n"
+  printf "➤ 'l': list\n"
+  printf "➤ 'q' quit\n"
+  printf "%s" "$prompt"
+  read -r dev
+}
 
 lsblk
-
-printf "\nEnsure you aren't still in the mounted device.\n"
-printf "%s" "$prompt"
-read -r dev
+inst_prompt
 
 while [ -n "$dev" ]
 do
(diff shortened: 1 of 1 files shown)


jarun:
CI fails due to shellcheck issues. Please fix.

jarun:
No, we can't replave pmount with udevil. Both of them are not installed by default on many distros.

Please detect which one is available and use it.

plugins/nmount
```diff
@@ -18,13 +16,17 @@
 # Shell: POSIX compliant
 # Author: Arun Prakash Jana
 
-prompt="device name ['l' lists]: "
+prompt="device name [eg. sdXn]: "
+inst_prompt () {
+  printf "\nEnsure you aren't still in the mounted device.\n"
+  printf "➤ 'l': list\n"
```
jarun, line 22: Please print a single comma-separated list of options instead of 2 options.

plugins/nmount
```diff
@@ -33,23 +35,25 @@ do
     elif [ "$dev" = "q" ]; then
         exit
     else
+        TO_MOUNT="/dev/$dev"
         if grep -qs "$dev " /proc/mounts; then
             sync
-            if pumount "$dev"
+            if udevil umount $TO_MOUNT
```
jarun, line 41: Detect which utility is present outside the loop and use it here.

Priority:

- pmount
- udevil

jarun (changes requested):

plugins/nmount
```diff
@@ -18,13 +16,36 @@
...
+inst_prompt () {
+  printf "\nEnsure you aren't still in the mounted device.\n"
+  printf "➤ 'l': list, 'q': quit\n"
+  printf "%s" "$prompt"
+  read -r dev
+}
 
-lsblk
+handle_mount () {
+  if type pmount >/dev/null 2>&1; then
```
jarun, line 28: Instead of running a costly `type` every time, please do the detection once and set a variable to indicate which utility should be used.

jarun (changes requested):

plugins/nmount
```diff
@@ -18,13 +16,36 @@
 # Shell: POSIX compliant
 # Author: Arun Prakash Jana
 
-prompt="device name ['l' lists]: "
+prompt="device name [eg. sdXn]: "
+inst_prompt () {
+  printf "\nEnsure you aren't still in the mounted device.\n"
```
jarun, line 21: And is it too difficult to see we use 4 spaces for alignment in all the plugins including this one?

When you contribute to an established project try not to increase maintenance overhead by making novice mistakes.

plugins/nmount
```diff
@@ -33,23 +54,26 @@ do
     elif [ "$dev" = "q" ]; then
         exit
     else
+        TO_MOUNT="/dev/$dev"
```
jarun, line 57: Did you test with `pmount`? if you did, you would have received the following error when trying to unmount a mounted device:

    Error: device /dev/sdb3 is already mounted to /media/sdb3

jarun:
I think instead of wasting so much time in reviewing 10 lines of change and finding issues in it, I should just make the changes myself.

jarun:
it appears udevil doesn't support LUKS:

```
device name [eg. sdXn] (l: list, q: quit): sdb1
udevil: denied 73: fstype 'crypto_LUKS' is not an allowed type
```

I don't think there's a good reason to add it.

geekodour:
> I think instead of wasting so much time in reviewing 10 lines of change and finding issues in it, I should just make the changes myself.

Yes i was having a hard time around because the the parameters that both of them expect seem to be little different and I was not able to get pmount running on my system that's why i resorted to udevil. udevil not supporting LUKS makes it incompatible anyway. sorry for messing around, i didn't test things properly. will keep things in mind next time around i make a pr. Thanks for reviewing :)

jarun:
Thanks for taking it positively! Great attitude!

Yes, be a little careful when raising PRs on projects with many users. I would have missed all of it if I didn't try it out.

PR #1033, opened 2021-05-30
Preview-tui-ext to preview-tui

luukvbaal:
Drop `preview-tui-ext` in favor of a single `preview-tui`. 

To users of `preview-tui`: the new extended `preview-tui` should behave identically to the old minimal `preview-tui` in the sense that all it does is add additional optional dependencies. Report back if you find behavior different from what you're used to.

Anyone still looking to adopt their own previewers in a more minimal version can use the last version of the minimal `preview-tui` [here](https://github.com/jarun/nnn/blob/b89c32b9fa9dbc40a3f98ff0c358d3e9d7c1946f/plugins/preview-tui). Although any previewers you can think of are welcome to be adopted in the new extended `preview-tui`.


jarun:
Thank you! Please update the Live Previews Wiki page as well.

luukvbaal:
Had already done so :)

Issue #1022, opened 2021-05-19
ToDo list


jarun:
Rolled from #935.

### Cooking

- [x] a patch management model for approved patches
- [x] multi-threaded disk usage calculation using pthreads and FTS
- [x] dynamic view update when calculating disk usage
- [x] <kbd>Bksp</kbd>/<kbd>Del</kbd> at empty filter prompt to refresh dir
- [x] Try to create new context with <kbd>Shift-TAB</kbd>, else reverse cycle
- [x] <kbd>Alt+Esc</kbd> to quit context from filter prompt 
- [x] fix zombies left behind after running plugins (#999)
- [x] named persistent sessions (mcchrish/nnn.vim#43)
- [x] consider nanosecond resolution when sorting by time (#978)
- [x] check external selection in `'c'urrent / 's'el` prompt (#976)
- [x] show number of files selected in local selection buffer, if any
- [x] `nnn` & picker plugin (e.g. `fzopen`) sync (mcchrish/nnn.vim#82)
- [x] make batch rename interactive (#971)
- [x] cached uid/gid for performance improvement
- [x] fixes for `nnn` with `netbsd-curses`, `musl-fts`, `musl` (#998)
- [x] script to statically compile `nnn` with `musl gcc` on Ubuntu
- [x] restore hovered file when plugin is chosen from plugin dir
- [x] support QuickLook on WSL in `preview-tui` (#959)
- [x] toggle `preview-tui` with the same _custom_ plugin key
- [x] smoother preview toggling in `preview-tui` (#966)
- [x] `listen_on` should be set in kitty.conf for `preview-tui` (#970)
- [x] minimal `bat` style in `preview-tui`, honors `$BAT_STYLE`
- [x] plugin `preview-tui-ext` supersedes `preview-tui` (#1033)
- [x] plugin `fzcd` can now fuzzy search multiple directories
- [x] plugin `imgview` supersedes `imgthumb` and `vidthumb`
- [x] plugin `umounttree`: unmount remote mntpoint from within
- [x] plugin `xdgdefault`: set the default app for hovered file type
- [x] plugin `fzplug`: fuzzy find, preview and run other plugins
- [x] plugin `cmusq`: queue/play music in `cmus`
- [x] plugin `mocplay` renamed to `mocq`
- [x] plugin `cleanfilename` renamed to `fixname`
- [x] go to last dir on <kbd>~</kbd> (HOME) or <kbd>`</kbd> (ROOT) key repeat
[...]

PR #1036, opened 2021-05-31
Fix preview-tui jpeg mime

luukvbaal:
Fix regression introduced in https://github.com/jarun/nnn/commit/bb37c9dd46216b261db6816b9159d82ba19d3002; don't generate previews for `jpeg` files as intended. You might want to clear out your `NNN_PREVIEWDIR` of `jpg` previews if it is not in `TMPDIR`. 

Should also prevent the recursive preview generation mentioned in https://github.com/jarun/nnn/issues/1022#issuecomment-851066265.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -197,7 +197,7 @@ print_bin_info() {
 
 handle_mime() {
     case "$2" in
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1024, opened 2021-05-21
Add git status column

crides:
Potentially en route to fix #802. This PR adds a git status column using [`libgit2`](https://github.com/libgit/libgit2). The implementation is derived from [`exa`](https://github.com/ogham/exa). 

Sample screenshot (with `nnn` above and `exa` below):

maim-2021 05 20 23 18 38

The colors are a hack. They are currently taken from the builtin filetype colors, which can change based on configs. 

I don't expect this to be merged (based on the conversation in #802; certainly not now). The current version has questionable variable/function naming, though this could be turned into a compile time configurable feature. If this is not going to be merged, this can just be a reference for those few of us that are looking for such a feature.

PR #1037, opened 2021-05-31
Add user patch framework

luukvbaal:
I went ahead and made an initial implementation of the patch framework @KlzXS @jarun. I added a few patches to start us off which need to be reviewed I guess.

Let me know what you think of the patch implementation though. I added additional make variables that apply the patches in `misc/patches` with `patch`, compile `nnn`, and reverse the applied patches.


jarun:
Thanks a lot!

@KlzXS will wait for you to have a look.

Makefile
```diff
@@ -27,6 +27,12 @@ O_NOSSN := 0  # enable session support
 O_NOUG := 0  # disable user, group name in status bar
 O_NOX11 := 0  # disable X11 integration
 
+# User patches
+O_DDMMYY := 0 # change dateformat to dd-mm-yy
+O_NAMEFIRST := 0 # change detail view to print namfirst, add guid column
+O_NOREVERSEIND := 0 # remove the reverse attribute from the selection indicator
+O_UNICODEARROW := - # change the entry overflow indicator to unicode arrow characters
```
jarun, line 34: @luukvbaal We should not have single line/tiny patches here to make the maintainer's life easy (also ensures the Makefile is not littered). 

For those, we can have inline documentation in the code with a fixed comment prefix and number. For example: 

```
/* NNN_TIP_001: change date format to dd-mm-yy */
/* printw("%s-%02d-%02d %02d:%02d",
            xitoa(t.tm_year + 1900), t.tm_mon + 1, t.tm_mday, t.tm_hour, t.tm_min); */
```

This mechanism will be documented in the Wiki as _Inline Easter Eggs_ above the `make options` section. ;)

jarun (changes requested):

Makefile
luukvbaal: Being able to simply supply a make option for it is much easier from the users' perspective.
It also keeps the actual source `nnn.c` free from those comments but if you think it's the better option, sure.

luukvbaal: I would argue that isolated patches make the maintainers life easier than making sure such comments stay up to date.

luukvbaal: Though if you don't want such tiny patches in at all I can understand that too.

luukvbaal:
Did we also want to add all the patch make options to the CircleCI `config.yml`? I think we considered this to make sure the stay compatible though I see it's also not done for the preprocessor macro make options, do we want to add those as well?

jarun:
> Did we also want to add the patch make options to the CircleCI config.yml?

I suggested that but I am not sure how frequently it will fail and how annoying it would get. This is also another reason behind keeping patches less.

Makefile
jarun: >  If it indeed only concerns one line changes/inline comments I suppose it's hard to miss so maybe it's fine.

Yes, only for the tiny ones. I saw 3 one-liners. We can take a call per patch going forward.

jarun: > Though if you don't want such tiny patches in at all I can understand that too. Just added a few for examples' sake.

Yes I got that. But I see a bad Makefile after 10 such patches.
This is inline with current make options. You won't find a single one liner.

jarun: I am adding the 3 one-liners in source now. You can remove them from here once i confirm.

luukvbaal: As you want, I still think isolated patch files with make options are easier to maintain/test than commented one-liners though.
I wouldn't consider a few additional if/else clauses in the Makefile "bad" per se.

jarun: > a few additional if/else clauses in the Makefile

1 for dd mm yy, 1 for single character change, one for 2 indicators...
I don't think you see that fact that those will only increase.

I am doing the in-code thing now. Will ping back.

luukvbaal: Sure they would increase, I thought that was the idea behind providing a framework for user submitted patches.
I'm fine with leaving them out if we don't want such small patches. 

I'm just not sure I see the point in providing them as in-code comments instead. Doubt anyone who could benefit from it (read person who could not have come up with the change in the first place) will go through the trouble of editing the source through anything other than make options.

jarun: > Doubt anyone who could benefit from it (read person who could not have come up with the change in the first place) will go through the trouble of editing the source through anything other than make options.

Did I ever claim I want to spoon-feed `nnn` to users? Users who can `make` but can't do that bare minimum would have to remain happy with what's exposed by `make`.

I am not Bill Gates touting Visual Basic as a programming language to generate generations of lame end-users (read revenue). My motto is - you want power? take what I can offer and empower yourself.

jarun: But there's a different problem!!!

I realized users have to rebase the code after each release and manually apply these patches. I don't want that repeat harassment. So your change stays. ;)

Makefile
```diff
@@ -27,6 +27,12 @@ O_NOSSN := 0  # enable session support
 O_NOUG := 0  # disable user, group name in status bar
 O_NOX11 := 0  # disable X11 integration
 
+# User patches
+O_DDMMYY := 0 # change dateformat to dd-mm-yy
```
jarun, line 31: dateformat -> date format

Makefile
```diff
@@ -27,6 +27,12 @@ O_NOSSN := 0  # enable session support
...
+O_NAMEFIRST := 0 # change detail view to print namfirst, add guid column
```
jarun, line 32: file name first, add uid and guid columns

Makefile
jarun, line 34: I think for the unicode arrow, for the sake of uniformity, we can also make the current file indicator (reversed block now) a similar arrow. 3 similar arrows would look good.

jarun (changes requested):

Makefile
luukvbaal: Again, I was just saying that if we don't provide these small patches as make options in view of maintainability (which I can agree with), we might as well leave them out all together as the more adventurous user can figure it out for themselves...

Though I see you are now reviewing the smaller patches, do you want them in or not?

jarun: Yes, I want them in as you have planned.

luukvbaal: Sure, good idea. 

This is where the trouble begins though. That change would make the namefirst and unicode patches incompatible (the namefirst patch gets rid of the indicator altogether in favor of reversing the entire line), not sure how/if we should account for that.

luukvbaal: Not removing the `markhovered()` function in the namefirst patch would keep the patches compatible but would result in an unused function compiler warning as a compromise...

jarun: Please drop the unicode arrow patch. They don't even show correct in my terminal.

luukvbaal: Yeah that's fine. We could include them in the `O_NERD/O_ICONS` options with their appropriate ascii strings instead of literal unicode strings if we want.

Makefile
```diff
@@ -141,10 +146,32 @@ DESKTOPFILE = misc/desktop/nnn.desktop
 LOGOSVG = misc/logo/logo.svg
 LOGO64X64 = misc/logo/logo-64x64.png
 
+DDMMYY = misc/patches/ddmmyy.diff
```
jarun, line 149: This should be ddmmyyyy.diff to be correct.

misc/patches/README.md
```diff
@@ -0,0 +1,15 @@
+<h1 align="center">User Patch Framework</h1>
+
+This directory contains user submitted patches containing more subjective features that were not deemed fit to merge into master. Nevertheless the patches are kept up to date with master at least each release and can be applied through make options.
```
jarun, line 3: containing more subjective features -> for more subjective features

merge into master -> merge mainline

Nevertheless the patches... -> The patches are maintained and adapted on each release. Each patch can be applied through its make option.

misc/patches/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## List of patches
+| Patch (a-z) | Description | Make variable |
+| --- | --- | --- |
+| ddmmyy | Changes datetime format to dd-mm-yy. | O_DDMMYY |
+| namefirst | Changes the detail view to keep names printed first. Also prints user/group columns when a directory contains different users/groups. | O_NAMEFIRST |
```
jarun, line 9: Prints filenames first in the detail view. Prints user/group columns when a directory contains different users/groups.

How do you detect files in a dir have different users/groups?

misc/patches/README.md
```diff
@@ -0,0 +1,15 @@
...
+| noreverseind | Removes the reverse attribute from the selection indicator. | O_NOREVERSEIND |
+
+To compile in the patches, supply the make variables listed above in the same manner as the [make options](https://github.com/jarun/nnn/wiki/Developer-guides#make-options) e.g.:
```
jarun, line 12: To compile... -> To apply the patches, use the corresponding make variables e.g.:

Keep this delinked from the Wiki section. We will have complete patch doc here (like plugins).

Makefile
luukvbaal: Besides looking gross, the patch actually prints dd-mm-yy, not dd-mm-yyyy.

misc/patches/README.md
luukvbaal: This [line](https://github.com/jarun/nnn/pull/1037/files#diff-bf328d0269ceda2a2149b4e87e9b67faf54ce9ab51895e9551384ecf22a5c7d5R20) in `getpwname()` sets a boolean to 1 if that is the case. Works because `printguid` and `maxuidln` are reset to 0 on each `redraw()`. If we just changed directory `printguid` will not be set because `maxuidln` is still 0. If a different guid is encountered for entries other than the first one `printguid` will be set to true, in which case the guid columns will be printed in `printent()`.

Makefile
jarun: I hope you are accounting for the 2 characters removed.

luukvbaal: If by accounting for you mean the column width in `adjust_maxcols()` etc, yeah I am.

luukvbaal: Sorry, only doing that in my namefirst patch. This patch by itself would need to be adjusted. 

How do you feel about moving to yy-mm-dd in mainline? If you don't want to do that I'll change the patch to dd-mm-yyyy I guess, although I prefer dd-mm-yy in view of screen estate.

jarun: > How do you feel about moving to yy-mm-dd in mainline?

No. I would follow the ISO thing as it is now. I told you earlier we received criticism for not following that.

luukvbaal: Fine, not sure what I'd like to do with this patch then. Do you think it's worth accounting for the time column width in both this patch(hard coded) and the namefirst patch(on runtime by i.e running xstrlen(print_time()) once and storing the result), or should I keep this a personal change.

jarun: Drop it.

Makefile
```diff
@@ -27,6 +27,11 @@ O_NOSSN := 0  # enable session support
 O_NOUG := 0  # disable user, group name in status bar
 O_NOX11 := 0  # disable X11 integration
 
+# User patches
+O_DDMMYY := 0 # change date format to dd-mm-yy
+O_NAMEFIRST := 0 # change detail view to file print name first, add uid and guid columns
+O_NOREVERSEIND := 0 # remove the reverse attribute from the selection indicator
```
KlzXS, line 33: Patches should be available to as wide a userbase as possible. Since this came in so late in the lifetime of `nnn` it means there will need to be at least _some_ amount of backporting. This also isn’t the last version of `nnn`, a patch made for this version needs to be available even when we are up to v4.5; users don’t always jump to the newest version, be it that they don't want to or that the new version isn't available for their distro. Why shouldn't they have easy access to a patch for their favorite version?

Of course we can’t force anyone to backport their patches, but we should encourage it.

That’s why I think the right directory tree for patches should look something like this: `misc/<patch_name>/<patch_name>-<ver/commit>.diff`. It would be nice if each patch included a short README that describes what the patch is for and maybe why it isn’t part of `nnn`’s core, though I guess that’s not strictly necessary. That can also fit in the main README in `misc/patches`.

With that in mind a binary switch in the Makefile just won’t cut it, it needs to be a bit more sophisticated than that. Also it would be nice if it were just one option accepting a comma separated list of values like `<patch_name>-<ver/commit>`. Maybe even having some mechanism to remember previously applied patches to make it easier to add a new one on top or remove a previously applied one. I don’t know how doable that is, I’m not too versed in the ways of Makefiles.

KlzXS:
@luukvbaal great to see that you started work on this! Check the comment I left in the Makefile.

@jarun I’m getting the vibe that you want to support these patches to some extent. Don’t get me wrong, I don’t want halfassed patches either, but I don’t think we should “support” them. Just make sure that we don’t deliberately break them, that’s all, or mark it as an “official patch” or something in that case. There really isn’t need for much more than that.

Makefile
luukvbaal: Seems like way too much work to me. Anyone who is not willing to update to the latest version should be expected to backport any patches they want to use if you ask me.

Don't see why it needs to be any more sophisticated than the other make options already present in mainline.

KlzXS: > Seems like way too much work to me. Anyone who is not willing to update to the latest version should be expected to backport any patches they want to use themselves if you ask me.

As I said, if you don't want to it's fine. Still we should use directory layout I suggested so that current patches stay available to people in the future.

> Don't see why it needs to be any more sophisticated than the other make options already present in mainline.

Simply to able to pick the correct file for the version of `nnn` you are patching.

luukvbaal: I could see starting the directory structure and maintaining the version history when we get to the next release but backporting to all previous releases seems like overkill to me.

KlzXS: > backporting to all previous releases seems like overkill to me.

That's fair. No forcing anyone, just mention that it would be nice to people who use older versions. But do start the directory structure.

luukvbaal: > No forcing anyone, just mention that it would be nice to people who use older versions. 

Not worth it considering updating to the latest version shouldn't be that much work in my opinion... The patches already assume the user is building `nnn` their selves, not depending on their distro's version. Why would they build any version other than the latest?

Would something like this be fine then?

    NAMEFIRST = misc/patches/namefirst/namefirst-$(VERSION).diff

This would work on the condition that we only accept patches on point releases, not commits.

I have a commit ready for that, curious to know how @jarun feels about it though.

KlzXS: > Why would they build any version other than the latest?

I for one prefer some older versions that had a bit different behavior, some users might also. I really haven't thought about it from that angle, I assumed that there will be people how would use older versions. There isn't always a reason to update to the newest version.

> Would something like this be fine then?

Yes. Patching based on commits is not good in my opinion. Things are prone to change fast and a patch might just pick the wrong thing to rely on.

jarun: @KlzXS @luukvbaal let's keep it simple. Please document that the patches will work and are supported nnn v4.1 onwards.

jarun: Let's not forget that the Makefile changes are also going to be released in v4.1.

jarun:
>  Just make sure that we don’t deliberately break them,

How? My understanding is you are going to check these post every release.

Makefile
KlzXS: Ok. There is no need to change anything then @luukvbaal.

luukvbaal: > @KlzXS @luukvbaal let's keep it simple. Please document that the patches will work and are supported nnn v4.1 onwards.

Alright. I take it that means you're in favor of the directory structure @KlzXS proposed with a folder per plugin containing the patch version history?

> Let's not forget that the Makefile changes are also going to be released in v4.1.

What do you mean?

KlzXS:
If you know there is a patch that touches a piece of code try to minimize the impact you'll have on it if possible. No need to make the author do more work.

Yes, patches will be tested before a release.

Makefile
luukvbaal: > > @KlzXS @luukvbaal let's keep it simple. Please document that the patches will work and are supported nnn v4.1 onwards.
> 
> Alright. I take it that means you're in favor of the directory structure @KlzXS proposed with a folder per patch containing the patch version history?

I still don't quite see the point in it myself. I know this is the way suckless.org maintains their patches but we are hosting the patches on github. Anyone who would like to apply the patch on an earlier release(not encouraged I would say) can go through the version history on github right?

luukvbaal: > @KlzXS @luukvbaal let's keep it simple. Please document that the patches will work and are supported nnn v4.1 onwards.

Never mind my previous comments, I see I read this wrong and I missed @KlzXS their comment. We will stay with the current implementation then.

KlzXS: Just ditch the thing. Forget about it.
I wanted to have something that seems nobody wants or needs.

You're the one submitting a patch after all.

luukvbaal: I was not alone in wanting to introduce a patch framework for `nnn` though. Discussion on the implementation doesn't hurt, as long as we come to a consensus :)

KlzXS: Now that I think about it you are completely right. If you want to compile for version x you need to git checkout to there. The patch will already be for the correct version. 🤦 

I guess I also forgot that I'm the weird one using old versions of nnn.

Anyway, good night.

luukvbaal:
Would we consider something like this an appropriate patch i.e. nordicons.diff providing the icons in nord theme colors?
```diff
diff --git a/src/icons.h b/src/icons.h
index aaefab63..a3c45fe3 100644
--- a/src/icons.h
+++ b/src/icons.h
@@ -25,23 +25,23 @@ struct icon_pair {
 #define ICON_PADDING_LEFT  ""
 #define ICON_PADDING_RIGHT " "
 
-#define COLOR_VIDEO        93  /* Purple */
-#define COLOR_AUDIO        220 /* Gold1 */
-#define COLOR_IMAGE        82  /* Chartreuse2 */
-#define COLOR_DOCS         202 /* OrangeRed1 */
-#define COLOR_ARCHIVE      209 /* Salmon1 */
-#define COLOR_C            81  /* SteelBlue1 */
-#define COLOR_JAVA         32  /* DeepSkyBlue3 */
-#define COLOR_JAVASCRIPT   47  /* SpringGreen2 */
-#define COLOR_REACT        39  /* DeepSkyBlue1 */
-#define COLOR_CSS          199 /* DeepPink1 */
-#define COLOR_PYTHON       227 /* LightGoldenrod1 */
-#define COLOR_LUA          19  /* LightGoldenrod1 */
-#define COLOR_DOCUMENT     15  /* WHITE */
-#define COLOR_FSHARP       31  /* DeepSkyBlue3 */
-#define COLOR_RUBY         160 /* Red3 */
-#define COLOR_SCALA        196 /* Red1 */
-#define COLOR_VIM          28  /* Green4 */
+#define COLOR_VIDEO        110
+#define COLOR_AUDIO        150
+#define COLOR_IMAGE        150
+#define COLOR_DOCS         173
+#define COLOR_ARCHIVE      179
+#define COLOR_C            150
+#define COLOR_JAVA         139
+#define COLOR_JAVASCRIPT   109
+#define COLOR_REACT        111
+#define COLOR_CSS          110
+#define COLOR_PYTHON       68
+#define COLOR_LUA          167
+#define COLOR_DOCUMENT     111
+#define COLOR_FSHARP       179
+#define COLOR_RUBY         150
+#define COLOR_SCALA        139
+#define COLOR_VIM          109
 
 /*
  * Using symbols defined in icons-in-terminal.h, or even using icons-in-terminal is not necessary.
```

jarun:
> If you know there is a patch that touches a piece of code try to minimize the impact you'll have on it if possible. No need to make the author do more work.

That's why we should have a few in number.

jarun:
@luukvbaal 

> Would we consider something like this an appropriate patch i.e. nordicons.diff providing the icons in nord theme colors?

I don't want a single other look and feel improvement patch.

luukvbaal:
I would have thought most of the patches that make sense for this framework are going to be subjective "look and feel improvements".
Other quality patches that are not "look and feel" in nature would probably make it to mainline.

But alright, lets wait and see what kind of patches we will approve then...

jarun:
> going to be subjective "look and feel improvements"

When I mentioned look and feel I meant look and feel related changes that don't provide much functional advantage/workflow improvement.

jarun:
@luukvbaal we missed https://github.com/jarun/nnn/pull/1024.

jarun:
Please add that one as well.

KlzXS:
That one still has issues doesn't it?

jarun:
> That one still has issues doesn't it?

Can you guys try to clean it up?

luukvbaal:
I am cleaning it up. 

What color is `C_MIS` supposed to be? It looks like this for me for some reason: st:
image
kitty:
image
Original PR:
image

Any ideas @jarun?

jarun:
Is there a bg or something in the color code? Behaves strangely... The original colors were looking good. Use something close.

luukvbaal:
Ah you're right, I guess all colors look different from the original PR. Not sure what's going on there.

luukvbaal:
It looks correct when I unset `NNN_FCOLORS`:
image

I think this is not related to the patch as it just calls `addch('-' | COLOR_PAIR(C_MIS))` I assume it would happen in other places too where `C_MIS` is used. Using the nord color theme listed on the [wiki](https://github.com/jarun/nnn/wiki/Themes#nord).

jarun:
> It looks correct when I unset NNN_FCOLORS
> I think this is not related to the patch as it just calls addch('-' | COLOR_PAIR(C_MIS))

Are you in 256-color mode or 8-color mode? In 8-color mode you should use `A_DIM`. `NNN_FCOLORS` is not used in 8-color mode.

jarun:
I would suggest apply the colors only if 256 color mode. Else, just print dimmed.
I think you need more than 8 colors for the different types of changes.

luukvbaal:
See [this](https://gist.github.com/luukvbaal/6144be2413c257b2ba81be3fa82c235f) for the cleanup diff @jarun @crides . I won't have time to look into the colors further any time soon.

jarun:
@crides please test.

@luukvbaal can you please use tabs instead of spaces?

luukvbaal:
I do use tabs, must have gotten lost in the diff generation. I'll check. 

Also feel like we should try to add compatibility patches between patches that contain merge conflicts if possible. Working on namefirst-gitstatus compatability.

luukvbaal:
Added a compatibility patch for namefirst/gitstatus. You can now do `make O_NAMEFIRST=1 O_GITSTATUS=1` for 
image

luukvbaal:
@jarun if you are still worried about "littering" the Makefile, do you have an alternative solution to apply the patches?

jarun:
Awesome!

Makefile
```diff
@@ -27,6 +27,14 @@ O_NOSSN := 0  # enable session support
 O_NOUG := 0  # disable user, group name in status bar
 O_NOX11 := 0  # disable X11 integration
 
+# User patches
+O_GITSTATUS := 0 # add git status to detail view
+O_NAMEFIRST := 0 # change detail view to file print name first, add uid and guid columns
```
jarun, line 32: print file name first in detail view, add uid and gid columns

Makefile
```diff
@@ -141,10 +149,29 @@ DESKTOPFILE = misc/desktop/nnn.desktop
 LOGOSVG = misc/logo/logo.svg
 LOGO64X64 = misc/logo/logo-64x64.png
 
+GITSTATUS = misc/patches/gitstatus
+NAMEFIRST = misc/patches/namefirst
+
 all: $(BIN)
 
 $(BIN): $(SRC) $(HEADERS)
+ifeq ($(strip $(O_NAMEFIRST)),1)
```
jarun, line 158: Can't we do this outside the targets? This really looks bad.

luukvbaal: If you know how. I tried but failed.

jarun: How about a new target?

Makefile
jarun: print file name first, add uid and gid columns in detail view

Makefile
luukvbaal: Not sure, perhaps move it to an external shell script `misc/patches/patch.sh`?

jarun: Yes, that's also fine.

luukvbaal: How about 
```Make
prebuild:
ifeq ($(strip $(O_NAMEFIRST)),1)
ifeq ($(strip $(O_GITSTATUS)),1)
	patch --forward --strip=1 --input=$(GITSTATUS)/namefirst.diff
endif
	patch --forward --strip=1 --input=$(NAMEFIRST)/mainline.diff
else ifeq ($(strip $(O_GITSTATUS)),1)
	patch --forward --strip=1 --input=$(GITSTATUS)/mainline.diff
endif

postbuild:
ifeq ($(strip $(O_NAMEFIRST)),1)
ifeq ($(strip $(O_GITSTATUS)),1)
	patch --reverse --strip=1 --input=$(GITSTATUS)/namefirst.diff
endif
	patch --reverse --strip=1 --input=$(NAMEFIRST)/mainline.diff
else ifeq ($(strip $(O_GITSTATUS)),1)
	patch --reverse --strip=1 --input=$(GITSTATUS)/mainline.diff
endif

$(BIN): $(SRC) $(HEADERS)
	@$(MAKE) --silent prebuild
	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o $@ $< $(LDLIBS)
	@$(MAKE) --silent postbuild
```

luukvbaal: Possibly moving pre/postbuild towards the bottom of the Makefile if you want.

luukvbaal: And/or name the targets pre/postpatch

jarun: > Possibly moving pre/postbuild towards the bottom of the Makefile if you want.

Yes, better.

misc/patches/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## List of patches
+| Patch (a-z) | Description | Make variable |
+| --- | --- | --- |
+| gitstatus | Add gitstatus column to the gitstatus view. Requires [libgit2](https://github.com/libgit2/libgit2). | O_GISTATUS |
+| namefirst | Prints filenames first in the detail view. Prints user/group columns when a directory contains different users/groups. | O_NAMEFIRST |
+
+To apply the patches, use the corresponding make variables, e.g.:
+
+	make O_NAMEFIRST=1
```
jarun, line 13: I think you just need 4 spaces here for markdown.

luukvbaal: Right.

Do you know if we can `clean` before building depending on whether any make option arguments are passed?

Currently if you build with eg `make` and after try to run `make O_NERD` you will first have to run make clean.

jarun: I think we do that in the packaging service. We have a single target. Otherwise .phony would have worked.
Anyway, if it is an existing issue, just leave it be.

jarun:
@KlzXS please do the final review and we'll pull this in.

KlzXS:
> do you have an alternative solution to apply the patches?

How about what I suggested earlier, having `O_PATCHES` accept a list of patch names to apply? To make in cleaner we might have a script to do that.

luukvbaal:
I think the current patch targets work fine to cleanup the Makefile actually. So if @jarun agrees I don't think we need to change anything anymore regarding that.

Having separate make options per patch instead of a single `O_PATCHES` is more in line with the already present make options in mainline I would say.

jarun:
Yes, I agree with @KlzXS. Let's just follow what we do now. If we really get to > 5 patches of this nature we will rethink this. But if we see we have 2 significant such patches in 4 years. ;)

KlzXS:
Both have their respectuve benefits. For now it's ok.

I can't review all this tonight, but I'll do it tomorrow. Want me to merge after that?

jarun:
> I can't review all this tonight, but I'll do it tomorrow. Want me to merge after that?

Thanks! That would be great.

jarun:
@KlzXS make sure you squash and merge.

luukvbaal:
Passing these `attrs` to the `git_render` function fixes the colors.

    		attrs = g_state.oldcolor ? (resetdircolor(ent->flags), A_DIM)
					 : (fcolors[C_MIS] ? COLOR_PAIR(C_MIS) : 0);

jarun:
:+1:

jarun:
Why is it required to call `resetdircolor(ent->flags)`?

luukvbaal:
It's called in `printent()`. Just passing the value to the git print function and substituting it where @crides had used `C_MIS`.

I thought that made sense, perhaps we can just hardcode the value in the git print function though, idk.

luukvbaal:
Not quite sure to be honest whats necessary in `print_gitstatus()` to be consistent with the color behavior in `nnn` regarding `g_state.oldcolor`. Latest commit should at least bug-free. See if you want to check it in the review @KlzXS, I'm done with this😅

jarun:
I am merging this.

@KlzXS please make sure you review this. I am planning to make a release today so need to get things sorted.

jarun:
@luukvbaal @KlzXS I have moved patches to the top level at commit aeb6a26f2b8688dc69e65346629becc9fb86402d so they are visible to interested users.

misc/patches/gitstatus/mainline.diff
```diff
@@ -0,0 +1,233 @@
...
++	}
++}
++
++static bool starts_with(const char *const s, const char *const start) {
++	const size_t start_len = strlen(start);
++
++	return (start_len <= strlen(s)) && (0 == strncmp(s, start, start_len));
++}
++
++static size_t mkpath(const char *dir, const char *name, char *out);
```
KlzXS, line 110: Move declaration up with other forward declarations.

misc/patches/gitstatus/mainline.diff
```diff
@@ -0,0 +1,233 @@
...
++		}
++
++		git_status_list_free(status_list);
++		git_repository_free(repo);
++	}
++	return statuses;
++}
++
++static void statuses_free(simple_git_statuses_t statuses) {
++	for (size_t i = 0; i < statuses.len; i ++)
```
KlzXS, line 150: `i ++`. We don't put spaces between unary operators. I think we use `++i`.

KlzXS:
Whoever is addressing these minor issues in gitstatus: remember to change both files.

luukvbaal:
Yeah, my fork contains two branches `gitstatus` and `gitstatus-namefirst` that can be used to `git rebase` upon future releases to fix any merge conflicts if necessary and to regenerate the diff. If you want you can make a PR there for the review you're doing.

KlzXS:
Fine by me. You also mentioned some bugs and odd behavior. What's that all about?

luukvbaal:
See https://github.com/jarun/nnn/pull/1024#issuecomment-852816549.

Also the git statuses are being stored in a separate struct containing the a path:
```
typedef struct {
	char *path;
	git_status_t status;
} simple_git_status_t;
```

This is being filled before hand and looped over in the `dentfill()` `do/while` loop. Seems to me we should try to generate the git status inside that loop and store it directly in the `pEntry` structure, not separately. Think we can avoid some redundant work that way.

luukvbaal:
The bug in https://github.com/jarun/nnn/pull/1024#issuecomment-852816549 seems to be due to the `starts_with()` function which I was commenting on earlier when Github comments didn't seem to working. Forgot to submit that comment.

Not sure why the original patch is checking for `starts_with()` instead of the full path but like I said I'll try to include the git_status in the `pEntry` structure directly to try and avoid the path comparison all together.

luukvbaal:
I think we can call [`git_status_file()`](https://libgit2.org/libgit2/#HEAD/group/status/git_status_file) in the `dentfill()` while loop instead.

luukvbaal:
Yep
```
		git_buf ret = { .ptr = 0, .asize = 0, .size = 0 };
  	git_repository *repo;
		git_repository_discover(&ret, path, false, NULL);
		git_repository_open(&repo, ret.ptr);
		git_buf_dispose(&ret);

    if (repo) {
			git_status_file(&dentp->status, repo, dentp->name);

      dentp->status_indexed = git_get_indexed_status(dentp->status);
      dentp->status_staged = git_get_staged_status(dentp->status);
    } else {
			dentp->status_indexed = GIT_COLUMN_STATUS_NONE;
			dentp->status_staged = GIT_COLUMN_STATUS_NONE;
    }
```

Fixes works while avoiding all the extra functions/structs while also fixing the bug.

jarun:
Please check the indentation.

crides:
> Not sure why the original patch is checking for starts_with() instead of the full path but like I said I'll try to include the git_status in the pEntry structure directly to try and avoid the path comparison all together. I think that should be possible though maybe I'm missing something and @crides is storing the git_statuses separately for some reason I'm not seeing.

@luukvbaal The git statuses for display are stored in the `pEntry` struct. The `git_statuses` are the ones returned from `libgit2` calls and simplified. 

The `starts_with` calls are needed for correct ignore detection. However they should've been path-wise not char-wise, though I'm not sure how I'd fix that (are there any path comparing utilities in `nnn` or `libc`?).

> I think we can call [`git_status_file()`](https://libgit2.org/libgit2/#HEAD/group/status/git_status_file) in the `dentfill()` while loop instead.

I think I used that originally, but at the time it seemed to me something is missing by not using the `gis_status_list_new` call. I can't recall what that is though. The docs does mention rename detection, which can be a part of it, but I can't be sure of what the output differences would be like between the 2.

luukvbaal:
Yeah idk, [this](https://gist.github.com/luukvbaal/2c20e8bc6d6d5f0ad76653969cf79864) patch using `git_status_file()` always shows the correct information however it seems to be much slower than your patch.

luukvbaal:
@KlzXS https://github.com/luukvbaal/nnn/tree/gitstatus now contains the style fixes you suggested and more.
@crides I think the code is still functionally identical to your initial patch, the output still looks very different from `exa` though. Still need to figure out how to parse the status entry list to get the correct output.

crides:
I'll try to see if `strtok` or something similar will fix the issue later today or tomorrow

luukvbaal:
Just updated https://github.com/luukvbaal/nnn/tree/gitstatus, please see if it gives the correct output.
Seems to me it does, but if you know of any edge cases that aren't covered by that implementation do let me know and make the PR if you can.

jarun:
> I'll try to see if strtok or something similar will fix the issue later today or tomorrow

@luukvbaal please watch out not to add new functions in the binary if not absolutely necessary. There is not strtok() call in `nnn` today.

PR #1038, opened 2021-05-31
Unicode arrow indicators for O_ICONS/NERD

luukvbaal:
Replace arrow indicators with unicode versions supplied by icons-in-terminal or nerd-font when compiled with `O_ICONS/NERD`.


jarun:
Thank you!

PR #1046, opened 2021-06-02
Further gitstatus cleanup/fixes

luukvbaal:
Pending review of @crides (check https://github.com/luukvbaal/nnn/commit/aca2ad3c56ca9a39a7fe3008b382b2566d6f9536 for the actual diff).

### DIFF
--- a/patches/gitstatus/mainline.diff
+++ b/patches/gitstatus/mainline.diff
@@ -3,26 +3,25 @@
 # Dependencies: libgit2
 #
 # Authors: @crides, Luuk van Baal
+
 diff --git a/src/nnn.c b/src/nnn.c
-index 562622f..96d946a 100644
+index d83d7f6..b23218e 100644
 --- a/src/nnn.c
 +++ b/src/nnn.c
-@@ -127,6 +127,8 @@
- #include "qsort.h"
- #endif
-
+@@ -105,6 +105,7 @@
+ #include <wchar.h>
+ #include <pwd.h>
+ #include <grp.h>
 +#include <git2.h>
-+
- /* Macro definitions */
- #define VERSION "4.0"
- #define GENERAL_INFO "BSD 2-Clause\nhttps://github.com/jarun/nnn"
-@@ -245,6 +247,18 @@ typedef unsigned char uchar_t;
+
+ #if !defined(alloca) && defined(__GNUC__)
+ /*
+@@ -244,6 +245,16 @@ typedef unsigned int uint_t;
+ typedef unsigned char uchar_t;
  typedef unsigned short ushort_t;
  typedef unsigned long long ulong_t;
-
 +typedef enum {
-+	GIT_COLUMN_STATUS_NONE = 0,
-+	GIT_COLUMN_STATUS_UNMOD,
++	GIT_COLUMN_STATUS_UNMOD = 0,
 +	GIT_COLUMN_STATUS_NEW,
 +	GIT_COLUMN_STATUS_MODIFIED,
 +	GIT_COLUMN_STATUS_DELETED,
@@ -31,11 +30,10 @@ index 562622f..96d946a 100644
 +	GIT_COLUMN_STATUS_IGNORED,
 +	GIT_COLUMN_STATUS_CONFLICTED,
 +} git_column_status_t;
-+
+
  /* STRUCTURES */
 
- /* Directory entry */
-@@ -258,6 +272,8 @@ typedef struct entry {
+@@ -258,6 +269,8 @@ typedef struct entry {
  		ulong_t blocks : 40; /* 5 bytes (enough for 512 TiB in 512B blocks allocated) */
  		ulong_t nlen : 16; /* 2 bytes (length of file name) */
  		ulong_t flags : 8; /* 1 byte (flags specific to the file) */
@@ -44,7 +42,7 @@ index 562622f..96d946a 100644
  	};
  #ifndef NOUG
  	uid_t uid; /* 4 bytes */
-@@ -362,7 +378,18 @@ typedef struct {
+@@ -362,7 +375,18 @@ typedef struct {
  } session_header_t;
  #endif
 
@@ -63,9 +61,9 @@ index 562622f..96d946a 100644
 
  /* Configuration, contexts */
  static settings cfg = {
-@@ -809,6 +836,92 @@ static void notify_fifo(bool force);
-
- /* Functions */
+@@ -3464,6 +3488,79 @@ static char *get_kv_val(kv *kvarr, char *buf, int key, uchar_t max, uchar_t id)
+ 	return NULL;
+ }
 
 +static git_column_status_t git_get_indexed_status(const uint32_t status) {
 +	if (status & GIT_STATUS_INDEX_NEW)         return GIT_COLUMN_STATUS_NEW;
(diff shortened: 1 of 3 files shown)


patches/gitstatus/mainline.diff
```diff
@@ -1,28 +1,21 @@
-# Description: Add git status column to detail mode.
-#
-# Dependencies: libgit2
-#
-# Authors: @crides, Luuk van Baal
```
KlzXS, line 5: I think you mistakenly removed this.

patches/gitstatus/mainline.diff
```diff
@@ -147,74 +133,67 @@ index 562622f..96d946a 100644
...
-+			mkpath(path, dentp->name, joined);
-+			char *real = realpath(joined, NULL);
-+			char *canon_path = real ? real : joined;
++			uint merged_status = 0;
 +
 +			for (size_t i = 0; i < git_statuses.len; i ++)
-+				if (((dentp->mode & S_IFMT) == S_IFDIR) ? starts_with(git_statuses.statuses[i].path, canon_path) :
-+						!xstrcmp(git_statuses.statuses[i].path, canon_path) || starts_with(canon_path, git_statuses.statuses[i].path))
++				if ((dentp->flags & DIR_OR_LINK_TO_DIR) ? starts_with(git_statuses.statuses[i].path, abspath(dentp->name, path))
++						: !xstrcmp(git_statuses.statuses[i].path, abspath(dentp->name, path)))
```
KlzXS, line 186: Do we really need a ternary operator as a condition for an if?
Also what is the point in using `starts_with()` for a dir and `xstrcmp()` for a file? Can these not be merged into a single string comparison?

Red alert! We are leaking memory. `abspath()` allocates a string and returns that pointer! It's not an in-place operation.

KlzXS (changes requested):
See comments. The memory leak needs to be fixed.

Sorry for mistakenly posting this on your fork, I forgot where I was. It did seem odd to me that it didn't offer me to start a review.

patches/gitstatus/mainline.diff
luukvbaal: > Do we really need a ternary operator as a condition for an if?

Need no, thought it was better than duplicating the if statement/assignment.

> Also what is the point in using `starts_with()` for a dir and `xstrcmp()` for a file? Can these not be merged into a single string comparison?

I've tried to but couldn't get the right behavior. Debugging the git status entries has proven difficult to me.

KlzXS: Well I don't have anything smart to say about the ternary operator. Let's see if I'll have more luck with merging the two cases.

The call to `abspath()` needs to be saved in a new `char*`, maybe make one for the other argument so that the whole if can fit on a single line at least.

crides: > Also what is the point in using starts_with() for a dir and xstrcmp() for a file? Can these not be merged into a single string comparison?

If it's a file, then either one of its parent directory is ignored, so it's ignored too; or its status is directly reported. 

If it's a directory, then (same applies) either one of its parent directory is ignored, so it's ignored too; or its status is directly reported; or, if its descendants has non default status, then those need to be squashed. 

These are needed as the `git_status_list_new()` call basically returns what `git status --short` shows.

luukvbaal: Yeah, trying one more time. Thinking we should also use the preexisting `is_prefix()` instead of `starts_with()`.

jarun:
@KlzXS please squash and merge when done.

KlzXS:
Will do,

patches/gitstatus/mainline.diff
luukvbaal: This seems to work as expected.
```
		if (git_statuses.len) {
			uint merged_status = 0;
			char *dentpath = abspath(dentp->name, path);

			for (size_t i = 0; i < git_statuses.len; i ++) {
				if (dentp->flags & DIR_OR_LINK_TO_DIR) {
					size_t dentlen = xstrlen(dentpath);

					if ((dentlen <= xstrlen(git_statuses.statuses[i].path)) && is_prefix(git_statuses.statuses[i].path, dentpath, dentlen))
						merged_status |= git_statuses.statuses[i].status;
				} else if (!xstrcmp(git_statuses.statuses[i].path, dentpath)) {
						merged_status |= git_statuses.statuses[i].status;
				}
			}

			dentp->status_indexed = git_get_indexed_status(merged_status);
			dentp->status_staged = git_get_staged_status(merged_status);
			free(dentpath);
		}
```

crides: Oh I didn't know that function is there. 

Still, I don't think that resolves the bug, as it still doesn't compare path-wise. I _think_ we need to check 1) if `prefix` is a prefix of `path`, and 2) if the character right after the prefix is `'/'` or 0 (so that the 2 are the same). I'll try that when I get home, or you can try it now

luukvbaal: What bug are we talking about now? I don't see the `dir/nnn.1` vs `dir/nnn` false positive with the latest commit.

luukvbaal:
I think we are good, seeing the same output as `exa --git` now. Can be merged as far as I'm concerned after you have done a final review @KlzXS.

KlzXS:
I'll test it myself a bit and look through the code one more time tomorrow. After that I'll merge.
@crides do you have anything to add?

crides:
No I don't think

KlzXS:
Everything seems fine. Merged.

KlzXS:
@luukvbaal the compatibility patch :facepalm: ...
Do you think we can make a script to automate generating mainline and compatibility patches together? Otherwise I see us regularly forgetting to generate them. Especially if we get a lot of conflicting patches.

luukvbaal:
Not sure what would be best but something to automate it would be good yeah.

KlzXS:
Better yet, let's nip the problem in the bud. Let's ditch these hardcoded numbers relating to line width and use something better. That way we can avoid making a bunch of compatibility patches each time a new patch adds something else to print. With just one more it would grow to 4, one more after that and we have 11.

luukvbaal:
Yeah if you have something in mind that will work, be my guest.

Alternatively we could auto-generate the diffs from the Github .diff [url](https://github.com/luukvbaal/nnn/commit/f8c98ecb07867a3df2ebbc759b68a1166ca5473c.diff). That would require maintaining the fork branches, either on the main repo or on the patch author's forks.

Not sure what we would present on the repo in that case, perhaps move the patches page to the wiki.

KlzXS:
> Yeah if you have something in mind that will work, be my guest.

I'm looking at `adjust_cols()` right now and this magical `n -= 32`. Where does this number even come from? But if each patch updated it to it needs along with all other numbers which are related to it (like the 36 just above, it must be related) it would work, right?

> Alternatively we could auto-generate the diffs from the Github .diff url. That would require maintaining the fork branches, either on the main repo or on the patch author's forks.

That seems like a lot of work. At some point you might even need a patch1-patch2-patch3-patch4.diff and then you can say goodbye to readability in the Makefile.

luukvbaal:
Yeah `adjust_cols()` always needs to be updated for patches adjusting the detail columns. The `n -= 32` comes from time + size + perms + spacing columns. Indeed the comment is out dated. 

> it would work, right?

Not quite sure what you have in mind.

> That seems like a lot of work. At some point you might even need a patch1-patch2-patch3-patch4.diff and then you can say goodbye to readability in the Makefile.

Could relegate the patching to a shell script that is passed the make options which `curl`s the diff urls and applies the patches. But if you have something better in mind be my guest.

KlzXS:
I'll think about it. Update gitstatus compatibility for now since you know how to do that.

luukvbaal:
It doesn't require an update currently I think?

PR #1048, opened 2021-06-03
Update gitstatus-namefirst patch

luukvbaal:


patches/gitstatus/namefirst.diff
```diff
@@ -173,26 +167,33 @@ index 22402fa..33be591 100644
...
-@@ -5322,6 +5436,18 @@ static int dentfill(char *path, struct entry **ppdents)
+@@ -5322,6 +5430,25 @@ static int dentfill(char *path, struct entry **ppdents)
  #endif
  		}
 
 +		if (git_statuses.len) {
 +			uint merged_status = 0;
++			char *dentpath = abspath(dentp->name, path);
++
++			for (size_t i = 0; i < git_statuses.len; i ++) {
```
KlzXS, line 178: I have no idea how to read a diff of a diff, but I think this means there is still a `i ++`?

KlzXS (changes requested):
See the review. I'll merge later.

patches/gitstatus/namefirst.diff
luukvbaal: You're right, must have changed this like 5 times now while working in branch/diff/PR...

luukvbaal:
Don't merge yet, we have a segfault when browsing directories with details disabled.

luukvbaal:
We also need to either always parse the gitstatus list in `dentfill()` or parse it when that has not been done and `cfg.showdetail` is toggled on. Otherwise you will be required to change directories after toggling it on.

luukvbaal:
The following diff fixes the segfault and will always fill the git_statuses array, even if `cfg.showdetail == 0`. Are we fine with that?

```diff
diff --git a/src/nnn.c b/src/nnn.c
index d064ab1..58aaf8d 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3559,6 +3559,7 @@ static void statuses_free(simple_git_statuses_t statuses) {
                free(statuses.statuses[i].path);

        free(statuses.statuses);
+       git_statuses.len = 0;
 }

 static void resetdircolor(int flags)
@@ -5223,8 +5224,7 @@ static int dentfill(char *path, struct entry **ppdents)

        if (git_statuses.len)
                statuses_free(git_statuses);
-       if (cfg.showdetail)
-               git_statuses = statuses_from_path(path);
+       git_statuses = statuses_from_path(path);

 #if _POSIX_C_SOURCE >= 200112L
        posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
```

KlzXS:
I think so. So is this ready now?

luukvbaal:
It is.

KlzXS:
Thank you for your contribution! I think this is all for now. You can go and rest a bit from this.

jarun:
Thanks guys!

luukvbaal:
Yes, it has been quite a struggle. For now it should be good. We can reconsider how the patch framework should work regarding organizing/applying the patches later.

PR #1049, opened 2021-06-03
Plugin imgview improvements

luukvbaal:
* Don't make symlinks to images in `.nthumbs` such that rename operations in `ucollage/sxiv` work.
* Instead reformat to open both images in working dir and .nthumbs
* Generate audio/video thumbnail for hovered file for `catimg/viu`.
* Other minor improvements...


jarun:
Thank you!

PR #1054, opened 2021-06-04
Preview-tui single previewpid file

luukvbaal:
Since image/pager previewers can't be running concurrently, drop separate pid files in favor of shared one.

Was also wondering if we can/should drop `pistol`/`scope.sh` from the plugin. Not sure how well they work with the current `preview-tui` but I would say the script doesn't need external scripts anymore to determine which previewers to use. @jarun @marioortizmanero (pinging because of https://github.com/jarun/nnn/pull/646)


jarun:
Thank you!

Issue #802, opened 2020-11-23
Feature request: status in git repos


phinzphinz:
Thank you for creating nnn!!! 

**Is your feature request related to a problem? Please describe.**

I would like to see at a glance some info when viewing git repos such as uncommitted files etc.

**Describe the solution you'd like**

It would be nice if this could be indicated via symbols/colors, preferably also with a toggle switch as done for the detail view in nnn.

**Describe alternatives you've considered**

Especially since I want to use nnn as a replacement of nerdtree in vim, it would be nice to have similar behaviour like here:
[Xuyuanp/nerdtree-git-plugin](https://github.com/Xuyuanp/nerdtree-git-plugin)

**Additional context**

Screenshot from the abohve nerdtree-git-plugin (see git status symbols on the left part of the image)
![grafik](https://user-images.githubusercontent.com/28191932/99957082-67308500-2d87-11eb-9990-f9cd987f6655.png)

jarun:
We can't scan every dir `nnn` loads to show git status. Non-file-management utility-specific stuff goes to plugins. You can easily run `git status` (or any other utility you prefer) using run cmd as plugin.

jarun:
BTW, I personally use an alias:

    alias gc='git status'

Whenever i want to check the git status of the current dir from `nnn`, I press <kbd>]</kbd> for the cmd prompt and type `gs`.

phinzphinz:
I understand. But then I have to do this manually `]gs` in every directory whenever I am interested and it would boost productivity to automatically have such info shown.

> We can't scan every dir nnn loads to show git status.

One does not have to, one could use tricks to make this possible: For example if git reports `not a git repository (or any parent up to mount point /mnt)`, we can use this info to avoid subsequent git calls in paret dirs. And if we have done status for some dir it covers also subdirs. If we save such info in a timed cache I think it would be doable without a severe performance hit. Maybe even we can scan manually for a .git dir whenever entering a directory to keep the info whether we are in a git dir as a state parameter. One would then use inotify for parent directories to keep this state parameter in sync. 

I do not know the details about the concepts I am talking about here, I only read about those things. Do you think what I propose would be possible to do in theory (despite violating against your minimalistic file manager design concept)?

jarun:
> One does not have to, one could use tricks to make this possible: For example if git reports not a git repository (or any parent up to mount point /mnt), we can use this info to avoid subsequent git calls in paret dirs.

[...]

Issue #654, opened 2020-06-16
Show git changes


amitlevy21:
Hi, 
I have opened the same issue in https://github.com/mcchrish/nnn.vim/issues/64 but was asked to ask this feature here.

**Is your feature request related to a problem? Please describe.**

When I change a file in vim, it would be nice if I could see which files were lastly modified since last commit.

**Describe the solution you'd like**

Basicly i'm looking for something like [this](https://github.com/Xuyuanp/nerdtree-git-plugin) for nnn.vim.

**Describe alternatives you've considered**

I'm not sure what else I can do besides asking nicely from the developers to implement this.

Thank you for considering this.

jarun:
`nnn` s a general purpose file manager and there are no plans to have git integration. The best you can have is to use sort by modification time.

akimdi:
@amitlevy21   
use https://github.com/SpaceVim/SpaceVim

jarun:
This is now available as user patch `gitstatus`: https://github.com/jarun/nnn/tree/master/patches

PR #1055, opened 2021-06-05
Add gitstatus patch -G flag

luukvbaal:
gitstatus patch change that allows showing the git status column also outside of detail view.

### DIFF
--- a/patches/README.md
+++ b/patches/README.md
@@ -7,7 +7,7 @@ The patches will be adapted on each release when necessary (v4.1 onwards). Each 
 ## List of patches
 | Patch (a-z) | Description | Make var |
 | --- | --- | --- |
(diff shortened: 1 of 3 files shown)


luukvbaal:
Lol @ these diff diffs, so readable.

jarun:
Oh @luukvbaal, it definitely looks like you lose track of your changes in nnn without this patch. ;)

Thank you!

luukvbaal:
Uhm not sure what that's supposed to mean(passive aggressive insult? if not never mind). Sure you can keep track of your changes without this patch. Also it wasn't my idea to begin with. I do feel it's nice to have though.

jarun:
I was just joking about your continued efforts to get it right. Nothing bad.

luukvbaal:
Cross-referencing #654 #802 for visibility as this PR should fix the issues.

PR #1065, opened 2021-06-11
Support opening regular text files on macOS

kevinsjoberg:
on macOS, `-I` does the equivalent of `-i` on other systems.

### DIFF
--- a/plugins/fzopen
+++ b/plugins/fzopen
@@ -52,14 +52,19 @@ if [ "$3" ]; then
 fi
 
 # Open the file (works for a single file only)
-case "$(file -biL "$entry")" in
+cmd_file=""
+cmd_open=""
+if uname | grep -q "Darwin"; then
+  cmd_file="file -bIL"
+  cmd_open="open"
+else
+  cmd_file="file -biL"
+  cmd_open="xdg-open"
+fi
+
+case "$($cmd_file "$entry")" in
     *text*)
         "${VISUAL:-$EDITOR}" "$entry" ;;
     *)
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

plugins/fzopen
```diff
@@ -52,14 +52,19 @@ if [ "$3" ]; then
 fi
 
 # Open the file (works for a single file only)
-case "$(file -biL "$entry")" in
+cmd_file=""
+cmd_open=""
+if uname | grep -q "Darwin"; then
+  cmd_file="file -bIL"
```
jarun, line 58: Seriously, guys! Please double check your indentation when you are contributing to an established project. I will take care of this. But we really don't have time for these silly stuff anymore.

kevinsjoberg: Oh, did I do something wrong? I'm terribly sorry if that's the case. I ran `shellcheck` on the file before submitting the pull request.

jarun: shellcheck doesn't check indentation. I think we need to find a way to automate that check. This is getting regular in plugin contributions.

kevinsjoberg: Oh, I see. I messed up the indentation. That's on me. Sorry. What do you think about introducing a `.editorconfig` that can enforce this for us?

jarun: No I think we should enable that on CI. We need to check that. No worries!

jarun:
Thanks for your contribution. If you have the time, please see how the other plugins do on macOS as well.

PR #1066, opened 2021-06-11
Remove opinionated options

kevinsjoberg:
The file picker becomes close to useless for me when used in a large enough project since I can only match against the file name, not the entire path. I think we can assume the majority of people wants multi selection, so this is set by default.

If one is interested in using other options, they can set `$FZF_DEFAULT_OPTS` accordingly.


jarun:
Plugins are indicative and supposed to be adapted fr use. However, I see there's a comment below that already specifies the params. So we are good.

PR #1068, opened 2021-06-12
Update preview-tui

Moelf:


jarun:
Thank you!

jarun:
I reverted the change on second thoughts. We don't list deps of deps.
kitty users should be aware of the deps of the tool they use.

Issue #1064, opened 2021-06-10
Should `nnn -p -` emit file details to stdout?


kevinsjoberg:
#### Environment details (Put `x` in the checkbox along with the information)

- [x] Operating System: macOS, `11.3.1`
- [x] Desktop Environment: macOS
- [x] Terminal Emulator: Kitty, `0.20.3`
- [x] Shell: bash, `5.1.8`
- [ ] Custom desktop opener (if applicable):
- [x] Program options used: `-p -`
- [ ] Configuration options set:
- [x] Issue exists on `nnn` master

#### Exact steps to reproduce the issue

When providing `-p -`, my assumption was that file selection would output to stdout, but when pressing `f` to get file details, this output is also outputted to stdout, which is kind of unexpected. If I instead, provide a regular file to `-p` like so `nnn -p output.txt`, pressing `f` does not output into it.

1. `nnn -p -`
2. Press `f`
3. Press `enter`
4. See output below

```
16777220 24380305 -rw-r--r-- 1 kevinsjoberg wheel 0 0 "Jun 10 18:14:55 2021" "Jun 10 18:14:55 2021" "Jun 10 18:14:55 2021" "Jun 10 18:14:55 2021" 4096 0 0 /private/tmp/test.txt


  empty

  inode/x-empty; charset=binary

/private/tmp/test.txt
```

1. `nnn -p output.txt`
2. Press `f`
3. Press `enter`
4. See output below

```
/private/tmp/test.txt
```

jarun:
> When providing -p -, my assumption was that file selection would output to stdout

That's correct!

> but when pressing f to get file details, this output is also outputted to stdout, which is kind of unexpected

No, this output is paged, `nnn` captures the output of `file` and shows it in PAGER. If not, please provide a screenshot or video of your observation.

> If I instead, provide a regular file to -p like so nnn -p output.txt, pressing f does not output into it.

Because `file` writes to stdout, it's not aware of the file you provided. However, as I mentioned above, `nnn` pages the `file` output.

jarun:
Relevant code: https://github.com/jarun/nnn/blob/master/src/nnn.c#L4175

0xACE:
Hejsan Kevin, nice catch :)

from `8238410d4c6c1164f6dcf2dadc834e1a2e0b9d9a` (master) I can recreate this issue.

```
$ nnn -p - | sed 's/^/CAPTURED STDOUT:/
[...]

PR #1070, opened 2021-06-12
Add force-tty capability to spawn and set pagers to it

annagrram:
Addresses issue #1064

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -210,15 +210,16 @@
 #define FILE_SELECTED 0x10
 
 /* Macros to define process spawn behaviour as flags */
-#define F_NONE    0x00  /* no flag set */
-#define F_MULTI   0x01  /* first arg can be combination of args; to be used with F_NORMAL */
-#define F_NOWAIT  0x02  /* don't wait for child process (e.g. file manager) */
-#define F_NOTRACE 0x04  /* suppress stdout and stderr (no traces) */
-#define F_NORMAL  0x08  /* spawn child process in non-curses regular CLI mode */
-#define F_CONFIRM 0x10  /* run command - show results before exit (must have F_NORMAL) */
-#define F_CHKRTN  0x20  /* wait for user prompt if cmd returns failure status */
-#define F_NOSTDIN 0x40  /* suppress stdin */
-#define F_PAGE    0x80  /* page output in run-cmd-as-plugin mode */
+#define F_NONE      0x00  /* no flag set */
+#define F_MULTI     0x01  /* first arg can be combination of args; to be used with F_NORMAL */
+#define F_NOWAIT    0x02  /* don't wait for child process (e.g. file manager) */
+#define F_NOTRACE   0x04  /* suppress stdout and stderr (no traces) */
+#define F_NORMAL    0x08  /* spawn child process in non-curses regular CLI mode */
+#define F_CONFIRM   0x10  /* run command - show results before exit (must have F_NORMAL) */
+#define F_CHKRTN    0x20  /* wait for user prompt if cmd returns failure status */
+#define F_NOSTDIN   0x40  /* suppress stdin */
+#define F_PAGE      0x80  /* page output in run-cmd-as-plugin mode */
+#define F_FORCE_TTY 0x100 /* Force stdout to go to tty if redirected to a non-tty */
 #define F_CLI     (F_NORMAL | F_MULTI)
 #define F_SILENT  (F_CLI | F_NOTRACE)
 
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -2081,6 +2082,13 @@ static int spawn(char *file, char *arg1, char *arg2, char *arg3, uchar_t flag)
 			dup2(fd, 1);
 			dup2(fd, 2);
 			close(fd);
+		} else if (flag & F_FORCE_TTY) {
+			/* If stdout has been redirecet to a non-tty, force output to tty */
+			if (!isatty(STDOUT_FILENO)) {
+				int fd = open("/dev/tty", O_WRONLY, 0200);
```
kevinsjoberg, line 2088: Could we use `freopen` here?

annagrram: We could, but I wanted to keep it in the same "API set" as the already written code in this function.

src/nnn.c
```diff
@@ -2081,6 +2082,13 @@ static int spawn(char *file, char *arg1, char *arg2, char *arg3, uchar_t flag)
 			dup2(fd, 1);
 			dup2(fd, 2);
 			close(fd);
+		} else if (flag & F_FORCE_TTY) {
+			/* If stdout has been redirecet to a non-tty, force output to tty */
```
kevinsjoberg, line 2086: Minor typo here. `redirecet` should be `redirected`.

annagrram: Thanks :)

src/nnn.c
```diff
@@ -2084,6 +2085,13 @@ static int spawn(char *file, char *arg1, char *arg2, char *arg3, uchar_t flag)
 			dup2(fd, STDOUT_FILENO);
 			dup2(fd, STDERR_FILENO);
 			close(fd);
+		} else if (flag & F_FORCE_TTY) {
+			/* If stdout has been redirected to a non-tty, force output to tty */
+			if (!isatty(STDOUT_FILENO)) {
+				int fd = open("/dev/tty", O_WRONLY, 0200);
```
kevinsjoberg, line 2091: Perhaps unnecessary here, but we could use `ctermid()` to avoid hard coding `/dev/tty`.

annagrram: Changed.
I'm not that strong in my POSIX functions :)

kevinsjoberg: No worries, thanks a lot! 👍

kevinsjoberg (approved):
LGTM. Thank you for this. 🎉

src/nnn.c
annagrram: My pleasure

jarun:
@annagrram CI fails. Please check.

luukvbaal:
CI fails due to https://github.com/jarun/nnn/commit/69926a8eff971655112d950150802f92f8f4423c @jarun

annagrram:
> @annagrram CI fails. Please check.

@jarun, not because of this PR. Check your last couple of commits

jarun:
Got it! Resolved the issue now.

src/nnn.c
```diff
@@ -800,7 +801,7 @@ static haiku_nm_h haiku_hnd;
 
 /* Forward declarations */
 static void redraw(char *path);
-static int spawn(char *file, char *arg1, char *arg2, char *arg3, uchar_t flag);
+static int spawn(char *file, char *arg1, char *arg2, char *arg3, uint16_t flag);
```
jarun, line 804: Please use `ushort_t` instead of `uint16_t`.

Both `uint16_t` and `uint32_t` are [optional](https://en.cppreference.com/w/cpp/types/integer).

jarun (changes requested):

src/nnn.c
```diff
@@ -2084,6 +2085,13 @@ static int spawn(char *file, char *arg1, char *arg2, char *arg3, uchar_t flag)
 			dup2(fd, STDOUT_FILENO);
 			dup2(fd, STDERR_FILENO);
 			close(fd);
+		} else if (flag & F_FORCE_TTY) {
```
jarun, line 2088: Hmm... I reduce lines, you keep adding.

Just kidding! This is fine!

jarun:
Thank you so much! :+1:

src/nnn.c
craigbarnes: FWIW, `uint16_t` is a required type on POSIX systems:

> The following types are required:
>
> int8_t
> int16_t
> int32_t
> uint8_t
> uint16_t
> uint32_t

-- https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdint.h.html

jarun: Thanks for the reference! I think we can reduce a few typedefs now. ;)

jarun: And thanks to out patch framework. Making changes all around the code is not so simple anymore.
I'm leaving these as they are.

PR #1073, opened 2021-06-15
File-explorer functionality

mizlan:
Adds flag `-F` that prevents nnn from exiting on selection. This uses the path given to `-p` and will append to the output on each selection.

I'm not exactly aware of what `seltofile` does but it was getting in the way so I commented it out. I am not that great at C so please point out if I'm doing anything incorrectly here.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -6428,7 +6428,7 @@ nochange:
 			if (g_state.picker && sel == SEL_OPEN) {
 				appendfpath(newpath, mkpath(path, pent->name, newpath));
 				// do not exit if in explorer mode
-				if (explorer) {
+				if (g_state.explorer) {
 					// add a newline to the file to make it easier to parse `tail -F`
 					strcat(pselbuf, "\n");
 					appendsel(pselbuf, selbufpos);
@@ -7817,7 +7817,7 @@ int main(int argc, char *argv[])
 
 	while ((opt = (env_opts_id > 0
 		       ? env_opts[--env_opts_id]
-		       : getopt(argc, argv, "aAb:cCdDeEfgHJKl:nop:P:QrRs:St:T:uUVwxh"))) != -1) {
+		       : getopt(argc, argv, "aAb:cCdDeEfFgHJKl:nop:P:QrRs:St:T:uUVwxh"))) != -1) {
 		switch (opt) {
 #ifndef NOFIFO
 		case 'a':
@@ -7853,6 +7853,9 @@ int main(int argc, char *argv[])
 #ifndef NORL
 			rlhist = TRUE;
 #endif
+			break;
+		case 'F':
+		  	g_state.explorer = 1;
 			break;
 		case 'g':
 			cfg.regex = 1;
@@ -8221,7 +8224,7 @@ int main(int argc, char *argv[])
 		if (selbufpos) {
 			fd = selpath ? open(selpath, O_WRONLY | O_CREAT, 0600) : STDOUT_FILENO;
 			// HACK: don't send seltofile if in explorer mode
-			if (!explorer && ((fd == -1) || (seltofile(fd, NULL) != (size_t)(selbufpos))))
+			if (!g_state.explorer && ((fd == -1) || (seltofile(fd, NULL) != (size_t)(selbufpos))))
 				xerror();
 
 			if (fd > 1)
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -339,6 +339,7 @@ typedef struct {
 	uint_t dirctx     : 1;  /* Show dirs in context color */
 	uint_t uidgid     : 1;  /* Show owner and group info */
 	uint_t prstssn    : 1;  /* Persistent session */
+	uint_t explorer   : 1;  /* Explorer mode: don't exit after selection is opened */
 	uint_t duinit     : 1;  /* Initialize disk usage */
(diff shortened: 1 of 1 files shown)


mizlan:
To try this yourself, if desired, run `nnn -F -p ~/testfile` and in another terminal window run `tail -F ~/testfile`.

jarun:
Ci fails. Please check.

@luukvbaal can you please help with the testing?

src/nnn.c
```diff
@@ -8208,7 +8238,8 @@ int main(int argc, char *argv[])
 	if (g_state.picker) {
 		if (selbufpos) {
 			fd = selpath ? open(selpath, O_WRONLY | O_CREAT, 0600) : STDOUT_FILENO;
-			if ((fd == -1) || (seltofile(fd, NULL) != (size_t)(selbufpos)))
+			// HACK: don't send seltofile if in explorer mode
+			if (!g_state.explorer && ((fd == -1) || (seltofile(fd, NULL) != (size_t)(selbufpos))))
```
mizlan, line 8242: Is it okay to not run this line?

jarun: Can you please explain what was the problem with this and what you are trying to achieve?

src/nnn.c
```diff
@@ -1392,6 +1393,21 @@ static void writesel(const char *buf, const size_t buflen)
 		printwarn(NULL);
 }
 
+static void appendsel(const char *buf, const size_t buflen)
+{
+	if (!selpath)
+		return;
+
+	FILE *fp = fopen(selpath, "a");
```
jarun, line 1401: if you see the latest code, I have removed every instance of `FILE *`. Please rebase and extend writesel() to take one more bool param and append instead of overwrite if it's true.

src/nnn.c
```diff
@@ -6411,6 +6427,17 @@ static bool browse(char *ipath, const char *session, int pkey)
 			/* If opened as vim plugin and Enter/^M pressed, pick */
 			if (g_state.picker && sel == SEL_OPEN) {
 				appendfpath(newpath, mkpath(path, pent->name, newpath));
+				// do not exit if in explorer mode
+				if (g_state.explorer) {
```
jarun, line 6431: Why is the explorer mode below picker mode? Do they have anything in common? If yes, case 'F' in line 7858 should also set picker mode.

src/nnn.c
```diff
@@ -6411,6 +6427,17 @@ static bool browse(char *ipath, const char *session, int pkey)
...
+					// add a newline to the file to make it easier to parse `tail -F`
```
jarun, line 6432: Please explain your design. What are you trying to achieve?

src/nnn.c
```diff
@@ -6411,6 +6427,17 @@ static bool browse(char *ipath, const char *session, int pkey)
...
+					appendfpath("\n", 2);
+					appendsel(pselbuf, selbufpos);
+					selbufpos = 0;
+					// is this necessary?
```
jarun, line 6436: Not needed.

src/nnn.c
jarun: So you are writing one file at a time to selection file newline separated.
When you read the selection file at the other end, how do you know which position you read last? Because user can keep pressing <kbd>Enter</kbd> on files one after another.

Why not read the FIFO like the previewer instead? We can have a different fifo for the same.

jarun (changes requested):

src/nnn.c
mizlan: > When you read the selection file at the other end, how do you know which position you read last? Because user can keep pressing Enter on files one after another.

Using `tail -F` and a callback on stdout.

> Why not read the FIFO like the previewer instead? We can have a different fifo for the same.

Ah, I was approaching this from the wrong direction. I think this will significantly simplify it. And I think we can use the same fifo right? Just attaching a different listener.

src/nnn.c
mizlan: I think I can remove it now. I was originally repurposing the picker file as a file to append filepaths to, but I'm going to switch to using NNN_FIFO.

jarun: Re-using `NNN_FIFO` may be an issue as it sends out the file path on hover. You only need it when user presses Enter/^M on the file.

BTW, please rebase your code.

mizlan: I have rebased. I may open a new PR as I'm going to scratch most of the changes I've made here. Are you okay with that?

mizlan: Could you refactor the code to make it easier to work with multiple fifos? For example having `notify_fifo` take a file descriptor as an argument, or I'm not too sure what's the best way to do it.

src/nnn.c
mizlan: Do I need to do this now that I plan on using a FIFO?

jarun: No. See `notify_fifo()`.

src/nnn.c
jarun: Just add a `bool` to `notify_fifo()` and follow the current code.

jarun: Yes, please close the current PR.

luukvbaal:
Sorry for the late response. 

This is supposed to provide a persistent picker mode mainly for `nnn.vim` right?
I agree that a separate FIFO makes sense.

jarun:
That's correct! @mizlan will be raising a new review with FIFO based implementation.

PR #1074, opened 2021-06-16
Update gitstatus patch

luukvbaal:
* Directly use `libgit2` `git_status_t` lookup in `print_gitstatus()`, getting rid of custom `git_column_status_t` `enum` and `git_get_indexed/staged_status()` functions. (Less code and should be more performant as well although I didn't benchmark)
* Don't print git status when folder contains only unchanged or ignored entries: (git status column showing up in my homefolder due to dotfile repository was bothering me:yum:)
out
Determined in `dentfill()` by setting boolean `git_statuses.show` to `TRUE` according to the described precondition; performance hit should be negligible.
In my opinion this is desired or at least reasonable behavior although if any user of the patch (now or in the future) disagrees we could reconsider.

@crides please review if you want/can and see if you agree with the change(don't know if you have looked at and/or actually ended up using the current patch)
@KlzXS as well if you can since you reviewed the initial patch.

* Also rebase the namefirst patch to master while we're at it.

### DIFF
--- a/patches/gitstatus/mainline.diff
+++ b/patches/gitstatus/mainline.diff
@@ -8,7 +8,7 @@
 # Authors: @crides, Luuk van Baal
 
 diff --git a/src/nnn.c b/src/nnn.c
-index 402d905..e10035b 100644
+index 67523fe6..c965bd1c 100644
 --- a/src/nnn.c
 +++ b/src/nnn.c
 @@ -105,6 +105,7 @@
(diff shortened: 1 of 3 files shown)


luukvbaal:
Force pushed fix for `getopt()` conflict assuming #1075 will be merged.

jarun:
Thank you!

jarun:
Good call too!

PR #1075, opened 2021-06-16
Refactor fifo code and add explorer

mizlan:
Work in progress, but let me know if anything doesn't look right.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -401,7 +401,8 @@ static context g_ctx[CTX_MAX] __attribute__ ((aligned));
 static int ndents, cur, last, curscroll, last_curscroll, total_dents = ENTRY_INCR, scroll_lines = 1;
 static int nselected;
 #ifndef NOFIFO
-static int fifofd = -1;
+static int hover_fifofd = -1;
+static int explorer_fifofd = -1;
 #endif
 static uint_t idletimeout, selbufpos, lastappendpos, selbuflen;
 static ushort_t xlines, xcols;
@@ -424,7 +425,8 @@ static char *plgpath;
 static char *pnamebuf, *pselbuf;
 static char *mark;
 #ifndef NOFIFO
-static char *fifopath;
+static char *hover_fifopath;
+static char *explorer_fifopath;
 #endif
 static unsigned long long *ihashbmp;
 static struct entry *pdents;
@@ -808,7 +810,7 @@ static void move_cursor(int target, int ignore_scrolloff);
 static char *load_input(int fd, const char *path);
 static int set_sort_flags(int r);
 #ifndef NOFIFO
-static void notify_fifo(bool force);
+static void notify_fifo(bool force, bool is_explorer); /* explorer has a separate fifo */
 #endif
 
 /* Functions */
@@ -2757,7 +2759,7 @@ try_quit:
 			} else {
 #ifndef NOFIFO
 				/* Send hovered path to NNN_FIFO */
-				notify_fifo(TRUE);
+				notify_fifo(TRUE, FALSE);
 #endif
 				escaped = TRUE;
 				settimeout();
@@ -5409,18 +5411,22 @@ static void populate(char *path, char *lastname)
 }
 
 #ifndef NOFIFO
-static void notify_fifo(bool force)
+static void notify_fifo(bool force, bool is_explorer)
 {
-	if (!fifopath)
+	/* refer to the explorer fifo instead of the hover fifo if is_explorer is true */
+	char **fifopath_ptr = is_explorer ? &explorer_fifopath : &hover_fifopath;
+	int *fifofd_ptr = is_explorer ? &explorer_fifofd : &hover_fifofd;
+
+	if (!(*fifopath_ptr))
 		return;
 
-	if (fifofd == -1) {
-		fifofd = open(fifopath, O_WRONLY|O_NONBLOCK|O_CLOEXEC);
-		if (fifofd == -1) {
+	if (*fifofd_ptr == -1) {
+		*fifofd_ptr = open(*fifopath_ptr, O_WRONLY|O_NONBLOCK|O_CLOEXEC);
+		if (*fifofd_ptr == -1) {
 			if (errno != ENXIO)
 				/* Unexpected error, the FIFO file might have been removed */
 				/* We give up FIFO notification */
-				fifopath = NULL;
+				*fifopath_ptr = NULL;
 			return;
 		}
 	}
@@ -5437,7 +5443,7 @@ static void notify_fifo(bool force)
 
 	path[len - 1] = '\n';
 
-	ssize_t ret = write(fifofd, path, len);
+	ssize_t ret = write(*fifofd_ptr, path, len);
 
 	if (ret != (ssize_t)len && !(ret == -1 && (errno == EAGAIN || errno == EPIPE))) {
 		DPRINTF_S(strerror(errno));
@@ -5473,7 +5479,7 @@ static void move_cursor(int target, int ignore_scrolloff)
 	curscroll = MAX(curscroll, MAX(cur - (onscreen - 1), 0));
 
 #ifndef NOFIFO
-	notify_fifo(FALSE);
+	notify_fifo(FALSE, FALSE);
 #endif
 }
 
@@ -6339,7 +6345,7 @@ nochange:
 					move_cursor(r, 1);
 #ifndef NOFIFO
 				else if (event.bstate == BUTTON1_PRESSED)
-					notify_fifo(TRUE);
+					notify_fifo(TRUE, FALSE);
 #endif
 				/* Handle right click selection */
 				if (event.bstate == BUTTON3_PRESSED) {
@@ -7761,7 +7767,7 @@ static void cleanup(void)
 	free(plug);
 #ifndef NOFIFO
 	if (g_state.autofifo)
-		unlink(fifopath);
+		unlink(hover_fifopath);
 #endif
 	if (g_state.pluginit)
 		unlink(g_pipepath);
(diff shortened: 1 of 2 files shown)


luukvbaal:
Passing `<hover/,explorer>_<fifo,path>` directly to `notify_fifo()` instead of a boolean makes more sense I think. Unless more changes in `notify_fifo()` are required for either case.

Also is `picker` instead of `explorer` not a more appropriate name? You're making a persistent picker mode right, I would consider `nnn` in and of itself a file "explorer"  already.

mizlan:
> Passing `<hover,explorer>_<fifo,path>` directly to `notify_fifo()` instead of a boolean

I like this better. @jarun What do you think?

> Also is picker instead of explorer not a more appropriate name?

Good point. Although, picker already has a meaning within the source code but this is not really related to the picker implementation-wise.

luukvbaal:
> I like this better. @jarun What do you think?

`notify_fifo(bool force, const int fifofd, const char *fifopath)` is what I meant. 
If the function stays exactly the same except for those two arguments I would say it makes more sense to pass those arguments rather than passing an arbitrary boolean that determines those arguments inside the function no, even if only two different cases exist. Whatever works though I guess.


> Good point. Although, picker already has a meaning within the source code but this is not really related to the picker implementation-wise.

Yeah I would name it `prstpicker` or something for persistent picker(`prstssn` is used for persistent session) because I thought that's what you were implementing, if not never mind.

mizlan:
The concept of persistence is different in the case of persistent session though (persisting across different instances of nnn) while here it is just "not closing." I am sort of against the idea.

luukvbaal:
The same terminology to describe the issue was used here https://github.com/mcchrish/nnn.vim/issues/61#issuecomment-633784351.
I think its an apt description, you would be using the `nnn` picker mode persistently right:yum:?
I don't mind though, its just a name after all...

mizlan:
Alright, I'll consider it :)

I am running into an issue, however. This code currently works when the FIFO passed to `-F` is created ahead of time. However, it doesn't create the FIFO, even adding this:

```diff
diff --git a/src/nnn.c b/src/nnn.c
index 3e874926..a5f4533e 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -8095,6 +8095,15 @@ int main(int argc, char *argv[])
 
 		sigaction(SIGPIPE, &(struct sigaction){.sa_handler = SIG_IGN}, NULL);
 	}
+
+	if (g_state.explorer && explorer_fifopath) {
+		if (mkfifo(explorer_fifopath, 0600) != 0 && !(errno == EEXIST && access(explorer_fifopath, W_OK) == 0)) {
+			xerror();
+			return EXIT_FAILURE;
+		}
+
+		sigaction(SIGPIPE, &(struct sigaction){.sa_handler = SIG_IGN}, NULL);
+	}
 #endif
 
 #ifdef LINUX_INOTIFY
```

Is there anything I should do other than adding that?

luukvbaal:
Going by your current PR, I think your `explorer_fifopath()` will be `NULL` because `realpath()` returns NULL for non-existing files.
Not sure if we have a way to dereference file paths passed as an argument, can't look into it right now.

mizlan:
@luukvbaal How about this? Here I assume that the path passed to `-F` is valid (no call to `realpath`)

mizlan:
Hmm, should I make this a patch instead of putting these changes in master?

luukvbaal:
> @luukvbaal How about this? Here I assume that the path passed to `-F` is valid (no call to `realpath`)

Yeah I think requiring an absolute path by the caller (e.g. `nnn.vim` or other scripts) for `-F` is fine.

> Hmm, should I make this a patch instead of putting these changes in master?

I think this is a welcome addition to master given how it was requested a few times in `nnn.vim. Don't know how @jarun feels about it.

src/nnn.c
```diff
@@ -7754,7 +7766,7 @@ static void cleanup(void)
 	free(plug);
 #ifndef NOFIFO
 	if (g_state.autofifo)
-		unlink(fifopath);
+		unlink(hover_fifopath);
 #endif
```
luukvbaal, line 7770: Don't forget
```suggestion
            if (g_state.explorer)
	        unlink(explorer_fifopath);
 #endif
```

luukvbaal:
Sorry, used spaces instead of tabs for that suggestion(idk how to insert tabs in this markdown window).
Please fix. Also off to bed now, gn.

src/nnn.c
mizlan: Thanks! I don't know how I missed this bit. I'm going to unresolve this because formatting and also, shouldn't this also be inside the ifndef NOFIFO part as well?

mizlan:
Should I change all occurrences of "explorer" to "filetree" or "prstpicker" instead? Not sure what the right terminology here is.

jarun:
Can you please share an image how it looks like?

mizlan:
Of nnn.vim or nnn? Do you want a side-by-side of the FIFO output and nnn?

jarun:
When you use it as an explorer... I think that would be nnn.vim.

mizlan:
Sorry, I'm not done with that yet.

jarun:
Okk. Why are you using the term _tree_?

mizlan:
I'm not sure to be honest. Emacs has a thing called treemacs, which is probably what made me think of "tree," but the naming is up to you. @luukvbaal said "explorer" doesn't really make sense because nnn is an explorer in itself.

jarun:
We don't have a tree view. I would say _prstpicker_ is the right term.

jarun:
On second thoughts, it's ok to have explorer. Stay with it. Easy to refer to and understand.

src/nnn.c
```diff
@@ -401,7 +402,8 @@ static context g_ctx[CTX_MAX] __attribute__ ((aligned));
 static int ndents, cur, last, curscroll, last_curscroll, total_dents = ENTRY_INCR, scroll_lines = 1;
 static int nselected;
 #ifndef NOFIFO
-static int fifofd = -1;
+static int hover_fifofd = -1;
```
jarun, line 405: Shorten these with comments: `hfifofd` and `efifofd`.

src/nnn.c
```diff
@@ -424,7 +426,8 @@ static char *plgpath;
 static char *pnamebuf, *pselbuf;
 static char *mark;
 #ifndef NOFIFO
-static char *fifopath;
+static char *hover_fifopath;
```
jarun, line 429: Shorten with comments - `hfifopath`, `efifopath`.

src/nnn.c
```diff
@@ -803,7 +806,7 @@ static void move_cursor(int target, int ignore_scrolloff);
 static char *load_input(int fd, const char *path);
 static int set_sort_flags(int r);
 #ifndef NOFIFO
-static void notify_fifo(bool force);
+static void notify_fifo(bool force, bool is_explorer); /* explorer has a separate fifo */
```
jarun, line 809: Just `explorer`. Drop the `is_`.

src/nnn.c
```diff
@@ -5402,18 +5405,22 @@ static void populate(char *path, char *lastname)
 }
 
 #ifndef NOFIFO
-static void notify_fifo(bool force)
+static void notify_fifo(bool force, bool is_explorer)
 {
-	if (!fifopath)
+	/* refer to the explorer fifo instead of the hover fifo if is_explorer is true */
+	char **fifopath_ptr = is_explorer ? &explorer_fifopath : &hover_fifopath;
```
jarun, line 5411: Use `ppath` and `pfd`.
We know we are dealing with fifos here.

jarun: Oh, why do we even need pointers here? fifofd and fifopath are both globals.

src/nnn.c
```diff
@@ -7821,6 +7835,12 @@ int main(int argc, char *argv[])
 		case 'f':
 #ifndef NORL
 			rlhist = TRUE;
+#endif
+			break;
+		case 'F':
```
jarun, line 7840: Use `-X` (capital) for `explorer` mode. We will document what we call `explorer mode`.

src/nnn.c
```diff
@@ -8068,9 +8088,18 @@ int main(int argc, char *argv[])
...
+	if (hover_fifopath) {
+		if (mkfifo(hover_fifopath, 0600) != 0 && !(errno == EEXIST && access(hover_fifopath, W_OK) == 0)) {
+			xerror();
+			return EXIT_FAILURE;
+		}
+
+		sigaction(SIGPIPE, &(struct sigaction){.sa_handler = SIG_IGN}, NULL);
+	}
+
+	if (g_state.explorer && explorer_fifopath) {
```
jarun, line 8101: Why do we need to check `explorer_fifopath`. The new option requires a value which comes through `optarg`.

jarun (changes requested):

src/nnn.c
mizlan: Yeah, I was a little bit confused at this part. I reassign the value of the path under some circumstances, so wouldn't it be necessary to use a pointer to update it?

src/nnn.c
mizlan: Will fix

src/nnn.c
jarun: Also drop the comment.

src/nnn.c
jarun: OK, I see why. You will need pointers. But keep those short as I mentioned.

luukvbaal:
We need a man page entry as well.

jarun:
Yes, we are good to go! Please add `-X` to man page.

mizlan:
Do I need to add a section on explorer mode or is the mention of the flag enough?

jarun:
Mention of the flag is enough. :+1: Thank you!

toddyamakawa:
It looks like this patch adds an "explorer fifo"? What is the difference between an explorer FIFO and a hover FIFO?

mizlan:
When nnn is embedded inside another text editor, the explore FIFO allows nnn to act as a project drawer, where selecting files opens them up in the editor, by having the editor listen to the explorer FIFO. Think of it as the picker mode, but persistent.

toddyamakawa:
Thanks for the explanation.
How do you send files to the explorer FIFO?
Will this allow you to send a path from the text editor to change nnn's directory?

mizlan:
1. Check the source code for `notify_fifo()`.
2. No, since the communication is one way. Also that functionality seems
outside the scope of explorer mode.

PR #1078, opened 2021-06-20
Replace fd dependency with find

KlzXS:
It's not the prettiest and might be slower than `fd` but it gets the job done.

### DIFF
--- a/plugins/fzcd
+++ b/plugins/fzcd
@@ -2,7 +2,7 @@
 
 # Description: Fuzzy search multiple locations read-in from a path-list file
 #              (or $PWD) and open the selected file's dir in a smart context.
-# Dependencies: fzf, fd (only for multi-location search)
+# Dependencies: fzf, find (only for multi-location search)
 #
 # Details: Paths in list file should be newline-separated absolute paths.
 #          Paths can be file paths; the script will scan the parent dirs.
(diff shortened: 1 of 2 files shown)


plugins/fzcd
```diff
@@ -55,11 +55,11 @@ if [ -n "$LIST" ]; then
             fi
         done < "$LIST"
 
-        sel=$(xargs -d '\n' -a "$tmpfile" fd -H . | fzf --delimiter / --tiebreak=begin --info=hidden)
+        sel=$(xargs -d '\n' -a "$tmpfile" -I{} find {} -type f -printf "%H//%P\n" | sed '/.*\/\/\(\..*\|.*\/\..*\)/d; s:/\+:/:g' | fzf --delimiter / --tiebreak=begin --info=hidden)
```
jarun, line 58: Probably keep the `fd` command commented below in case someone wants to use it.

jarun:
> might be slower

On SSD I don't see any difference. In a lot of cases, at least for _reasonable_ workloads, basic utilities are not remotely as bad as they are presented to be by newer tools, esspecially from the rust community.

I am merging it. Will add the commented hint.

jarun:
Thank you!

KlzXS:
What I meant by slower is that there isn't really a way to prune the search tree while keeping the desired behavior. So it will search everything and then hand it to `sed` to remove unwanted stuff. A program that is written in rust like `fd` could prune it right there and then.

People likely won't even notice it or will just think it's normal, but still, I know that it's there and will be acutely aware of it.

jarun:
Got it!

PR #1084, opened 2021-06-27
Imgview open hovered image and reuse NNN_PREVIEWDIR

### LINKED ISSUE #1083: imgview does not open sxiv on the image that was hovered
https://youtu.be/mDO9ks_D4OI

The first image I open is what should come up when using imgview. But as you can see, a different image comes up.

luukvbaal:
Should fix #1083

### DIFF
--- a/plugins/imgview
+++ b/plugins/imgview
@@ -1,8 +1,7 @@
 #!/usr/bin/env sh
 
 # Description: Open hovered or current directory in image viewer.
-#              Supported image viewers open in thumbnail mode when
-#              the hovered entry is a directory.
+#              Generates media thumbnails with optional dependencies.
 #
 # Dependencies:
 #   - imv (https://github.com/eXeC64/imv) or,
@@ -16,8 +15,10 @@
 #
 # Shell: POSIX compliant
 # Author: Arun Prakash Jana, Luuk van Baal
-
-target="$(readlink -f "$1")"
+#
+# Consider setting NNN_PREVIEWDIR to $XDG_CACHE_HOME/nnn/previews
+# if you want to keep media thumbnails on disk between reboots.
+NNN_PREVIEWDIR="${NNN_PREVIEWDIR:-${TMPDIR:-/tmp}/nnn/previews}"
 
 exit_prompt() {
     [ -n "$1" ] && printf "%s\n" "$1"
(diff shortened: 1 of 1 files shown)


plugins/imgview
```diff
@@ -83,31 +91,31 @@ elif type lsix >/dev/null 2>&1; then
...
-      nthumb_cleanup && exit_prompt
+		    nthumb_cleanup && exit_prompt
     fi
     nthumb_cleanup
     exit
 elif type sxiv >/dev/null 2>&1; then
     make_thumbs 0
     if [ -f "$target" ]; then
-        view_files sxiv -a >/dev/null 2>&1 &
+        view_files sxiv "$target" >/dev/null 2>&1 &
```
tonijarjour, line 101: Please add the -a flags so sxiv plays gifs automatically

tonijarjour:
nvm its all good

plugins/imgview
```diff
@@ -83,31 +91,31 @@ elif type lsix >/dev/null 2>&1; then
 elif type ucollage >/dev/null 2>&1; then
     make_thumbs 1
     if ! UCOLLAGE_EXPAND_DIRS=1 ucollage . .nthumbs; then
-      nthumb_cleanup && exit_prompt
+		    nthumb_cleanup && exit_prompt
     fi
     nthumb_cleanup
     exit
 elif type sxiv >/dev/null 2>&1; then
     make_thumbs 0
```
jarun, line 99: If I understand the logic correctly, do we really need to create a directory and show the image even when only a file is passed? I would like avoid that on e.g. read-only storage. In general, we should not issue writes if not absolutely necessary.

jarun (changes requested):

plugins/imgview
tonijarjour, line 99: For what its worth, if you could make this only happen if a video was passed, and if ffmpegthumbnailer is actually installed. It would make the program a lot faster.

luukvbaal: My reasoning was that there is no reason the behavior should be any different when a file or a directory is hovered after merging the thumbnail plugin.

When only a file is passed we still want to be able to view other images (or thumbnails) in the same directory (i.e. by pressing <kbd>n</kbd> in `sxiv`), thus thumbnails for media files in the directory are created in this case as well.

I do agree that it should continue gracefully for read only memory, would that be fine or do you still want it out altogether?

luukvbaal: Whether a video is passed or not is not sufficient to decide not to run `make_thumbs` with the current logic. We are creating thumbnails for all audio/video files in the same directory as the passed argument.

We are also using `ffmpeg` to generated audio thumbnails. I doubt much people would have  a system without `ffmpeg` so running `make_thumbs` when either `ffmpeg` or `ffmpegthumbnailer` is installed would likely result in it always being ran.

Is the slowdown really that noticeable? All thats being run is `file`.
You can also just comment it out if you don't need media thumbnails.

tonijarjour: The slowdown is noticeable but its not a big deal. Im working with over 300 files.
Since as you said only `file` is being run, there's nothing to do there. But in the case of sxiv, you passed `0` to `make_thumbs`, which means you aren't making audio thumbails (from what I understand from the code). In that case you can check if `ffmpegthumbnailer` is installed and if not, skip the iterator all together.

luukvbaal: You are correct, added the `ffmpegthumbnailer` check for `sxiv` case.
The reason for passing `0/1` to `make_thumbs` btw is that `sxiv` creates audio thumbnails by itself, but not video, whereas `ucollage` creates video thumbnails, but not audio🤣

jarun: Yes, we should detect read-only and work accordingly. Maybe try /tmp as fallback else jist show the image.

jarun: Please ping once this is done and we can have this in.

plugins/imgview
```diff
@@ -27,47 +28,50 @@ exit_prompt() {
     exit
 }
 
-nthumb_cleanup() {
```
luukvbaal, line 30: Get rid of cleanup function since we are in `$TMDIR` or `$XDG_CACHE_HOME`

plugins/imgview
```diff
@@ -16,8 +15,10 @@
 #
 # Shell: POSIX compliant
 # Author: Arun Prakash Jana, Luuk van Baal
-
-target="$(readlink -f "$1")"
+#
+# Consider setting NNN_PREVIEWDIR to $XDG_CACHE_HOME/nnn/previews
+# if you want to keep media thumbnails on disk between reboots.
+NNN_PREVIEWDIR="${NNN_PREVIEWDIR:-$TMPDIR/nnn/previews}"
```
luukvbaal, line 21: Reuse `preview-tui` variable for preview directory

plugins/imgview
```diff
@@ -27,47 +28,50 @@ exit_prompt() {
...
+              ret="$NNN_PREVIEWDIR/$target.jpg" ;;
+          *) ret="$target" ;;
         esac
     fi
     for file in *; do
-        case "$(file -bL --mime-type -- "$file")" in
-          audio/*) [ "$1" -ne 0 ] && ffmpeg -i "$file" ".nthumbs/${file%%.*}.jpg" -y >/dev/null 2>&1 ;;
-          video/*) [ "$1" -ne 1 ] && ffmpegthumbnailer -i "$file" -o .nthumbs/"${file%%.*}".jpg 2> /dev/null ;;
-        esac
+        if [ ! -f "$NNN_PREVIEWDIR$PWD/$file.jpg" ]; then
```
luukvbaal, line 48: Don't regenerate previews

plugins/imgview
```diff
@@ -27,47 +28,50 @@ exit_prompt() {
...
+            case "$(file -bL --mime-type -- "$file")" in
+                audio/*) [ "$1" -ne 0 ] && ffmpeg -i "$file" "$NNN_PREVIEWDIR$PWD/$file.jpg" -y >/dev/null 2>&1 ;;
+                video/*) [ "$1" -ne 1 ] && ffmpegthumbnailer -i "$file" -o "$NNN_PREVIEWDIR$PWD/$file.jpg" 2> /dev/null ;;
+            esac
+        fi
+    done
+    for file in "$NNN_PREVIEWDIR$PWD"/*; do
```
luukvbaal, line 55: Delete files not present in source directory from preview directory

luukvbaal (commented):
Reuse `NNN_PREVIEWDIR` from `preview-tui` to store the thumbnails instead of (trying to) create a `.nthumbs` in working directory.

jarun:
@tonijarjour can you please test and confirm the changes?

tonijarjour:
@jarun Tested with [this](https://github.com/jarun/nnn/blob/5e3c90751c6ff1e9c4395c8fe2b8250395ceacd4/plugins/imgview). Using imgview on a directory fails. Works correctly on a file.

luukvbaal:
> @jarun Tested with [this](https://github.com/jarun/nnn/blob/5e3c90751c6ff1e9c4395c8fe2b8250395ceacd4/plugins/imgview). Using imgview on a directory fails. Works correctly on a file.

Could you elaborate? Seems to work fine on my end.

tonijarjour:
The problem is that when `type ffmpegthumbnailer >/dev/null 2>&1 && make_thumbs 0` fails, the `previews` folder made by `make_thumbs` never comes into existence. So `find` in `listimages` fails as well and garbage gets passed to `sxiv`.

Thats about it.

luukvbaal:
Right, thanks.

tonijarjour:
Alright tested with and without ffmpegthumbnailer. Everything works perfectly and it is much faster now. :+1:

luukvbaal:
Yeah should be good now sorry.

tonijarjour:
Wait there is a problem. `mkdir` fails with a permission denied error. (trying to make `/nnn`)

luukvbaal:
Yeah `TMPDIR` variable was missing, try now.

tonijarjour:
It isn't generating thumbnails with `ffmpegthumbnailer`

tonijarjour:
`video/*) [ "$1" -ne 1 ] && ffmpegthumbnailer -i "$file" -o "$NNN_PREVIEWDIR$dir/$file.jpg" 2> /dev/null ;;`
to
`video/*) [ "$1" -ne 1 ] && ffmpegthumbnailer -i "$file" -o "$NNN_PREVIEWDIR$file.jpg" 2> /dev/null ;;`

luukvbaal:
Thanks for testing!
I think it's good now.

jarun:
Thank you!

PR #1086, opened 2021-06-29
Persistent selection

KlzXS:
Added a patch for persistent selection markers across directory change.

It is not a smart implementation, it simply searches the selection for every entry found in a directory an sets its selected flag if found. Works surprisingly smoothly even with 100k+ selected files while browsing across directories with up to 10k files. When I tried to enter a 100k directory with 100k file selection I had to wait a few seconds. Do test it on an older system.
Where it falls apart is deselection. It searches for the path to be deselected and moves everything in front of it in the buffer back. Which means that removing older selections is more expensive than removing new ones.

It doesn't allocate any extra memory compared to master.
It does use `memmem()` which is not part of the C standard, but should be available on the majority of systems.

I was thinking of also making a version with a [trie](https://en.wikipedia.org/wiki/Trie), but now that I see that only deselection is slow (on a PC bought last year) I'm not really sure. It would use more memory and more CPU over master, but at a certain size of selection it _should_ use less CPU than this brute. But nothing is for sure until benched properly. Let me hear what you think.

Fixes #alotofissues


jarun:
> Where it falls apart is deselection

When you are working with too many files, I think the best solution is to allocate an uninitialized new buffer equal to the current one and copy the files which still stand selected. Then free the old one and use the new one as selbuf. Also set the new file append position correctly.

KlzXS:
I don't know about that. For individual files that is worse. For bulk block deselection `E` is better. I would let it be as is with the note about the behavior. 

Come to think of it there is no way to deselect something once you exit the directory without this patch. You would have to either erase everything or use `E`. I don't think this would be too big of a disturbance for anyone.

patches/persistsel/mainline.diff
```diff
@@ -0,0 +1,73 @@
...
+ 
++static char *findinsel(int len)
++{
++	if (!selbufpos)
++		return FALSE;
++
++	/* memmem(3):
++	 * This function is not specified in POSIX.1, but is present on a number of other systems.
++	 */
++	return memmem(pselbuf, selbufpos, g_buf, len);
```
jarun, line 26: How do you handle this case:

```
/tmp/new/file/hello
/new/file/hello
```

Say the first one is selected. An you enter `/new/file`.

I think either you should check the file is at the start of the selection buffer or the `startpos - 1` should have a NULL.

jarun (changes requested):

jarun:
> For individual files that is worse

If already know `nselected`. Pick a reasonable number. If less, use memcpy, else allocate a new buffer.

jarun:
> Come to think of it there is no way to deselect something once you exit the directory without this patch

Neither do we show anything selected when we directory today. Once you show, it's too tempting to deselect.

patches/persistsel/mainline.diff
KlzXS: Good catch! I think your that's just about how I'll do it.

KlzXS:
So each time I deselect a single file it is better to `memmove()` the current buffer over. We can test this, but it is.
But when I do an invert I would need to keep track of where all those paths are in the selection buffer and their length. That array of (start, length) pairs would need to be dynamic. After all of that I would need to sort all those pairs and then marge them to be able to `memcpy()` over whole blocks of memory at a time.
That doesn't sound appealing to me.

The initial plan was for this to be a brute force implementation that doesn't use extra memory while sacrificing CPU time. The trie implementation would be the one to sacrifice memory for a negligible increase of CPU time. The recommended one would be the trie. I think that might be best to pursue after all.

jarun:
When you entered the directory you already marked the files which were selected earlier.
On invert, why not make it 2 pass? In first pass find all that's marked and remove them using memmove/memcpy, invert the bits for all files and then in the next pass add the selected ones (which is pretty fast).

KlzXS:
> In first pass find all that's marked and remove them using memmove/memcpy

This is the slow part. Shifting everything after a path marked for removal is slow if the path is at the start of a long buffer. There simply is no great way to remove elements from the middle of an array.

jarun:
> This is the slow part. Shifting everything after a path marked for removal is slow if the path is at the start of a long buffer. There simply is no great way to remove elements from the middle of an array.

That's why using a new buffer would be much faster when combined with the inode check I mentioned above.

KlzXS:
Doesn't that fail across different filesystems? Or something like that. I think that's what you said the last time you were playing with inode numbers.

jarun:
> Doesn't that fail across different filesystems?

A dir inode is unique only within the filesystem. We can store the device major minor along with it. Still not too much of a memory.

KlzXS:
> using a new buffer would be much faster 

How? Now instead of moving memory from `selbuf+startpos+len` to `selbuf+startpos` you're moving it from `selbuf+startpos+len` to `newselbuf`.

> A dir inode is unique only within the filesystem. We can store the device major minor along with it. Still not too much of a memory usage.

Sure, we'll try it. Could someone create a filesystem(s) with files to test performance on? I don't really know what's the best directory structure to use for that.

KlzXS:
I'll do a double pass on inversion. That will help. But I won't allocate a new buffer, I'm not convinced that's better.

jarun:
> How? Now instead of moving memory from selbuf+startpos+len to selbuf+startpos you're moving it from selbuf+startpos+len to newselbuf.

Yes, double buffer is not required. In the first pass memmove/memcpy and remove the selected entries in the dir. Also invert the file selected bit. In the second pass append all that's marked.

jarun:
> Sure, we'll try it. Could someone create a filesystem(s) with files to test performance on? I don't really know what's the best directory structure to use for that.

I am not sure what you mean. stat() on a dir will give something like `Device: 802h/2050d`. Inode is 64/128-bit. So have fixed length fields and match when you enter a dir.

KlzXS:
> I am not sure what you mean.

I mean I don't know how to create the best directory structure for measuring performance.

Now can someone please test this double pass change. I am getting it being slower. I select 100k files (00000-99999) then 100 (00-99) with `a` then invert the 100k with `A`. I keep them in different directories.

patches/persistsel/mainline.diff
```diff
@@ -8,60 +8,155 @@
...
+ 
+@@ -1488,33 +1495,95 @@ static size_t appendslash(char *path)
  	return len;
  }
  
+-static void invertselbuf(char *path, bool toggle)
 +static char *findinsel(int len)
-+{
+ {
+-	selbufpos = lastappendpos;
```
jarun, line 42: Why do we use lastappendpos?

jarun: I think when we change a directory we remember the position of the selection buffer. If we invert we go back that position and append now selected ones.

patches/persistsel/mainline.diff
```diff
@@ -8,60 +8,155 @@
...
+ }
+ 
++/* removes g_buf from selbuf */
++static void rmfromselbuf(size_t len)
++{
++	char *found = findinsel(len);
++	if (!found)
++		return;
++
++	memmove(found, found+len, selbufpos - ((found+len) - pselbuf));
```
jarun, line 135: do we need both memmove and memmem? I think we should use only one.

patches/persistsel/mainline.diff
KlzXS: We don't use it in `findinsel()` that's git's awesome diff of a diff at work. I also removed it from `invertselbuf()` because with these changes it no longer servers a purpose (now we look at the whole buffer not just things that have changed since entering a directory). Should it be removed from elsewhere?

patches/persistsel/mainline.diff
KlzXS: Well they do different things. `memmem()` is `strstr()` but not for strings. And `memmove()` moves memory around which _can overlap_, so we can't use `memcpy()` there.

patches/persistsel/mainline.diff
jarun: I think in that case we can remove it completely.

patches/persistsel/mainline.diff
0xACE: I'd reckon it'd be even faster if you put a NULL infront of the `needle`. Though that'd have a problem with the first row...

patches/persistsel/mainline.diff
jarun: sorry, i meant memcpy. This is fine. I thought we can do a memcpy because we are _pulling_ memory up. memcpy is a bit faster.

KlzXS: `memcpy()` doesn't do overlaps, at least not officially.

> memcpy is a bit faster.

Funny thing. I looked into this a bit earlier. Turns out most of the time yes, sometimes no.

jarun:
> Now can someone please test this double pass change. I am getting it being slower.

You can try a single pass also. Move up when you find a file that's already set and append when you find one that's not set already.

jarun:
I think it's a good test to have 50% selected and 50% unselected in the same dir. You get the average performance with whatever number you take. Note that pulling up is costlier than appending. So if you have like 100 files and the top 99 are selected, when you invert you move the memory up 99 times and then append only 1 file. That's the costliest case.

jarun:
Also, note that there's something called practicality. I am totally OK if you show a msg to edit manually and block popup if total selected files are > 500.

KlzXS:
> You can try a single pass also

I was already doing that. Maybe that wrong order of operands and leftover lastappendpos had something to do with it. I'll try it again.

> I think it's a good test to have 50% selected and 50% unselected in the same dir.

I think it would be even better to do it in a zipper-like fashion. Perhaps I'll remove some limitations of selection editing to do that also. But tomorrow. I had enough of looking at 100k directories for a day. It's been 2 years since I've done that.

> I am totally OK if you show a msg to edit manually

I'll play around with it a bit more.

patches/persistsel/mainline.diff
KlzXS: That would be convenient and I do have a hack in mind. Just put a `NULL` at `pselbuf` and `++pselbuf`. Dirty, horrible, but it works. Not really sure if it's worth it.

patches/persistsel/mainline.diff
```diff
@@ -0,0 +1,186 @@
...
++}
++
+ static int entrycmp(const void *va, const void *vb)
+ {
+ 	const struct entry *pa = (pEntry)va;
+@@ -5283,6 +5360,9 @@ static int dentfill(char *path, struct entry **ppdents)
+ 			entflags = 0;
+ 		}
+ 
++		if (findinsel(mkpath(path, dentp->name, g_buf)) != NULL)
```
jarun, line 163: This is costly. To optimize, before entering the loop we should check if the path (prefix) is present or not. If not, skip this. Else check for each file.

KlzXS: That would tie in with your inode+device number suggestion. But it could also be done without that.

jarun: Extra memory vs. processing. Let's first try the memmem with the path and see.

patches/persistsel/mainline.diff
jarun: > Not really sure if it's worth it.

Nope. I am considering it for mainline. Please keep the implementation clean.

patches/persistsel/mainline.diff
jarun: > memcpy() doesn't do overlaps, at least not officially.

You are right. We need `memmove()`.

patches/persistsel/mainline.diff
```diff
@@ -0,0 +1,186 @@
+# Description: Make selection markers persistent across directory change.
+#
+# Note: This is a brute version. It simply searches for a path in the selection
+#       to determine if a file is selected. Having large selections while browsing
+#       directories with lots of files can be slow.
+#       The earlier an entry was selected the slower the deselection will be!
+#
+# Author: KlzXS
```
KlzXS, line 8: Note to @KlzXS at the end place this note back in.

KlzXS:
Note the split into `mainline-1pass.diff` and `mainline-2pass.diff`. Use `O_PERSISTSEL=1` and `O_PERSISTSEL=2` respectively. Once the better one is determined it will replace `mainline.diff` which is currently not used.

patches/persistsel/mainline.diff
jarun: Concentrate on the best possible practical implementation :). As I said I am targeting this for mainline. I would still suggest just show a msg to edit selection if nselected > some reasonable value and user wants to invert.

jarun: If user doesn't prefer to edit he can clear and re-select as well.

KlzXS: To me 10000 comes as a natural cutoff point. Considering a pretty decent 512 byte average path length it would take up 5MiB, small enough to fit into most CPU caches, but still large enough to handle big selections. My favorite is 1pass, but I'll wait for some test results to see which will come out on top.

Offer a prompt for more than 10000 "This operation might be slow. Continue? (y)es, (n)o, (e)dit manually [default: n]" perhaps worded more tersely. Best of both worlds.

jarun: Too long for a prompt ;). We'll reword. When ready I will test it on a mid-range Android I have.

pvx-:
Just a small report for me: I was using your first patch yesterday whole day without any problems or delays (apart from initial testing on a few thousand files mostly with small selections). Today I was at work whole day, now I pulled your latest patch. No difference for a few thousand files, then I tried on one of our servers with 30k files, with following result:

1pass: 5s to select all, 2s to invert to one file; when 30k are selected, it takes 5s to unselect single file, both at start and at the end of selection
2pass: no delay to select all, 5s to invert to one file, no delay to unselect single items when 30k are selected

On my home computer on test directory with 30k the times where almost the same, only inverting to one file for 1pass was almost immediate for some reason (instead of 2s). For 100k files everything that takes 5s for 30k takes around 40s, still no delay for selecting and one-file operations for 2pass.

If you give me an Android binary, I can test it on my "ancient" Motorola Photon Q ;).

jarun:
@pvx- thank you! You can definitely help us with the performance tests.

jarun:
Regarding the Android binary, you'll have to build it on your device. Install Termux and follow [this](https://gist.github.com/jarun/660fd1d5e79e1a4038de8ba5f2e1a513) to use it like another computer.

KlzXS:
> 5s to select all
> no delay to select all,

This discrepancy makes no sense to me. I didn't touch logic for selection only deselection. If you had a selection and then inverted on an empty directory than yes, with `a` no.

> For 100k files everything that takes 5s for 30k takes around 40s, still no delay for selecting and one-file operations for 2pass.

I guess that makes sense. I just realized I did a stupid and forgot another thing that can be done for 2pass to not write over memory when it's not necessary.

Does anyone actually have the famed Pentium I we are targeting lying around?

pvx-:
> Regarding the Android binary, you'll have to build it on your device. Install Termux and follow [this](https://gist.github.com/jarun/660fd1d5e79e1a4038de8ba5f2e1a513) to use it like another computer.

Thanks, I use ssh a lot in many ways :), my problem was with compilation itself :). But after a bit of searching I managed to make it work on my testing phone (Moto Z), for 500 files it works without delay. I don't feel so good about creating 10k+ files in phone memory (and AFAIK you cannot access tmpfs on android), now I have some work, but tomorrow I will try to find some old SD and use that, also test it on Photon Q :).

> > 5s to select all
> > no delay to select all,
> 
> This discrepancy makes no sense to me. I didn't touch logic for selection only deselection. If you had a selection and then inverted on an empty directory than yes, with `a` no.

Sorry, I wrote it wrong and didn't correct this part; "select all" here means selecting 1 file and then invert (on 30k dir). This is still fast on 2pass but slow on 1pass.

> Does anyone actually have the famed Pentium I we are targeting lying around?

[Motorola Photon Q](https://www.gsmarena.com/motorola_photon_q_4g_lte_xt897-4885.php) is probably slower than that ;), will try tomorrow and let you know.

pvx-:
After spending a few hours making it work (termux no longer supports Android 6, some packages had to be updated to install clang, but had unresolvable conflicts for which the only solution I found was just "upgrade termux", etc. ;) I finally tested 2pass on Photon Q, results:

3000 files:

select all: 0s
invert while all selected: 2s
invert while 1 selected: 0s
invert while all but 1 selected: 2s

10000 files:

select all: 0s
invert while all selected: 15s
invert while 1 selected: 0s
invert while all but 1 selected: 15s

If you think testing 1pass is worthwile, let me know. I will try to test any new patch of course.

Btw. compile took really long time (I didn't time it but it felt like more than a minute) and showed this warning, which I didn't get anywhere else (including Moto Z which I tried yesterday):

```
src/nnn.c:3652:11: warning: comparison of integers of different signs: 'off_t' (aka 'long') and 'uint_t' (aka 'unsigned int') [-Wsign-compare]
                if (rem == comp) {
                    ~~~ ^  ~~~~
```

KlzXS:
Based on those results I think we'll be going forward with 2pass. Thank you dedicating your to this!

The warning is not that important. Different architectures means different type widths. But yes, seems as if phones were never made to compile programs.

Kabouik:
For the record, compilation took 15 seconds on a F(x)tec Pro1 running Sailfish (SOC: msm8998) and there was no such warning.

jarun:
@pvx- compiling on phones with modern compilers is slow. Expected.

> 10000 files:
> invert while all selected: 15s

@KlzXS say "To me 10000 comes as a natural cutoff point" again. ;)

I will try to take care of the warning.

pvx-:
> Based on those results I think we'll be going forward with 2pass. Thank you dedicating your time to this!

No problem, glad to help. I just realized ;) that we are leaving for a family holiday tomorrow, so no Photon Q benchmarks for a week :). But hopefully I will still be online in the evenings and can run tests on my computer & servers; any speedups will probably be proportional.

> For the record, compilation took 15 seconds on a F(x)tec Pro1 running Sailfish (SOC: msm8998) and there was no such warning.

Pro1 is my primary phone ;), I just thought it would be too fast to be useful for testing this.

> > 10000 files:
> > invert while all selected: 15s
> 
> @KlzXS say "To me 10000 comes as a natural cutoff point" again. ;)

Unless you want to do some quick benchmark on startup ;) I think leaving the limit to the user (using env param or command option) is the best way (then you can have some ultra-safe default, like 1000, which would still be enough for most cases).

jarun:
> Unless you want to do some quick benchmark on startup

I don't understand what you mean.

> I think leaving the limit to the user

That's the last thing we'll do.

pvx-:
> > Unless you want to do some quick benchmark on startup
> 
> I don't understand what you mean.

Somehow measure/detect how fast the computer is and set the limit accordingly :). Probably overkill, unless that information could be useful in other ways.

> > I think leaving the limit to the user
> 
> That's the last thing we'll do.

I think userbase of nnn would manage :), but do what you think is best. For my use even 1000 would be enough ;).

jarun:
> Unless you want to do some quick benchmark on startup

It depends on several parameters including the CPU and RAM. We need to find something else.

> I think userbase of nnn would manage

Yes, I would like to avoid _readily configurable_ options for features like these. One should really know what she is doing before tinkering here because this varies from device to device.

@KlzXS probably have the max as 1000 in the code as a macro. And let a make variable with the same name override it.

jarun:
Also, please have it mainline. It gets rid of several compromises/assumptions in the current implementation. This can't be a patch.

KlzXS:
> It gets rid of several compromises/assumptions in the current implementation

What would those be? What needs to be changed?

jarun:
> What would those be? What needs to be changed?

You don't need to change anything. The current _mainline_ implementation has drawbacks which get fixed by this implementation:

- no way to edit the selection after you change the directory (or reload the same dir)
- can't handle the case where user revisits the same dir multiple times and selects, then inverts
- other known selection issues

Your implementation handles these cases. So let's have it mainline.

KlzXS:
Ah yes, the famous double selection of an item. Alright, will make the changes in the morning.

src/nnn.c
```diff
@@ -821,6 +834,8 @@ static int set_sort_flags(int r);
 #ifndef NOFIFO
 static void notify_fifo(bool force);
 #endif
+static int markcmp(const void *va, const void *vb);
+static int editselection(void);
```
KlzXS, line 838: Perhaps I should somehow reorder the functions to not have this here. But I do want to keep them in a somewhat sensible order as they are right now. Any suggestions?

src/nnn.c
```diff
@@ -645,6 +657,7 @@ static const char * const messages[] = {
 	"remove tmp file?",
 	"invalid key",
 	"unchanged",
+	"continue with large selection? 'y'es, 'n'o, 'e'dit",
```
KlzXS, line 660: Clear enough? Short enough?

src/nnn.c
jarun: Please move forward selection at the top with others.

src/nnn.c
jarun: Yes. If we are allowing them to continue on `y`, I think we can remove `O_LARGESEL` from `Makefile`.

jarun: Take a call. I am fine anyway. Some distro maintainer may want to set it higher.

KlzXS: I'll leave it in. Some may even want to set it lower if they expect low-power devices. Who knows.

src/nnn.c
KlzXS: You mean above the NOFIFO macro?

jarun: Sorry. I meant forward declarations.

See line number 828:

```
 828 /* Forward declarations */
 829 static void redraw(char *path);
 830 static int spawn(char *file, char *arg1, char *arg2, char *arg3, ushort_t flag);
 831 static void move_cursor(int target, int ignore_scrolloff);
 832 static char *load_input(int fd, const char *path);
 833 static int set_sort_flags(int r);
 834 #ifndef NOFIFO
 835 static void notify_fifo(bool force);
 836 #endif
 837 static int markcmp(const void *va, const void *vb);
 838 static int editselection(void);
```

jarun:
For the CI failure, I think you'll have to add a NOLINT:

    5472                                 mkpath(path, namep, buf);

jarun:
`buf` is allocated under `cfg.blkorder` check along with guard but still this false positive keeps coming.

KlzXS:
Did I add it in a wrong place?

src/nnn.c
```diff
@@ -1512,31 +1525,135 @@ static size_t appendslash(char *path)
...
-			if (toggle) { /* Toggle selection status */
-				pdents[i].flags ^= FILE_SELECTED;
-				pdents[i].flags & FILE_SELECTED ? ++nselected : --nselected;
+	if (nselected > LARGESEL) {
+		int r = get_input(messages[MSG_LARGESEL]);
+		if (r == 'n')
+			return;
+
+		if (r == 'e') {
+			/* Add all entires for manual inversion */
```
jarun, line 1559: By doing this, we are actually adding files she has never selected. I don't think we should do any processing in the selection buffer on behalf of the user.

I think we shouldn't even show the edit option. If the user wants to edit, she can always select everything and do an edit. We don't select twice now so we are good there.

She can also clear everything and start from scratch.

jarun (changes requested):

src/nnn.c
KlzXS: Well if there wasn't a large selection those files would be selected as part of the inversion. The already selected ones would be automatically removed. This just offloads the unselecting to the user in the hope that it would be faster then shuffling memory around.

The workflow is a bit awkward even if fast with `vim`. Your choice.

jarun: Let's not change user selection proactively. Just change the prompt to `this may be slow, continue?`

yes and no are imperative in similar other prompts so leave those out as well?

jarun: OR, `operation may be slow, continue?`

KlzXS:
In `dentfill()` why do we use `buf` instead of `g_buf`? Switching to the global buffer solves CI fail.

jarun:
Right. Make buf point to g_buf then.

src/nnn.c
```diff
@@ -1512,31 +1525,110 @@ static size_t appendslash(char *path)
...
 
-		for (int i = 0; i < ndents; ++i) {
-			if (toggle) { /* Toggle selection status */
-				pdents[i].flags ^= FILE_SELECTED;
-				pdents[i].flags & FILE_SELECTED ? ++nselected : --nselected;
-			}
+static void invertselbuf(char *path)
+{
+	if (!nselected) {
+		writesel(NULL, 0);
```
jarun, line 1546: If nothing is selected, we should select all, right? Why are we returning from here?

src/nnn.c
```diff
@@ -5140,7 +5240,7 @@ static int dentfill(char *path, struct entry **ppdents)
 	uchar_t entflags = 0;
 	int flags = 0;
 	struct dirent *dp;
-	char *namep, *pnb, *buf = NULL;
+	char *found, *namep, *pnb;
```
jarun, line 5243: If the usage of `found` is like a bool, use a bool.

Also, retain buf and point to g_buf. Somewhere I have seen using local reference to a global pointer saves on binary size.

src/nnn.c
```diff
@@ -1512,31 +1525,110 @@ static size_t appendslash(char *path)
...
+}
+
+/* removes g_buf from selbuf */
+static void rmfromselbuf(size_t len)
+{
+	char *found = findinsel(len);
+	if (!found)
+		return;
+
+	memmove(found, found+len, selbufpos - ((found+len) - pselbuf));
```
jarun, line 1628: Add space in both places:

    found + len

src/nnn.c
```diff
@@ -1512,31 +1525,110 @@ static size_t appendslash(char *path)
...
+		pdents[i].flags ^= FILE_SELECTED;
+		pdents[i].flags & FILE_SELECTED ? ++nselected : --nselected;
+	}
 
-			if (pdents[i].flags & FILE_SELECTED)
-				appendfpath(path,
-					len + xstrsncpy(path + len, pdents[i].name, PATH_MAX - len));
+	for (int i = 0; i < ndents; ++i){
+		/* Find no longer selected */
+		if(pdents[i].flags & FILE_SELECTED)
```
jarun, line 1567: Add space after `if`.

src/nnn.c
```diff
@@ -1512,31 +1525,110 @@ static size_t appendslash(char *path)
...
+		writesel(NULL, 0);
+		return;
+	}
+
+	if (nselected > LARGESEL && !xconfirm(get_input(messages[MSG_LARGESEL])))
+		return;
+
+	size_t len, endpos, offset = 0;
+	char *found;
+
```
jarun, line 1555: Redundant newline.

jarun (changes requested):

src/nnn.c
KlzXS: Because it was written weirdly on master. Okay these are the last changes from me for... well it's almost not night anymore.

> Somewhere I have seen using local reference to a global pointer saves on binary size.

Now that I think about it, yes, that makes the tiniest bit of sense.

src/nnn.c
jarun: I think you missed this one.

src/nnn.c
```diff
@@ -1512,31 +1525,104 @@ static size_t appendslash(char *path)
...
+			marked[nmarked].len = len;
+			++nmarked;
 		}
+	}
 
-		if (len > 1)
-			--len;
-		path[len] = '\0';
+	/* Merge non-selected into blocks */
+	qsort(marked, nmarked, sizeof(selmark), &markcmp);
```
jarun, line 1575: Why can't we move `markcmp()` above this function?

jarun: And what are we trying to do here with `qsort`? The entries in the directory are already selected in some order.

src/nnn.c
```diff
@@ -1512,31 +1525,104 @@ static size_t appendslash(char *path)
...
+		if(pdents[i].flags & FILE_SELECTED)
+			continue;
+
+		len = mkpath(path, pdents[i].name, g_buf);
+		found = findinsel(len);
 
-			if (pdents[i].flags & FILE_SELECTED)
-				appendfpath(path,
-					len + xstrsncpy(path + len, pdents[i].name, PATH_MAX - len));
+		if (found) {
```
jarun, line 1567: What are we doing here? Why is this needed?

src/nnn.c
```diff
@@ -1512,31 +1525,104 @@ static size_t appendslash(char *path)
...
+		return;
+
+	size_t len, endpos, offset = 0;
+	char *found;
+	int nmarked = 0, prev = 0;
+	selmark *marked = malloc(nselected * sizeof(selmark));
+
+	for (int i = 0; i < ndents; ++i) {
+		 /* Toggle selection status */
+		pdents[i].flags ^= FILE_SELECTED;
```
jarun, line 1555: Why can't we check and memmove here? Weren't you planning for 1-pass?

src/nnn.c
jarun: This file was marked selected in your earlier loop. Which means it wasn't selected before this and shouldn't be present in selection. Why are we searching for it in selection?

src/nnn.c
```diff
@@ -1512,31 +1525,104 @@ static size_t appendslash(char *path)
...
-			--len;
-		path[len] = '\0';
+	/* Merge non-selected into blocks */
+	qsort(marked, nmarked, sizeof(selmark), &markcmp);
 
-		nselected ? writesel(pselbuf, selbufpos - 1) : writesel(NULL, 0);
-	} else
-		writesel(NULL, 0);
+	for (int i = 1; i < nmarked; ++i) {
+		if (marked[i].startpos == marked[prev].startpos + marked[prev].len)
```
jarun, line 1578: Please explain what we are doing here.

src/nnn.c
jarun: Also, we'll be doing all this even if only 1 file is selected and we want to invert.

src/nnn.c
```diff
@@ -1512,31 +1525,104 @@ static size_t appendslash(char *path)
...
+		memmove(found, found + len, endpos - (found + len - pselbuf));
+		offset += len;
+	}
+
+	selbufpos -= offset;
+
+	free(marked);
+
+	/* Add newly selected */
+	for (int i = 0; i < ndents; ++i) {
```
jarun, line 1604: We have 5 loops and a qsort in this function. This is performant??? Just because we are running on modern hardware we think we are fast. We need to optimize much more.

src/nnn.c
```diff
@@ -1512,31 +1525,104 @@ static size_t appendslash(char *path)
...
+	selmark *marked = malloc(nselected * sizeof(selmark));
+
+	for (int i = 0; i < ndents; ++i) {
+		 /* Toggle selection status */
+		pdents[i].flags ^= FILE_SELECTED;
+		pdents[i].flags & FILE_SELECTED ? ++nselected : --nselected;
+	}
+
+	for (int i = 0; i < ndents; ++i){
+		/* Find no longer selected */
```
jarun, line 1560: Is this comment correct? You entered this function with some files selected, inverted the flag. So the files selected earlier would have their flags not selected anymore. I don't see how this goes with the next line.

src/nnn.c
KlzXS: We can move it up. I just placed it next to the similar function for entry comparison. Perhaps to later also use the custom qsort.

We are sorting it because the order the files that were previously selected (non became unselected) can be in arbitrary order.

To optimize moving memory around it is best to group all chunks of no longer selected memory buffer together to do a minimal number of `memmove()`s. For that we need to sort.

src/nnn.c
KlzXS: When we are moving memory back to fill the no longer selected gaps we only want to move selected chunks. This ensures we are only moving around each selected chunk only once.

Maybe I should drawn this whole process?

jarun: We need more comments.

src/nnn.c
KlzXS: The forst two can be merged. The others can't.

> This is performant???

Well you've seen how it compares to 1pass which is basicaly master with `foundinsel()` and `memmove()`.
If you can think of something better than 1 pass to invert and mark all no longer selected chunks of memory, 1 qsort to sort _marked_, 1 pass _through the marked items_ to merge them, 1 pass to move memory back never moving the same memory twice and finally 1 pass to add all the selected files to the end.

src/nnn.c
KlzXS: There's a continue under, skip the ones that were unselected, now selected. We are skipping currently selected.

jarun: Please remove the continue and do the processing only for unselected. I missed that. If you do not intend to process anything else, just do that if the condition check passes.

src/nnn.c
jarun: I think the first 2 loops in line 1553 and 1559 can be combined.

src/nnn.c
jarun: I think the first 2 loops in line 1553 and 1559 can be combined.

src/nnn.c
jarun: Yes, please move it up.

src/nnn.c
KlzXS: I think you read wrong what this loop does. It only runs for files which we are undelecting.

To unselect them we need to know _where_ they were in the selection. So yes I guess that it should never fail.

KlzXS: Removed the check.

src/nnn.c
```diff
@@ -1550,30 +1558,34 @@ static void invertselbuf(char *path)
...
+	/* Files marked for deselection could be found in arbitrary order.
+	 * Sort by appearance in selection buffer.
+	 * With entries sorted we can merge adjacent ones allowing us to
+	 * move them in a single go.
+	 */
 	qsort(marked, nmarked, sizeof(selmark), &markcmp);
 
+	/* Some files might be adjacent. Merge them into a single entry */
 	for (int i = 1; i < nmarked; ++i) {
 		if (marked[i].startpos == marked[prev].startpos + marked[prev].len)
```
jarun, line 1590: Where are we incrementing `prev`?

src/nnn.c
```diff
@@ -1550,30 +1558,34 @@ static void invertselbuf(char *path)
 	int nmarked = 0, prev = 0;
 	selmark *marked = malloc(nselected * sizeof(selmark));
 
+	/* First pass: inversion */
 	for (int i = 0; i < ndents; ++i) {
 		 /* Toggle selection status */
 		pdents[i].flags ^= FILE_SELECTED;
 		pdents[i].flags & FILE_SELECTED ? ++nselected : --nselected;
```
jarun, line 1565: You can combine this with the next condition check.

src/nnn.c
```diff
@@ -645,6 +657,7 @@ static const char * const messages[] = {
 	"remove tmp file?",
 	"invalid key",
 	"unchanged",
+	"operation may be slow, continue?",
```
jarun, line 660: Instead of `operation` maybe we should say `inversion`? We are showing this at only one place.

KlzXS: Sure. If there is another operation that behaves this way we can change it back.

src/nnn.c
KlzXS: Combined.

jarun:
I think we are good to go here.

@pvx- some benchmarks please.

@luukvbaal please take a look at the change if you have the time.

jarun:
I see a lockup happening every time in the following sequence:

```
1. Detail mode, go to /usr/bin
2. Select some files
3. Press `Up`
4. Select some files in /usr
5. Press `Up`
```

jarun:
Might be a lockup in `dentfill()` -> `findinsel()`. Please have a look.

KlzXS:
It must be there. I'll check it when I'm home.

jarun:
Sure!

pvx-:
Inverting speed test for current code (on my computer and server, again no significant difference):

30k : 2s  (was 5s)
50k : 8s
100k: 36s (was 40s)

I tested 50k when I saw the speedup for 30k, but not really for 100k; it seems for smaller (if you can call 30+k that ;) number it speeded up, but slows down more when going into really high amounts - which is OK IMHO, probably very few people will make use of 50+k selections.

The small hotel we are in has really poor internet connection, so I ran all tests more times than usual; it should be reasonably reliable. I will test on Photon Q as soon as we return home (probably Saturday night).

Also I suggest replacing the question (may be slow, continue?) with some progress message (like 'Inverting ...') after confirm, now the whole time just the question is shown and I wasn't even sure it registered the answer the first time :).

KlzXS:
> I see a lockup happening every time in the following sequence:

Fixed the infinite loop by adding an infinite loop.

KlzXS:
> but slows down more when going into really high amounts

By this you mean the bigger the selection the less significant the speed up, right? Which is expected, after we can no longer fit in cache everything should become slower.

> Also I suggest replacing the question (may be slow, continue?) with some progress message (like 'Inverting ...') after confirm, now the whole time just the question is shown and I wasn't even sure it registered the answer the first time :).

Maybe do both, but asking is still recommended. There is no way to cancel the operation as it currently stands and even if there were the only way to go back to a consistent state is to clear selection without eating up even more time. Which is not exactly ideal.

pvx-:
> By this you mean the bigger the selection the less significant the speed up, right? Which is expected, after we can no longer fit in cache everything should become slower.

Yes, exactly; thanks for the explanation.

> > after confirm
> 
> Oh... I should have read that more closely.

I should have probably worded it more clearly :). The point is to show the user his confirmation was acknowledged and something is going on (because of course no input has any effect while invert is running).

luukvbaal:
Haven't had time to fully check it out yet but I think you need to take another look at your fix for the infinite while loop. Seems to me that https://github.com/jarun/nnn/pull/1086/commits/40321eb921de29d66f98928fd53648f33c40056f is the latest commit that works properly.

Selection flag is not set for selected entries in commit https://github.com/jarun/nnn/pull/1086/commits/35b21cb1fc7143e9b934d56c72142e8c420a908b after changing directories.

jarun:
@KlzXS when I switch to another context and return back to the earlier one, the selection markers are lost. The selection is still there.

jarun:
@luukvbaal @0xACE @pvx- this needs much more testing. Need your help.

KlzXS:
Thanks for the feedback both of you. I'll get to it as soon as I have some time. There is an off-by-one error that I seem to have fixed in the wrong direction.

KlzXS:
> when I switch to another context and return back to the earlier one,

I can't check this out now, had a pretty annoying headache the whole day, but does this have anything to do with the way contexts are redrawn?

Do we call dentfill on context switch?

jarun:
No hurry. This is going to take time anyway. I am using the latest patch for my regular file management already.

Yes, when we switch the context we load the target context's path and go to begin. And we don't save any snapshot either. So `dentfill()` must be called.

pvx-:
> @luukvbaal @0xACE @pvx- this needs much more testing. Need your help.

I noticed the issue as well - selection marks are lost, as in non-patched nnn - I didn't post because I couldn't add anything more; if there is something else useful to test, just let me know.

jarun:
Just keep using it in your regular workflows. Be careful with deletion! ;)

0xACE:
> ... @0xACE ... this needs much more testing. Need your help.

This thread grew fast, and I'm not sure i understand what the current status is. anyhow checked out KlzXS latest branch (patches) and `make O_LARGESEL=1000` doesnt seem to activate the changes :S

I can't promise to try it all the time (because my version is still different from `master), but I  can occasionally experiment with it...

KlzXS:
@0xACE Hmmm... it's almost as if that's the default :)

KlzXS:
> when I switch to another context and return back to the earlier one, the selection markers are lost.

This should be fixed.

> but I think you need to take another look at your fix for the infinite while loop.

@luukvbaal in what way does it not work properly? I just updated what is fed to it in `dentfill()` maybe that resolves it.

pvx-:
> > when I switch to another context and return back to the earlier one, the selection markers are lost.
> 
> This should be fixed.

I can confirm that it works like before, benchmark numbers are the same as well. I cannot do much testing now, since I'm online only for a short time in the evening, but I will as soon as I'm back home.

0xACE:
> @0xACE Hmmm... it's almost as if that's the default :)

Hahaha

Turns out this was the issue i was running into:

> > when I switch to another context and return back to the earlier one, the selection markers are lost.
> 
> This should be fixed.

Shot myself in the foot by not reading the thread, I guess it works now.

luukvbaal:
> @luukvbaal in what way does it not work properly? I just updated what is fed to it in `dentfill()` maybe that resolves it.

Yeah it's working now.
This was the behavior I was talking about:
[asciicast](https://asciinema.org/a/aWPzL5U3I3RgqVMtDd2uSGMRe)

jarun:
@KlzXS please confirm if you are done with the changes.

I am hoping other s are using it daily. I am planning to merge this by the next weekend.

src/nnn.c
```diff
@@ -5199,6 +5320,9 @@ static int dentfill(char *path, struct entry **ppdents)
 	}
 #endif
 
+	/* We don't include NULL since we only want to know if dir path is present */
+	found = findinsel(xstrsncpy(buf, path, xstrlen(path) + 1) - 1) != NULL;
```
jarun, line 5324: Can you please append a `/` to the path and check? That will ensure something inside the dir is actually selected and not just the dir.

I understand this still doesn't guard against the parent tree directories matching e.g. `/usr/bin/` matching `/usr/bin/some/file`. But still it's a bit better.

jarun (changes requested):

KlzXS:
This is it featurewise. If you find any problems that's the only changes I planned on making.

src/nnn.c
jarun: I didn't get it. I am asking this specifically for `dentfill()` - instead of searching for `/usr/bin`, if we search for `/usr/bin/` we can ensure that some file within the directory is in fact selected and not the directory itself.

jarun: The intention is to avoid the case where the directory itself is selected and no file within it is selected. Currently even in that case `found` will be `TRUE` and we will scan the selection buffer for every file within it which we can avoid.

jarun: At this point in `dentfill()`, `path` can only be a directory.

jarun: Something in the lines of:

```c
        /* Append a '/' to dir path to ensure files within are selected */
        off = xstrsncpy(buf, path, PATH_MAX);
        buf[off - 1] = '/';
        found = findinsel(off) != NULL;
```

KlzXS: That is brilliant. But I like this:

```c
off = xstrsncpy(buf, path, xstrlen(path) + 1);
if (buf[1] != '\0') /* path should always be at least two bytes (including NULL) */
	buf[off - 1] = '/';
found = findinsel(off) != NULL;
off = 0;
```

src/nnn.c
```diff
@@ -5199,6 +5320,15 @@ static int dentfill(char *path, struct entry **ppdents)
 	}
 #endif
 
+	/* We don't include NULL since we only want to know if dir path is present */
+	off = xstrsncpy(buf, path, xstrlen(path) + 1);
```
jarun, line 5324: No need to do `xstrlen(path) + 1`. Please pass `PATH_MAX`.

src/nnn.c
```diff
@@ -5199,6 +5320,15 @@ static int dentfill(char *path, struct entry **ppdents)
...
+	if (buf[1] != '\0') /* path should always be at least two bytes (including NULL) */
+		buf[off - 1] = '/';
+	else
+		--off;
+	found = findinsel(off) != NULL;
```
jarun, line 5329: We can directly set `found = TRUE` for `/`. No need to call `findinsel()`. Hence better to set `found = TRUE` at declaration.

src/nnn.c
jarun: And also if we detect `/` we can skip the `xstrsncpy()` call as well.

I think (`buf[1] != '\0'`) is enough to check for `/`. We always pass absolute path to `dentfill()`.

jarun: `path[1] != '\0'` - pre-copy check.

src/nnn.c
jarun: So we can have:

```c
5265         bool found = TRUE;
...
5268         size_t off, namebuflen = NAMEBUF_INCR;
...
5323         if (path[1] != '\0') { /* Skip checking for `/` which will always be present in absolute path */
5324                 /* Append a '/' to dir path to ensure files within are selected */
5325                 off = xstrsncpy(buf, path, PATH_MAX);
5326                 buf[off - 1] = '/';
5327                 found = findinsel(off) != NULL;
5328         }
5329 
5330         off = 0;
```

KlzXS: I did exactly that sans the comments.

src/nnn.c
```diff
@@ -5199,6 +5320,13 @@ static int dentfill(char *path, struct entry **ppdents)
 	}
 #endif
 
+	if (buf[1] != '\0') { /* path should always be at least two bytes (including NULL) */
```
jarun, line 5323: This should be `path` nt `buf` as we haven't copied `path` to `buf` yet.

src/nnn.c
```diff
@@ -5199,6 +5320,13 @@ static int dentfill(char *path, struct entry **ppdents)
...
+		off = xstrsncpy(buf, path, PATH_MAX);
+		buf[off - 1] = '/';
+		found = findinsel(off) != NULL;
+		off = 0;
```
jarun, line 5327: I wanted to avoid setting the value of `off` to 0 at declaration because in most cases we will not be in `/`. So we'll have to use `off` and reset the value anyway.

KlzXS: I don't get it. As far as I can see `off` is only ever modified inside the loop. Before it it's always 0. This all happens right before the loop so using it and resetting it should have no effect on it.

jarun:
```c
size_t off = 0; /* 2 operations: make space for a variable in stack and assign 0 to it */
size_t off = 0; /* 1 operation: make space for a variable in stack, assigns nothing */
```

You enter the if condition in non-`/` cases which is 99% of the time. So 99% of the time you know you will be resetting `off` to 0 i.e. assigning 0 it post declaration.

So why do you still want to do the assignment at the time of declaration?

KlzXS: I see. I'll remove it if you wish. But you just made me want to fire up a disassembler just to see if it really matters in the end.

I'll back back home in a couple of hours.

jarun: Try https://godbolt.org/. You'll see the extra `mov`.

#### No assignment

```c
void f(void) {
    int i;
}
```

```asm
f():
        push    rbp
        mov     rbp, rsp
        nop
        pop     rbp
        ret
```

#### Assignemnt

```c
void f(void) {
    int i = 0;
}
```

```asm
f():
        push    rbp
        mov     rbp, rsp
        mov     DWORD PTR [rbp-4], 0
        nop
        pop     rbp
        ret
```

jarun: In fact we should do the same with `found` as well.

src/nnn.c
```diff
@@ -5199,6 +5320,13 @@ static int dentfill(char *path, struct entry **ppdents)
 	}
 #endif
 
+	if (path[1] != '\0') { /* path should always be at least two bytes (including NULL) */
```
jarun, line 5323: if (path[1])

src/nnn.c
KlzXS: I know for a fact that it can do better than that.

```asm
xor rax, rax
push rax
```

Is better in every way. But fine, we can do it that way.

jarun:
BTW @KlzXS this is an epic change in `nnn`. ;)

KlzXS:
It is a nice one and a highly requested one at that. I don't know if we received a more requested feature.
That being said you were opposed to it for a long time. What changed? I was fine with keeping in a patch.

jarun:
> That being said you were opposed to it for a long time. What changed?

The most important reason is - I couldn't decide on a performant implementation and I didn't have the time to make changes and benchmark it myself (thanks @pvx- for that).

Secondly, given how frequently we make changes in `nnn` I had to hold till the time is favourable. This is it. The team has enough time to concentrate on a much-requested but complex feature... and stabilize it through tests. We understand what we are doing and now it's done the way it should be.

I can't say I didn't want it... but I was also worried how long it would take to contain regressions. But we are there! Thanks all for your invaluable contribution!

Coincidentally England too made history after 59 years just now! In case you are watching Euro 2020.

src/nnn.c
0xACE: FYI: gcc `-O0` only adds this single instruction for the initialization `mov QWORD PTR -576[rbp], 0` on `-O{1,2,3}` that initialization is optimized out.

jarun: In fact @KlzXS is also correct at `-O3`:

```
xor rax, rax
push rax
```

But it's always better to write self-documented code.

src/nnn.c
```diff
@@ -1520,32 +1532,143 @@ static size_t appendslash(char *path)
...
 {
-	selbufpos = lastappendpos;
+	if (!selbufpos)
+		return FALSE;
 
-	if (toggle || nselected) {
-		size_t len = appendslash(path);
+	char *found = pselbuf;
+	while (1) {
+		/* memmem(3):
```
jarun, line 1542: Please use the following format for multi-line comments:

```c
/*
 * comment line 1
 * comment line 2
 * /
```

src/nnn.c
```diff
@@ -1520,32 +1532,143 @@ static size_t appendslash(char *path)
...
+			--nselected;
+			offset += len; /* buffer size adjustment */
+		} else
+			++nselected;
+	}
 
-			if (pdents[i].flags & FILE_SELECTED)
-				appendfpath(path,
-					len + xstrsncpy(path + len, pdents[i].name, PATH_MAX - len));
+	/* Files marked for deselection could be found in arbitrary order.
```
jarun, line 1598: Please use the following format for multi-line comments:

```c
/*
 * comment line 1
 * comment line 2
 * /
```

src/nnn.c
```diff
@@ -1520,32 +1532,143 @@ static size_t appendslash(char *path)
...
+		if (marked[i].startpos == marked[prev].startpos + marked[prev].len)
+			marked[prev].len += marked[i].len;
+		else {
+			++prev;
+			marked[prev].startpos = marked[i].startpos;
+			marked[prev].len = marked[i].len;
 		}
+	}
+
+	/* Number of entries is increased by encountering a non-adjacent entry
```
jarun, line 1616: Please use the following format for multi-line comments:

```c
/*
 * comment line 1
 * comment line 2
 * /
```

src/nnn.c
```diff
@@ -1520,32 +1532,143 @@ static size_t appendslash(char *path)
...
 
-		for (int i = 0; i < ndents; ++i) {
-			if (toggle) { /* Toggle selection status */
-				pdents[i].flags ^= FILE_SELECTED;
-				pdents[i].flags & FILE_SELECTED ? ++nselected : --nselected;
-			}
+		if (found == pselbuf || *(found - 1) == '\0')
+			return found;
+
+		/* We found g_buf as a substring of a path. Move forward */
```
jarun, line 1552: /* We found g_buf as a substring of a path, move forward */

src/nnn.c
```diff
@@ -1520,32 +1532,143 @@ static size_t appendslash(char *path)
...
+		 */
+		found = marked[i].startpos;
+		endpos = (i + 1 == nmarked ? selbufpos : marked[i + 1].startpos - pselbuf);
+		len = marked[i].len;
+
+		/* Move back only selected entries. No selected memory is moved twice */
+		memmove(found, found + len, endpos - (found + len - pselbuf));
+	}
+
+	/* buffer size adjustment */
```
jarun, line 1641: /* Buffer size adjustment */

src/nnn.c
```diff
@@ -1520,32 +1532,143 @@ static size_t appendslash(char *path)
...
+		nmarked = prev + 1;
+
+	/* Using merged entries remove unselected chunks from selection buffer */
+	for (int i = 0; i < nmarked; ++i) {
+		/*
+		 * found: points to where the current block starts
+		 *        variable is recycled from previous for readability
+		 * endpos: points to where the the next block starts
+		 *         area between the end of current block (found + len)
+		 *         and endpos is selected entries. This is what we are
```
jarun, line 1630: Redundant space at the end of the line.

src/nnn.c
jarun: Also, there is a redundant space at the end of the line.

jarun (changes requested):

jarun:
Please update the man page section on selection as well.

KlzXS:
```man
.Nm
clears the selection after an operation with the selection. Plugins are allowed
to define the behaviour individually.
.Pp
.Nm
doesn't match directory entries for selected files after a redraw or after the
user navigates away from the directory. An attempt to do so will increase
memory consumption and processing significantly as
.Nm
allows selection across directories. So the selection marks are cleared. The
selection can still be edited in the same instance.
```

Should I just remove the whole thing or explicitly state that this is allowed now?

nnn.1
```diff
@@ -293,12 +293,12 @@ clears the selection after an operation with the selection. Plugins are allowed
...
 .Nm
-doesn't match directory entries for selected files after a redraw or after the
-user navigates away from the directory. An attempt to do so will increase
-memory consumption and processing significantly as
-.Nm
-allows selection across directories. So the selection marks are cleared. The
-selection can still be edited in the same instance.
+matches directory entries for selected files after a redraw or after the
+user navigates away from the directory. Keep in mind that the larger the
+selection the slower operations on it will be, including entering a directory.
```
KlzXS, line 298: How true is this line? I mean they are slower, but just how much. The only time I experience slowdowns is when entering a directory is at 100000, at 10000 there is a delay, but in reality it's about as long as the startup time of ranger.

And most other selection operations are either bound by the shell, disk speeds, editor performance or some combination of those.

jarun:
> Should I just remove the whole thing or explicitly state that this is allowed now?

Just remove it. Note that the Wiki section would also change now.

nnn.1
jarun: Remove these as well. I think our prompt does the job of warning about possibly slow inversion now. Rest isn't required anymore.

nnn.1
```diff
@@ -292,14 +292,6 @@ use the selection in the other pane.
...
 .Pp
-.Nm
-matches directory entries for selected files after a redraw or after the
-user navigates away from the directory. Keep in mind that the larger the
-selection the slower operations on it will be, including entering a directory.
-If the selection exceeds a 1000 entries the user will be prompted to proceed
-before inverting entries in the current directory. No such prompt is provided
-when deselecting a single entry.
-.Pp
 To edit the selection use the _edit selection_ key. Use this key to remove a
```
jarun, line 295: > Use this key to remove a
> file from selection after you navigate away from its directory or to remove
> duplicates.

I think we should remove this.

> You can add more files to
> the selection and edit the list again.

You mentioned somewhere that this doesn't work. In that case we can remove this as well.

KlzXS: > I think we should remove this.

I think so too.

> You mentioned somewhere that this doesn't work. In that case we can remove this as well.

It got you too I see. It's saying that you should exit the edit, select more and then go back. Perhaps a bit poorly worded, but that's what it's trying to say.

pvx-: > > Keep in mind that the larger the selection the slower operations on it will be, including entering a directory.

> How true is this line? I mean they are slower, but just how much

I did a quick benchmark for entering directory with selected files on my computer and it seems to be roughly the same as inverting (at least for all files selected):

10k files, all selected: <0.5s
30k files, all selected: 2s
100k files, all selected: 37s
100k files, 10k of them selected: 2s

So maybe leaving a line about slower operations for large selections is not a bad idea.

I would also like to repeat my suggestion to let the user know that something is going on - now after positive answer to "inversion may be slow" the question is displayed all the time until inverting finishes which may be very confusing. Just replace it with some "in progress" text (like "Handling selection...") then - maybe show this as well when entering directory with selected items when selection is larger than MAX, because now nnn looks "frozen" during that time :).

KlzXS: > So maybe leaving a line about slower operations for large selections is not a bad idea.

Nah. Truth be told I don't expect anyone to browse around with 100k files. I think you just select them in that dir and go to town on them. And these other times are way better than just acceptable.

> I would also like to repeat my suggestion

Oh right that. @jarun we haven't addressed this. I think it would be a nice touch.

jarun: > Oh right that. @jarun we haven't addressed this. I think it would be a nice touch.

:+1:

jarun: I am exclusively using the patched version for some days now and it seems stable.
Please wrap up the `in progress...` message so it's in a state to release to users for wider usage.

jarun: For the quick msg see `dirwalk()`.

src/nnn.c
```diff
@@ -5199,6 +5323,15 @@ static int dentfill(char *path, struct entry **ppdents)
 	}
 #endif
 
+	if (path[1]) { /* path should always be at least two bytes (including NULL) */
+		off = xstrsncpy(buf, path, PATH_MAX);
+		buf[off - 1] = '/';
+		found = findinsel(off) != NULL;
```
jarun, line 5329: Another scope for a good optimization here. If found is `TRUE`, we can remember the location at which we found the first occurance of the path and do the subsequent searches from there instead of the start of the selection buffer, In the most common workflow all files in a directory will be selected together and so we can save a lot on the subsequent `find()`s because they will be found towards near this offset.

jarun (changes requested):

src/nnn.c
jarun: We may have to pass the address to start checking from to `findinsel()` for this. But I think that's an improvement. If NULL is passed, use selbuf, else use the passed address. The length of the memory to be scanned should also need to be adjusted... I think that would be `addr - selbuf`.

KlzXS: It can be done.

jarun:
@luukvbaal please confirm the patches are still working once I merge this.

src/nnn.c
KlzXS: Unfortunately files in a directory are listed in a n arbitrary order so we can only start from the first appearance of the dir. Too bad.

luukvbaal:
> @luukvbaal please confirm the patches are still working once I merge this.

Sure, current PR has no conflicts with the patches.

src/nnn.c
```diff
@@ -1576,6 +1580,10 @@ static void invertselbuf(char *path)
 	int nmarked = 0, prev = 0;
 	selmark *marked = malloc(nselected * sizeof(selmark));
 
+	tolastln();
+	addstr(" in progress...\n");
```
jarun, line 1584: Please drop the space in the beginning both here and in dirwalk msg.
Also, please make the string `processing...`.

src/nnn.c
```diff
@@ -1532,27 +1532,31 @@ static size_t appendslash(char *path)
...
 {
 	if (!selbufpos)
 		return FALSE;
 
-	char *found = pselbuf;
+	if (!startpos)
+		startpos = pselbuf;
+
+	char *found = startpos;
+	size_t buflen = (selbuflen + pselbuf) - startpos;
```
jarun, line 1544: Please make it:

    size_t buflen = selbuflen - (startpos - pselbuf);

And add a newline below.

jarun (changes requested):

src/nnn.c
jarun: Still better than beginning at the start of the selection buffer. :+1:

jarun:
Thank you!

jarun:
@KlzXS I see the following issues.

My start dir is `nnn` directory. I have 1662 files in `/usr/bin`.

#### Issue 1

1. Go to `plugins` dir and select some files with <kbd>^J</kbd>
2. Go to `/usr/bin` dir and select some files with <kbd>^J</kbd>
3. Press <kbd>a</kbd>
4. Press <kbd>A</kbd>
5. Press <kbd>y</kbd> to continue
6. Go to parent `/usr`
7. Go back to `/usr/bin`

Observation: all files are selected.

#### Issue 2

1. Go to `plugins` dir and select a file with <kbd>^J</kbd>
2. Go to `/usr/bin` dir and select the first file with <kbd>^J</kbd>
3. Go back to `nnn` dir with <kbd>@</kbd>
4. Go back to `/usr/bin` dir with <kbd>-</kbd>
5. Go to the end of the dir and select the last file with <kbd>^J</kbd>
6. Press <kbd>a</kbd>
7. Press <kbd>A</kbd>
8. Press <kbd>y</kbd> to continue

Observation: `nnn` quits.

Please take a look.

pvx-:
Finally home, I will be testing more, but just a quick post that something is indeed wrong with multi-dir select (I don't think it was before, but I rarely used it) - even without inversion just randomly selecting files in different dirs ends up with different files being shown as selected than there are in selection file; I will try to find a simple step-by-step to check.

I would suggest dumping the whole selection (memory buffer) into debug file (or some extra file) on each selection change for O_DEBUG=1, that should help to find the problem (and any other :).

Btw. it also seems to be a bit slower again, selecting 100k now consistently takes over 41s (was 40s in original 2step and 36s last time), selecting 30k is also slower by a few tenths.

KlzXS:
@pvx- we can continue this in #1104 

But both of you experience these issue on the latest commit right? Are they present on c94e6cd77810b167ee79c9, I'll take a look at it a bit later today myself.

pvx-:
I tested the latest version on Photon Q and the invert times were almost exact the same as before (2s for 3k, 15s for 10k). Entering a directory with this many selected files takes roughly the same time. Btw. I timed the compilation this time and it was a lot more than one minute, almost 4 minutes in fact ;).

Given that entering directory with selected files takes roughly as much time as inverting such selection I would suggest removing the "slow, continue?" question (which would be annoying for someone with fast computer doing a lot of those) and just showing a message (like "selection update in progress...") whenever this happens. Someone using large selection on a slow computer would quicky find out and most people will never even notice ;).

But as I wrote multiple times now, I personally rarely use more than 1000 selected files and I don't remember inverting such number ever (apart from testing ;)).

jarun:
> Entering a directory with this many selected files takes roughly the same time.

How much time does entering the directories without anything selected take in each case?

pvx-:
> > Entering a directory with this many selected files takes roughly the same time.
> 
> How much time does entering the directories without anything selected take in each case?

When entering directory without any selected files, there is no noticeable delay (even on Photon Q, which sometimes takes a few tenths to enter a directory without any selection at all :). Even if you have a huge selection in one directory and small in other, entering the directory with small selection is fast.

Only when you enter a directory with lots of files selected in it the slowdown occurs; as I wrote above, when all files are selected, entering directory is roughly as slow as inverting full selection, if a big portion is selected, it slows down (on my computer):

10k selected out of 100k: 2s
20k selected out of 100k: 9s
30k selected out of 100k: 16s

These are all the first N files selected, if I selected the *last* 10k out of 100k, it took 9s. It also depends on the length of path, tests above are in /tmp/test, when tested with 200-character long path, first 10k out of 100k took 6s, last 10k out of 100k took 16s.

To be clear, all this times are how long it takes from pressing `l` on a directory until its contents is displayed.

jarun:
Got it. Yes, I think we can drop the prompt.

sikmir:
> It does use memmem() which is not part of the C standard, but should be available on the majority of systems.

`memmem` is not available on Haiku, so nnn-4.2 is broken on Haiku now.

jarun:
https://review.haiku-os.org/c/haiku/+/61

If it's not there. Haiku would have 4.1 as the last available till the devs add it to Haiku at some point.

sikmir:
> https://review.haiku-os.org/c/haiku/+/61
> 
> If it's not there. Haiku would have 4.1 as the last available till the devs add it to Haiku at some point.

Oh, I see. `memmem` is alread there, we just need `LDFLAGS = -lgnu` for nnn build. See https://github.com/haikuports/haikuports/pull/6121.

jarun:
Thanks! Please raise the PR.

PR #1091, opened 2021-07-02
Move adjust_cols and no else braces

luukvbaal:
Move `adjust_cols()` call down a few lines in `redraw()` such that the optimization from https://github.com/jarun/nnn/commit/513b4f24a69ae53ee68053e50c701fe93b5593e4 can be used in the namefirst patch, is this fine?

Also remove braces from single line else statement in `adjust_cols()`.

### DIFF
--- a/patches/namefirst/mainline.diff
+++ b/patches/namefirst/mainline.diff
@@ -4,10 +4,10 @@
 # Author: Luuk van Baal
 
 diff --git a/src/nnn.c b/src/nnn.c
-index eabe30f0..d8dc5925 100644
+index c4bced5..d1bc16d 100644
 --- a/src/nnn.c
 +++ b/src/nnn.c
-@@ -363,6 +363,10 @@ typedef struct {
+@@ -368,6 +368,10 @@ typedef struct {
  } session_header_t;
  #endif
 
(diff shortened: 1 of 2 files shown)


jarun:
I don't know really. I thought of it a few times myself but left it there because adjust_cols() and redraw() aren't the areas I like to revisit often.

jarun:
I think with sessions it may be relevant.

luukvbaal:
Hmm, okay. Not sure what sessions have got to do with it to be honest.

I just noticed the for loop comparison had changed from when I first wrote the namefirst patch and wanted to adopt that change to the namefirst for loop. I just wanted to move the `adjust_cols()` call down a few lines to below the `onscreen` assignment such that the namefirst patch change could be applied more cleanly.

Should I close this then?

jarun:
Sorry, I got confused with `move_cursor()`. I am not sure why you would even ask moving `adjust_cols()` is fine or not. It doesn't change any global.

jarun:
Thank you!

luukvbaal:
Just thought I would ask whether you would be fine with a change in mainline for patch compatibility I suppose. But as you say the change doesn't affect anything logically, sorry for the confusion.

PR #1094, opened 2021-07-05
Allow tmux inside kitty without KITTY_LISTEN_ON set

### LINKED ISSUE #1093: preview-tui with kitty+tmux no longer works.
**Problem description**
If I run nnn inside tmux inside kitty and try to use the "preview-tui" pluging, I get this error:

```
$KITTY_LISTEN_ON not set!
Please read Usage in preview-tui.
```

This did not happen before.

**Possible solution**
In commit d99894311941498b9bf93c9f3ff9b55a533ed6cf this change was made:

image

The thing is that, because of the new condition inside the red rectangle, I can no longer use the preview-tui plugin when using tmux inside kitty (when I do, the error a few lines below is shown and "togglepreview" is never called).

The reason is that KITTY_WINDOW_ID is defined inside tmux.

What I did is to change the original line:
```
elif [ "$KITTY_WINDOW_ID" ] && [ -z "$KITTY_LISTEN_ON" ]; then
```
...into this:
```
elif [ "$KITTY_WINDOW_ID" ] && [ -z "$KITTY_LISTEN_ON" ] && [ -z "$TMUX" ]; then
```

Would this be an acceptable solution?

Thanks.

luukvbaal:
Fix #1093

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -412,7 +412,7 @@ else
         clear
         printf "No FIFO available! (\$NNN_FIFO='%s')\nPlease read Usage in preview-tui." "$NNN_FIFO"
         cfg=$(stty -g); stty raw -echo; head -c 1; stty "$cfg"
-    elif [ "$KITTY_WINDOW_ID" ] && [ -z "$KITTY_LISTEN_ON" ]; then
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1095, opened 2021-07-06
Remove gitstatus libgit2 dependency

luukvbaal:
Haven't been able to get rid of the size discrepancy for the merged gitstatus and namefirst binary but the same discrepancy is present in the current implementation so I thought I would submit the PR:
image

Don't know how others(@crides @CantoroMC) feel about it but I'm mostly fine with directly displaying the [`git status --porcelain`](https://git-scm.com/docs/git-status#_short_format) representation. I'm just replacing the spaces with a dash which looks slightly better in my opinion.
I don't feel there's any need to replace `!!` and `??` with `-I` and `-N` from the current implementation. It might look better (subjectively) but is also less correct as it disregards the underlying index/working tree column representation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,10 +31,6 @@ O_NOX11 := 0  # disable X11 integration
 O_GITSTATUS := 0 # add git status to detail view
 O_NAMEFIRST := 0 # print file name first, add uid and guid to detail view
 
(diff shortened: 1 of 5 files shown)


jarun:
Thank you!

PR #1111, opened 2021-07-20
pseudo-ability to run binary files for nuke

BlyatOrange:
It was becoming quite annoying to have to open the shell each time I wanted to execute a file, so I did this. This can be made to also run shell scripts and the sort (application/x-shellscript is different from text/x-shellscript), but i'm still not sure about it.  

A more elegant way of doing this would be to just create a separate plugin which would handle running things from within nnn, and then binding that to a key (like how e is bound to open in editor) or possibly have nuke use it as an opener like it does mocq.

### DIFF
--- a/plugins/nuke
+++ b/plugins/nuke
@@ -63,8 +63,9 @@
 #   1. Adapt, test and enable all mimes
 #   2. Clean-up the unnecessary exit codes
 
-# set to 1 to enable GUI apps
+# set to 1 to enable GUI apps and/or BIN execution
 GUI="${GUI:-0}"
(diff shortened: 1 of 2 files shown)


jarun:
Thank you! This is indeed a very good improvement.

jarun:
I have added a check to run executables only and force pushed.
Note that it's easy to make a file executable with <kbd>*</kbd>.

jarun:
@BlyatOrange can you add another option `a` (along with y and n) which would allow the user to add some arguments and execute the binary?

jarun:
Took care of it. Thanks!

PR #1112, opened 2021-07-23
xdgdefault plugin: add dmenu support

N-R-K:


jarun:
CI breaks. Please fix that.

Also, dmenu needs X. Please use a flag like `GUI` (see `nuke`) to enable it.

plugins/xdgdefault
```diff
@@ -36,7 +45,7 @@ app=$(find "$@" -iname '*.desktop' -exec grep '^Name=' {} + \
     | sed -E 's;.+/(.+desktop):Name=(.+);\2:\1;' \
     | sort -t ':' -k 1,1 \
     | column -t -s ':' -o "$(printf '\t')" \
-    | fzf -e --tiebreak=begin \
+    | $menu \
```
jarun, line 48: Can you confirm you have executed this with `fzf` and everything works fine?

jarun: @luukvbaal @KlzXS please review once.

jarun (changes requested):

jarun:
Also please update the requirement in the plugin README to `xdg-utils, fzf/dmenu`

plugins/xdgdefault
N-R-K: > Can you confirm you have executed this with `fzf` and everything works fine?

Yes, I have, with both fzf and dmenu.

jarun: Thanks for confirming!

jarun:
Thank you!

PR #1118, opened 2021-07-28
xdgdefault: make dmenu case-insensitive, fix style

N-R-K:
Most *.desktop entries have same name as their application name so this
is not an issue most of the time. However in the case of Neovim, the
application name is "Neovim" while the desktop entry is "nvim.desktop"

Since dmenu is case sensitive by default this means that searching
"neovim" will not show any results since the N is not capitalized and
the desktop entry name is "nvim"

fzf doesn't have this issue since its case-insensitive/fuzzy by
default. Making dmenu case-insensitive solves this.

Also fix the indentation to be consistent with the rest of the script.

### DIFF
--- a/plugins/xdgdefault
+++ b/plugins/xdgdefault
@@ -13,10 +13,10 @@
 # set to 1 to enable GUI apps
 GUI="${GUI:-0}"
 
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1119, opened 2021-07-29
fix misleading comment in Makefile

N-R-K:
reading the source code, setting NOSSN to 1 _disables_ session support,
not enable it.


jarun:
Thank you!

PR #1124, opened 2021-08-15
Add finder history/bookmarks

luukvbaal:
Modify finder plugin to record find commands in a history file and allow user to enter these by index.
Semi-solution to ToDo item "support applying filter by key (like bookmarks/plugins)".

### DIFF
--- a/plugins/finder
+++ b/plugins/finder
@@ -1,28 +1,90 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 # Description: Run custom search and list results in smart context
 #
-# Note: To enable multi select in fzf, export the following:
-#       - export FZF_DEFAULT_OPTS='--bind ctrl-a:select-all,ctrl-d:deselect-all,ctrl-t:toggle'
+# Usage:
+#   Run plugin and enter e.g. "-size +10M" to list files in current
+#   directory larger than 10M. By default entered expressions are
+#   interpreted as arguments to find. Results have to be NUL
+#   terminated which is done by default for find. Alternatively one
+#   can prepend a '$' to run a custom search program such as fd or
+#   ripgrep. Entered expressions will be saved in history file to
+#   be listed as bookmarks and and can be entered by index and edited.
 #
-# Shell: POSIX compliant
-# Author: Arun Prakash Jana
+# Shell: Bash
+# Author: Arun Prakash Jana, Luuk van Baal
+TMPDIR="${TMPDIR:-/tmp}"
+HISTFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/finderbms"
+HISTLEN=10000
 
-. "$(dirname "$0")"/.nnn-plugin-helper
+printexamples() {
+    printf ". -maxdepth 1 -name pattern
+. -maxdepth 1 -size +100M
+\$fd pattern -0
+\$fd -d 2 -S +100M -0
+\$grep -rlZ pattern
+\$rg -l0 pattern
+\$fzf -m | tr '\\\n' '\\\0'\n"
+}
 
-printf "Examples:\n"
-printf "  find . -name \"pattern*\" -print0\n"
-printf "  fd pattern -0\n"
-printf "  find -maxdepth 1 -size +100M -print0\n"
-printf "  fd -d 2 -S +100M -0\n"
-printf "  grep -rlZ pattern\n"
-printf "  rg -l0 pattern\n"
-printf "  fzf -m | tr %s %s\n\n" "'\n'" "'\0'"
+printexprs() {
+    for i in "${!fexprs[@]}"; do
+        printf '%s\t%s\n' "$((i + 1))" "${fexprs[$i]}"
+    done
+}
 
-printf "cmd: "
-read -r cmd
+mapexpr() {
+    if [ "$fexpr" -eq "$fexpr" ] 2>/dev/null; then
+        fexpr=${fexprs[$((fexpr - 1))]}
+        read -r -e -p "Search expression: " -i "$fexpr" fexpr
+    fi
+}
 
-if [ -n "$cmd" ]; then
-    printf "%s" "+l" > "$NNN_PIPE"
-    eval "$cmd" > "$NNN_PIPE"
+readexpr() {
+    case "$fexpr" in
+        [1-9]|10)
+             mapexpr
+             readexpr ;;
+        h)   clear
+             printf "Examples:\n"
+             mapfile -t fexprs < <(printexamples)
+             printexprs
+             read -r -p "Search expression or index: " fexpr
+             mapexpr
+             readexpr ;;
+        \$*) cmd="${fexpr:1}" ;;
+        *)   cmd="find $fexpr -print0" ;;
+    esac
+}
+
+clear
+[ -f "$HISTFILE" ] || printexamples > "$HISTFILE"
+
+mapfile -t fexprs < <(sort "$HISTFILE" | uniq -c | sort -nr | head -n5 |\
+  awk '{for (i=2; i<NF; i++) printf $i " "; print $NF}')
+printf "Most used earch expressions:\n"
+printexprs
+i=$((i + 1))
+
+mapfile -t -O"$i" fexprs < <(tail -n20 "$HISTFILE" | tac | uniq | head -n5)
+printf "Most recently used search expressions:\n"
+for ((j = i; j < ${#fexprs[@]}; j++)); do
+    printf '%s\t%s\n' "$((j + 1))" "${fexprs[$j]}"
+done
+read -r -p "Search expression or index (h for help): " fexpr
+
+mapexpr
+
+if [ -n "$fexpr" ]; then
+    printf "+l" > "$NNN_PIPE"
+    while :; do
+        readexpr
+        eval "$cmd" > "$NNN_PIPE" && break
+        read -r -e -p "Search expression: " -i "$fexpr" fexpr
+    done
+    if [ -n "$fexpr" ]; then
+        tail -n"$HISTLEN" "$HISTFILE" > "$TMPDIR/finderbms"
+        printf "%s\n" "$fexpr" >> "$TMPDIR/finderbms"
+        mv "$TMPDIR/finderbms" "$HISTFILE"
+    fi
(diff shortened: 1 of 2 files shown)


plugins/finder
```diff
@@ -1,28 +1,60 @@
...
+#   a '$' to run a custom search program such as fd or ripgrep.
+#   Entered expressions will be saved in history file to be listed
+#   as bookmarks and and can be entered by index and edited.
 #
-# Shell: POSIX compliant
-# Author: Arun Prakash Jana
+# Shell: Bash
+# Author: Luuk van Baal
+TMPDIR="${TMPDIR:-/tmp}"
+HISTFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/finderbms"
```
jarun, line 16: Please document very clearly that this plugin stores search history.

In the README as well.

jarun (changes requested):

plugins/finder
```diff
@@ -1,28 +1,60 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 # Description: Run custom search and list results in smart context
 #
-# Note: To enable multi select in fzf, export the following:
-#       - export FZF_DEFAULT_OPTS='--bind ctrl-a:select-all,ctrl-d:deselect-all,ctrl-t:toggle'
+# Usage:
+#   Run plugin and enter i.e. "-size +10M" to list files in current
```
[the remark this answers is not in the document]
luukvbaal: Added the note

plugins/finder
```diff
@@ -1,28 +1,60 @@
...
+# Shell: Bash
+# Author: Luuk van Baal
+TMPDIR="${TMPDIR:-/tmp}"
+HISTFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/finderbms"
+HISTLEN=10000
 
-. "$(dirname "$0")"/.nnn-plugin-helper
+clear
 
-printf "Examples:\n"
```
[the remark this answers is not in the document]
luukvbaal: These queries are now added to the history file for the initial list when ran the first time and can be displayed by pressing <kbd>h</kbd>.

plugins/finder
```diff
@@ -1,28 +1,90 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 # Description: Run custom search and list results in smart context
 #
-# Note: To enable multi select in fzf, export the following:
-#       - export FZF_DEFAULT_OPTS='--bind ctrl-a:select-all,ctrl-d:deselect-all,ctrl-t:toggle'
+# Usage:
+#   Run plugin and enter e.g. "-size +10M" to list files in current
```
jarun, line 6: This should be `-size +10M -print0` as the plugin is not adding it.

luukvbaal: It is, see https://github.com/jarun/nnn/pull/1124/files#diff-68fd6d7776c348e7e02e85cd49e9903e4d4fc5d24a9e83a91c342906dd302066R56

jarun: In this comment also we should add `-print0`. The reason is - running just `-size +10M` wouldn't list anything as the results are not NUL-terminated.

jarun: OK. I see we add it automatically for `find`.

jarun:
Thank you!

plugins/finder
luukvbaal: Entering `-size +10M` in the plugin will run `find -size +10M -print0".
Adding `-print0` to the comment will instruct the user to type a redundant `-print0`, not sure I agree.

luukvbaal:
Is it worth it to just use `tr` to NUL-separate to allow easier expressions for all programs??
Potential problem is filenames containing newlines but this is already not considered for the `fzf` example.
```diff
@@ -18,13 +17,13 @@ HISTFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/finderbms"
 HISTLEN=10000

 printexamples() {
-    printf ". -maxdepth 1 -name pattern
-. -maxdepth 1 -size +100M
-\$fd pattern -0
-\$fd -d 2 -S +100M -0
-\$grep -rlZ pattern
-\$rg -l0 pattern
-\$fzf -m | tr '\\\n' '\\\0'\n"
+    printf "-maxdepth 1 -name pattern
+-maxdepth 1 -size +100M
+\$fd pattern
+\$fd -d 2 -S +100M
+\$grep -rl pattern
+\$rg -l pattern
+\$fzf -m"
 }

 printexprs() {
@@ -53,7 +52,7 @@ readexpr() {
              mapexpr
              readexpr ;;
         \$*) cmd="${fexpr:1}" ;;
-        *)   cmd="find $fexpr -print0" ;;
+        *)   cmd="find $fexpr" ;;
     esac
 }

@@ -79,7 +78,7 @@ if [ -n "$fexpr" ]; then
     printf "+l" > "$NNN_PIPE"
     while :; do
         readexpr
-        eval "$cmd" > "$NNN_PIPE" && break
+        eval "$cmd | tr '\n' '\0'" > "$NNN_PIPE" && break
         read -r -e -p "Search expression: " -i "$fexpr" fexpr
     done
     if [ -n "$fexpr" ]; then
```

jarun:
I would say we skip the extra processing. We are already showing examples.

One issue I see is - we are adding duplicate entries. maybe we should not add the command to history if the last executed command from history matches.

luukvbaal:
We have to add duplicate entries if we want to keep the most used expression list accurate.

Alternatively we could drop the most used list and provide an actual user defined bookmarks list instead.

jarun:
> We have to add duplicate entries if we want to keep the most used expression list accurate.

Just compare with the last entry and skip adding if it's the same.

luukvbaal:
Yeah like I said if you skip recording entries that would make the most used list inaccurate.
What are you saving by doing that? A few bytes.
I could see it make sense if you drop the most used list in favor of static bookmarks but otherwise I'm not sure its worth it.

luukvbaal:
>  I would say we skip the extra processing. We are already showing examples.

Doubt the the extra processing is noticeable or even measurable, why not make it more user friendly?

jarun:
> What are you saving by doing that? A few bytes.

Not that. After a while, in the suggestions I see only 1 entry because everything towards the end in history is the same.

Please make the change.

luukvbaal:
> Not that. After a while, in the suggestions I see only 1 entry because everything towards the end in history is the same.

Yeah that is worth fixing but is not fixed by only comparing the most recent entry.
Identical adjacent entries are already filtered out of the list by `uniq`:
`tail -n20 "$HISTFILE" | tac | uniq | head -n5`

I'll see if an elegant solution exists for not having duplicate entries show up in the list of 5.

jarun:
In the examples screen, if I press `Enter`, a search is issued which lists everything. If there's not expression, search should not be issued.

luukvbaal:
`tail -n20 "$HISTFILE" | tac | awk '!a[$0]++' | head -n5` works.

> > I would say we skip the extra processing. We are already showing examples.
> 
> Doubt the the extra processing is noticeable or even measurable, why not make it more user friendly?

Do you still disagree on this?

jarun:
Don't know.

@KlzXS what do you think?

luukvbaal:
> In the examples screen, if I press `Enter`, a search is issued which lists everything. If there's not expression, search should not be issued.

Missed this, I'll check.

KlzXS:
> `tail -n20 "$HISTFILE" | tac | uniq | head -n5`

Still if at the end of the file there is more that 16 (= 20 - 4) of the same entry less than 5 lines will be printed.
I think that could be eliminated by using `uniq "$HISTFILE"  | tac | head -n5`. I wrote that on the spot, try it out.

Maybe `tac "$HISTFILE" | uniq | head -n5` is a more efficient pipeline. Doubt that it matters.

> `awk '!a[$0]++'`

Now I'm not that good with awk, but isn't this just uniq implemented in awk?

KlzXS:
Also isn't `$HISTFILE` the variable that stores the bash command history file? We can't use that.

luukvbaal:
> I think that could be eliminated by using `uniq "$HISTFILE" | tac | head -n5`. I wrote that on the spot, try it out.

`uniq` only filters out adjacent identical entries, this would still show for example alternating entries.

`awk '!a[$0]++'` doesn't print duplicate lines by keeping track of lines in a associative array.

> Still if at the end of the file there is more that 16 (= 20 - 4) of the same entry less than 5 lines will be printed. 

If this is seen as a problem (the script doesn't break when less than 5 entries are shown) we can increase the size of the history checked (e.g. `tail -n100`) or indeed improve it by throwing in a `uniq`.

luukvbaal:
> Also isn't `$HISTFILE` the variable that stores the bash command history file? We can't use that.
> You do override it. You decide.

It is indeed overridden, however we could use e.g. `NNN_FINDHIST` and `NNN_FINDHISTLEN` for those who want to set alternative values in their profile.

luukvbaal:
> I would say we skip the extra processing. We are already showing examples.

> > Doubt the the extra processing is noticeable or even measurable, why not make it more user friendly?

@KlzXS thoughts on using `tr '\n' '\0'` globally on the search commands instead of their respective program options (`find -print0, fd -0, grep -Z`)?

KlzXS:
> for example alternating entries.

Right! In that case I think ` tac "$HISTFILE" |  awk '!a[$0]++' | head -n5` would do it. You could limit the number of lines read with `tail`, but then you run the risk of not printing 5 entries and I think you should. Even 10000 lines is not that much to process, it shouldn't be slow.

As for NULL terminating gloablly I'm indifferent. I took a look at how `tr` is used across the project. 90% of the usage is translating the selection file from '\0' to `'\n`' . The only translations that go the other way is when we have output from `sed`. IIRC that is because POSIX sed doesn't have the flag for printing NULL seperated entries. Maybe some other program in another place. In general we don't do that it seems.

jarun:
How do I use pipes? I want to run something like:

    eval find | sort | uniq

0xACE:
@jarun why eval? just `find -iname "*.mp4" | sort | uniq`, or what are you referring to something else? We can chat live via gitter if you prefer.

jarun:
OK. I need to use `$find | uniq | tr '\n' '\0'`

PR #1130, opened 2021-08-17
Decide string length at compile time

0xACE:
I run into many premature optimizations in our codebase which are
unnecessary.

In this particular case `strlen()` is optimized at compile time even at
`-O0` with `gcc`.

I would value higher code quality than dealing with these things in our
future endeavours. If this is accepted I may supply some more
readability patches.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7972,12 +7972,12 @@ static bool setup_config(void)
 			return FALSE;
 		}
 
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

jarun:
Made some minor changes and force-pushed at commit 9d4330e382e5b3542bd5f51690c8bb0bdbff1fd8.

PR #1132, opened 2021-08-17
icon for zsh, nix, and matlab files(nerdfonts), lua color to lua file

CantoroMC:
I added some icons mainly for nerdfonts, they should probably extended for icons-in-terminal and .iconlookup plugin.
I would also like to suggest to add color to *.*sh icon (COLOR_JAVASCRIPT seems suitable).

### DIFF
--- a/src/icons.h
+++ b/src/icons.h
@@ -199,7 +199,7 @@ static const struct icon_pair icons_ext[] = {
 	/* L */
 	{"lha",      FA_FILE_ARCHIVE_O,    COLOR_ARCHIVE},
 	{"log",      FA_FILE_TEXT_O,       0},
-	{"lua",      FILE_LUA,             0},
+	{"lua",      FILE_LUA,             COLOR_LUA},
 	{"lzh",      FA_FILE_ARCHIVE_O,    COLOR_ARCHIVE},
 	{"lzma",     FA_FILE_ARCHIVE_O,    COLOR_ARCHIVE},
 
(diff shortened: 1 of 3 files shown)


src/icons.h
```diff
@@ -404,14 +404,16 @@ static const struct icon_pair icons_ext[] = {
 	{"lha",        ICON_EXT_LHA,       COLOR_ARCHIVE},
 	{"lhs",        ICON_EXT_LHS,       COLOR_VIM},
 	{"log",        ICON_EXT_LOG,       0},
-	{"lua",        ICON_EXT_LUA,       0},
+	{"lua",        ICON_EXT_LUA,       COLOR_LUA},
 	{"lzh",        ICON_EXT_LZH,       COLOR_ARCHIVE},
 	{"lzma",       ICON_EXT_LZMA,      COLOR_ARCHIVE},
 
 	/* M */
+	{"m",          ICON_EXT_C,         COLOR_C},
```
luukvbaal, line 412: Should be `ICON_EXT_MAT`.
LGTM otherwise.

CantoroMC: The goal was to have to different icons (one for source code *.m and one for tables *.mat) unfortunately there is no an icon that is similar for m file, so i though that the icon for c was suitable since matlab is derived from c.
Having the same icon for mat and m doesn t allow to immediately distinguish between the two.

If you find a better solution, tell me.

luukvbaal: Ah I see, a more general code icon makes more sense in that case I would say.

luukvbaal:
Also feel free to propagate the changes to `.iconlookup` if you want.
As for `icons-in-terminal.h`, I think its still behind some other additions to `nerdfont.h`.
Those are best picked up by anyone who misses them and actually uses icons-in-terminal.

src/icons.h
luukvbaal: Perhaps this percentage file icon `"\ufd1c"` image? 
I think matlab's percentage comments are quite distinctive.

CantoroMC: This seems clever, feel free to add the color you think more appropriate.

jarun:
@luukvbaal please squash merge when this is ready.

PR #1137, opened 2021-08-20
Macos 10.11

elder-n00b:


0xACE:
I doubt `README_FORK.md` should be included

elder-n00b:
I am sorry, I did not even intend to make a pull request yet. I botched something. And I put a bug in the makefile -- fixed.
I'll also remove the README_FORK.md, I put it there just in case someone comes across my fork.
It's clearly unnecessary now that I bothered you with the PR, sounds like it would be accepted.

0xACE:
I practically didn't review the changes at all, just scrolled and saw `README_FORK.md`.

I don't know if it will get accepted. In my experience: there are very tight restrictions on what gets merged into `nnn`...

Quick glance:
- some of your files are missing a newline character at the end of the file.
- why are `#include` directors given before include guards?

jarun:
Please explain the issue and the fix.

0xACE:
> Please explain the issue and the fix.

Just to speed things up im making a guess: I think he is supplying a patch for old macos versions that didnt support `clock_gettime` or w/e... I think we had discussions about this in the past.

Have a good weekend XOXO

elder-n00b:
Thanks for head up about the newlines.

About the includes, I only added this:
```
#ifdef MACOS1011
#include "mach_gettime.h"
#endif
```
where `MACOS1011` is defined in the makefile (didn't want to bother with the AccessibilityMacros).

Yes, as mentioned above, it's just a small workaround to make it compile on old systems.
I thought it could be useful to others on legacy hardware.

jarun:
Got it!

KlzXS:
I think this is fine. CI needs to be addressed first, though.

It's complaining about something in the added files and the command in Makefile. The command (`sw_vers`) is not available on linux. You should first check that the system has it using something that works on both linux and macos. And the BSDs.

I can't remember of the top of my head what was the portable solution we used. You can find an example somewhere in the project I'm sure.

elder-n00b:
I confirm, I put that bug in the makefile. It's already fixed -- in theory. I'll double check though.

elder-n00b:
>You can find an example somewhere in the project I'm sure.
I noticed the misc/haiku dir, I think I'll do something similar as it's more self-contained.
TBH I'm not very happy that I put the two mach_gettime files in src, they're not properly part of the project.

jarun:
Haiku has a dedicated Makefile. For macOS I would suggest using the same Makefile.

jarun:
See if this helps: https://stackoverflow.com/questions/8024961/c-macro-for-os-x-version-lion-or-not-detection

jarun:
Please do not add new files.

jarun:
Closing as we have a new PR.

PR #1138, opened 2021-08-21
MacOSX legacy

elder-n00b:
Hopefully it's now OK.

- Makefile now checks that `sw_vers` is present, also does version check a little better and uses (hopefully) better variable names.
- Added files are out of the way, in `misc/macosx-legacy`.
- (edit) Renamed the branch to `macosx-legacy` to reflect it's not 10.11 specific
- (edit) Note that the `clock_gettime` implementation used is not written by me, I grabbed it from [this gist](https://gist.github.com/alfwatt/3588c5aa1f7a1ef7a3bb) by [Alf Watt](github.com/alfwatt).

I thought I could run the CI workflow before PR but I haven't found how, even enabling workflows for the fork -- crossing fingers...


jarun:
@luukvbaal @KlzXS doesn't this belong to the patch framework? I don't want to add new files under `misc` for macOS.

elder-n00b:
I had the doubt too when I noticed the new patches framework, but the description didn't rhyme.
Also, it seems to me there's no way to patch the makefile from there -- I could be wrong.

jarun:
> Also, it seems to me there's no way to patch the makefile from there -- I could be wrong.

You should be able to patch any file in the repo. Would wait for others to help you out.

luukvbaal:
The current patch framework indeed does not allow patching the Makefile since we are applying the patches from inside the Makefile itself. At least the Makefile change will have to be merged.

The new files and `nnn.c` change could be adopted into a patch if you prefer although it doesn't really fit our current patch description.

luukvbaal:
In principle you can patch the Makefile but with the current framework that would just apply and revert the diff without affecting the already running ` make` process.

jarun:
@elder-n00b I think @luukvbaal intends to convey that this is related to OS support. I agree.

In that case I think we are good to go.

misc/macosx-legacy/mach_gettime.c
```diff
@@ -0,0 +1,42 @@
...
+#define MT_GIGA UINT64_C(1000000000)
+
+// TODO create a list of timers,
+static double mt_timebase = 0.0;
+static uint64_t mt_timestart = 0;
+
+int clock_gettime(clockid_t clk_id, struct timespec *tp)
+{
+    kern_return_t retval = KERN_SUCCESS;
+    if( clk_id == TIMER_ABSTIME) {
```
jarun, line 14: space after `if`

misc/macosx-legacy/mach_gettime.c
```diff
@@ -0,0 +1,42 @@
...
+#define MT_NANO (+1.0E-9)
+#define MT_GIGA UINT64_C(1000000000)
+
+// TODO create a list of timers,
+static double mt_timebase = 0.0;
+static uint64_t mt_timestart = 0;
+
+int clock_gettime(clockid_t clk_id, struct timespec *tp)
+{
+    kern_return_t retval = KERN_SUCCESS;
```
jarun, line 13: We use tabs.

misc/macosx-legacy/mach_gettime.c
```diff
@@ -0,0 +1,42 @@
...
+            mt_timebase = tb.numer;
+            mt_timebase /= tb.denom;
+            mt_timestart = mach_absolute_time();
+        }
+
+        double diff = (mach_absolute_time() - mt_timestart) * mt_timebase;
+        tp->tv_sec = diff * MT_NANO;
+        tp->tv_nsec = diff - (tp->tv_sec * MT_GIGA);
+    }
+    else { // other clk_ids are mapped to the coresponding mach clock_service
```
jarun, line 27: } else {

misc/macosx-legacy/mach_gettime.c
```diff
@@ -0,0 +1,42 @@
...
+#include <mach/mach_time.h>
+
+#define MT_NANO (+1.0E-9)
+#define MT_GIGA UINT64_C(1000000000)
+
+// TODO create a list of timers,
+static double mt_timebase = 0.0;
+static uint64_t mt_timestart = 0;
+
+int clock_gettime(clockid_t clk_id, struct timespec *tp)
```
jarun, line 11: static int

Makefile
```diff
@@ -153,11 +153,23 @@ LOGO64X64 = misc/logo/logo-64x64.png
 GITSTATUS = patches/gitstatus
 NAMEFIRST = patches/namefirst
 
+MACOSX_VERSION := $(strip $(shell command -v sw_vers >/dev/null && [ "`sw_vers -productName`" = "Mac OS X" ] && sw_vers -productVersion))
```
jarun, line 156: Please add a comment above to clarify what we are doing here.

jarun (changes requested):

luukvbaal:
> @elder-n00b I think @luukvbaal intends to convey that this is related to OS support. I agree.

No I was just conveying that IF we want to support legacy MacOS with this PR, the change to the Makefile is required in `master` due the the limitation of the patch framework.

I see no problem in placing this in `misc` but I was just saying that alternatively the required changes (outside of the Makefile change) could be placed in a patch if you really wanted.

elder-n00b:
Going to apply requested changes and commit.
Re patch, I guess it could also be done as a patch that interested users would apply manually before make.
Let me know.

luukvbaal:
That could work as well accompanied with a note on the Wiki/patches Readme yeah. Whatever @jarun prefers.

misc/macosx-legacy/mach_gettime.c
elder-n00b: I get compile error with `static int` in `mach_gettime.c`.
```
misc/macosx-legacy/mach_gettime.c:11:12: error: static declaration of 'clock_gettime' follows non-static declaration
```
It becomes warnings if I also declare it `static int` in `mach_gettime.h`.
```
misc/macosx-legacy/mach_gettime.c:11:12: warning: unused function 'clock_gettime' [-Wunused-function]
In file included from src/nnn.c:112:
src/../misc/macosx-legacy/mach_gettime.h:24:12: warning: function 'clock_gettime' has internal linkage but is not defined [-Wundefined-internal]
```
```
static int clock_gettime(clockid_t clk_id, struct timespec *tp);
           ^
src/nnn.c:5701:2: note: used here
        clock_gettime(CLOCK_REALTIME, &ts1); /* Use CLOCK_MONOTONIC on FreeBSD */
        ^
```
One of those c subtleties I never really understood... I would have though that meant "return a static int" but apparently static applies to the function, not the return value. 
I read that "static means that the function can be called only in the .c file where it is defined" but I don't really know. 

Should I drop the static?

elder-n00b:
The patch would be simply `git diff origin/master > macosx-legacy.patch`, is that correct?

misc/macosx-legacy/mach_gettime.c
jarun: Yes, please drop it then.

jarun:
No the misc thing is fine. It's a legacy OS support.

misc/macosx-legacy/mach_gettime.h
```diff
@@ -0,0 +1,28 @@
+#include <sys/types.h>
+#include <sys/_types/_timespec.h>
+#include <mach/mach.h>
+#include <mach/clock.h>
+
+#ifndef mach_time_h
+#define mach_time_h
+
```
0xACE, line 8: cap
It's probably not important in our limited usecase but incase someone else picks it up...

elder-n00b: Ops. Done. I'll be back ~6hrs.
Note: ideally I would have applied the fixup commits to the initial branch from v4.2 not HEAD but I'm not very git savvy and I wasn't sure not to wreck something with the open PR... happy swapping and squashing. Apologies. Or I'll do it tomorrow morning if it's a safe operation.

jarun:
Thank you!

PR #1141, opened 2021-08-23
Added theme support for bat in preview-tui plugin

f4z3r:


jarun:
Plugins are indicative. Personal preferences vary widely so we leave it upto the users. Also, tools like bat are really too fancy for nnn. Users who are really doing some serious work are probably well off with _less_ already.

jarun:
Reopened as discussed in #1160.

jarun:
Thank you!

PR #1143, opened 2021-08-25
fix fzhist plugin - get fish_history from the right place

joelazar:
The history file is at this location (`~/.local/share/fish/fish_history`) since fish 2.3.0. [ref](https://github.com/fish-shell/fish-shell/issues/862#issuecomment-19138822)


jarun:
Thank you!

PR #1148, opened 2021-08-29
[draft] signal CWD change to terminal via OSC-7

### LINKED ISSUE #1147: Emit OSC-7 escape when changing directory
Originally reported here: https://codeberg.org/dnkl/foot/issues/696; some users expect the terminal's builtin "spawn a new window" feature to open the new terminal window in nnn's current directory.

This works "automagically" on some terminals (which most likely dig through `/proc` to find out the current CWD of the leaf child process).

Others implement OSC-7, and escape that let's the application control which directory the terminal should consider to be the "current" one. Shells are often configured to emit this as part of rendering their prompts.

OSC-7 is pretty simple; just URL encode the directory and write `\033]7;file://$HOSTNAME$CWD\033\\` (for example, `\033]7;file://mini.la.casa/home/user/foobar\033\\`). See e.g. https://codeberg.org/dnkl/foot/wiki#user-content-how-to-configure-my-shell-to-emit-the-osc-7-escape-sequence for a shell implementation. It is also documented here: https://iterm2.com/documentation-escape- codes.html.

It is implemented by at least VTE terminals, iTerm2 and foot.

Note that well behaving terminals ignore unrecognized OSC sequences, meaning it should be safe to emit OSC-7 unconditionally.

Note: avoid using `localhost` as hostname. The terminal matches the hostname to be able to ignore OSC-7 escapes from remote hosts (e.g. SSH).

Is it a "standardized" escape? It's not a VT escape, and not part of ECMA-48. So no, in that sense it's not standardized. There's an open issue over at [terminal-wg](https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/20), and one unmerged [PR](https://gitlab.freedesktop.org/terminal-wg/specifications/-/merge_requests/7) (unmerged due to too much bikeshedding). With that, I fully understand if this is a feature you want to pass :)

dnkl:
The path still needs to be URL encoded, but this shows the gist of it. Tested in foot only so far. Will hopefully be able to try a few more tomorrow.

**TODO**

* [ ] URL encode path
* [ ] Test more terminals

Closes #1147

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -6424,7 +6424,7 @@ static bool browse(char *ipath, const char *session, int pkey)
 	char newpath[PATH_MAX] __attribute__ ((aligned)),
 	     rundir[PATH_MAX] __attribute__ ((aligned)),
 	     runfile[NAME_MAX + 1] __attribute__ ((aligned));
-	char *path, *lastdir, *lastname, *dir, *tmp;
+	char *path, *lastdir, *lastname, *dir, *tmp, hostname[_POSIX_HOST_NAME_MAX];
 	pEntry pent;
 	enum action sel;
 	struct stat sb;
@@ -6500,6 +6500,11 @@ begin:
 		valid_parent(path, lastname);
 		setdirwatch();
 	}
+
+	/* Signal CWD change to terminal */
+	gethostname(hostname, sizeof(hostname));
+	printf("\033]7;file://%s%s\033\\", hostname, path);
+	fflush(stdout);
 
 #ifndef NOX11
 	if (cfg.x11 && !g_state.picker) {
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -6424,7 +6424,7 @@ static bool browse(char *ipath, const char *session, int pkey)
 	char newpath[PATH_MAX] __attribute__ ((aligned)),
 	     rundir[PATH_MAX] __attribute__ ((aligned)),
 	     runfile[NAME_MAX + 1] __attribute__ ((aligned));
-	char *path, *lastdir, *lastname, *dir, *tmp, hostname[_POSIX_HOST_NAME_MAX];
+	char *path, *lastdir, *lastname, *dir, *tmp;
 	pEntry pent;
 	enum action sel;
 	struct stat sb;
@@ -6504,8 +6504,8 @@ begin:
 #ifndef NOX11
 	if (cfg.x11 && !g_state.picker) {
 		/* Signal CWD change to terminal */
-		gethostname(hostname, sizeof(hostname));
-		printf("\033]7;file://%s%s\033\\", hostname, path);
+		gethostname(newpath, sizeof(newpath));
+		printf("\033]7;file://%s%s\033\\", newpath, path);
 
 		/* Set terminal window title */
 		r = set_tilde_in_path(path);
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -6501,6 +6501,11 @@ static bool browse(char *ipath, const char *session, int pkey)
 		setdirwatch();
 	}
 
+	/* Signal CWD change to terminal */
```
jarun, line 6504: I think we should add it under the check in line 6510 below. 

Then users can enable it using option `-x`. The code is compiled by default.

Program options: https://github.com/jarun/nnn/wiki/Usage#program-options

The reason we did this is there were some cases where purist users complained about stray prints (like for xterm title print).

jarun (changes requested):

src/nnn.c
dnkl: That's perfect! Fixed, thanks!

src/nnn.c
```diff
@@ -6503,6 +6503,10 @@ static bool browse(char *ipath, const char *session, int pkey)
 
 #ifndef NOX11
 	if (cfg.x11 && !g_state.picker) {
+		/* Signal CWD change to terminal */
+		gethostname(hostname, sizeof(hostname));
```
jarun, line 6507: You can use `newpath` (which works like a scratchbuffer in each iteration) here to avoid additional memory usage for hostname. It's `PATH_MAX` long.

jarun (changes requested):

src/nnn.c
jarun: I think there was a oversight on my part. Given the hostname doesn't change often (an it's non-fatal) if it does, we should have allocated the memory and set it only once in browse and then printed it here.

jarun: Once you complete the tests, please update. I'll make the change to use a single call instead of each time. Thanks!

jarun:
Please confirm commit 0556ac14ecd6ae028dafe9645ac3bcefa8c0faf4 works fine in your tests.

dnkl:
That commit works fine in both foot, and in gnome-terminal.

Technically, the path needs to be URL encoded. In reality, not doing will work in probably 99% of all cases. I tested switching to a directory named `åöä&>`, and it worked in gnome-terminal as well as in foot.

I also tested a directory that included a literal `ESC`, and that did **not** work. That's expected, since an ESC will typically reset the VT state machine.

It would be fairly simple to implement a URL encoder that allows plain ASCII but `%NN` encodes all other characters. Such an encoder would be overly eager, compared to what's actually necessary, but would produce "safe" output.

On the other hand, if we want to keep things simple, then let's skip URL encoding. The worst thing that happens is the new terminal doesn't spawn in the correct directory.

jarun:
> I tested switching to a directory named åöä&>, and it worked in gnome-terminal as well as in foot.

That's because by default `nnn` is linked to ncursesw (with wide char support) so wide chars workas they should under the hood. Wide chars should work as expected.

jarun:
BTW, given the restrictions in `nnn`, no one would agree to have an encoder for a single-line printf. Practically speaking, there's no ROI.

dnkl:
Then it's all good from my side.

jarun:
:+1:

PR #1156, opened 2021-09-04
fix jump cd to use seperate params

assaf758:
the shellcheck CI failure (SC2086) was false-positive - as actually we want word splitting here.


jarun:
Thank you!

PR #1166, opened 2021-09-17
New plugin: goto - quick directory navigation with key combos

ichirou2910:
Quick directory jump with user-defined key combos. An alternative to `cdpath`.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -29,6 +29,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [fzopen](fzopen) | Fuzzy find file(s) in subtree to edit/open/pick | sh | fzf, xdg-open/open |
 | [fzplug](fzplug) | Fuzzy find, preview and run other plugins | sh | fzf |
 | [getplugs](getplugs) | Update plugins to installed `nnn` version | sh | curl |
+| [goto](goto) | Use key combos for quick directory navigation | sh | - |
 | [gpg\*](gpg\*) | Encrypt/decrypt files using GPG [✓] | sh | gpg |
 | [gutenread](gutenread) | Browse, download, read from Project Gutenberg | sh | curl, unzip, w3m<br>[epr](https://github.com/wustho/epr) (optional) |
 | [imgresize](imgresize) | Batch resize images in dir to screen resolution | sh | [imgp](https://github.com/jarun/imgp) |
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -29,7 +29,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [fzopen](fzopen) | Fuzzy find file(s) in subtree to edit/open/pick | sh | fzf, xdg-open/open |
 | [fzplug](fzplug) | Fuzzy find, preview and run other plugins | sh | fzf |
 | [getplugs](getplugs) | Update plugins to installed `nnn` version | sh | curl |
-| [goto](goto) | Use key combos for quick directory navigation | sh | - |
+| [goto](goto) | Use key combos for quick directory navigation | bash | - |
 | [gpg\*](gpg\*) | Encrypt/decrypt files using GPG [✓] | sh | gpg |
 | [gutenread](gutenread) | Browse, download, read from Project Gutenberg | sh | curl, unzip, w3m<br>[epr](https://github.com/wustho/epr) (optional) |
 | [imgresize](imgresize) | Batch resize images in dir to screen resolution | sh | [imgp](https://github.com/jarun/imgp) |
--- a/plugins/goto
+++ b/plugins/goto
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 # Description: Use key combos for quick changing directory
 #
@@ -18,7 +18,7 @@
 #     Eg: For 'cf' and 'cff', 'cf' will be selected
 #
 #
-# Shell: POSIX compliant
+# Shell: Bash
 # Author: Ichirou2910
 
 prompt_exit() {
--- a/plugins/goto
+++ b/plugins/goto
@@ -46,7 +46,7 @@ fi
 
 printf "Input keys: "
 
-while [ -z "$target" ]
+while [ -z "$match" ]
 do
     read -rn 1 key
 
@@ -66,7 +66,7 @@ do
 done
 
 printf "\n"
-target=$(echo "$match" | cut -d ' ' -f 2)
+target=$(cut -d ' ' -f 2- <<< "$match")
 
 if [ -n "$target" ] && [ -d "$target" ]; then
     # Go to target directory
--- a/plugins/goto
+++ b/plugins/goto
@@ -1,0 +1,79 @@
+#!/usr/bin/env sh
+
+# Description: Use key combos for quick changing directory
+#
+# Usage:
+#   1. Create a gotos file and set $GOTOS_LIST accordingly
+#      By default, $GOTOS_LIST is set to: ${XDG_CACHE_HOME:-$HOME/.cache}/nnn/gotos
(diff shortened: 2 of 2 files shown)


ichirou2910:
I forgot to put a demo usage. Here it is.
Note: at 0:24 there was a typo in my goto list, you can see I fixed it right after

https://user-images.githubusercontent.com/51302893/133742626-93d1dd3c-a419-4f2c-95bf-84feb5fb169c.mp4

luukvbaal:
I like this way of navigation but in my opinion it would make more sense to extend `nnn`'s bookmark interface to allow for multi-key bookmarks rather than yet another bookmark plugin.

i.e. allowing `NNN_BMS=cf:/home/user/.config` and pressing <kbd>bcf</kbd>.

ichirou2910:
You got a point there. This plugin can be treated as a quick solution for that bookmark problem.

Something I find this also better than `nnn`'s bookmark system is that it only reads the bookmark list when needed so you can add and use more bookmarks without re-opening `nnn`.

Also my rather personal reason is that I don't kinda like the idea of putting all of my many bookmarks (I have like tens of it) inside a single looooong string environment variable.

akimdi:
> I like this way of navigation but in my opinion it would make more sense to extend `nnn`'s bookmark interface to allow for multi-key bookmarks rather than yet another bookmark plugin.
> 
> i.e. allowing `NNN_BMS=cf:/home/user/.config` and pressing bcf.

I totally agree with this

Kabouik:
On the one hand, I quite like this system because it is more flexible than the default (more bookmarks possible), but on the other hand I also like the default nnn system because it is quicker when the user has few bookmarks and does not need multiple-key sequences, as it does not require <kbd>Return</kbd> to confirm.

ichirou2910:
> and does not need multiple-key sequences, as it does not require Return to confirm.

@Kabouik Oh this plugin doesn't require `Return` on the key combo itself. It's only required when you encounter an error.

It works by continously looking for matches and potential matches at each new key the user types in. User won't have to confirm the key combo with `Return` at all.

Kabouik:
But let's say you have `c` for `~/.config` and `cn` for `~/.config/nnn`, how do you go to `~/.config` without <kbd>Return</kbd> to interrupt the prompt and tell it you're done?

ichirou2910:
You won't be able to reach `~/.config/nnn` but always `~/.config` cuz the plugin will always its keys first.
I also put it in the note section of the plugin. It will automatically navigate to the first full match.
I would map `~/.config` to `cc`. That way you can reach both directory.
Yes indeed there must be an extra key in this case, but imo it shouldn't matter too much.

N-R-K:
>  it would make more sense to extend nnn's bookmark interface to allow for multi-key bookmarks rather than yet another bookmark plugin.

> I don't kinda like the idea of putting all of my many bookmarks (I have like tens of it) inside a single looooong string environment variable.
 
imo, the new directory/symlink bookmark system already takes care of this usecase.

Also the env var can be split into chunks. This is what I'm doing with my plugins for example.

```sh
NNN_PLUG="${NNN_PLUG}d:drag-drop-instant;"
NNN_PLUG="${NNN_PLUG}s:sudo-edit;"
NNN_PLUG="${NNN_PLUG}x:xdgdefault;"
export NNN_PLUG
```

jarun:
@KlzXS @luukvbaal we are not going to implement multi-key in-program. With that in mind and the bookmark plugins, do we need this?

> imo, the new directory/symlink bookmark system already takes care of this usecase.

@N-R-K can you explain a bit how?

N-R-K:
> can you explain a bit how?

Going off the `bcf` example, you can just have the symlink be named cf and then (assuming no NNN_BMS) pressing `b` should take you to the bookmark dir, and if you're in type to nav mode then you can now just press `cf` to enter that bookmark.

jarun:
Thanks! I forgot we can rename the symlinks.

ichirou2910:
> Going off the `bcf` example, you can just have the symlink be named cf and then (assuming no NNN_BMS) pressing `b` hould take you to the bookmark dir, and if you're in type to nav mode then you can now just press `cf` to enter that bookmark.

Then you have to manually switch to type-to-nav mode right? Is there a way to automatically enter type-to-nav mode and when done navigating to target directory automatically exit it?

> Also the env var can be split into chunks. This is what I'm doing with my plugins for example.

Wow thanks, I've never thought doing this.

N-R-K:
> Is there a way to automatically enter type-to-nav mode and when done navigating to target directory automatically exit it?

In mainline nnn, no. On my personal build I have it patched to automatically enter type-to-nav mode when going into the bookmark dir. Doesn't auto exit however.

lawnowner:
The built-in bookmarking system is good, but an alternative can also be provided, like the built-in renamer and the plugin `renamer`, to offer differences that may be more preferable to some users, e.g. when `dir` is bookmarked, navigating to `/path/to/dir` instead of `~/.config/nnn/bookmarks/dir`. As a suggestion, this PR's `enter`-less navigation can be integrated with a switch such as an environment variable into a plugin that supports goto/add/remove functionality:

    #!/usr/bin/env sh

    #  Up to 99 bookmarks are supported, and bookmark names can't include '/'.
    
    clear
    
    bms=$(sed -En "s:'\\\\'':':g; s:^[[:blank:]]+'(.+)'\) bmval=.+:/\1/:p" "$0" \
        | sort -f \
        | pr -t -W 60 -i' 1' -o 4 -n' 2' -3)
    printf 'Bookmarks:\n\n%s\n\nName|Number|?: ' "$(echo "$bms" | tr -d '/')"
    IFS= read -r bmkey
    
    case $bmkey in \?)
        l1="     <name|number>  : Go to the bookmark.\n"
        l2="     /<name>        : Bookmark the current entry.\n"
        l3="     <name|number>/ : Remove the bookmark.\n"
        l4="     /              : Remove all single-character bookmarks.\n"
        printf '%b' "\n$l1$l2$l3$l4\nName|Number|?: "; IFS= read -r bmkey
    esac
    
    case $bmkey in
        [0-9]|[0-9][0-9])
            bmkey=$(echo "$bms" | sed -En "s:.*\<$bmkey /([^/]+)/.*:\1:p") ;;
        [0-9]/|[0-9][0-9]/)
            bmkey=$(echo "$bms" | sed -En "s:.*\<${bmkey%/} /([^/]+/).*:\1:p") ;;
        /*/*|*/*/)
            clear; exit 1
    esac
    
    case $bmkey in
        */*)
            sanitize () {
                printf '%s' "$1" | sed -E "s/'/'\\\\''/g; "'s/\\/\\\\/g
                                           s/(\.|\*|\[|\^|\$|\+|\?|\{|\||\(|\))/\\\1/g'
            }
            case $bmkey in
                /?*)
                    bmkey=$(sanitize "${bmkey#/}")
                    bmval=$(echo "$2/$1" | sed -E 's/(&|\\[1-9])/\\\1/g')
                    if echo "$bms" | grep -Pq "\b(\d|\d\d) ${bmkey}/"; then
                        sed -Ei "s:(^[[:blank:]]+'$bmkey'\) bmval=).+:\1'$bmval' ;;:" "$0"
                    else
                        sed -Ei "s:^([[:blank:]]+)\*\):\1'$bmkey'\) bmval='$bmval' ;;\n\1\*\):" "$0"
                    fi
                    clear; exit 0 ;;
                ?*/)
                    sed -Ei "/^[[:blank:]]+'$(sanitize "${bmkey%/}")'\)/d" "$0"
                    clear; exit 0 ;;
                /)
                    sed -Ei "/^[[:blank:]]+('.'|''\\\\''')\)/d" "$0"
                    clear; exit 0 ;;
            esac ;;
        'home') bmval=~ ;;
        *) clear; exit 1
    esac
    
    [ -p "$NNN_PIPE" ] && printf '%s' "0c$bmval" > "$NNN_PIPE"
    
    clear

jarun:
I think we are done with bookmarks.

PR #1171, opened 2021-09-20
Close previewer when opening file

luukvbaal:
Write opcode to FIFO when opening/editing files to let previewers know that a file has been opened.
Makes the assumption that the previewer is no longer necessary which I think is fair. One can always reopen the previewer when returning to `nnn`.


jarun:
CI fails. Please check.

jarun:
Thank you! I'll take care of the lint.

Issue #1170, opened 2021-09-19
Make preview-tui using kitty terminal splits close the split before opening text files to edit


David-Else:
@luukvbaal Currently the preview is not closed and the split remains, is this a bug or a feature request? It seems like a bug, why keep the preview?

My `.bashrc` has:
```
alias nnn="nnn -xe"
export NNN_FIFO=/tmp/nnn.fifo

export EDITOR="nvim"
```
```
$ nnn -V
4.2
Source       : nnn-4.2-1.3.src.rpm
```

## select file with preview to the right
![Screenshot from 2021-09-19 23-41-23](https://user-images.githubusercontent.com/12832280/133945311-948c6e5b-58fb-475f-a692-e4baf874e4fe.png)
## open file (in Neovim) and preview remains
![Screenshot from 2021-09-19 23-41-32](https://user-images.githubusercontent.com/12832280/133945312-7a1f9c12-0e2d-4a3b-8288-d3d7c7562f95.png)

luukvbaal:
Not really a bug but I do agree it is unwanted behavior. The problem is that there is currently no way for `preview-tui` to know that a file was opened by `nnn`.

Would require something to be written to the FIFO when opening/editing a file.
Might be worthwhile addition. Currently the only two plugins that read from the FIFO are the preview plugins so it shouldn't be too much trouble.

What if we write some sort of opcode to the FIFO when opening/editing a file that can be checked in the preview plugins @jarun?

luukvbaal:
Like this I would say
```diff
diff --git a/plugins/preview-tui b/plugins/preview-tui
index b6788fc4..9f58e821 100755
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -389,6 +389,7 @@ preview_fifo() {
         if [ -n "$selection" ]; then
             kill "$(cat "$PREVIEWPID")"
             [ -p "$FIFO_UEBERZUG" ] && ueberzug_remove
+            [ "$selection" = "open" ] && break
             preview_file "$selection"
             printf "%s" "$selection" > "$CURSEL"
         fi
diff --git a/src/nnn.c b/src/nnn.c
index 44f297f4..642edd04 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -846,7 +846,7 @@ static char *load_input(int fd, const char *path);
 static int set_sort_flags(int r);
 static void statusbar(char *path);
 #ifndef NOFIFO
-static void notify_fifo(bool force);
[...]

PR #1172, opened 2021-09-21
Re-open previewer when internal editor is closed

luukvbaal:
Tried to implement what was discussed in #1170.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -367,7 +367,8 @@ typedef struct {
 	uint_t stayonsel  : 1;  /* Disable auto-proceed on select */
 	uint_t trash      : 2;  /* Use trash to delete files 1: trash-cli, 2: gio trash */
 	uint_t uidgid     : 1;  /* Show owner and group info */
-	uint_t reserved   : 7;  /* Adjust when adding/removing a field */
+	uint_t previewer  : 1;  /* Run state of previewer */
+	uint_t reserved   : 6;  /* Adjust when adding/removing a field */
 } runstate;
 
 /* Contexts or workspaces */
@@ -512,6 +513,11 @@ static char g_tmpfpath[TMP_LEN_MAX] __attribute__ ((aligned));
 
 /* Buffer to store plugins control pipe location */
 static char g_pipepath[TMP_LEN_MAX] __attribute__ ((aligned));
+
+#ifndef NOFIFO
+/* Buffer to store preview plugins control pipe location */
+static char g_ppipepath[TMP_LEN_MAX] __attribute__ ((aligned));
+#endif
 
 /* Non-persistent runtime states */
 static runstate g_state;
@@ -687,12 +693,13 @@ static const char * const messages[] = {
 #define NNN_FCOLORS 5
 #define NNNLVL      6
 #define NNN_PIPE    7
-#define NNN_MCLICK  8
-#define NNN_SEL     9
-#define NNN_ARCHIVE 10
-#define NNN_ORDER   11
-#define NNN_HELP    12 /* strings end here */
-#define NNN_TRASH   13 /* flags begin here */
+#define NNN_PPIPE   8
+#define NNN_MCLICK  9
+#define NNN_SEL     10
+#define NNN_ARCHIVE 11
+#define NNN_ORDER   12
+#define NNN_HELP    13 /* strings end here */
+#define NNN_TRASH   14 /* flags begin here */
 
 static const char * const env_cfg[] = {
 	"NNN_OPTS",
@@ -703,6 +710,7 @@ static const char * const env_cfg[] = {
 	"NNN_FCOLORS",
 	"NNNLVL",
 	"NNN_PIPE",
+	"NNN_PPIPE",
 	"NNN_MCLICK",
 	"NNN_SEL",
 	"NNN_ARCHIVE",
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -5131,15 +5139,20 @@ static bool run_cmd_as_plugin(const char *file, uchar_t flags)
...
 	g_pipepath[len - 1] = '/';
-	len = xstrsncpy(g_pipepath + len, "nnn-pipe.", TMP_LEN_MAX - len) + len;
-	xstrsncpy(g_pipepath + len - 1, xitoa(getpid()), TMP_LEN_MAX - len);
+	xstrsncpy(g_ppipepath, g_pipepath, TMP_LEN_MAX);
+	len += xstrsncpy(g_pipepath + len, "nnn-pipe.", TMP_LEN_MAX - len);
+	xstrsncpy(g_pipepath + len - 1, xitoa(pid), TMP_LEN_MAX - len);
+	len = xstrsncpy(g_ppipepath + lenbuf, "nnn-ppipe.", TMP_LEN_MAX - lenbuf) + lenbuf;
+	xstrsncpy(g_ppipepath + len - 1, xitoa(pid), TMP_LEN_MAX - len);
 	setenv(env_cfg[NNN_PIPE], g_pipepath, TRUE);
+	setenv(env_cfg[NNN_PPIPE], g_ppipepath, TRUE);
```
luukvbaal, line 5155: Also set `NNN_PPIPE` for preview pipe. Would have wanted to re-use the `NNN_PIPE` fifo but I didn't see a way to elegantly include it into the current logic. The problem being that we `waitpid()` for the plugins but the main `preview-tui` shell returns before it can write to the pipe(written to by subshell).

src/nnn.c
```diff
@@ -6900,6 +6938,12 @@ static bool browse(char *ipath, const char *session, int pkey)
 				notify_fifo(FALSE, TRUE);
 #endif
 				spawn(editor, newpath, NULL, NULL, F_CLI);
+#ifndef NOFIFO
+				if (g_state.previewer) {
+					pkey = previewkey;
+					goto run_plugin;
```
luukvbaal, line 6944: Set `pkey` and run `SEL_PLUGIN` case if run status is 1.

src/nnn.c
```diff
@@ -5168,6 +5181,28 @@ static ssize_t read_nointr(int fd, void *buf, size_t count)
...
+		if (read_nointr(fd, buf, 1) != 1)
+			return NULL;
+		g_state.previewer = (uint_t)buf[0];
+	}
+
+	close(fd);
+	unlink(g_ppipepath);
+	return NULL;
+}
+#endif
```
luukvbaal, line 5204: I think we need to do this concurrently so I put it in a function pointer for `pthread_create()` although not complete sure it's necessary. This is what worked though.

src/nnn.c
```diff
@@ -7598,6 +7646,14 @@ static bool browse(char *ipath, const char *session, int pkey)
 					goto nochange;
 				}
 
+#ifndef NOFIFO
+				if (xstrcmp(tmp, "preview-tui") == 0) {
```
luukvbaal, line 7650: Set `previewkey` to the plugin key corresponding to `preview-tui` and run `previewpipe()` to get previewer run status.

Currently only for `preview-tui` and not `preview-tabbed` as I'm not sure what is the desired open/close behavior for `preview-tabbed` users, would require feedback.

luukvbaal:
Please review if you can @jarun @KlzXS @0xACE.

jarun:
It's a big change for a minor problem. I am not OK with this.

luukvbaal:
Unfortunate. You should probably apply
```diff
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -133,10 +133,6 @@ previewer_loop () {
             continue
         fi

-        if [ "$FILE" = "close" ] ; then
-            break
-        fi
-
         kill_viewer

         MIME="$(file -bL --mime-type "$FILE")"

--- a/src/nnn.c
+++ b/src/nnn.c
@@ -6951,9 +6995,6 @@ nochange:
                        }

                        /* Invoke desktop opener as last resort */
-#ifndef NOFIFO
-                       notify_fifo(FALSE, TRUE);
-#endif
                        spawn(opener, newpath, NULL, NULL, opener_flags);

                        /* Move cursor to the next entry if not the last entry */
```
 or revert #1171 altogether, was still waiting on feedback when you merged that.

The current implementation is only valid for `preview-tui` until we know what is desired behavior for `preview-tabbed` users and for `preview-tui` we don't want to close the preview when using xdg-open.

jarun:
Going forward, please use the _work in progress_ label when you do not want something to be merged immediately.

I will revert 1171. We can't keep making heavy changes for minor behavioural inconveniences.

jarun:
I will revert #1171.

luukvbaal:
Sure, I thought the fact that it was in progress was obvious from the discussion in #1170.

I must say I really like the improvement to this "minor behavioral inconvenience". Having the preview pane close and re-open automagically when opening/closing the internal editor feels relly nice. 
Any chance it could be included as a user patch?

jarun:
> Sure, I thought the fact that it was in progress was obvious from the discussion in #1170.

I can't always guarantee that I will be in a position to read long threads of discussions. That why I expect bold single-liners or appropriate labels which would tell me that something needs more attention (read "time").

> Any chance it could be included as a user patch?

Sure thing!

Kabouik:
Great news about the patch, I admit I was a bit frustrated that the change would not be merged after @luuvkvaal finished implementing it. This patch will probably be my first nnn patch, and a good excuse to finally try them! This change will be a good QOL improvement for me, so a big thank to @luukvbaal for all the sweat put into it.

jarun:
@Kabouik from the history of your interaction with nnn, I haven't seen a single instance where you thought a new change in not required in nnn. That comes from the fact that you don't really have to maintain the project or fix issues that might come from those changes. Having an opinion without any responsibility is easy.

Kabouik:
> I haven't seen a single instance where you thought a new change in not required in nnn. That comes from the fact that you don't really have to maintain the project or fix issues that might come from those changes. Having an opinion without any responsibility is easy.

I won't pretend the opposite (although I actually surprised myself two days ago when I posted in favor of having `goto` as a patch instead of merging it with the `nnn` builtin system). I think this is true and am aware that one reason why I am enthusiastic about any improvement is I don't have to deal with the associated overheads, but I only post as a user, not expecting opinion to be taken as that of a developer.

I didn't post here to discuss the decision, I wouldn't have posted if a user patch hadn't been suggested. I did because I truly thought this was good news and would be my first patch, and thought showing there would be a user base could maintain @luukvbaal's motivation for the patch.

Perhaps I shouldn't have mentioned my frustration before that point, but I really meant no offense. I wouldn't want the hard job of filtering through the numerous PRs and user requests to maintain such an active project, but I think the end quality of `nnn` speaks for itself.

PR #1174, opened 2021-09-22
Improve restorepreview race conditions

luukvbaal:
* Send runstate as soon as possible (updating g_state.previewer earlier avoiding some cases where the previewer is left open).
* Sleep longer before breaking from the preview loop such that editor has had longer to start up and will receive `SIGWINCH`.

My editor (neovim) only uses half the screen when the preview pane is closed while it is starting up, prolonging the sleep fixes it:
out
Slower machines might need a longer sleep and faster machines could do with shorter probably. On my machine `0.05` seems to work for most files but some larger files require `0.15`.


jarun:
Thank you!

PR #1178, opened 2021-09-24
Update preview-tui

tax79:
Removed the color option from bat due to a bug while displaying any text files.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -77,7 +77,7 @@ USE_PISTOL="${USE_PISTOL:-0}"
 ICONLOOKUP="${ICONLOOKUP:-0}"
 PAGER="${PAGER:-less -P?n -R}"
 TMPDIR="${TMPDIR:-/tmp}"
-BAT_STYLE="${BAT_STYLE:-numbers}"
+BAT_STYLE="${BAT_STYLE:-plain}"
 BAT_THEME="${BAT_THEME:-ansi}"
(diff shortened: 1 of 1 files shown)


luukvbaal:
What "bug" are you talking about? It's best to elaborate if you make a PR. 
The bug is more likely to be in your environment than in `preview-tui`.

If you can't be bothered to fix your environment, feel free to make these changes locally, no need to make a PR...

tax79:
Try it yourself. lol
Now the preview looks like this:
[https://imgur.com/a/BfUvq2S](url)
If you change 
`--color=always`
to 
`--color=auto`
the problem is solved.

luukvbaal:
Like I said, the preview looks like that due to problems with your local environment.
https://github.com/sharkdp/bat/issues/1113 https://github.com/sharkdp/bat/issues/1063 seem related but those are supposed to have been fixed so I'm not sure.

In any case I used `preview-tui` every day and the problem is not present for me, I don't need to "try it myself", lol.

tax79:
Changing 'most' with 'less' in this line solves the issue:
`[ "$PAGER" = "most" ] && PAGER="less -R"`
Now everything works as supposed to.
Thanks.

luukvbaal:
The fact that this is necessary for you must mean you have `PAGER` set in your environment without the `-R` flag. Adding that should be enough, see `PAGER="${PAGER:-less -P?n -R}"` in `preview-tui`.

PR #1179, opened 2021-09-26
Enter type-to-nav on bookmark key

luukvbaal:
This provides the bookmark workflow I think @N-R-K was looking for, entering type-to-nav mode when pressing <kbd>b</kbd> to enter bookmark dir, and exiting type-to-nav mode if you were not in type to nav mode before pressing <kbd>b</kbd>.

At the same time, following the suggestion by @N-R-K of renaming your bookmark symlinks to the desired keycombos #1166, you can effectively achieve the workflow from the @ichirou2910's `goto` plugin by simply pressing <kbd>b</kbd> followed by the keycombo (assuming `NNN_BMS` is unset).

As it shouldn't break any existing workflows, I hope it can be merged. Otherwise I suppose a user-patch will have to do.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4952,6 +4952,8 @@ static size_t handle_bookmark(const char *bmark, char *newpath)
 
 	if (!r && chdir(newpath) == -1)
 		r = MSG_ACCESS;
+	else if (fd == '\r')
+		r = -1;
 
 	return r;
 }
@@ -6475,7 +6477,7 @@ static bool browse(char *ipath, const char *session, int pkey)
 	pEntry pent;
 	enum action sel;
 	struct stat sb;
-	int r = -1, presel, selstartid = 0, selendid = 0;
+	int r = -1, presel, selstartid = 0, selendid = 0, waitbm = 0;
 	const uchar_t opener_flags = (cfg.cliopener ? F_CLI : (F_NOTRACE | F_NOSTDIN | F_NOWAIT));
 	bool watch = FALSE;
 	ino_t inode = 0;
@@ -6785,7 +6787,13 @@ nochange:
 				goto begin;
 
 			pent = &pdents[cur];
-			mkpath(path, pent->name, newpath);
+			if (waitbm) {
+				cfg.filtermode = waitbm - 1;
+				waitbm = FALSE;
+				if (!realpath(pent->name, newpath))
+					mkpath(path, pent->name, newpath);
+			} else
+				mkpath(path, pent->name, newpath);
 			DPRINTF_S(newpath);
 
 			/* Visit directory */
@@ -7005,9 +7013,12 @@ nochange:
 		case SEL_BMOPEN:
 			if (sel == SEL_BMOPEN) {
 				r = (int)handle_bookmark(mark, newpath);
-				if (r) {
+				if (r > 0) {
 					printwait(messages[r], &presel);
 					goto nochange;
+				} else if (r == -1) {
+					waitbm = cfg.filtermode + 1;
+					cfg.filtermode = 1;
 				}
 
 				if (strcmp(path, newpath) == 0)


jarun:
@luukvbaal the filter key is `/` and is a per-directory thing. User can press `/` in the bookmark directory and select the bookmark. Why do we need to add 3 heavy functions and a global to achieve that?

I am not OK with adding heavy code for tiny conveniences. Why do we have to mimic the `/` keypress in code and change a workflow? What happens to the users who are already accustomed to press the `/` key to filter whenever needed?

jarun:
Also, we have put in a lot of effort to make nav-to-type as smooth as possible. It just takes some time to adapt to. Why can't users set that by default?

N-R-K:
> This provides the bookmark workflow I think @N-R-K was looking for, entering type-to-nav mode when pressing b to enter bookmark dir, and exiting type-to-nav mode if you were not in type to nav mode before pressing b.

Half-half. I said that I don't auto exit type-to-nav after selecting a bookmark, that was intended since I end up using that mode anyways. Also even if this were to be merged, I'd say this should've been a separate keybinding (maybe <kbd>ctrl+b</kbd>) so it doesn't disrupt existing workflow.

> we have put in a lot of effort to make nav-to-type as smooth as possible.

Have been using type-to-nav (with `-A`) quite a lot recently and I can attest to this.

luukvbaal:
> I am not OK with adding heavy code for tiny conveniences. Why do we have to mimic the / keypress in code and change a workflow?

I suppose it's a matter of definition. I wouldn't call this small diff "heavy code" and pressing 3 extra keystrokes a "minor inconvenience"... In my opinion any improvement made to the bookmark system should be considered as it is a convenience feature by definition.

> What happens to the users who are already accustomed to press the / key to filter whenever needed?

Like I said, no workflows are changed by the PR as it restores the nav-to-type mode to the same state it was in as before pressing <kbd>b</kbd>.

It's just that this way of navigating bookmarks (multi-key keystrokes prefixed by the bookmark key) is really convenient when u have a somewhat large list of bookmarks in my opinion and is what I used to use on other file managers.

I guess I'll just stop with the PR's though as anything I propose seems to be dismissed.

jarun:
> I guess I'll just stop with the PR's though as anything I propose seems to be dismissed.

Please discuss once before raising them. I have mentioned this multiple times.
I don't want to add code and keep increasing the binary size without discussions.

jarun:
> no workflows are changed by the PR as it restores the nav-to-type mode to the same state it was in as before pressing b

Not that way, due to muscle memory someone who doesn't use nav-to-type would press `/` in the bookmark directory now and actually lose the filter.

I think the middle ground is to set presel=FILTER when you enter the bookmark directory or plugins directory. Does that make sense?

jarun:
At least I am not sure why we need to enable the nav-to-type mode.

luukvbaal:
> Please discuss once before raising them. I have mentioned this multiple times.

 Isn't a PR the appropriate place to discuss if I have the patch ready anyways. If you immediately close the PR without discussion it feels dismissive but now with your follow up questions we are discussing it anyways...

> Not that way, due to muscle memory someone who doesn't use nav-to-type would press / in the bookmark directory now and actually lose the filter.

The bookmark directory is a new feature so there should be no muscle memory to speak of. The bookmark system should be a quick way to move around the system so entering type-to-nav mode makes sense there imo.

> At least I am not sure why we need to enable the nav-to-type mode.

Such that pressing e.g. <kbd>bcf</kbd> with a bookmark named `cf` to `$HOME/.config` will bring you to that bookmark.

It's fine though, I'll just use it locally.

jarun:
> Isn't a PR the appropriate place to discuss if I have the patch ready anyways.

It takes time to look into the code and make changes. Which is something I prefer contributors to `nnn` avoid. A prior discussion is always helpful.

> The bookmark directory is a new feature so there should be no muscle memory to speak of.

No, jumping to plugin directory is an existing use case and no one complained over years.

N-R-K:
> Such that pressing e.g. bcf with a bookmark named `cf` to `$HOME/.config` will bring you to that bookmark, without having to press // ENTER or Control-N x2.

There's one assumption here, we're assuming that the user isn't using `-A`. If `-A` is used (I find the auto cd a bit jarring so I have `A` on my `NNN_OPTS`), then he'll need to press <kbd>Enter</kbd> anyways.

jarun:
Yes, we are mixing things... a user who doesn't even use nav-to-type is now also seeing auto-dir selection. He is used to press Enter all the time!

luukvbaal:
> It takes time to look into the code and make changes. Which is something I prefer contributors to `nnn` avoid. A prior discussion is always helpful.

Sure but I was going to use this locally anyways so might as well open the PR if I have the patch.

> No, jumping to plugin directory is an existing use case and no one complained over years.

We are talking **bookmark** directory not plugin, it's literally listed in the current ToDo list of new features.

> There's one assumption here, we're assuming that the user isn't using `-A`. If `-A` is used (I find the auto cd a bit jarring so I have `A` on my `NNN_OPTS`), then he'll need to press Enter anyways.

Sure, in my opinion it would make sense to enable both `nav-to-type` AND `autoselect` when going into the bookmarks directory when pressing the <kbd>b</kbd> as it is the fastest way to navigate and you should know your bookmark keystrokes by head expecting to enter the directory. Subsequently `nav-to-type` and `autoselect` can be restored to their previous values from before pressing <kbd>b</kbd>.

> Yes, we are mixing things... a user who doesn't even use nav-to-type is now also seeing auto-dir selection. He is used to press Enter all the time!

I see no problem with introducing this behavior in a new feature(the bookmark directory and <kbd>b</kbd> key) but like I said I was ready to drop it.

N-R-K:
I have type-to-nav on my local build, but that's entirely personal and opinionated. If we want to have this feature in mainline nnn, *imo it shouldn't be implicit.* I'd prefer a configurable and explicit behavior. Maybe a user wants to use `regex` filters, or maybe just the vanilla `filter` instead of `type-to-nav`.

luukvbaal:
I don't think it's that opinionated if you restore the config(`nav-to-type`,`autoselect`). 

I do think a configurable behavior is overkill for this and probably best left in our local builds.

N-R-K:
> I do think a configurable behavior is overkill for this and probably best left in our local builds.

If we want to make it configurable at _runtime_ then yes, I can see this being too much code.

> By pressing b you expect the next keystrokes to indicate the bookmark you want to go to(same for `NNN_BMS` just with a single key), hence toggling `type-to-nav` makes sense. This PR was just meant to allow multi-key bookmark strokes in combination with the new symlink bookmarks. `type-to-nav` and `autoselect` were an easy way to achieve that without too much new code.

@jarun Do you think it'll be acceptable if this is done on a separate keybinding, e.g <kbd>ctrl+b</kbd> ? My concern about implicit behavior won't be valid anymore if it's a separate keybind.

luukvbaal:
I think the concern about implicit behavior is null if we document it since we are talking about a yet unreleased feature and <kbd>b</kbd> will do nothing if there is no bookmark directory. 

It makes sense to have the behavior be as convenient as possible and I think its hard to disagree that `type-to-nav` with `autoselect` makes sense for the bookmark directory🤷🏻‍♂️. In a sense the bookmark directory is then just an interface/prompt for the bookmarks if you name them accordingly.

But again if it's too much code or still too opinionated I'll just use it locally.

luukvbaal:
@N-R-K https://github.com/luukvbaal/nnn/commit/cc1590bdb6745523b5d050ddb74dedc5940514bd takes into account the `-A` flag and restores `cfg.autoselect` as well when leaving bookmark dir.

jarun:
If you press `Enter` at the plugin prompt, it takes you to the plugin dir. Same workflow.

jarun:
Anyway, when someone is not using nav-to-type in regular directories, I doubt he would care pressing the filter key when he is going into the bookmark/plugin dir. It's basically just another directory to that user and he would use his regular workflow. I don't want to add a special meaning to these directories.

luukvbaal:
> If you press `Enter` at the plugin prompt, it takes you to the plugin dir. Same workflow.

Sure, it will after the next release. This case also doesn't have to be affected by this proposal though.

My intention is that when <kbd>b</kbd> is pressed and it brings you directly to the bookmark directory i.e. ONLY IF NNN_BMS is unset and no directory is marked, not by pressing <kbd>Enter</kbd>; only then make sure `type-to-nav` mode is enabled and that it auto-selects the last match. Which is what this latest commit https://github.com/luukvbaal/nnn/commit/7dc9c205caf253bfa77cd483f9a8c52243a83a38 does while restoring `cfg.filtermode` after entering a directory and leaving `cfg.autoselect` unaffected.

quantum-booty:
I think the majority of the users will need to use / filter as the number of bookmarks expands. So it would be nice to give user the option to automatically enter filter or type-to-nav on bookmark key.

PR #1183, opened 2021-09-30
Disable e on picker mode

N-R-K:
Mainly so that pressing 'e' due to muscle memory on nnn.vim doesn't nest
a new editor instance inside the embedded vim terminal.

However invoking nnn with picker mode implies that the intention is to
pick file(s), there shouldn't be any business trying to edit things on
the fly. And if editing a file while in picker mode is desirable, then
'l' can be used for that instead.


jarun:
Thanks!

PR #1186, opened 2021-10-03
Added recursive capabilities to .nmv

### LINKED ISSUE #1184: Recursive Rename - Change filenames and folder structure in one place
**Is your feature request related to a problem? Please describe.**

Batch renaming files in subfolders - I'd like to be able to rename files in a way that also allows me to change the directories they live in, automatically make the directories etc, while being able to do this in a text editor (i.e. vim) to make certain renaming and moving tasks much simpler 

**Describe the solution you'd like**

An option in the batch renaming system to list all files recursively

**Describe alternatives you've considered**

I tend to do this via bash scripts when I need to rename files while also moving them

I'm happy to contribute a plugin if somebody can point me towards a guide on working with them, especially regarding how dependency management works. Trying to do this manually in bash feels like a complex problem.

KlzXS:
Resolves #1184


jarun:
Thank you!

PR #1189, opened 2021-10-03
Remove preview-tui winch tmux workaround

### LINKED ISSUE #1188: nnn closes previews when a window is opened and closed within same workspace

- [ ] Operating System: Solus (most recent)
- [ ] Desktop Environment: BSPWM
- [ ] Terminal Emulator: kitty
- [ ] Shell: bash
- [ ] Program options used: U, e, a, alias nnn='nnn -P p' (p:preview-tui)
```
# nnn config
[ -n "$NNNLVL" ] && PS1="nnn$NNNLVL $PS1"
export NNN_PLUG='p:preview-tui'
export NNN_FIFO=/tmp/nnn.fifo
export NNN_OPTS="Uea"
export LC_COLLATE="C"
export SPLIT='v'
ACTIVE_PLUGINS="p"
alias nnn='nnn -P $ACTIVE_PLUGINS'
# TODO: use the symlink bookmarks when nnn -v is 4.3

n ()
{
    # Block nesting of nnn in subshells
    if [ -n $NNNLVL ] && [ "${NNNLVL:-0}" -ge 1 ]; then
        echo "nnn is already running"
        return
    fi

    # To cd on quit only on ^G, either remove the "export" 
    export NNN_TMPFILE="${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.lastd"

    nnn "$@"

    if [ -f "$NNN_TMPFILE" ]; then
            . "$NNN_TMPFILE"
            rm -f "$NNN_TMPFILE" > /dev/null
    fi
}


```

Open kitty. Run ```n``` with above in .bashrc. nnn opens with preview-tui working as expected. Open any new window in same desktop window. Then close it. The preview is gone in the original window

luukvbaal:
Resolves #1188

This was here to prevent a zombie preview pane when starting `nnn` with `preview-tui` open with a command such as `tmux new-session nnn -Pp`. Starting `tmux` in this way seems to mess with the sigwinch behavior for some reason. This was however a dumb workaround and also implemented incorrectly (should have been `&&` not `||`) causing `preview-tui` to close when started by nnn without the `-a autofifo` flag.

I'll see if I can find a proper fix for the `tmux` problem.


luukvbaal:
I would like to keep some sort of workaround in. This will work for a single `nnn` instance but a zombie preview pane will be left over if multiple `nnn` instances are running.

jarun:
Thank you!

PR #1190, opened 2021-10-07
Save session after browse()

luukvbaal:
`save_session()` after returning from `browse()` since this was skipped by the picker mode `SEL_OPEN` control flow: https://github.com/jarun/nnn/blob/2435263052661afbfd433ffc20b0519affc23866/src/nnn.c#L6827 

I think this fixes https://github.com/mcchrish/nnn.vim/issues/109 without any side-effects.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7734,11 +7734,6 @@ nochange:
 					break; // fallthrough
 			}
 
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

Issue #1133, opened 2021-08-17
ToDo list


jarun:
Rolled from #1122.

### Cooking

- [x] add `bookmarks` directory for flexible symlinked bookmarks
- [x] new key <kbd>B</kbd> to add a symlinked bookmark for current dir
- [x] sync multiple instances of `nnn` after operation on selection
- [x] save complete per-context filter when saving sessions
- [x] disable symlink resolution for paths in `NNN_BMS` and arg `PATH`
- [x] do not end selection mode before running plugins/prompt/shell
- [x] plugin `finder` will now retain search filter history
- [x] plugin `bookmarks` removed with symlinked bookmarks support
- [x] list open locations in active contexts in help page
- [x] make option `O_MATCHFLTR` to discard filter key if no match
- [x] configurable `NNN_TMPFILE` to _cd on quit_
- [x] disable auto marking directories (use <kbd>-</kbd>)
- [x] picker mode improvements
  - open tty for input if `STDIN` is non-tty
  - truncate output file before writing
  - do not double select a file on <kbd>Enter</kbd>
- [x] no redraw during du calculation, show only current dir name
- [x] plugin `xdgdefault`: add dmenu support

### Up for grabs

- [ ] notcurses UI backend (#1014) [@KlzXS]
- [ ] open new terminal window in spawned shell dir (#759)

For anything else please discuss in this thread.

**Contribution [guideline](https://github.com/jarun/nnn/blob/master/misc/CONTRIBUTING.md).**

N-R-K:
Small nitpick, but I think having the ascii logo above the version/license would be better. Reasoning is that the help screen should prioritize showing useful information like keybinding more.

jarun:
Nah. I want to show it off.

CantoroMC:
Hello,
this is are some of my entries for NNN_PLUG 
- e:-!&emacs $nnn
- f:-!&pickOne (this is a zsh script that open a random file with xdg-open)
- d:-!git diff

The first two doesn't work anymore (but they used to); probably something has changed but i'm unable to find the documentation.

jarun:
Please try master. If they do not work, please find the commit that may have broke them. I'm sorry but I don't have the time to find that out for you.

CantoroMC:
I'm on master, it may have happen in the last two days, i will take a look which is the guilty commit and report back.

I've asked to know if plugin syntax have changed without being documented.

CantoroMC:
[...]

PR #1191, opened 2021-10-07
Fix gitstatus regression and restorepreview conflict

luukvbaal:
https://github.com/jarun/nnn/issues/1133#issuecomment-938463943


jarun:
Thank you!

PR #1196, opened 2021-10-14
Support OSX iterm2 in preview-tui by using split pane

anphung:
Support OSX iterm2 in preview-tui by using split pane. Tmux doesn't work great as it's showing pixelated images.

#### Demo:

Screen Shot 2021-10-14 at 01 31 53
.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -67,6 +67,8 @@
 #   has to be set to `all` or `splits` (the former is the default value).
 #   This terminal is also able to show images without extra dependencies.
 #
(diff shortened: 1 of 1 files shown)


jarun:
Nice improvement! Please see the review comment.

jarun:
Also, why can't users use the `-a` option?

anphung:
A `get_terminal()` function could be introduced, but it's out of scope that I don't want to make this PR too be, so leave it as it.

luukvbaal:
Thanks again!

PR #1202, opened 2021-10-20
Added %j and %J for using selection in shell

KlzXS:
Example usage:
`echo %s` - list every file in selection, one per line
`echo %S` - list every file in selection, all at once.

%s and %S cannot be combined inside a single command.
Multiple %s can be used in a single command.
We should probably print a warning instead of silently ignoring a command, but I didn't know exactly where that would go so I left it out.

Perhaps we should implement some sort of escape. Currently you aren't able to run something like `printf "%s" "some string"` as the `%s` would be replaced.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5391,12 +5391,12 @@ static bool prompt_run(void)
 
 		cnt_J = 0;
 		next = cmdline;
-		while (!cnt_j && (next = strstr(next, "%J"))) {
+		while ((next = strstr(next, "%J"))) {
 			++cnt_J;
 
 			tmplen = xstrsncpy(tmpcmd, cmdline, next - cmdline + 1) - 1;
-			tmplen += xstrsncpy(tmpcmd + tmplen, "${0} ${@}", sizeof("${0} ${@}"));
-			xstrsncpy(tmpcmd + tmplen, next + 2, len - (next - cmdline + 2));
+			tmplen += xstrsncpy(tmpcmd + tmplen, "${0} ${@}", sizeof("${0} ${@}")) - 1;
+			xstrsncpy(tmpcmd + tmplen, next + 2, len - (next - cmdline + 2) + 1);
 
 			++next;
 		}


jarun:
I think the double shell is fine. We are running commands at the prompt.

Regd. the space thing, we can go with the proposal from @N-R-K and document the limitation on double quotes. It's a reasonable trade off for the power we are giving here to the users. I don't want to change to single quote because we use that more often for possessive nouns.

N-R-K:
This could fix the double quote problem and give us a list of files on which `eval` could be run.

```sh
sed 's/"/\\" /g'  ~/.cache/nnn/selection | xargs -0 printf '"%s" ' 
```

Completely untested.

KlzXS:
I'm not getting round to testing this at all. I'll get to it soon. Perhaps after the other issue where you taggee me. Somehow I missed that.

jarun:
Sure thing!

KlzXS:
If you have an example of `"%j"` failing to do the proper thing please share. I don't see it.
The first shell is because the code invokes whatever `$SHELL` is, the second one is needed in case of `%J` so that I can control where all the files are placed. If we mandate it be only at the end of the command we can simply use `xargs` without an extra shell and it will automatically properly esace spaces and everything else.

You would think that you could tell `xargs` to place more than one thing at `{}` but it seems that's not the case. So the double shell it was.

jarun:
I don't see any issue with `%j` or the double shell.

But we need to take care of the `%J`. Filenames having spaces in them do not work.

KlzXS:
I don't see a good way of dealing with `%J` and escapes. If we limit it to appearing only at the end of the command then we can let `xargs` take care of it. If there is a way to do it in the middle of a command (which I will admit it a bit of an odd place to put it) it would be very hacky.

We could also remove the double shell that way.

jarun:
Yes, it's ok to limit `%J` at the end and document. Please raise the PR.

jarun:
@KlzXS did you get time for this?

PR #1204, opened 2021-10-21
fix: no mouse support inside vim terminal

N-R-K:
we need to signal mouse event by printing these escape codes. the X10
escape code seems to be legacy format only used as fallback incase SGR
extended coordinates aren't supported.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2079,6 +2079,10 @@ static bool initcurses(void *oldmask)
 	//intrflush(stdscr, FALSE);
 	keypad(stdscr, TRUE);
 #ifndef NOMOUSE
+#ifndef NOX11
+	printf("\x1b[?1000h\n"); /* xterm X10 mouse mode */
+	printf("\x1b[?1006h\n"); /* extended SGR mouse mode */
+#endif
 #if NCURSES_MOUSE_VERSION <= 1
 	mousemask(BUTTON1_PRESSED | BUTTON1_DOUBLE_CLICKED | BUTTON2_PRESSED | BUTTON3_PRESSED,
 		  (mmask_t *)oldmask);


luukvbaal:
Oops you're right, my bad.

jarun:
> Hmm I suppose now it makes sense to let the user send files to explorer fifo/picker output with SEL_CLICK.

Do we do that currently? I'm all in for the enhancement though I never use the mouse. ;)

luukvbaal:
No currently the file is just opened, we need to check for `g_state.picker/fifomode` in `SEL_CLICK`. I don't use the mouse either but since we went through the trouble of fixing this users might now expect it to work inside the vim plugins.

luukvbaal:
Yeah probably, it looks like we're already checking for explorer mode in `SEL_CLICK` but it's not working for me:
https://github.com/jarun/nnn/blob/5f76b69c36b971de1c6baafb24072dbac2f940a2/src/nnn.c#L6755

luukvbaal:
I think it makes sense to also open the file in picker mode yeah.

luukvbaal:
Something like this I think for explorer mode
```diff
diff --git a/src/nnn.c b/src/nnn.c
index fed3aaf1..4bcf3e79 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -6752,7 +6752,8 @@ nochange:
                                if (r != cur)
                                        move_cursor(r, 1);
 #ifndef NOFIFO
-                               else if ((event.bstate == BUTTON1_PRESSED) && !g_state.fifomode)
+                               else if ((event.bstate == BUTTON1_PRESSED) &&
+                                               (!g_state.fifomode || !(pdents[cur].flags & DIR_OR_DIRLNK)))
                                        notify_fifo(TRUE); /* Send clicked path to NNN_FIFO */
 #endif
                                /* Handle right click selection */
```

jarun:
Look for

    6773                                 /* Single click just selects, double click falls through to SEL_OPEN */

and it falls thorugh to sel_open for double click.

jarun:
In fact, I think in picker mode, single click should add to selection.

jarun:
So if it is double click, change `sel` to SEL_OPEN.

> Hmm, why if right light already does that?

Because there is no right on mobiles. ;)

luukvbaal:
Yeah 
```diff
@@ -6788,6 +6788,7 @@ nochange:
                                copycurname();
                                goto nochange;
                        }
+                       sel = SEL_OPEN;
 #endif
                        // fallthrough
                case SEL_NAV_IN: // fallthrough
```
does the trick.

luukvbaal:
I'm not sure how you want to change the mouse click select behavior, I'll open a PR with the change required to have mouse multiple selection picker mode working in combination with explorer mode.

PR #1215, opened 2021-11-01
plugin: fzopen: print full path for picker mode

N-R-K:
Closes: https://github.com/mcchrish/nnn.vim/issues/133

@jarun Since I don't use the plugin IDK if this might have any side-effects.

### DIFF
--- a/plugins/fzopen
+++ b/plugins/fzopen
@@ -42,7 +42,7 @@ fi
 if [ "$3" ]; then
     if [ "$entry" ]; then
         case "$entry" in
-            /*) fullpath="${2}/$entry" ;;
+            /*) fullpath="$PWD/$entry" ;;
             *)  fullpath="$entry" ;;
         esac
         if [ "-" = "$3" ]; then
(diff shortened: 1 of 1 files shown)


N-R-K:
Am also assuming that `$entry` is only filename and not fullpath. If entry is fullpath somehow then this wouldn't work.

plugins/fzopen
```diff
@@ -42,9 +42,9 @@ fi
 if [ "$3" ]; then
     if [ "$entry" ]; then
         if [ "-" = "$3" ]; then
-            printf "%s\n" "$entry"
+            printf "%s\n" "${2}/$entry"
```
jarun, line 45: I would say don't assume anything and check if the path starts with `/` or not. If not, assume it's relative to the current directory (where fzopen is issued).

jarun (changes requested):

plugins/fzopen
N-R-K: > If not, assume it's relative to the current directory (where fzopen is issued).

Should I use `$2` or `$PWD` then?

jarun: I think `$PWD` would be the right thing to use.

N-R-K:
Okay, we should be good. Just need a regular user of this plugin to test and confirm this doesn't break anything.

jarun:
Thank you!

PR #1217, opened 2021-11-02
nuke: port sxiv-rifle performance improvement over

N-R-K:
this commit is mostly porting over some recent performance improvements
from `sxiv-rifle`: https://github.com/ranger/ranger/pull/2411

there's one "bug-fix" in this commit. currently a file named "afilejpeg"
would be matched due to the `-iregex` in listimages. this commit changes
that so only extensions would match, so for example "afile.jpeg" would
match but not "afilejpeg".

As for performance, there's a couple things this commit does:

* store the result of listimages into a tmp file instead of calling that
  function twice, this is probably the biggest performance improvement.
  especially when loading large directories.

* abspath now sets the var abs_target instead of calling printf. since
  abspath is only called from load_dir, we can go one step further and
  inline it. but i haven't done that since the function might be useful
  later on.

* avoid call to dirname and use parameter subsitution instead inside
  `listimages`

* use grep instead of `-iregex`, it's POSIX compliant and can be faster.

i've tested this out with sxiv and everything seems to be working as
expected.

### DIFF
--- a/plugins/nuke
+++ b/plugins/nuke
@@ -259,27 +259,32 @@ handle_extension() {
     esac
 }
 
+# sets the variable abs_target, this should be faster than calling printf
 abspath() {
     case "$1" in
-        /*) printf "%s\n" "$1";;
-        *)  printf "%s\n" "$PWD/$1";;
+        /*) abs_target="$1";;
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1218, opened 2021-11-02
Resolve symlinks one level for .nmv when in list mode

KlzXS:
`NNN_LIST` is set automatically if we are in list mode and we are in the tmp directory.

Resolving is done using `readlink(1)`. This utility reads out what the symlink points to (the thing `ls -l` would display). We always use absolute paths so that's fine. The resolving is done while reading in the input, this way most of the script can stay untouched.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7269,8 +7269,9 @@ nochange:
 			case SEL_RENAMEMUL:
 				endselection(TRUE);
 				setenv("INCLUDE_HIDDEN", xitoa(cfg.showhidden), 1);
-				if (listpath && is_prefix(path, listpath, xstrlen(listpath)))
-					setenv("NNN_LIST", "1", 1);
+				if (listpath)
+					setenv("NNN_LIST",
+						xitoa(is_prefix(path, listpath, xstrlen(listpath))), 1);
 
 				if (!(getutil(utils[UTIL_BASH])
 				      && plugscript(utils[UTIL_NMV], F_CLI))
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -7269,6 +7269,8 @@ static bool browse(char *ipath, const char *session, int pkey)
 			case SEL_RENAMEMUL:
 				endselection(TRUE);
 				setenv("INCLUDE_HIDDEN", xitoa(cfg.showhidden), 1);
+				if (listpath && is_prefix(path, listpath, xstrlen(listpath)))
+					setenv("NNN_LIST", "1", 1);
```
jarun, line 7273: Do we need to clear `NNN_LIST` after the operation? What happens if the same operation is run once again?

jarun (changes requested):

src/nnn.c
jarun: I mean multiple rename from a regular directory? As `NNN_LIST` remains set, the plugin will treat it as a case of listing.

KlzXS: We do.

KlzXS:
CI fails on `browse()` complexity. Can't we do something to silence that for good? This is like the 10th time I've seen this that I can remember.

src/nnn.c
KlzXS: I used the return value of `is_prefix()` for the value in the latest commit. Please check the indentation, I'm not on my usual system. That's why all the commits are unverified.

jarun:
Thank you!

> CI fails on browse() complexity.

I don't want the function to cross 900 lines. I will reduce somewhere.

PR #1220, opened 2021-11-05
Add nerd icons to gitstatus patch

luukvbaal:
Changes:
* gitstatus nerd icons when built with `O_NERD=1`
* use `git status` flag `--norenames` as this detection is not working when providing `pathspec` (which we do to improve performance when not showing hidden files) and does not work at all anyways when renaming with nnn (on my system). Git only detects the rename when I use `git mv` so not worth checking for renames imo.
* logic for the non NERD case is probably less efficient now but should be insignificant and I don't see the need to differentiate/maintain both cases.


jarun:
Thank you!

PR #1221, opened 2021-11-06
Added plugin usermount

chrimbo:
The plugin for easy mounting/unmounting of devices/network shares which are in /etc/fstab and have the user option set.

I use this for my network shares on my NAS. They are not automatically mounted on bootup because the NAS isn't always available.

### DIFF
--- a/plugins/usermount
+++ b/plugins/usermount
@@ -1,0 +1,58 @@
+#!/usr/bin/python3
+#
+# Description: Mount and unmount user filesystems in /etc/fstab.
+#              If the filesystem is mounted, it will be unmounted.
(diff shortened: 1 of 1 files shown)


jarun:
Please try to write this in bash. Python has tons of dependencies.

jarun:
@luukvbaal @N-R-K can you guys try this in sh? Also, do we need it?

luukvbaal:
I don't see the need for such a plugin myself.
I would look into https://wiki.archlinux.org/title/fstab#Automount_with_systemd.

Will allow you to automatically mount a network share when you try to access it.

jarun:
Thanks! Closing this.

PR #1223, opened 2021-11-07
Fix invisible description

akrifari:
### Summary 

One of the inline command descriptions is invisible due to unscape pipe character in code block.

#### Before
select-1636264300

### After
select-1636275865


jarun:
Thank you!

PR #1226, opened 2021-11-09
Proper fix for preview-tui zombie-pane

luukvbaal:
Think I finally found a proper fix for the zombie-pane issue described in https://github.com/jarun/nnn/pull/1189#issuecomment-933029985.

By sleeping for a short time we can make sure the sub-shell spawned by `SIGWINCH` is killed when exiting the preview loop.


jarun:
Thank you!

Issue #1228, opened 2021-11-10
Kitty preview image dimension is cut in half


ahloiscreamo:
Hello,
i'm reporting a possibly bug on kitty icat 

- [x ] Operating System: `Linux artix 5.14.16-zen1-1-zen #1 ZEN SMP PREEMPT Wed, 03 Nov 2021 07:49:34 +0000 x86_64 GNU/Linux`
- [x ] Desktop Environment:
`i3 version 4.20.1 (2021-11-03) © 2009 Michael Stapelberg and contributors`
- [x ] Terminal Emulator:`kitty 0.23.1 created by Kovid Goyal`
- [x ] Shell:`fish, version 3.3.1`
- [x ] Program options used:`nnn -r -e -x -C`
- [x ] Configuration options set: via config.fish
```
#NNN
export NNN_FIFO="/tmp/nnn.fifo nnn"
export NNN_OPENER="xdg-open"
export NNN_OPENER_DETACH=1
export EDITOR="nvim"
export NNN_EDITOR="/usr/bin/nvim"
export NNN_VISUAL="nvim"
export NNN_COLORS="453"
export NNN_FCOLORS="0B0B04060006060009060B06"
export NNN_BMS='m:/media;g:/gallery;d:~/Downloads;p:~/Pictures;t:~/Documents;f:~/Desktop'
export NNN_PLUG='j:autojump;p:preview-tui;l:launch;r:renamer;w:wallpaper;o:organize;x:xdgdefault'                  

#Bat
export BAT_THEME="Nord"

#Alias
alias nnn="nnn -r -e -x -C" 
alias icat="kitty +kitten icat"
```
- [x] Issue exists on  `nnn 4.3-1`

#### Exact steps to reproduce the issue

Launch nnn, activate preview-tui and see the image loaded but the dimension is wrong, sometimes the dimension is small, sometimes it cut half on terminal especially when the image is landscape.

However i did not track on when this behavior  happen in the first place, because i usually view images by opening straight with sxiv.
so i don't know on which version it started to happen. All the preview script are updated to the latest version.

video for references -

windowed terminal

https://user-images.githubusercontent.com/9318372/141144988-959957aa-0996-4bbe-a246-8a3da9649018.mp4

full-screen terminal

https://user-images.githubusercontent.com/9318372/141145790-74a20fc3-7312-4f1f-893b-ac5646976b9c.mp4

So that is the behavior.
[...]

PR #1231, opened 2021-11-10
Preview-tui workaround for ncurses tput regression

luukvbaal:
Workaround for `tput` regression introduced by `ncurses-6.3` as discussed in #1228.
Write terminal dimensions to tmpfile 🤷🏻‍♂️


jarun:
Thank you!

PR #1234, opened 2021-11-11
Preview-tui ensure pts device for tput in subshell

luukvbaal:
~~We can pass the tty device of the main shell in an env variable rather than writing to a tmpfile.~~

Removing redirection of sub-shells restores `tput` behavior without other workarounds for `ncurses-6.3`.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -91,12 +91,11 @@ FIFOPID="$TMPDIR/nnn-preview-tui-fifopid.$NNN_PARENT"
 PREVIEWPID="$TMPDIR/nnn-preview-tui-pagerpid.$NNN_PARENT"
 CURSEL="$TMPDIR/nnn-preview-tui-selection.$NNN_PARENT"
 FIFO_UEBERZUG="$TMPDIR/nnn-preview-tui-ueberzug-fifo.$NNN_PARENT"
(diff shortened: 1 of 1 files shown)


plugins/preview-tui
```diff
@@ -451,6 +447,7 @@ else
         printf "\$KITTY_LISTEN_ON not set!\nPlease read Usage in preview-tui."
         cfg=$(stty -g); stty raw -echo; head -c 1; stty "$cfg"
     else
-        toggle_preview "$1" &
+        TTY="$(tty)"
```
N-R-K, line 450: On zsh `$TTY` seems to be predefined. Could be worth checking.

luukvbaal: Yeah I'm aware, didn't think it mattered though.

luukvbaal: It's not predefined on `sh` or `bash`, but even if it was we would be overwriting it to the same value.

luukvbaal:
I noticed that the missing file descriptor for the pts device is not created due to error redirection. I assume because both stdin and stdout are redirected for the sub-shell, the pts device is not "inherited" from the main shell.

When removing error redirection,`tput` works as expected (and image previews are the correct dimensions).

Adding WIP label to see if I can find a more elegant fix.

jarun:
Thank you!

PR #1235, opened 2021-11-12
Preview-tui pass pts device in env variable

luukvbaal:
Seems I was too quick in my testing of #1234... That commit makes `tput` return other than default dimensions, but NOT the correct dimensions. This actually fixes the issue.


jarun:
Thank you!

PR #1240, opened 2021-11-15
Enforce %J only at the end of command

KlzXS:


jarun:
Just thinking... a problem with this limitation is, many common commands like cp, mv etc. can't be used.

N-R-K:
> Just thinking... a problem with this limitation is, many common commands like cp, mv etc. can't be used.

Should be able to just use `"%j"` for those instead.

jarun:
Can we do the following?

If `%J` is not at the end, even before processing anything can we replace it with `%j`?
Is there any case where this wouldn't work?

KlzXS:
I think there are quite a lot of places where that would produce an unexpected result. I remember that `convert` (ImageMagik tool) needed inputs to be placed in the middle of the command. `ffmpeg` also has a similar thing.

jarun:
Got it! But keep thinking of a better solution if possible. Please document the limitation as well!

Thank you!

KlzXS:
The wiki is updated. I will report if I stumble upon a way to escape everything properly.

jarun:
Thank you!

PR #1241, opened 2021-11-23
Guard open_with getreadline for picker

luukvbaal:
Copied the [guard](https://github.com/jarun/nnn/blob/256f0d008fe9d28699aebb0155ab31664c06d683/src/nnn.c#L5386) already present in [`prompt_run()`](https://github.com/jarun/nnn/blob/256f0d008fe9d28699aebb0155ab31664c06d683/src/nnn.c#L5374) which I assume is there for the same reason.

fix luukvbaal/nnn.nvim#25

Edit: Yeah looked through git blame to find the original relevant issue: https://github.com/mcchrish/nnn.vim/issues/39.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7519,10 +7519,13 @@ nochange:
 				tmp = xreadline(tmp, messages[MSG_ARCHIVE_NAME]);
 				break;
 			case SEL_OPENWITH:
-#ifdef NORL
-				tmp = xreadline(NULL, messages[MSG_OPEN_WITH]);
-#else
-				tmp = getreadline(messages[MSG_OPEN_WITH]);
+#ifndef NORL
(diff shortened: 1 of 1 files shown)


jarun:
This is because we exit the curses mode when we call `getreadline()`. We can't do that when we are inside a plugin e.g the (n)vim plugins.

jarun:
Thank you!

luukvbaal:
> This is because we exit the curses mode when we call `getreadline()`. We can't do that when we are inside a plugin e.g the (n)vim plugins.

Yeah I figured `exitcurses()` was the issue here. Although the issue was present for picker mode in general (i.e. `nnn -p -` in terminal), not just inside the plugins. Guess the underlying reason is just the picker mode redirection.

PR #1243, opened 2021-11-24
misc sed tweaks

omar-polo:
Hello,

re-reading #1210 I noticed that @0xACE highlighted a chunk of code in `archive_selection` that is no more necessary.  If I've understood the discussion correctly, `gsed` is now mandatory, so why don't use it?  It makes archive_selection slightly more robust on every platform.  Tested on OpenBSD and the `z` key seems to be working fine.

While there I've also wrapped the ifdef block that defines SED in another ifndef, this way porters can override the `SED` constant without patching.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2729,12 +2729,7 @@ static void archive_selection(const char *cmd, const char *archive, const char *
 	}
 
 	snprintf(buf, CMD_LEN_MAX,
-#ifdef __linux__
 		SED" -ze 's|^%s/||' '%s' | xargs -0 %s %s", curpath, selpath, cmd, archive
-#else
-		"tr '\\0' '\n' < '%s' | "SED" -e 's|^%s/||' | tr '\n' '\\0' | xargs -0 %s %s",
-		selpath, curpath, cmd, archive
-#endif
 		);
 	spawn(utils[UTIL_SH_EXEC], buf, NULL, NULL, F_CLI | F_CONFIRM);
 	free(buf);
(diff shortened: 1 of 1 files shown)


jarun:
Is GNU sed installed by default on OpenBSD?

omar-polo:
of course not, but from what I'm understanding GNU sed is a dependency of nnn, or at least that's what I presume from the ifdef that defines SED as gsed on Darwin and the BDSs.

jarun:
Makes sense.

jarun:
Thank you!

omar-polo:
cheers :)

N-R-K:
Wondering if it might be worthwhile to try to do these things internally and avoid GNU-sed as dep. As far as I'm seeing we're depending on it for the `-z` and `-i` flags, correct? Or are there more reasons?

omar-polo:
for one, I'd be happier to see less `xargs -0 -n2 sh -c '...'` sprinkled around :D

jarun:
Yes, `-z` and `-i` are the two. But we don't want to spawn a shell for every file either.

KlzXS:
> Wondering if it might be worthwhile to try to do these things internally

I thinks most of the sed scripts don't do overly complicated string manipulations, but each one of the 10ish of them does something different.

It was simply faster to just offload the work to the shell than to implement separate functions. Also it takes up less space.

> I'd be happier to see less xargs -0 -n2 sh -c '...' sprinkled around

I know at least one of them is necessary to pass through `/dev/tty` so that we can have interactive prompts for overwriting (`see P_CPMVRNM`). This could be avoided if we were to use GNU xargs, since the POSIX version doesn't support the needed `-o`.

PR #1244, opened 2021-11-24
fix the malloc size in get_archive_cmd

N-R-K:
since the tr hack was removed in deead97, the format string is no longer
70 chars. also removes unnecessary malloc casting.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2719,10 +2719,10 @@ static void get_archive_cmd(char *cmd, const char *archive)
 	xstrsncpy(cmd, archive_cmd[i], ARCHIVE_CMD_LEN);
 }
 
+#define CMD_FMT " -ze 's|^%s/||' '%s' | xargs -0 %s %s"
 static void archive_selection(const char *cmd, const char *archive, const char *curpath)
 {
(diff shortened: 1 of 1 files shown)


N-R-K:
The 2nd commit (578e6d1aaa00b8f051d65fc5617400529e7bd52c) makes things a bit more robust as changing the format string would automatically change the malloc size. Though I assume it might be a bit controversial, can revert it if that's not OK @jarun

src/nnn.c
```diff
@@ -2719,23 +2719,22 @@ static void get_archive_cmd(char *cmd, const char *archive)
 	xstrsncpy(cmd, archive_cmd[i], ARCHIVE_CMD_LEN);
 }
 
+#define CMD_FMT " -ze 's|^%s/||' '%s' | xargs -0 %s %s"
```
jarun, line 2722: Could you add this to `patterns[]` so we have all of these at the same location?

jarun (changes requested):

src/nnn.c
N-R-K: done.

jarun:
Thank you!

PR #1247, opened 2021-11-26
Remove -E from sed in plugins

KlzXS:
Please pass it through some light tests to see if I missed some spot.

### DIFF
--- a/plugins/dups
+++ b/plugins/dups
@@ -28,17 +28,17 @@ printf "\
 " > "$tmpfile"
 
 # shellcheck disable=SC2016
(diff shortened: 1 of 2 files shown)


jarun:
@N-R-K can you please check?

N-R-K:
The `xdgdefault` changes are fine, tested it out a bit. @KlzXS if you got GNU-sed try to use the `--posix` flag to disable all extension and see if everything works as expected or not. If it does then it's most likely ok.

jarun:
@KlzXS please confirm once good to go.

KlzXS:
I tried it out. Everything seems in order. Both with `--posix` and without.

jarun:
Thank you!

Issue #1256, opened 2021-12-14
Hide `preview-tui` plugin when editing a file


UtkarshVerma:
**Is your feature request related to a problem? Please describe.**

I am planning to use `preview-tui` plugin permanently. However, one issue I'm facing is the fact that the preview pane doesn't close even if I go to edit a file.

**Describe the solution you'd like**
When the user wants to edit a file(in vim), the preview pane should be hidden until nnn gains control of the terminal again.

>  Also, before exiting, `nnn` should close the preview pane itself as `tmux` closing the unclosed preview pane looks sloppy.

**Describe alternatives you've considered**
Maybe give the users an option to configure what happens when an edit/open action is invoked through *edit hooks*.

**Additional context**
![image](https://user-images.githubusercontent.com/31820255/146020667-c1a3d692-6ee7-4e49-8cad-d5380e06806b.png)

luukvbaal:
> I _think_ the [restorepreview](https://github.com/jarun/nnn/tree/master/patches#list-of-patches) patch does what you're asking for.

Yeah it does.

> Also, before exiting, `nnn` should close the preview pane itself as `tmux` closing the unclosed preview pane looks sloppy.

What do you mean by this?

UtkarshVerma:
Thanks for the quick reply. I tried the `restorepreview` patch but I still don't see any difference. Is there any way I can verify if it's working. I can confirm that the patch got applied during `make`.

![image](https://user-images.githubusercontent.com/31820255/146025551-f87e4e4a-ac58-418c-ac7b-378c727ba5be.png)


Regarding the other thing, when I close `nnn`, it exits first and the preview-pane exits just after that. This results in the terminal being blank for a short duration and it doesn't look good.
It's not a show-stopper but from UX point of view, things like this annoy me.

luukvbaal:
Well, the patch writes `"close\n"` to the fifo path when editing a file. If you have the latest version of `preview-tui`, it will listen for this "opcode" and exit.

I agree it would be better in terms of UX but I doubt @jarun will welcome the required logic overhaul to the plugin system for such a use case.

UtkarshVerma:
Apparently, the patch was working, well kind of.
The preview pane is properly hidden if I edit a Makefile or LICENSE file. But for some reason, it doesn't go away for Markdown or python files. Could you please let me know how I can get this to work.

luukvbaal:
[...]

PR #1257, opened 2021-12-14
Restorepreview patch improvements

luukvbaal:
* Replace workaround
* Also close/restore preview for subshell/prompt

As discussed in https://github.com/jarun/nnn/issues/1256#issuecomment-996443636.


jarun:
Thank you!

PR #1271, opened 2021-12-23
add support for previewing pngs in preview-tui

PatrickF1:
tested this locally

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -225,7 +225,7 @@ print_bin_info() {
 
 handle_mime() {
     case "$2" in
-        image/jpeg) image_preview "$cols" "$lines" "$1" ;;
+        image/jpeg | image/png) image_preview "$cols" "$lines" "$1" ;;
         image/gif) generate_preview "$cols" "$lines" "$1" "gif" ;;
         image/*) generate_preview "$cols" "$lines" "$1" "image" ;;
         video/*) generate_preview "$cols" "$lines" "$1" "video" ;;


luukvbaal:
Have you tried previewing a png without this change? Should work fine as it's handled by https://github.com/jarun/nnn/blob/bd6e7e81479823fdf28cdc507027982011039c9f/plugins/preview-tui#L230

PatrickF1:
Interesting. Maybe I'm missing one of the deps my screen is definitely empty!

luukvbaal:
The point of `generate_preview` is that it converts images to jpg as it is faster to preview. This does indeed require a dependency (ImageMagick) however it will fall back to the exact same line executed by this PR so I'm not sure why it wouldn't work for you:
https://github.com/jarun/nnn/blob/bd6e7e81479823fdf28cdc507027982011039c9f/plugins/preview-tui#L363-L367

luukvbaal:
Do you perhaps have a `convert` command in your `PATH` that is not the one from ImageMagick? Only thing I can think of.

PatrickF1:
convert isn't on my path

N-R-K:
You can try removing the null redirections `>/dev/null` around that area to see if anything interesting shows up.

PatrickF1:
Sure I'll try that, thanks!

PatrickF1:
Well, I have truly no idea what happened but after deleting the /dev/null it worked. So then I ran the plugin installer again and it worked. Maybe I had accidentally modified it before and broke it. 🤷‍♂️

PR #1275, opened 2021-12-25
Pedantic cleanups

N-R-K:
Fixes a couple small warnings when compiled with additional warning flags (`-Wpointer-arith`, `-Wstrict-prototypes`).

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1322,7 +1322,7 @@ static void msg(const char *message)
 }
 
 #ifdef KEY_RESIZE
-static void handle_key_resize()
+static void handle_key_resize(void)
 {
 	endwin();
 	refresh();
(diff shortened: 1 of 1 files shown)


KlzXS:
Genuine curiosity. Why is `void f(void)` considered better than `void f()`?

I've seen that warning before, but not with `gcc` and `clang` I think. Never understood why anyone would bother with that.

Other changes are reasonable.

N-R-K:
> Genuine curiosity. Why is `void f(void)` considered better than `void f()`?

The first one declares function f which takes no arguments, so if one tries to pass argument to it the compiler will detect that.

The second one declares function f with ***unspecified*** arguments, the compiler cannot check anything in this case.

```c
void fun1() { return; }
void fun2(void) { return; }

int main(void) {
    fun1(10, "string"); // should compile fine
    fun2(1); // error
}

jarun:
Thank you!

Issue #1283, opened 2022-01-04
[preview-tui] add option for pdf text preview


sweenu:
Pdf image previews are a good thing with a terminal that supports graphics, but with half-blocks it's not as useful as you cannot read the pdf.

Therefore, it'd be a good thing to have an option to use `pdftotext` and preview the text of the pdf instead of the pdf itself.

luukvbaal:
Be aware that the `ueberzug` option for preview-tui works regardless of whether your terminal supports graphics and displays the images in original quality. Closing as per #1284.

PR #1284, opened 2022-01-04
preview-tui: allow pdf text preview

sweenu:
Closes https://github.com/jarun/nnn/issues/1283

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -25,6 +25,7 @@
 #   - optional: ffmpeg for audio thumbnails
 #   - optional: libreoffce for opendocument/officedocument preview
 #   - optional: pdftoppm(poppler) for pdf thumbnails
+#   - optional: pdftotext(poppler) for pdf text preview. Export PDFTOTEXT=1 to use this
 #   - optional: gnome-epub-thumbnailer for epub thumbnails (https://gitlab.gnome.org/GNOME/gnome-epub-thumbnailer)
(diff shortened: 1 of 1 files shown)


plugins/preview-tui
```diff
@@ -246,7 +251,11 @@ handle_mime() {
 handle_ext() {
     case "$2" in
         epub) generate_preview "$cols" "$lines" "$1" "epub" ;;
-        pdf) generate_preview "$cols" "$lines" "$1" "pdf" ;;
+        pdf) if [ "$PDFTOTEXT" = 0 ]; then
```
N-R-K, line 254: Should use `-eq` instead.
```suggestion
        pdf) if [ "$PDFTOTEXT" -eq 0 ]; then
```

luukvbaal:
It is recommended to use `preview-tui` with one of the image preview options that displays images in original quality.

I would say original quality pdf preview for the first page is more useful than `pdftotext`. If you do prefer `pdftotext` I would suggest to make the change locally. We can't make an envvar for each change in previewer I would say.

sweenu:
It's up to you, but the way I see it, preview-tui is a genric preview plugin, not only for images but also for texts, markdown, code, etc. so many might want to use the plugin for its text previews primarily, not caring for the images.
I already made the change locally, you can go ahead and close this if you think no one else would want that :)

luukvbaal:
I'm not saying no one else would want that, but I don't see the point in adding a new preview option locked behind an envvar. Plugins are indicative.

PR #1287, opened 2022-01-08
Remove trailing blanks to assure .desktop suffix for xdg-mime.

lawnowner:
I remember this plugin worked without any problems when I last used it, but now trailing blanks in the name of the .desktop file seem to cause xdg-mime to give "Malformed argument" error.

### DIFF
--- a/plugins/xdgdefault
+++ b/plugins/xdgdefault
@@ -49,5 +49,5 @@ app=$(find "$@" -iname '*.desktop' -exec grep '^Name=' {} + \
     | cut -f 2)
 
 if [ -n "$app" ]; then
-    xdg-mime default "$app" "$ftype"
+    xdg-mime default "${app%%[[:blank:]]*}" "$ftype"
 fi


plugins/xdgdefault
```diff
@@ -49,5 +49,5 @@ app=$(find "$@" -iname '*.desktop' -exec grep '^Name=' {} + \
     | cut -f 2)
 
 if [ -n "$app" ]; then
-    xdg-mime default "$app" "$ftype"
+    xdg-mime default "${app%%[[:blank:]]*}" "$ftype"
```
N-R-K, line 52: If you're just trying to remove trailing blanks, then a single `%` should be used, `%%` is the greedy version.

lawnowner: Unfortunately, this is where greedy does it: it's a wildcard pattern, '*' is not a regex quantifier.

N-R-K: Can a valid `$app` contain spaces?

lawnowner: The name of a .desktop file should be a valid D-Bus well-known name, and thus it can't contain blanks, that's why there's no need to isolate the file name (e.g. with `sed`), but with multiple trailing spaces, for example, one '%' removes only one space.

lawnowner: `$app` potentially contains trailing blanks because of the `column`'s tabulation (`-t/--table`).

lawnowner: `sed` or `awk` can be used instead of `cut`, in which case `-o "$(printf '\t')"` becomes unnecessary too, e.g. `awk '{ print $NF }'`, or cut's output can be piped to `tr -d '[:blank:]'`, which are unnecessary.

N-R-K: > The name of a .desktop file should be a valid D-Bus well-known name, and thus it can't contain blanks

If it's not valid to have spaces between `$app`, then this should be fine. But could also try this instead, just in case

```sh
${trim%${trim##*[![:space:]]}}
```

Untested. But it's from the [pure sh bible](https://github.com/dylanaraps/pure-sh-bible#trim-leading-and-trailing-white-space-from-string) so I assume it should work.

lawnowner: `${trim%${trim##*[![:space:]]}}` isn't fine with leading blanks, which isn't a problem here since the columns are left aligned. But `"${app%%[[:blank:]]*}"` does the job cheaply enough for now.

jarun:
Thank you!

PR #1288, opened 2022-01-08
add support for imv when named imv in nuke

Anomalocaridid:
While the `imv` binary is named `imvr` in some Linux distros, such as Ubuntu, it is named `imv` in others, such as Arch Linux and its derivatives.


jarun:
Thank you!

Issue #1292, opened 2022-01-23
ToDo list


jarun:
Rolled from #1219.

### Cooking

- disable filter info if file details (option `-i`) enabled
- open previous active context on context quit
- switch <kbd>^J</kbd> and <kbd>+</kbd> functionality:
  - <kbd>+</kbd>: toggle file selection
  - <kbd>^J</kbd>: toggle auto-jump on file open
- allow symlink creation with name `@` to a single file (#1345)
- clear selection on successful operation at native prompt with "%j" (#1330)
- reverse timestamps of entries modified/created within 5 minutes
- avoid using non-portable `xargs` flags on macOS (#1299)
- quitcd script for Elvish shell > 0.17.0 (#1344)
- plugin `openall` to open selected files together (#1333)
- plugin `gitroot` to jump to git root directory from a subtree
- plugin `gsconnect` to send the selected files to Android using gsconnect
- icon for opus and webp files
- `preview-tui` - fix gif conversion and whitespace name
- `preview-tui` - add support for windows terminal split
- `preview-tui` - djvu file previews
- `nuke` - add support for `imv` when named _imv_
- `gsconnect` - support connection to multiple devices
- export `NNN_INCLUDE_HIDDEN` to plugins (#1308)
- respect `NNN_TRASH` in `.nmv` (#1306)
- add GNU sed as a dependency with support for env var `SED`
- use bold `>` to point at current entry in detail mode
- add 2 spaces after icons for better visibility
- documentation refresh
- make option `O_NOSORT` to load directories unsorted on entry

### Up for grabs

None open at the time.

For anything else please discuss in this thread.

**Contribution [guideline](https://github.com/jarun/nnn/blob/master/misc/CONTRIBUTING.md).**

rieje:
Is it possible to save nnn session automatically without `Q`, e.g. if closed terminal without manually exiting nnn?

Also would like the option of `-S` (persistent session) to not save the session if all contexts are closed. That would work nicely for my wrapper script where first argument would be checked if it's a directory/file and if so, open it with nnn, otherwise, create a session. Sometimes I realize I don't want the session saved anymore after working with it, so it would be nice to quit all contexts to prevent the session from being saved instead. I don't see any benefits of saving a session when all contexts are closed and this would be an option anyway.

Would be nice to be able to "redraw" a filtered list.

[...]

PR #1306, opened 2022-02-13
Respect NNN_TRASH in .nmv

KlzXS:
In response to https://github.com/jarun/nnn/issues/1292#issuecomment-1037334056

I'm not familiar with the functioning of the trash programs so I don't know if any flags are needed.
`rm` has an interactive flag, but since trashing doesn't automatically delete files I take it that isn't even needed.

### DIFF
--- a/plugins/.nmv
+++ b/plugins/.nmv
@@ -23,6 +23,15 @@ INCLUDE_HIDDEN="${INCLUDE_HIDDEN:-0}"
 VERBOSE="${VERBOSE:-0}"
 RECURSIVE="${RECURSIVE:-0}"
 
+case "$NNN_TRASH" in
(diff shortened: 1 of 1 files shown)


N-R-K:
`trash-put` doesn't need any flags, idk about `gio trash`. Also should we grep through all the plugins to see if there's any other plugin where it makes sense to respect `NNN_TRASH` ?

KlzXS:
`NNN_TRASH` isn't used anywhere else.
`dups` removes files. But it removes identical files, what's the point in putting it in the trash if you have an identical copy elsewhere? Most other result of `grep rm` are tmpfile removals.

jarun:
Thank you!

PR #1307, opened 2022-02-15
don't leak inotify_fd to plugins

N-R-K:
Note: there still might be fds leaking away since I only *skimmed* through the open() calls, didn't inspect them thoroughly (there also might be functions other than `open` which are opening fds). But this seems to be sufficient to solve the issue with `tree`.


jarun:
Thank you!

PR #1313, opened 2022-02-20
dragdrop: use "dragon-drop" if it exists

N-R-K:
to avoid name conflict some distros renamed dragon to `dragon-drop`
https://github.com/mwh/dragon/issues/17#issuecomment-889878960


jarun:
👍

PR #1314, opened 2022-02-20
Added a user-script/plugin to send selected files using GSConnect.

Darukutsu:
It is questionable if this should be added since only GNOME users will profit from this.
Supports multiple files at once transfer(somehow with kdeconnect-cli it wasn't working for me)


plugins/gsconnect
```diff
@@ -0,0 +1,25 @@
...
+# Shell: POSIX compliant
+# Author: Darukutsu 
+
+selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
+gsconnect=$HOME/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js
+id=$($gsconnect -l)
+
+if [ -s "$selection" ]; then
+    IFS=$'\n'
+    for file in $(cat -t $selection | sed "s/\^\@/\n/g"); do
```
N-R-K, line 16: This won't be able to deal with filenames with newlines in them. Use `xargs -0` instead.

plugins/gsconnect
```diff
@@ -0,0 +1,25 @@
...
+    	$gsconnect -d "$id" --share-file="$file"
+    done 
+else
+    notify-send -a "GSConnect" "No files selected"
+fi
```
N-R-K, line 21: Shouldn't it just use the hovered file instead? Also think it's better to just print the msg instead of using notify-send, not everyone has notifications enabled.

N-R-K (changes requested):

plugins/gsconnect
```diff
@@ -0,0 +1,19 @@
...
+#              GSConnect must be configured on the Android device and the PC.
+#
+# Shell: POSIX compliant
+# Author: Darukutsu 
+selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
+gsconnect=$HOME/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js
+id=$($gsconnect -l)
+
+if [ -s "$selection" ]; then
+	xargs -0 -a $selection -i $gsconnect -d "$id" --share-file="{}"
```
N-R-K, line 14:
```suggestion
    xargs -0 < "$selection" -I{} "$gsconnect" -d "$id" --share-file="{}"
```

plugins/gsconnect
```diff
@@ -0,0 +1,19 @@
...
+    # Clear selection
+    printf "-" > "$NNN_PIPE"
+else
+	$gsconnect -d "$id" --share-file="$(pwd)/$1"
```
N-R-K, line 18:
```suggestion
    "$gsconnect" -d "$id" --share-file="$2/$1"
```

N-R-K (changes requested):
Looks better. Suggested some style and quoting issue below, also should probably use `$2` rather than pwd, nnn sends the directory as the 2nd argument.

N-R-K (approved):

N-R-K:
@jarun The plugin code looks ok to me, though I haven't run it since Im not using gnome, assuming @Darukutsu has and it works for him. The ci seems to be failing due to something else.

jarun:
Thank you!

Issue #1308, opened 2022-02-17
Detecting whether hidden files are active when running a plugin


gzagatti:
**Is your feature request related to a problem? Please describe.**

I have written a basic plugin that looks for directories and then switches to the selected one, similar to `fzcd`

**Describe the solution you'd like**

I would like the plugin to know whether hidden files are active or not. If so, then I would include hidden directories in my search.

**Describe alternatives you've considered**

Mapping two different keys to two variations of the plugin.

**Additional context**

My plugin is very simple. It can be found [here](https://github.com/gzagatti/dotfiles/blob/refactor-kitty/nnn/plugins/fcd).

KlzXS:
I see you've already prepared it for accepting `INCLUDE_HIDDEN`. @jarun I think it would be simple enough to always set the variable when calling the function (and showing hidden files), currently we only do it on batch rename. Are we okay with that?

N-R-K:
If we do decide to make it available to all plugins, I think it'd make sense to prefix it with `NNN_`. So `$NNN_INCLUDE_HIDDEN` , just in case it doesn't collide with other preexisting env var.

gzagatti:
Having an environment variable like `$NNN_INCLUDE_HIDDEN` that toggles according to whether hidden files are visible or not in `nnn` would completely solve my issue.

jarun:
@KlzXS yes, please add the support.

PR #1315, opened 2022-02-22
export NNN_INCLUDE_HIDDEN to plugins

N-R-K:
Closes: https://github.com/jarun/nnn/issues/1308

---

Think this should do the trick. Plugins can check if hidden files is active or not via `[ "$NNN_INCLUDE_HIDDEN" -eq 1 ]`.


jarun:
Thank you!

PR #1321, opened 2022-03-08
preview-tui: make debugging easier

N-R-K:
i've noticed that there's a non-trival amount of issues user face
related to preview-tui. allowing all the null redirection to be easily
switched on/off can make debugging much easier, both for the users and
for us.


plugins/preview-tui
```diff
@@ -96,6 +97,12 @@ PREVIEWPID="$TMPDIR/nnn-preview-tui-pagerpid.$NNN_PARENT"
 CURSEL="$TMPDIR/nnn-preview-tui-selection.$NNN_PARENT"
 FIFO_UEBERZUG="$TMPDIR/nnn-preview-tui-ueberzug-fifo.$NNN_PARENT"
 
+if [ "$DEBUG_LOG" -eq 0 ]; then
+    logfile="/dev/null"
```
luukvbaal, line 101: Great idea :) I think `LOGFILE` instead of `logfile` is more consistent with our shellscript variable style, at least in preview-tui.

N-R-K: Went with `DEBUG_LOGFILE`; a tad bit more consistent with the variable `DEBUG_LOG` itself.

luukvbaal (approved):

jarun:
Thanks guys!

PR #1327, opened 2022-03-21
added cddir plugin

muppi090909:
Add the plugin cddir
# Dependencies
1. fzf

This plugin is different from cdpath as this cd's into any child dir and not just the ones in the CDPATH enviorment variable.


muppi090909:
The latest commit combines the directories under CDPATH with the ones under PWD.

jarun:
No more fzf dependent trivial plugins please! Have your local copy.

PR #1329, opened 2022-03-22
avoid unnessary heap allocation

N-R-K:
the binary size shouldn't change since it'll most likely go into the bss
section.


jarun:
Thank you!

PR #1330, opened 2022-03-23
native prompt: clear selection on successful operation

N-R-K:
have been using the %j substitution for a bit now, it's quite nice :)
but it's also quite annoying when doing something like

	mv "%j" ./

in this case, the files have been moved but the selection is incorrectly
points to them.

when there's a successful operation made on the selection via the native
prompt, it makes sense to clear the selection.


jarun:
Thank you!

PR #1338, opened 2022-04-09
Allow gsconnect to handle multiple devices

JingMatrix:
When gsconnect is connected to multiple devices, the original
script won't work. Now the code will share files to all devices.


jarun:
Thank you!

PR #1341, opened 2022-04-12
plugin(dragdrop): unquote $all to prevent possible empty argument

jonathannerat:
When using the `dragdrop` plugin to select multiple files (via selection / current directory), it's possible to specify an optional argument to drag all files at once, or separately. If we leave this option empty, the line that calls `$dnd` passes an empty argument (`"$all"`), which results in an error as displayed here:

image

This PR just unquotes that `$all` so the shell ignores it if we leave it empty.

### DIFF
--- a/plugins/dragdrop
+++ b/plugins/dragdrop
@@ -54,10 +54,10 @@ fi
 
 if [ "$resp" = "s" ]; then
     use_all
-    sed -z 's|'"$PWD/"'||g' < "$selection" | xargs -0 "$dnd" "$all" &
+    sed -z 's|'"$PWD/"'||g' < "$selection" | xargs -0 "$dnd" $all &
 elif [ "$resp" = "d" ]; then
     use_all
-    "$dnd" "$all" "$PWD/"* &
+    "$dnd" $all "$PWD/"* &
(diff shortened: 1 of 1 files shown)


N-R-K:
> If we leave this option empty, the line that calls `$dnd` passes an empty argument (`"$all"`), which results in an error as displayed here:

I can't reproduce this, could you provide more details on how to trigger this?

jonathannerat:
Sure, I have the plugin mapped to `d` (`NNN_PLUG="d:dragdrop"`), and `NNN_FIFO="$HOME/.cache/nnn.fifo"`.

- Open `nnn` without any arguments:

  Open nnn

- Select multiple folders / files:

  desktop and documents folders selected

- Trigger the `dragdrop` plugin, the options show up:

  Options for dragdrop

- Choose "Drag selection (s)" (`s<Enter>`), then select the default option for "mark --all" (i.e. just `<Enter>`):

  image

- Error shows up:

  image

jonathannerat:
I had an outdated version (1.1.1) because the [AUR package](https://aur.archlinux.org/packages/dragon-drop) name changed (was `dragon-drag-and-drop`), my bad. Thanks for the help!

Issue #1346, opened 2022-04-19
[Feature] use emojis as a simple solution for icons


eirinnm:
**Is your feature request related to a problem? Please describe.**

Problem: getting icons in nnn requires jumping through hoops with third party packages: either patching your system font with nerdfont, or using icons-in-terminal, which appears to be abandoned (no updates in 5 years, PRs getting ignored).

**Describe the solution you'd like**

nnn should have an option to use emojis as icons. Emojis are widely supported now. For instance, ubuntu 18.04 supports colour emojis. 

Obviously, we can't have unique icons for every filetype. But here are some examples, which would make nnn really nice to use:

📄 file  
📂 folder  
📑 executable file  
🔗 symbolic link
📦️ archive

Since emojis are just unicode codepoints and the actual artwork is provided by the operating system, this feature should not cause any bloat.

N-R-K:
> Emojis are widely supported now. 

Untrue.

1. Many DIY distros don't include emoji font by default
2. Any terminal using XFT will either crash trying to display color emojies or will ignore color emoji font in order to avoid crashing. https://git.suckless.org/st/file/FAQ.html#l224
3. I don't believe (most ?) bitmap fonts, which are quite common among terminal users, support emojies.

> Since emojis are just unicode codepoints and the actual artwork is provided by the operating system, this feature should not cause any bloat.

Depends on the font. `noto-emoji` is 165Mb for example. Just because it comes preinstalled on some systems doesn't make it non bloated.

MuhammedZakir:
Not specfically related to nnn: if you want to avoid patching or other headaches, you can use a terminal that support user-configured font fallback option, such as WezTerm and foot. This makes it really easy to use your favorite font along with, for example, nerd icons.

cainnz:
thank you for all of you guys response, @MuhammedZakir I will go ahead and try one of those terminals. I really love nnn! thank you for all you guys do.

eirinnm:
>Do you see the contradiction that arises from these two statements? Yes, this way or that your system font (or rather the one your terminal uses) has to support displaying the desired symbol. 

I don't see the contradiction; in one case, I have to patch my fonts, in the other case, it's already been done. 

[...]

PR #1135, opened 2021-08-18
change and add some icons-nerdfont

CantoroMC:
# More Icons

1. These icons are more pleasant in my opinion

| Name            | Old        | New        |
| :-------------- | :--------: | :--------: |
| ICON_VIDEOFILE  | ﳜ "\ufcdc" |  "\uf72a" |
| ICON_ARCHIVE    |  "\uf53b" | 遲"\ufac3" |
| ICON_JAVASCRIPT |  "\ue74e" |  "\uf81d" |
| ICON_EXT_CSS    |  "\ue614" |  "\ue749" |

2. These instead are some additions and a "fix".<br>
  Added icon for zst archives and cabal files, added COLOR_C to cpp icon

PR #1356, opened 2022-05-05
Icons using emoji

dkabus:
I have implemented icons using emoji such that it works in my version of the suckless terminal st. This addresses the feature request #1346.


Makefile
```diff
@@ -21,6 +21,7 @@ O_NOFIFO := 0  # no FIFO previewer support
 O_CTX8 := 0  # enable 8 contexts
 O_ICONS := 0  # support icons-in-terminal
 O_NERD := 0  # support icons-nerdfont
+O_EMOJI := 1  # support emoji
```
N-R-K, line 24: Should probably be 0 by default.

Makefile
```diff
@@ -227,6 +234,9 @@ static:
 	# static binary with patched nerd font support
 	make O_STATIC=1 O_NERD=1 strip
 	mv $(BIN) $(BIN)-nerd-static
+	# static binary with emoji support
+	make O_STATIC=1 O_EMOJI=1 strip
+	mv $(BIN) $(BIN)-icons-static
```
N-R-K, line 239:
```suggestion
	mv $(BIN) $(BIN)-emoji-static
```

src/icons.h
```diff
@@ -24,8 +26,13 @@ struct icon_pair {
 
 #define ICON_PADDING_LEFT  ""
 #define ICON_PADDING_LEFT_LEN  (0)
+#if defined(EMOJI)
+#define ICON_PADDING_RIGHT " "
+#define ICON_PADDING_RIGHT_LEN (2)
+#else
 #define ICON_PADDING_RIGHT "  "
 #define ICON_PADDING_RIGHT_LEN (2)
+#endif
```
N-R-K, line 35: Why the conditional? Both branches seem to be doing the same thing.

N-R-K (dismissed):
Glancing over the patch, here's a couple issues that stands out.

src/icons.h
dkabus: I am padding the emoji with two spaces vs one. This is kind of the ugly part of my implementation. Honestly, I do not understand how emoji are rendered well enough to know why this works, but it does.

Makefile
dkabus: I need to set O_EMOJI to 1 for it to properly compile. If I do it via `make O_EMOJI=1 clean all`, the emoji are not properly displayed.

src/icons.h
N-R-K: Oh, my bad. I didn't notice that the emoji branch had different amount of spaces. But in that case the `*_LEN` should've been adjusted accordingly as well.

Makefile
N-R-K: That's weird, I don't see anything obvious to why that might be happening. Could you double check?

dkabus: Nevermind. It does in fact compile properly that way 🤦

src/icons.h
dkabus: Well its complicated. I think it works like this: Some of the glyphs are two characters wide, others are just one character. I made sure to add spaces where nessessary such that the strings stored in the compiler variables `EMOJI_*` all contain two characters. In the drawing process, to all of them one additional space is appended, namely `ICON_PADDING_RIGHT`. In total we now have three characters (`[emoji][maybe emoji][space]`). With the other icon sets, the icons are represented by only one character each (`[icon][space][space]`). So, this is why `ICON_PADDING_RIGHT` and `ICON_PADDING_RIGHT_LEN` do not match up here. This can probably done in a more elegant way, but this should probably be done by someone more familiar with the source code.

N-R-K: The following seems to be working fine for me.

```diff
diff --git a/src/icons.h b/src/icons.h
index 7bfa7c8..4f3f243 100644
--- a/src/icons.h
+++ b/src/icons.h
@@ -26,11 +26,12 @@ struct icon_pair {
 
 #define ICON_PADDING_LEFT  ""
 #define ICON_PADDING_LEFT_LEN  (0)
-#define ICON_PADDING_RIGHT_LEN (2)
 #if defined(EMOJI)
 #define ICON_PADDING_RIGHT " "
+#define ICON_PADDING_RIGHT_LEN (1)
 #else
 #define ICON_PADDING_RIGHT "  "
+#define ICON_PADDING_RIGHT_LEN (2)
 #endif
 
 #define COLOR_VIDEO        93  /* Purple */
```

This is in XTerm

image

Not sure if it's a good idea to mix different width emojies though.

dkabus: This (at least for me) does not work with long filenames that have to be truncated.

On the left: your proposed patch, on the right, my proposal:
screenshot_20220506_165626

N-R-K: Aright, I've digged into this a bit. The patch I sent earlier is correct, the `LEN` should indeed be 1.

The problem is elsewhere, I think it's because `nnn` expects all the icons to be one cell wide.

I think the nerd icons also suffers similar problem, which is why we rejected any double cell wide icons : https://github.com/jarun/nnn/pull/1135#issuecomment-901298575

N-R-K: https://github.com/jarun/nnn/pull/1356/commits/936d6fb0a825a0d05a9b096c8b9f3951a92b4174 Should fix the issue.

N-R-K:
The cell width issue should be fixed by https://github.com/jarun/nnn/pull/1356/commits/936d6fb0a825a0d05a9b096c8b9f3951a92b4174, but still not entirely sure if mixing single and double width emojies are a good idea or not.

Looks ugly IMO.

image

src/icons.h
```diff
@@ -60,6 +72,10 @@ static const struct icon_pair exec_icon = {"", FA_COG,    0};
 static const struct icon_pair dir_icon  = {"", ICON_DIRECTORY, 0};
 static const struct icon_pair file_icon = {"", ICON_FILE,      0};
 static const struct icon_pair exec_icon = {"", ICON_EXEC,      0};
+#elif defined(EMOJI)
+static const struct icon_pair dir_icon  = {"", EMOJI_FOLDER, 0};
+static const struct icon_pair file_icon = {"", EMOJI_FILE, 0};
+static const struct icon_pair exec_icon = {"", EMOJI_EXE,    0};
```
N-R-K, line 78: Let's respect the existing naming convention, so `EMOJI_FOLDER` -> `EMOJI_DIRECTORY` and `EMOJI_EXE` -> `EMOJI_EXEC`.

src/icons.h
```diff
@@ -93,6 +109,29 @@ static const struct icon_pair icons_name[] = {
...
+	{"Videos",       EMOJI_MOVIE,      0},
+	{"Uni",          EMOJI_UNI,        0},
+	{"University",   EMOJI_UNI,        0},
+	{"Work",         EMOJI_WORK,       0},
+	{"Games",        EMOJI_GAME,       0},
+	{"Cloud",        EMOJI_CLOUD,      0},
+	{"Nextcloud",    EMOJI_CLOUD,      0},
+	{"Literature",   EMOJI_BOOKS,      0},
+	{"Code",         EMOJI_CODE,       0},
+	{"Notes",        EMOJI_NOTES,      0},
```
N-R-K, line 130: The existing icons don't have these rules and some of these ("Uni", "Nextcloud" etc) seems better suited toward a personal build.

src/icons-emoji.h
```diff
@@ -0,0 +1,66 @@
+#ifndef ICONS_EMOJI
+#define ICONS_EMOJI
+
```
N-R-K, line 3: Add a comment explaining that double width emojies need to be manually right padded with space.

N-R-K (changes requested):

src/icons.h
dkabus: Sure, I've changed `EXE` -> `EXEC`. Nerd uses "directory" and icons uses "folder". I will just leave it.

src/icons.h
dkabus: Yes, indeed. I will keep "Games", "Cloud", "Literature", "Code", and "Notes" though, as these seem reasonably general.

N-R-K (approved):
Other than the mixing of single and double width emojies, looks good.

jarun:
Thank you so much!
Personally I like it very much.

jarun:
The game and code emoji are not showing correctly on xubuntu 22.04, xfce4-terminal. Can you please have a look?

jarun:
@dkabus At commit af94c7704fcb3c9549fec2397b79c1c799346f51 I removed all the additional emojis you added. I think I made it very clear in the defect that I wanted something minimal, not with additional bells and whistles. This is a public project, we do not want non-standard stuff. If you need those for your custom distro, please patch downstream.

Makefile
```diff
@@ -269,14 +280,19 @@ upload-local: sign static musl
 	curl -XPOST 'https://uploads.github.com/repos/jarun/nnn/releases/$(ID)/assets?name=$(BIN)-nerd-static-$(VERSION).x86_64.tar.gz' \
 	    -H 'Authorization: token $(NNN_SIG_UPLOAD_TOKEN)' -H 'Content-Type: application/x-sharedlib' \
 	    --upload-file $(BIN)-nerd-static-$(VERSION).x86_64.tar.gz
+	# upload emoji compiled static binary
+	tar -zcf $(BIN)-emoji-static-$(VERSION).x86_64.tar.gz $(BIN)-icons-static
```
N-R-K, line 284: @jarun Just noticed this while trying to test the released emoji static binary. It `tar`-ed the icons-in-terminal binary instead of the emoji binary.

PR #1364, opened 2022-05-13
Reference dependency needed for kitty preview

cain-dev:


N-R-K:
Should probably go at the top where all the deps are listed.

cain-dev:
As long as it's declared as a dependency for kitty's image preview it's fine.

jarun:
Thank you!

PR #1366, opened 2022-05-23
Improve preview-tui kitty documentation

### LINKED ISSUE #1365: I think the documentation of preview-tui with kitty should be changed.
I met a problem with using kitty to preview images, finally, I found the problem is setting kitty's `listen_on "unix:$TMPDIR/kitty"`, where the double quotes should be removed! I stuck there for a long time, I hope the documentation can be modified！

luukvbaal:
Resolves #1365.

Edit: Oops, I see I accidently branched on the main repo instead of my fork. Don't usually use github directly for PRs, but I'm away from home sorry.


jarun:
Thank you!

PR #1373, opened 2022-05-29
Enter type-to-nav for bookmarks dir

luukvbaal:

PR #1374, opened 2022-05-29
Enter type-to-nav for plugin dir

luukvbaal:
RE https://github.com/jarun/nnn/pull/1373#issuecomment-1140461536: I think this covers everything, without too many extra checks. Please check if you agree. I think this is a nice alternative to the `fzplug` plugin.

Also enables auto-enter for `g_state.selbm` and `g_state.runplugin`.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -367,12 +367,13 @@ typedef struct {
 	uint_t rangesel   : 1;  /* Range selection on */
 	uint_t runctx     : 3;  /* The context in which plugin is to be run */
 	uint_t runplugin  : 1;  /* Choose plugin mode */
-	uint_t selbm      : 2;  /* Select bookmark dir and store type-to-nav */
+	uint_t selbm      : 1;  /* Select a bookmark from bookmarks directory */
+	uint_t filtermode : 2;  /* Store type-to-nav when entering plugin/bookmarks directory */
 	uint_t selmode    : 1;  /* Set when selecting files */
 	uint_t stayonsel  : 1;  /* Disable auto-jump on select */
 	uint_t trash      : 2;  /* Trash method 0: rm -rf, 1: trash-cli, 2: gio trash */
 	uint_t uidgid     : 1;  /* Show owner and group info */
-	uint_t reserved   : 5;  /* Adjust when adding/removing a field */
+	uint_t reserved   : 4;  /* Adjust when adding/removing a field */
 } runstate;
 
 /* Contexts or workspaces */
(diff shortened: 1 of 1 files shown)


N-R-K:
Without looking at the code, I'm wondering what usecase does this serve? Having ttn mode on in bookmarks dir allows for quickly navigating the bookmarks. I haven't seen anyone ask for ttn mode on plugins dir.

luukvbaal:
It only enters ttn in plugin selection mode(i.e. pressing <kbd>;Enter</kbd>, not when navigating there menually).

I think it is a fair assumption to make that the user would like to easily filter/select the plugins and ttn provides a native way to do so.

Basically it is a builtin alternative to `fzplug` like I said. Do you think it is undesired behavior?

N-R-K:
> It only enters ttn in plugin selection mode(i.e. pressing ;Enter, not when navigating there menually).
> 
> I think it is a fair assumption to make that the user would like to easily filter/select the plugins and ttn provides a native way to do so.
> 
> Basically it is a builtin alternative to `fzplug` like I said. Do you think it is undesired behavior?

Ah, I misunderstood what this was doing then. And no, I think this is a fair improvement and makes sense.

src/nnn.c
```diff
@@ -6683,6 +6681,12 @@ static bool browse(char *ipath, const char *session, int pkey)
 		setdirwatch();
 	}
 
+	if (g_state.filtermode) {
```
jarun, line 6684: A new check every time we jump to begin. This is what I was talking about. I have to think about it.

jarun (changes requested):

jarun:
This adds open condition checks in flows. That's what I was indicating. I'll think about it.

src/nnn.c
luukvbaal: Sure, I understand why you would be apprehensive of adding a check there. I just hoped the convenience factor would outweigh the performance consideration :)

Let me know and I'll raise the PR for the `restorepreview` patch conflict(dependent on this PR) and patch conflict resolve instructions later.

luukvbaal: Keep in mind that this block will almost never execute, the first check will only evaluate to true when the user presses <kbd>b/;Enter</kbd>.

jarun: It's just a matter of pressing /.

jarun:
Guys I have decided not to go with this. I will also revert the bm hack. Can't litter the code with roundabout checks to save on a single keypress (`/`).

PR #1377, opened 2022-05-30
Add patch conflict instructions

luukvbaal:
Add instructions and `--merge` flag in `Makefile` to make the steps required somewhat simpler.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -312,28 +312,28 @@ clean:
 
 prepatch:
 ifeq ($(strip $(O_NAMEFIRST)),1)
-	patch --forward --strip=1 --input=$(NAMEFIRST)/mainline.diff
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

N-R-K:
If `--merge` requires GNU patch, wouldn't it be better not to use it in the Makefile?

luukvbaal:
Yeah I guess so. We can revert and update the instructions.

N-R-K:
Have the following macro in the Makefile, `PATCH_OPTS` and have it be empty by default. And then in the ci invoke make as `make PATCH_OPTS="--merge"`. Which I think should do what we're trying to do without making GNU patch a build time dep.

PR #1380, opened 2022-05-31
Update patch conflict instructions/workflow

luukvbaal:
Per suggestion from @N-R-K: https://github.com/jarun/nnn/pull/1377#issuecomment-1141561291. I think it is a good solution.


N-R-K (approved):

jarun:
Thank you!

PR #1381, opened 2022-05-31
Added 'Repository'-icon

cmachmeier75:
As a daily nnn-user, I found it helpful to add an icon for my repositories-folder.


cmachmeier75:
1. The directory is called "Repositories"!
2. XDG **Defaults** are just that!
It's not a standard that specifies how the $HOME-subdirectory structure has to be like.
3. You wrote "most"? So, not all! :)

N-R-K:
Scratch the custom icons idea, we do icons based on name as well as extension. Trying to do it through env var will be messy.

cmachmeier75:
Alright, I will patch downstream instead

PR #1384, opened 2022-06-02
fix incorrect usage of sizeof

N-R-K:
`sizeof(cmd)` here would give the sizeof a char pointer, not the sizeof
the cp/mv array.


jarun:
Thank you!

PR #1386, opened 2022-06-05
Add support for more archive formats by using bsdtar by default.

eNV25:
### why

I created this PR because I couldn't extract a `.tar.zst` file with nnn. `bsdtar` supports many more formats than `atool`, the current default.

### issues

We let the command error out for unsupported formats. I think this is fine as long as we show the error to the user.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -776,7 +776,12 @@ static const char * const patterns[] = {
 	SED" -i 's|^\\(\\(.*/\\)\\(.*\\)$\\)|#\\1\\n\\3|' %s",
 	SED" 's|^\\([^#/][^/]\\?.*\\)$|%s/\\1|;s|^#\\(/.*\\)$|\\1|' "
 		"%s | tr '\\n' '\\0' | xargs -0 -n2 sh -c '%s \"$0\" \"$@\" < /dev/tty'",
-	"\\.(bz|bz2|gz|tar|taz|tbz|tbz2|tgz|z|zip)$",
+	// https://man.archlinux.org/man/atool.1#ARCHIVE_TYPES
+	// https://man.archlinux.org/man/libarchive-formats.5
+	// https://github.com/libarchive/libarchive/blob/master/tar/creation_set.c#L69
+	// https://github.com/libarchive/libarchive/blob/master/tar/creation_set.c#L90
+	// https://github.com/libarchive/libarchive/blob/master/tar/creation_set.c#L111
+	"\\.(tar|zip|jar|rar|lha|7z|alz|ace|a|ar|arj|arc|rpm|deb|cab|cpio|iso|mtree|xar|warc|t?(gz|grz|bz|bz2|Z|lzma|lzo|lz|lz4|lrz|xz|rz|uu|zst))",
 	SED" -i 's|^%s\\(.*\\)$|%s\\1|' %s",
 	SED" -ze 's|^%s/||' '%s' | xargs -0 %s %s",
 };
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -776,7 +776,12 @@ static const char * const patterns[] = {
 	SED" -i 's|^\\(\\(.*/\\)\\(.*\\)$\\)|#\\1\\n\\3|' %s",
 	SED" 's|^\\([^#/][^/]\\?.*\\)$|%s/\\1|;s|^#\\(/.*\\)$|\\1|' "
 		"%s | tr '\\n' '\\0' | xargs -0 -n2 sh -c '%s \"$0\" \"$@\" < /dev/tty'",
-	"\\.(bz|bz2|gz|tar|taz|tbz|tbz2|tgz|z|zip)$",
+	// https://man.archlinux.org/man/atool.1#ARCHIVE_TYPES
+	// https://man.archlinux.org/man/libarchive-formats.5
+	// https://github.com/libarchive/libarchive/blob/master/tar/creation_set.c#L69
+	// https://github.com/libarchive/libarchive/blob/master/tar/creation_set.c#L90
+	// https://github.com/libarchive/libarchive/blob/master/tar/creation_set.c#L111
+	"\\.(tar|zip|jar|rar|lha|7z|alz|ace|a|ar|arj|arc|rpm|deb|cab|cpio|iso|mtree|xar|warc|t?(gz|grz|bz|bz2|Z|lzma|lzo|lz|lz4|lrz|xz|rz|uu|zst))",
```
eNV25, line 784: Perhaps, I should remove `jar|rpm|deb`. Probably also `arc` because it's ambiguous. Not sure about `iso`.

eNV25: I decided to keep `iso` because with this patch, you can list contents and mount with archivemount.

I am not sure about `jar|rpm|deb`. Users most probably want to install packages with their system's package manager or run jar files with java, not extract or list or mount. On the other has there is already an `o` option to open with the default handler.

eNV25:
> Are you aware there is an [`NNN_ARCHIVE`](https://github.com/jarun/nnn/wiki/Usage) environment variable available to you to this end?

I wasn't aware. But that's not the only change.

I can remove the regex change if the maintainer doesn't like it.

jarun:
`bsdtar` is a wrapper over libarchive. It has issues with the popular format `rar`:

https://github.com/libarchive/libarchive/issues/151

jarun:
Other advantage of using atool is extraction to the current directory with a parent directory created automatically. bsdtar doesn't support this.

eNV25:
@jarun That case is already handled in your code. It already supports bsdtar, and zip. This change just tries bsdtar before atool.

jarun:
Not to my liking. With atool I just use `.` and it works. With bsdtar there is some quirk. I would prefer it to be simple to use.
Also, the `rar` issue is a deal-breaker. It's one of the most popular formats.

eNV25:
rar is supported by libarchive already...

jarun:
Did you check the Ubuntu version? I remember rar was only supported upto a certain version.

jarun:
http://manpages.ubuntu.com/manpages/kinetic/man5/libarchive-formats.5.html

```
RAR
     Libarchive has limited support for reading RAR format archives.  Currently, libarchive can
     read RARv3 format archives which have been either created uncompressed, or compressed using
     any of the compression methods supported by the RARv3 format.  Libarchive can also read
     self-extracting RAR archives.
```

jarun:
Also, it can't create `rar` archives, right?

eNV25:
No but it can extract. This change only affects listing and extracting.

jarun:
Why should anyone have to use 2 utilities to read/extract and create rar?

eNV25:
I have updated my branch to remove the regex change. It will show up in the PR if you open it again.

eNV25:
I can understand if you don't want to merge this change. How about we keep it as a patch instead?

eNV25:
> Why should anyone have to use 2 utilities to read/extract and create rar?

AFAIK there are some licensing issues that prevent libarchive from including RAR create support. Atool just uses rar command-line programs under the hood. I guess bsdtar could do the same as atool for creation only.

But this has nothing to do with listing and extracting archives.

jarun:
Since you have pointed out the issue that Arch packages atool, I am OK to have this on second thoughts. Please -re-raise the PR.

Please revert back the extra archive formats you have added as the current default is a minimal set that is supported on systems where there is no additional utilities like bsdtar or atool installed.

jarun:
Note that it's only the changes in this PR. I do not want #1392. Not our problem.

PR #1387, opened 2022-06-05
Go to last dir on bookmark key repeat

ggustafsson:
Hello,

This patch makes `SEL_BMOPEN` toggle between bookmarks directory and last directory, just like `SEL_CDHOME` and  `SEL_CDROOT`. 

Disclaimer: I don't know what the hell I am doing. I have not played around with C for more than 7 years.... BUT it doesn't segmentation fault on me (anymore) and it works 😃 

Best regards,
Göran Gustafsson

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7138,8 +7138,10 @@ nochange:
 					goto nochange;
 				}
 
-				if (strcmp(path, newpath) == 0)
-					break;
+				if (strcmp(path, newpath) == 0) {
+					dir = lastdir; /* Go to last dir on bookmark key repeat */
+					xstrsncpy(newpath, dir, PATH_MAX);
+				}
 			}
 
 			/* In list mode, retain the last file name to highlight it, if possible */


src/nnn.c
```diff
@@ -7138,8 +7138,10 @@ static bool browse(char *ipath, const char *session, int pkey)
 					goto nochange;
 				}
 
-				if (strcmp(path, newpath) == 0)
-					break;
+				if (strcmp(path, newpath) == 0) {
```
jarun, line 7141: I don't think this is very intuitive. Too many keypresses and since bookmark keys can be many, it's not very easy to remember that the user is in the bookmarked dir and he has to press the same key again. We have

ggustafsson: Ah, I don't use `NNN_BMS` so I didn't realise it affects that too. This was not intentional at all and it should be disabled for those. It is when you go straight to the bookmarks directory that it is convenient and intuitive.

ggustafsson:
You are absolutely right about how it makes no sense in combo with `NNN_BMS`. This patch fixes it:
`https://github.com/ggustafsson/nnn/commit/64ebcc97592d8a8186491b6823cfe80e8bfd848c`

Please give it a go without `NNN_BMS` set.

jarun:
> Please give it a go without NNN_BMS set.

I use the `type-to-nav` mode all the time. So for me the select bookmark key is either `^/` or `Esc-B` followed by the bookmark key. To return to previous directory, shall I press `-` or the earlier sequence again (while also having to remember that I am still in the bookmarked directory and its key)?

ggustafsson:
Sorry I don't quite follow... This makes `b` a toggle just like `~` and nothing more (in the last revised patch), is that not also the case for `^/`?
(It is not possible to type `^/` on a Swedish keyboard so it is very possible I'm missing something here)

jarun:
Please think of the case where `NNN_BMS` is defined. User presses `Enter` at the bookmark prompt and reaches the bookmark directory. Let's say now he wants to open another bookmark from `NNN_BMS`. If pressing `b` takes him to last visited dir, that use case breaks. I.e., you cannot open a bookmark from bookmark dir,

ggustafsson:
Ah yes the initial patch is broken as you have pointed out, my bad! I'm talking about the link in the last post, sorry if that was not clear. (It is not possible to update code in a closed PR hence a link that is easy to miss)

This patch:
[https://github.com/ggustafsson/nnn/commit/64ebcc97592d8a8186491b6823cfe80e8bfd848c](https://github.com/ggustafsson/nnn/commit/64ebcc97592d8a8186491b6823cfe80e8bfd848c)

If `NNN_BMS` is set nothing changes. If `NNN_BMS` is not set then "b" is a toggle.

jarun:
Makes sense. Please raise the PR!

PR #1388, opened 2022-06-05
Selections in current context color

ggustafsson:
Hello,

This patch makes selections use the current context color.

Screenshot 2022-06-05 at 20 56 34

Best regards,
Göran Gustafsson

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4190,7 +4190,7 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
 
 	uchar_t color_pair = get_color_pair_name_ind(ent, &ind, &attrs);
 
-	addch((ent->flags & FILE_SELECTED) ? '+' | A_REVERSE | A_BOLD : ' ');
+	addch((ent->flags & FILE_SELECTED) ? '+' | A_REVERSE | A_BOLD | COLOR_PAIR(cfg.curctx + 1) : ' ');
 
 	if (g_state.oldcolor)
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -4190,7 +4190,7 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
 
 	uchar_t color_pair = get_color_pair_name_ind(ent, &ind, &attrs);
 
-	addch((ent->flags & FILE_SELECTED) ? '+' | A_REVERSE | A_BOLD : ' ');
+	addch((ent->flags & FILE_SELECTED) ? '+' | A_REVERSE | A_BOLD | COLOR_PAIR(cfg.curctx + 1) : ' ');
```
jarun, line 4193: Did you try the following option with this:

    -D      dirs in context color

When it is used, I believe simple reverse is much more visible, bright and distinct.

jarun: I think selection should be uniformly presented in all contexts.

ggustafsson: Agreed. This must either be one setting for all contexts or not implemented at all. What do you think about using the last entry of `NNN_COLORS`? Basically `CTX_MAX` + 1.

I'm unsure about this whole thing right now. Since it is a "nice to have" thing it should only be implemented if there are no downsides to the implementation.

ggustafsson:
Closing this one as `-D` makes this patch a no go and I don't see a good solution either.

PR #1389, opened 2022-06-05
Use -x flag with BSD stat

ggustafsson:
Hello,

This patch makes `stat` start with flag `-x` on FreeBSD, NetBSD, OpenBSD and macOS.

```-x      Display information in a more verbose way as known from some Linux distributions.```

Screenshot 2022-06-05 at 22 24 49

`stat` above and `stat -x` below.

Best regards,
Göran Gustafsson


jarun:
Thank you!

PR #1392, opened 2022-06-07
Fix UI inconsistency when not using atool to extract archive.

eNV25:
This PR follows up on discussion in #1386.

@jarun, you pointed out how zip and bsdtar don't create a sub-directories automatically when extracting to the current directory.

The way this is currently handled doesn't make sense.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4641,7 +4641,7 @@ static bool handle_archive(char *fpath /* in-out param */, char op)
 	bool is_atool = getutil(utils[UTIL_ATOOL]);
 
 	if (op == 'x') {
-		outdir = xreadline(is_atool ? "." : xbasename(fpath), messages[MSG_NEW_PATH]);
+		outdir = xreadline(".", messages[MSG_NEW_PATH]);
 		if (!outdir || !*outdir) { /* Cancelled */
 			printwait(messages[MSG_CANCEL], NULL);
 			return FALSE;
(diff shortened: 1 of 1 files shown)


jarun:
> The way this is currently handled doesn't make sense.

What do you mean by this?

jarun:
When you provide `.`, atool smartly creates a new directory if required. bsdtar will extract all files as packed, even if they are directly under the root directory of the archive. Please try out the behaviour before raising reviews.

eNV25:
> > The way this is currently handled doesn't make sense.
> 
> What do you mean by this?

If one does the same input on a computer with a tool and one without atool, different things will happen without any feedback.

eNV25:
If I were to use `tmp/` as the output directory on a system with atool, nnn would cd to `tmp/` and extract to a sub-directory of `tmp/`, but in a system without atool the files would be with there in `tmp/`. This is an inconsistency.

jarun:
> If one does the same input on a computer with a tool and one without atool, different things will happen without any feedback.

> If I were to use tmp/ as the output directory on a system with atool, nnn would cd to tmp/ and extract to a sub-directory of tmp/, but in a system without atool the files would be with there in tmp/. This is an inconsistency.

That's because of how the tools differ, not because of `nnn`. And it's documented by atool how it behaves. I don't want to maintain code to make that uniform and then see it break one day because `bsdtar` actually does the sensible thing.

Users generally use a single tool and get accustomed to it.

jarun:
Also, if you have so much against `atool`, why are you even installing `atool`. Not having it would use `bsdtar` as fallback. I think it's too trivial an issue to spend so much time on. And unfortunately I do not have any time to spend on these discussions.

jarun:
Please do NOT raise any further issue/PR on the atool/bsdtar issue without discussing in the ToDO list. If devs think it's interesting enough they would respond.

You can always choose to have your own fork and have the changes there.

PR #1398, opened 2022-06-10
Revisit of Go to last dir on bookmark key repeat

ggustafsson:
This fixes issues with original PR:
[Go to last dir on bookmark key repeat #1387](https://github.com/jarun/nnn/pull/1387)

Best regards,
Göran Gustafsson

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7142,8 +7142,13 @@ nochange:
 					goto nochange;
 				}
 
-				if (strcmp(path, newpath) == 0)
-					break;
+				if (strcmp(path, newpath) == 0) {
+					if (bookmark)
+						break;
+
+					dir = lastdir; /* Go to last dir on bookmark key repeat */
+					xstrsncpy(newpath, dir, PATH_MAX);
+				}
 			}
 
 			/* In list mode, retain the last file name to highlight it, if possible */


src/nnn.c
```diff
@@ -7142,8 +7142,14 @@ static bool browse(char *ipath, const char *session, int pkey)
 					goto nochange;
 				}
 
-				if (strcmp(path, newpath) == 0)
-					break;
+				if (strcmp(path, newpath) == 0) {
+					if (!bookmark) {
```
jarun, line 7146: Please change to:

```c
if (bookmark)
    break;

dir = lastdir; /* Go to last dir on bookmark key repeat */
xstrsncpy(newpath, dir, PATH_MAX);
```

Fewer brackets and more readable code.

ggustafsson: Done!

jarun:
Thank you!

ggustafsson:
I introduced a bug here... If `nnn` is started from `~/.config/nnn/bookmarks` and you press `b` it will segfault.
This solves it:
```diff --git a/src/nnn.c b/src/nnn.c
index 0a291251..fddcf153 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7143,7 +7143,7 @@ nochange:
                                }

                                if (strcmp(path, newpath) == 0) {
-                                       if (bookmark)
+                                       if (bookmark || !*lastdir)
                                                break;

                                        dir = lastdir; /* Go to last dir on bookmark key repeat */
```
Does this look OK?

jarun:
Pushed at commit 49f21ae0139e54c0c8467ee439269f720f62c9bb. Thanks!

jarun:
Now I see the problem with this change. Strange you didn't encounter this.

The keys like `-`, ``` etc. to go back on repeat are symbols. Those are interpreted specially in _type-to-nav_ mode:

https://github.com/jarun/nnn/blob/40f75e9c38268cd205dedf83413a36f7a1d9925f/src/nnn.c#L3384

`b` works to go back as long as you are not in _type-to-nav_ mode. However, in _type-to-nav_ (which you were looking for, or even with just filter enabled), it will just filter with the string `b`. How is this inconsistent behaviour working for you? Please confirm if you see something else. I am planning to revert it back.

ggustafsson:
No issues for me. Works fine. I don't see a problem with `b` not working in type-to-nav, same as key `d` except that this bookmark key has an extra key that works with type-to-nav so it even less of a problem (Or does that not work? I still can't type `^/` :slightly_smiling_face:). I might be missing something here.

Type-to-nav on:  `^/` is a toggle
Type-to-nav off: `b` & `^/` is a toggle

The patch is a little bit pointless in combo with bookmarknav though. Not sure if I will use that or not yet.

jarun:
Key `d` does not have a special meaning in type-to-nav mode. Please stick to keys with similar functionality like ``` and `-`.

What I am saying is, when type-to-nav is disabled and NNN_BMS is not set:

- press b to go to bookmark dir
- filter gets enabled

Can you please `b` to go back to the earlier dir now?

jarun:
BTW, it has also broken another use case:

- without this change you can go back to the original dir (skipping the bookmarks dir) from the bookmark you have opened by pressing `-`. With this change you can only go back to  the bookmarks dir.

ggustafsson:
I’ve not seen any regressions but I am absolutely fine with a reversal if anyone does so no problem at all! 👍

jarun:
Yes, because you are not aware of the behaviour that's broken.

PR #1400, opened 2022-06-11
Use bsdtar by default if installed v2

eNV25:
Following #1386.

Changes:
- use `bsdtar` by default if installed (for **both creating and extraction**, unlike the previous patch)
- **adds `rar` and `unrar`**, because `bsdtar` / libarchive doesn't support it well
- handle_archive: early `return`s replaced with `goto END`

Summary of archiver choice:
- if `bsdtar` is installed, choose it
- else if `atool` is installed, choose it
- if archive is `.rar` and `atool` is not chosen, use `unrar`
- else if use choice
- else if archive is `.zip`, use `unzip`
- else use `tar`

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -535,24 +535,25 @@ static runstate g_state;
 #define UTIL_OPENER    0
 #define UTIL_ATOOL     1
 #define UTIL_BSDTAR    2
-#define UTIL_UNZIP     3
-#define UTIL_TAR       4
-#define UTIL_LOCKER    5
-#define UTIL_LAUNCH    6
-#define UTIL_SH_EXEC   7
-#define UTIL_BASH      8
-#define UTIL_SSHFS     9
-#define UTIL_RCLONE    10
-#define UTIL_VI        11
-#define UTIL_LESS      12
-#define UTIL_SH        13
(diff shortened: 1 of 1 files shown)


jarun:
I am not sure why you have added a new direct dependency of `unrar`. As per my knowledge both latest bsdtar and atool can handle rar in the backend.

I mentioned that you should raise the PR verbatim as the earlier one except the archive config change. You added new changes. This is exactly the problem. I do not have the time you have to discuss irrelevant (to my real-life situation) stuff right now.

I will make the changes as per the earlier PR and push it. Thanks!

eNV25:
@jarun You're right bsdtar can handle `rar`. I added `unrar` too since `rar` was added, for completeness. It would have been used if atool was not installed. But I guess it is not too important.

jarun:
Pushed at commit e104b749be2fd6dc22a89404ffb9e5bd155cbab1. Thanks!

eNV25:
Sorry for not re-raising, I deleted the original branch after the PR was closed.

Should I make other PRs for the other changes?

Anyway thanks for your time.

jarun:
Not right now. I don't want to add a hard dependency on unrar.

PR #1401, opened 2022-06-11
Add bookmarknav patch

luukvbaal:


jarun:
Thank you!

jarun:
As an alternative, can you please have a look if the following minimal patch works for you?

```diff
diff --git a/src/nnn.c b/src/nnn.c
index df326430..7929be79 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7150,6 +7150,7 @@ nochange:
                                        dir = lastdir; /* Go to last dir on bookmark key repeat */
                                        xstrsncpy(newpath, dir, PATH_MAX);
                                }
+                               presel = FILTER;
                        }
 
                        /* In list mode, retain the last file name to highlight it, if possible */
@@ -7788,6 +7789,8 @@ nochange:
                        }
                        setdirwatch();
                        clearfilter();
+                       if (g_state.runplugin == 1)
+                               presel = FILTER;
                        goto begin;
                case SEL_SHELL: // fallthrough
                case SEL_LAUNCH: // fallthrough
```

luukvbaal:
Well yeah it works for what it is but it's not quite the same as this patch as I'm sure you realize. To me the added logic is worth not having to press enter in the bookmark workflow.

jarun:
OK. I get it. Interesting that you are convinced users who don't use type-to-nav are so well versed in its features.

jarun:
But I think once they see the benefits, they will get used to it.

jarun:
Ok, so auto-enter... that skipped my mind. I thought you want the users to benefit from the nav-to-type mode to pick the right bookmark.

In that case, how about this - currently we tie auto-enter to _type-to-nav_ only. What if we allow auto-enter in filter mode also (as long as `-A` is not used)?

luukvbaal:
Yeah that could potentially yield a similar result in combination with the `presel = FILTER` patch you suggested earlier. I'm not sure whether that would result in a simplified patch (or mainline change), hadn't looked into that route. I think you're right that it might reduce the required checks as FILTER will already be disabled once you leave the directory in the current codebase.

I think you would only want to tie auto-enter to regular filter mode inside the bookmark(and plugin) dir though, so not sure if you actually can get away with less checks. I can look into it later, or you can if you already had something in mind.

jarun:
> I think you would only want to tie auto-enter to regular filter mode inside the bookmark(and plugin) dir though

Please explain a bit more. I am thinking of enabling auto-enter in regular filters (along with current type-to-nav) if user hasn't used `-A`.

jarun:
@luukvbaal see is commit 56f96c3d0a0cf23154948bffd376ad2ad7e65240 plus commit 581e629c20e4540c63443211318b48b555a41053 work for you. If not, I will revert both.

luukvbaal:
> Please explain a bit more. I am thinking of enabling auto-enter in regular filters (along with current type-to-nav) if user hasn't used `-A`.

Disregard, on second thought the behavior you proposed and introduced in https://github.com/jarun/nnn/commit/581e629c20e4540c63443211318b48b555a41053 is fine!

PR #1402, opened 2022-06-13
Fix false positive selbm

luukvbaal:
Avoid unintentionally entering symlink target when leaving bookmark dir through `CDHOME/CDBEGIN/CDLAST/CDROOT`.

For example pressing <kbd>b\`</kbd> -> `/bin` enters `/usr/bin` instead of `/bin`.


jarun:
Thank you!

PR #1404, opened 2022-06-14
Replace return with goto END in handle_archive

eNV25:
Replace return with goto END in handle_archive


jarun:
What is the problem with the current code?

eNV25:
It looks like `outdir` isn't `free`ed in some paths.

jarun:
`nnn` has integrated static analysis in CI. Also, `xreadline` returns a pre-allocated buffer.

PR #1406, opened 2022-06-20
account for realpath failure and update link

N-R-K:


jarun:
Thank you!
Great to see you back!

PR #1411, opened 2022-06-29
Account for ueberzug offset in preview-tui

luukvbaal:
Allows to draw ueberzug images correctly when running inside an embedded terminal such as with nnn.nvim (https://github.com/luukvbaal/nnn.nvim/commit/99da2cfa6437ce71415caa122f616df6e0cd1756).

Not that useful but better than seeing images drawn incorrectly when you do hover over images inside an embedded terminal 🤷🏻‍♂️


jarun:
Thank you!

PR #1412, opened 2022-06-29
Misc cleanups

N-R-K:
A couple small changes to get rid of some unnecessary code, no changes in functionality intended:

* Fix an incorrect comment on `xmemrchr()`.
* Use designated initializer instead of initializing `cfg` manually.
* Initialize to 0 instead of `memset`-ing to 0.

### DIFF
--- a/patches/gitstatus/mainline.diff
+++ b/patches/gitstatus/mainline.diff
@@ -6,13 +6,13 @@
 # Authors: Luuk van Baal
 
 diff --git a/src/nnn.c b/src/nnn.c
-index 508ee915..77d56d63 100644
+index f8a2c58..157cea5 100644
 --- a/src/nnn.c
 +++ b/src/nnn.c
 @@ -265,6 +265,25 @@
  #define FREE     0
  #define CAPACITY 1
-
+ 
 +/* Git icons */
 +#ifdef NERD
 +#define GIT_ADD ""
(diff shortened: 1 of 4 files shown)


N-R-K:
I'll fix the patches in a bit. Also don't think it's a good idea for the patches CI to abort on first failure, we want it to print out all the failing patches before aborting. Will look into fixing that as well.

jarun:
Neat!
Patching fails on CI. Please have a look.

luukvbaal:
> I'll fix the patches in a bit. Also don't think it's a good idea for the patches CI to abort on first failure, we want it to print out all the failing patches before aborting. Will look into fixing that as well.

Definitely. Would be great if we could expose the patch ci to the developer in the Makefile somehow. I.e. running `make patches` will ideally compile all the patches (not aborting on fail, if possible). Might make it easier to confirm patches are still working locally.

N-R-K:
> Patching fails on CI. Please have a look.

Done.

>  Would be great if we could expose the patch ci to the developer in the Makefile somehow. I.e. running make patches will ideally compile all the patches (not aborting on fail, if possible).

Yeah, that sounds great. I think it can be done easily by extracting the script from the CI's yaml to a bash script so that it can be invoked both from the Makefile and the run in the ci.

jarun:
I am fine with what you guys decide.

jarun:
Make it easy for the dev to replicate the failure, but please don't remove the CI failure. Otherwise, the issue might be ignored.

jarun:
Thank you!

N-R-K:
Note that we need to run `git restore src` so that it can clean up the src after a patch fails so that subsequent patches don't automatically fail due to dirty src.

N-R-K:
Updated version which exits with the proper exit code: 

```bash
#!/bin/bash

export PATCH_OPTS="--merge"
patches=("O_GITSTATUS" "O_NAMEFIRST" "O_RESTOREPREVIEW")
z=$(( 1 << ${#patches[@]} ))
pid=$$
ret=0
trap 'ret=1' SIGUSR1

for ((n=1; n < z; ++n)); do
    for ((i=0; i < ${#patches[@]}; ++i)); do
        printf "%s=%d " "${patches[$i]}" "$(( (n & (1 << i)) != 0 ))"
    done | tee "/dev/stderr" | (
        make clean -s
        xargs make 2>&1
        if [ "$?" -ne 0 ]; then
            echo "[FAILED]" >&2
            kill -SIGUSR1 "$pid"
        else
            echo "[SUCCESS]" >&2
        fi
        git restore src
    ) >/dev/null
done
exit "$ret"
```

We can try to put this into the Makefile if you prefer that @luukvbaal . But afaik by default make invokes `/bin/sh` which might not be bash, so we'll need to avoid any bash-ism if we want the Makefile to be portable. Which is why I opted for a separate script which invokes bash specifically.

PR #1413, opened 2022-06-29
make it easy to check for failing patches locally

N-R-K:
adds a script `check-patches.sh` to check for patch failures and also
adds a make target `checkpatches` which will invoke the check-patches
script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -310,6 +310,9 @@ upload-local: sign static musl
 clean:
 	$(RM) -f $(BIN) nnn-$(VERSION).tar.gz *.sig $(BIN)-static $(BIN)-static-$(VERSION).x86_64.tar.gz $(BIN)-icons-static $(BIN)-icons-static-$(VERSION).x86_64.tar.gz $(BIN)-nerd-static $(BIN)-nerd-static-$(VERSION).x86_64.tar.gz $(BIN)-emoji-static $(BIN)-emoji-static-$(VERSION).x86_64.tar.gz $(BIN)-musl-static $(BIN)-musl-static-$(VERSION).x86_64.tar.gz
 
+checkpatches:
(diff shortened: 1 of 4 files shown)


luukvbaal:
LGTM. BTW it would look like this embedded in the Makefile:
```Make
SHELL := /bin/bash
checkpatches:
	export PATCH_OPTS="--merge"; \
	patches=("O_GITSTATUS" "O_NAMEFIRST" "O_RESTOREPREVIEW"); \
	z=$$(( 1 << $${#patches[@]} )); \
	pid=$$; \
	ret=0; \
	trap 'ret=1' SIGUSR1; \
	for ((n=1; n < z; ++n)); do \
	    for ((i=0; i < $${#patches[@]}; ++i)); do \
	        printf "%s=%d " "$${patches[$$i]}" "$$(( (n & (1 << i)) != 0 ))"; \
	    done | tee "/dev/stderr" | ( \
	        make clean -s; \
	        xargs make 2>&1; \
	        if [ "$$?" -ne 0 ]; then \
	            echo "[FAILED]" >&2; \
	            kill -SIGUSR1 "$$pid"; \
	        else \
	            echo "[SUCCESS]" >&2; \
	        fi; \
	        git restore src; \
	    ) >/dev/null; \
	done; \
	exit "$$ret" \
```

Involves a lot of escaping, idk if it's worth it. Up to jarun.

Makefile
```diff
@@ -338,4 +338,7 @@ endif
 
 skip: ;
 
+checkpatches:
+	./misc/test/check-patches.sh
```
jarun, line 342: Why not add this under the `patches` directory?

jarun (changes requested):

jarun:
Thank you!

PR #1414, opened 2022-06-30
add z.lua to autojump

shuizhongyueming:
add support for [z.lua](https://github.com/skywind3000/z.lua) in autojump


jarun:
Thank you!

PR #1415, opened 2022-06-30
fix buggy clang warning on older versions

N-R-K:
This warning flag is buggy on older clang versions but fixed on `v9` [^0]. Disable the warning for older versions.

[^0]: https://godbolt.org/z/6cGqWoeP7

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,8 @@ jobs:
             ls -l nnn
             make clean
             echo
+            # NOTE: `-Wmissing-braces` is buggy on clang versions below 9
+            export CFLAGS="-Werror -Wno-missing-braces"
             echo "########## clang-6 ##########"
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

jarun:
Somehow it still fails. Bringing back the old `mousetimings` initialization.

N-R-K:
> Somehow it still fails. Bringing back the old `mousetimings` initialization.

hmm, weird. But in any case, reverting the change is fine by me.

N-R-K:
Ahh, I see the problem

    clang-6.0  -Werror -Wno-missing-braces -std=c11 -Wall -Wextra -Wshadow

The `-Wno-missing-braces` is getting overriden by the `-Wall -Wextra` flags later on.

Anyways let's just leave it be, not a big of a deal.

PR #1418, opened 2022-07-03
Add saucenao plugin

GospodinHoroshiy:
New plugin for searching hovered images with saucenao
Tested locally

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -57,6 +57,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [renamer](renamer) | Batch rename selection or files in dir [✓] | sh | [qmv](https://www.nongnu.org/renameutils/)/[vidir](https://joeyh.name/code/moreutils/) |
 | [ringtone](ringtone) | Create a variable bitrate mp3 ringtone from file | sh | date, ffmpeg |
 | [rsynccp](rsynccp) | Gives copy-paste verbose progress percentage [✓] | sh | rsync |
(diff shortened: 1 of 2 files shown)


plugins/README.md
```diff
@@ -57,6 +57,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [renamer](renamer) | Batch rename selection or files in dir [✓] | sh | [qmv](https://www.nongnu.org/renameutils/)/[vidir](https://joeyh.name/code/moreutils/) |
 | [ringtone](ringtone) | Create a variable bitrate mp3 ringtone from file | sh | date, ffmpeg |
 | [rsynccp](rsynccp) | Gives copy-paste verbose progress percentage [✓] | sh | rsync |
+| [saucenao](saucenao) | Upload file to saucenao and scrap links | sh | curl |
```
jarun, line 60: Can you mention that it is a reverse image search engine?

jarun (changes requested):

jarun:
Can you explain the benefit of this plugin for regular users?
Why should one use this plugin instead of opening the reverse image search engine in a web browser where she can visually see the images?

jarun:
Closing this as I am not seeing any real use case for regular users.
Also, the script is to scrape the results which is unrelated to day-to-day file management.

N-R-K:
A more useful version might have been to download all the images to some `/tmp/` folder and take the user there on a new context. Then they can use `nnn` to delete the entries they don't like and move/copy the ones they do.

jarun:
@N-R-K we are not aware of the terms of usage and the copyrights of the images.

PR #1419, opened 2022-07-08
Add icons for common elixir extensions

skbolton:
eex and heex are for elixirs templating languages
exs is for elixir scripts
ex is the common extension for files


jarun:
Sorry, please patch locally.

skbolton:
I can go about this route but I think its worth calling out that the codebase has references to icons that it doesn't inject. Should I PR something that removes a reference to an elixir icon?

N-R-K:
This got added in https://github.com/jarun/nnn/pull/1437

PR #1421, opened 2022-07-09
Add colemak patch

luukvbaal:


jarun:
Thank you!

@anjerukare please confirm this works as intended.

anjerukare:
Checked, works as intended!

But there's one thing. I didn't emphasize that these patches are for Colemak modification, which is called Colemak-DH.
Here are changes from Colemak: https://colemakmods.github.io/mod-dh/#colemak-changes.

To make patches for default Colemak layout, as I see it, we need to swap <kbd>m</kbd> with <kbd>h</kbd>.

jarun:
@anjerukare can you please raise the PR to fix it? We are really busy.

luukvbaal:
I think the question is do we want 2 new patches, 1 for "colemak" and 1 for "colemak-DH"...

jarun:
Only one - "colemak".

anjerukare:
> @anjerukare can you please raise the PR to fix it? We are really busy.

Same here. All I can do to help right now: [colemak-patches.tar.gz](https://github.com/jarun/nnn/files/9079126/colemak-patches.tar.gz).

jarun:
@luukvbaal I think @anjerukare has forgotten that (s)he came up with the request.

I leave it up to you to update the patch or discard. If you want I can revert the earlier commit.

luukvbaal:
I don't care, this seems a relatively free patch as in I don't think many conflicts will occur. I imagine it will be nice to have for colemak users but for that we need a colemak user to define the layout. I would say wait for a PR to update but if you want to make a release then idk.

jarun:
I think it's fine not to update. I am sure users would have different preferences.

patches/check-patches.sh
```diff
@@ -18,7 +18,7 @@ for ((n=1; n < z; ++n)); do
     done | tee "/dev/stderr" | (
         make clean -s
         xargs make 2>&1
-        if [ "$?" -ne 0 ]; then
+        if ! xargs make 2>&1; then
```
ggustafsson, line 21: Line 20 should be removed, xargs make is executed twice.

luukvbaal: Whoops, thanks.

PR #1426, opened 2022-07-17
Escape call to `nnn` so `nnn` can be used as an alias to `n`

Anomalocaridid:
Currently, if you use quitcd and set `nnn` to be an alias of `n` in bash or zsh, you will get the following error if you try to run nnn: `maximum nested function level reached; increase FUNCNEST?`. Escaping the call to `nnn` in quitcd lets you use `nnn` as an alias to `n` without raising this error.

### DIFF
--- a/misc/quitcd/quitcd.bash_zsh
+++ b/misc/quitcd/quitcd.bash_zsh
@@ -19,7 +19,7 @@ n ()
     # stty lwrap undef
     # stty lnext undef
 
-    nnn "$@"
+    \nnn "$@"
 
     if [ -f "$NNN_TMPFILE" ]; then
             . "$NNN_TMPFILE"
--- a/misc/quitcd/quitcd.bash_zsh
+++ b/misc/quitcd/quitcd.bash_zsh
@@ -19,7 +19,7 @@ n ()
     # stty lwrap undef
     # stty lnext undef
 
-    # The backslash allows one to alias nnn to n if desired without making an
+    # The backslash allows one to alias n to nnn if desired without making an
     # infinitely recursive alias
     \nnn "$@"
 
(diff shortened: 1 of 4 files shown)


jarun:
For completeness, can you please check if that's achievable for the other shells with `quitcd` script as well?

misc/quitcd/quitcd.bash_zsh
```diff
@@ -19,7 +19,7 @@ n ()
     # stty lwrap undef
     # stty lnext undef
 
-    nnn "$@"
+    \nnn "$@"
```
jarun, line 22: Please add a comment on why this is done.

jarun (changes requested):

Anomalocaridid:
How's this? Also, as far as I was able to figure out, quitcd.elv already has the Elvish equivalent of this implemented, so I just put a comment in it.

misc/quitcd/quitcd.bash_zsh
```diff
@@ -19,6 +19,8 @@ n ()
     # stty lwrap undef
     # stty lnext undef
 
+    # The backslash allows one to alias nnn to n if desired without making an
```
N-R-K, line 22: You can already alias `nnn` to `n`, think this should've been the other way around (?)

Anomalocaridid:
Sorry about that. Looks like I failed to notice a typo I made. How's this?

misc/quitcd/quitcd.fish
```diff
@@ -27,7 +27,9 @@ function n --wraps nnn --description 'support nnn quit and change directory'
     # stty lwrap undef
     # stty lnext undef
 
-    nnn $argv
+    # The command function allowa one to alias n to nnn if desired without
```
N-R-K, line 30:
```suggestion
    # The command function allows one to alias this function to `nnn` if desired without
```

jarun: I'll take care of this. Merging.

jarun:
Thank you!

PR #1429, opened 2022-07-21
Allow specifying preview width

UtkarshVerma:
This PR allows the user to specify the split width for the nnn pane in percentage.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -80,6 +80,7 @@
 # Authors: Todd Yamakawa, Léo Villeveygoux, @Recidiviste, Mario Ortiz Manero, Luuk van Baal, @WanderLanz
 
 #SPLIT="$SPLIT"  # you can set a permanent split here
+SPLIT_WIDTH="${SPLIT_WIDTH:-50}"
 #TERMINAL="$TERMINAL"  # same goes for the terminal
 DEBUG_LOG=0  # set to 1 to enable logging for debug purposes
 USE_SCOPE="${USE_SCOPE:-0}"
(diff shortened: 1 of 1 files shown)


jarun:
@N-R-K @luukvbaal please review.

jarun:
Looks OK to me. Merging. Thanks!

jarun:
Not sure why CI didn't run. There are some issues.

jarun:
I am reverting the change. Please raise after fixing the shellcheck issues: https://app.circleci.com/pipelines/github/jarun/nnn/4295/workflows/04f78340-ed5d-4447-8c42-e26ece98bb84/jobs/4366

plugins/preview-tui
```diff
@@ -80,6 +80,7 @@
 # Authors: Todd Yamakawa, Léo Villeveygoux, @Recidiviste, Mario Ortiz Manero, Luuk van Baal, @WanderLanz
 
 #SPLIT="$SPLIT"  # you can set a permanent split here
+SPLIT_WIDTH="${SPLIT_WIDTH:-50}"
```
luukvbaal, line 83: This will also determine the split height for a vertical split. Should be named `SPLIT_SIZE` I think. We should also try to include it for the other preview options.

plugins/preview-tui
```diff
@@ -334,7 +336,7 @@ preview_file() {
     ext="${1##*.}"
     [ -n "$ext" ] && ext="$(printf "%s" "${ext}" | tr '[:upper:]' '[:lower:]')"
     lines=$(tput lines <"$TTY")
-    cols=$(tput cols <"$TTY")
+    cols=$(( $(tput cols <"$TTY") * (100 - $SPLIT_WIDTH) / $SPLIT_WIDTH ))
```
luukvbaal, line 339: Are you sure this is necessary? `tput cols` should already return the actual size of the split.

UtkarshVerma: This value of cols is used to determine the with of the preview pane. I've currently written the patch such that SPLIT_WIDTH corresponds to the width of the nnn pane.

plugins/preview-tui
UtkarshVerma: Yeah, that would make sense. Will do once I'm back on my machine.

plugins/preview-tui
luukvbaal: Yes but `cols` should be the width of the preview pane, which is the value that `tput cols` returns since it runs inside the split. There is should be no need to do any kind of percentage multiplication here.

Also makes more sense to set `SPLIT_WIDTH` from the perspective of the preview pane, not nnn imo.

UtkarshVerma: Yeah, that's what I thought as well, but I noticed that the `tput cols` command was being evaluated in the nnn pane. You can test this patch on your end to verify this as well. Because having this percentage multiplication works for me.

UtkarshVerma:
I have pushed your requested changes to my fork. Please have a look.

luukvbaal:
> but I noticed that the tput cols command was being evaluated in the nnn pane. 

Right, this is an unfortunate side-effect of the `TTY` passing workaround. I wish we could do away with the workaround. Might make a PR with this (https://github.com/jarun/nnn/pull/1234/files) change back in as I see no issues with it testing it now...

PR #1430, opened 2022-07-21
icons: Add support for FontAwesome6

UtkarshVerma:
This PR adds support for icons through FontAwesome 6. This has been done because Nerd font and icons-in-terminal are pretty old and don't include Font Awesome 6 glyphs.

There are only a few glyphs which aren't covered by Font Awesome 6 (followed by // in icons-fontawesome.h), for which a fallback of Nerd font symbols is expected. I use the following fontconfig rule myself:

```xml
    <alias>
        <family>monospace</family>
        <prefer>
            <family>JetBrainsMono</family>
            <family>icons</family>
        </prefer>
    </alias>
    <alias binding="same">
        <family>icons</family>
        <prefer>
            <family>Font Awesome 6 Free Solid</family>
            <family>Font Awesome 6 Brands</family>
            <family>Symbols Nerd Font</family>
        </prefer>
    </alias>
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ O_NOFIFO := 0  # no FIFO previewer support
 O_CTX8 := 0  # enable 8 contexts
 O_ICONS := 0  # support icons-in-terminal
 O_NERD := 0  # support icons-nerdfont
+O_FONTAWESOME := 0  # support icons-fontawesome
 O_EMOJI := 0  # support emoji
 O_QSORT := 0  # use Alexey Tourbin's QSORT implementation
 O_BENCH := 0  # benchmark mode (stops at first user input)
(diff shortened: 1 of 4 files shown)


N-R-K (changes requested):
> This PR adds support for icons through FontAwesome 6. This has been done because Nerd font and icons-in-terminal are pretty old and don't include Font Awesome 6 glyphs.

I'm pretty sure Nerd font supports Font Awesome glyphs, [in their README](https://github.com/ryanoasis/nerd-fonts) they advertise Font Awesome support. And if some glyphs are missing then the correct action would be to open an issue on the Nerd Font repo.

UtkarshVerma:
> > This PR adds support for icons through FontAwesome 6. This has been done because Nerd font and icons-in-terminal are pretty old and don't include Font Awesome 6 glyphs.
> 
> I'm pretty sure Nerd font supports Font Awesome glyphs, [in their README](https://github.com/ryanoasis/nerd-fonts) they advertise Font Awesome support. And if some glyphs are missing then the correct action would be to open an issue on the Nerd Font repo.

They do have FontAwesome glyphs, but from two years back. Hence it doesn't have the new FontAwesome 6 glyphs.

N-R-K:
> They do have FontAwesome glyphs, but from two years back. Hence it doesn't have the new FontAwesome 6 glyphs.

Then that should be a bug report on the Nerd Font repo. It's better if they add it instead of us having to create new icon set for every single icon-font.

N-R-K:
P.S: There's already an open issue and there seems to be some work done as well. https://github.com/ryanoasis/nerd-fonts/issues/789

UtkarshVerma:
> > They do have FontAwesome glyphs, but from two years back. Hence it doesn't have the new FontAwesome 6 glyphs.
> 
> Then that should be a bug report on the Nerd Font repo. It's better if they add it instead of us having to create new icon set for every single icon-font.
> 
> 

Yeah, that makes sense.

PR #1431, opened 2022-07-22
Preview-tui improvements

luukvbaal:
* Do away with all-encompassing error stdout/err redirection, resulting in ncurses tput problems.
* Add some helper functions
* Add `SPLIT_SIZE` option for tmux and windows terminal(untested), iterm/kitty don't seem to support percentage split size(#1429).

@UtkarshVerma please test.


UtkarshVerma:
Thanks for the PR. I've got bombarded with work so I'll test this after a couple of days.

jarun:
Thank you!

PR #1432, opened 2022-07-22
Revise and optimize icons handling

N-R-K:
This pretty much reworks the entire icon system. Some notable changes:

* The extensions are put into a statically generated hash-table instead
  of a sorted array. We use Robin-Hood insertion to reduce the max probe
  length. Currently we need to probe only 2 slots for `O_EMOJI` and only
  3 for `O_NERD`/`O_ICONS`.
* I've opted not to use a perfect-hash since the perfect hashes
  generated by [`gperf`](https://www.gnu.org/software/gperf) used some
  huge lookup table. The hash function also wasn't as minimal as I'd
  like.
* Colors are now using X-Macros. This should speed up startup since we
  don't have to search `icons_ext` linearly to find unique colors.
* The hash-table generator outputs a more space optimized `struct
  icon_pair` using a char array instead of char pointer. This brings
  down the binary size from `145KiB` when using `O_NERD` down to
  `137KiB`.
* Some unnecessary duplication and indirection has been reduced by using
  the `ICON_STR()` macro.


src/icons-hash.c
```diff
@@ -0,0 +1,203 @@
...
+	#define log(...) ((void)0)
+#endif
+
+static uint16_t icon_ext_hash(const char *s);
+
+/* change ICONS_TABLE_SIZE to increase the size of the table */
+static struct icon_pair table[1u << ICONS_TABLE_SIZE];
+static uint16_t seen[ARRLEN(table)];
+static uint16_t hash_start = 7; /* arbitrarily picked. change if needed. but ensure it's above 1 */
+static uint16_t hash_mul = 251; /* same as above ^ */
```
KlzXS, line 42: I think it would be nice to mention that they should be prime. It _could_ work otherwise, but primes are better.

jarun:
Simply awesome!!! :+1:

jarun:
@N-R-K Can you please update the ToDo list?

jarun:
If I try to bypass `make` (to use `musl-gcc`) I get the following error:

```
musl-gcc -march=native -O3 -DNORL -DNOLC -DNOMOUSE -DNOBATCH -DTOURBIN_QSORT -DNOSSN -DNOUG -DNOX11 -DEMOJI -std=c11 -Wall -Wextra -Wshadow -I../netbsd-curses/libcurses -I../musl-fts -o nnn src/nnn.c -Wl,-Bsymbolic-functions -lpthread -L/opt/nnn-libs -lcurses -lterminfo -lfts -static
src/nnn.c:128:10: fatal error: icons-generated.h: No such file or directory
  128 | #include "icons-generated.h"
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

What should I use instead of `-DEMOJI`?

jarun:
The following works:

```
make O_EMOJI=1
musl-gcc -march=native -O3 -DNORL -DNOLC -DNOMOUSE -DNOBATCH -DTOURBIN_QSORT -DNOSSN -DNOUG -DNOX11 -DEMOJI -std=c11 -Wall -Wextra -Wshadow -I../netbsd-curses/libcurses -I../musl-fts -o nnn src/nnn.c -Wl,-Bsymbolic-functions -lpthread -L/opt/nnn-libs -lcurses -lterminfo -lfts
```

N-R-K:
> If I try to bypass `make` (to use `musl-gcc`) I get the following error:
> 
> ```makefile
> musl-gcc -march=native -O3 -DNORL -DNOLC -DNOMOUSE -DNOBATCH -DTOURBIN_QSORT -DNOSSN -DNOUG -DNOX11 -DEMOJI -std=c11 -Wall -Wextra -Wshadow -I../netbsd-curses/libcurses -I../musl-fts -o nnn src/nnn.c -Wl,-Bsymbolic-functions -lpthread -L/opt/nnn-libs -lcurses -lterminfo -lfts -static
> src/nnn.c:128:10: fatal error: icons-generated.h: No such file or directory
>   128 | #include "icons-generated.h"
>       |          ^~~~~~~~~~~~~~~~~~~
> compilation terminated.
> ```
> 
> What should I use instead of `-DEMOJI`?

First the "generator" needs to be built:

```console
$ gcc -DICONS_GENERATE -DEMOJI -o src/icons-hash-gen src/icons-hash.c
```

Then the hash-table needs to be generated with the generator:

```console
$ ./src/icons-hash-gen > src/icons-generated.h
```

Now you should be able to run the compile command manually as long as both the generator and `nnn` is build with the same icon set (`-DEMOJI`).

The makefile already takes care of this. So just doing `make O_EMOJI=1 src/icons-generated.h` to generate the hash-table and then manually compiling `nnn` with `-DEMOJI` would also work.

jarun:
Settled with:

```
CC=musl-gcc make O_EMOJI=1 src/icons-generated.h
musl-gcc -march=native -O3 -DNORL -DNOLC -DNOMOUSE -DNOBATCH -DTOURBIN_QSORT -DNOSSN -DNOUG -DNOX11 -DEMOJI -std=c11 -Wall -Wextra -Wshadow -I../netbsd-curses/libcurses -I../musl-fts -o nnn src/nnn.c -Wl,-Bsymbolic-functions -lpthread -L/opt/nnn-libs -lcurses -lterminfo -lfts -static
```

Thank you!

N-R-K:
[Updated the performance section](https://github.com/jarun/nnn/wiki/Performance/_compare/a5f6f5b446fc04676279fdf5b6e3fd45252121b2...b7e68bc840cd1ca24c3f0a596e90c06e863e83fa) on the wiki :)

Also note that I still see some room for improvements. For example a lot of the extensions use the same icon. So instead of wasting 4 bytes (for `O_NERD` and `O_ICONS`) or 9 bytes (for `O_EMOJI`) per slot, we can store the unique icons in a separate array and store a single byte index into that array.

There's also the possibility of using a *minimal-perfect-hash*, which can reduce memory usage by shrinking the table size but might make the hash-function/lookup slower. I still have lot more to learn and research on this topic.

I'll see if I can make some time next weekend to implement and benchmark some of these ideas. If they turn out to be overall improvement, then I'll open the PR.

jarun:
Thank you! Keep it up!

jarun:
With the above options and musl-gcc , dynamically linked binary size:

    -rwxrwxr-x 1 vaio vaio 116632 Jul 22 23:33 nnn*

Static:

    -rwxrwxr-x 1 vaio vaio 399504 Jul 22 23:04 /usr/local/bin/nnn*

Great improvement!

PR #1436, opened 2022-07-24
icons: use a compact array

N-R-K:
a lot of the extension use the same icon. this can be exploited via
having an array with all the unique icons and then storing a single byte
index into the unique array.

when using `O_EMOJI` this results in around ~1.7KiB drop in the total
table size. `O_NERD` and `O_ICONS` get roughly ~0.5KiB savings.


N-R-K:
Btw, on master I don't see any drop in binary size. But on my local branch (where I use additional compiler optimizations) I see some binary size drop.

jarun:
:+1:

PR #1443, opened 2022-07-30
Update patches

luukvbaal:
* Resolve patch conflict
* Update patch instructions to use `git diff` instead of `diff`. I tried to avoid using git diff before but devs having git installed is a given. The current patches were all generated with `git diff` and the current `Makefile` expects the git format.


jarun:
Thank you!

PR #1451, opened 2022-08-02
Improved chunk allocation logic

KlzXS:
I'm much more satisfied with this check. No useless calculations. Also I saw that there was a way to `free(paths)` before we even allocated it so I patched that. Also it seems we were reading one chunk too few, not that anyone would notice without digging in too deep.

There is also the modified check for exhausting the input @N-R-K wanted to see.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -190,6 +190,7 @@
 #define ASCII_MAX       128
 #define EXEC_ARGS_MAX   10
 #define LIST_FILES_MAX  (1 << 14) /* Support listing 16K files */
+#define LIST_INPUT_MAX  ((size_t)LIST_FILES_MAX * PATH_MAX)
 #define SCROLLOFF       3
 #define COLOR_256       256
 
@@ -7948,7 +7949,7 @@ static char *make_tmp_tree(char **paths, ssize_t entries, const char *prefix)
 
 static char *load_input(int fd, const char *path)
 {
-	ssize_t i, chunk_count = 1, chunk = (ssize_t)(512 * 1024) /* 512 KiB chunk size */, entries = 0;
+	size_t i, chunk_count = 1, chunk = 512UL * 1024 /* 512 KiB chunk size */, entries = 0;
 	char *input = malloc(sizeof(char) * chunk), *tmpdir = NULL;
 	char cwd[PATH_MAX], *next;
 	size_t offsets[LIST_FILES_MAX];
@@ -7969,9 +7970,12 @@ static char *load_input(int fd, const char *path)
 	} else
 		xstrsncpy(cwd, path, PATH_MAX);
 
-	while ((chunk_count) < 512 && !msgnum) {
+	while (chunk_count < LIST_INPUT_MAX / chunk && !msgnum) {
 		input_read = read(fd, input + total_read, chunk);
 		if (input_read < 0) {
+			if (errno == EINTR)
+				continue
+
 			DPRINTF_S(strerror(errno));
 			goto malloc_1;
 		}
@@ -7980,7 +7984,6 @@ static char *load_input(int fd, const char *path)
 			break;
 
 		total_read += input_read;
-		++chunk_count;
 
 		while (off < total_read) {
 			if ((next = memchr(input + off, '\0', total_read - off)) == NULL)
@@ -8001,29 +8004,23 @@ static char *load_input(int fd, const char *path)
 			off = next - input;
 		}
 
-		if (chunk_count == 512) {
+		/* We don't need to allocate another chunk */
+		if (chunk_count > (total_read + chunk - 1) / chunk)
+			continue;
+
+		/* We can never need more than one additional chunk */
+		++chunk_count;
+		if (chunk_count > LIST_INPUT_MAX / chunk) {
 			msgnum = MSG_SIZE_LIMIT;
 			break;
 		}
 
-		/* We don't need to allocate another chunk */
-		if (chunk_count == (total_read - input_read) / chunk)
-			continue;
-
-		chunk_count = total_read / chunk;
-		if (total_read % chunk)
-			++chunk_count;
-
-		input = xrealloc(input, (chunk_count + 1) * chunk);
+		input = xrealloc(input, chunk_count * chunk);
 		if (!input)
-			return NULL;
+			goto malloc_1;
 	}
 
-	/* Read off the extra data if limits exceeded */
-	if (msgnum) {
-		char buf[512];
-		while (read(fd, buf, 512) > 0);
-	}
+	/* We close fd outside this function. Any extra data is left to the kernel to handle */
 
 	if (off != total_read) {
 		if (entries == LIST_FILES_MAX)
@@ -8067,7 +8064,6 @@ static char *load_input(int fd, const char *path)
 		if (!paths[i]) {
 			entries = i; // free from the previous entry
 			goto malloc_2;
-
 		}
 
 		DPRINTF_S(paths[i]);
@@ -8087,7 +8083,7 @@ static char *load_input(int fd, const char *path)
 		tmpdir = make_tmp_tree(paths, entries, listroot);
 
 malloc_2:
-	for (i = entries - 1; i >= 0; --i)
+	for (i = 0; i < entries; ++i)
 		free(paths[i]);
 malloc_1:
 	if (msgnum) { /* Check if we are past init stage and show msg */
@@ -8099,6 +8095,7 @@ malloc_1:
 			usleep(XDELAY_INTERVAL_MS << 2);
 		}
 	}
+
 	free(input);
 	free(paths);
 	return tmpdir;


src/nnn.c
```diff
@@ -8099,8 +8097,11 @@ static char *load_input(int fd, const char *path)
 			usleep(XDELAY_INTERVAL_MS << 2);
 		}
 	}
+
 	free(input);
-	free(paths);
+	if (paths)
+		free(paths);
```
N-R-K, line 8103: `free(3)` is defined to be a no-op on NULL arg. So checks like these don't really have any effect.

src/nnn.c
```diff
@@ -8001,28 +8001,26 @@ static char *load_input(int fd, const char *path)
...
 		if (!input)
-			return NULL;
+			goto malloc_1;
 	}
 
 	/* Read off the extra data if limits exceeded */
 	if (msgnum) {
 		char buf[512];
-		while (read(fd, buf, 512) > 0);
+		while (read(fd, buf, 512) != 0);
```
N-R-K, line 8023: Would prefer using `sizeof` instead of hard coding.

```suggestion
		while (read(fd, buf, sizeof buf) != 0);
```

There's a lot of other places in our code that uses hard-coded sizes where sizeof would've sufficed (and would've been more robust against future changes).

src/nnn.c
KlzXS: Huh. I must have forgotten that. It's been a while since I've worked with dynamic allocation in C.

N-R-K (changes requested):

src/nnn.c
N-R-K: Now that I'm looking at it again, not sure if this is correct either.. This *might* end up going into an infinite loop where read fails for some reason and keeps returning `-1`.

N-R-K: I suppose something like this would be more robust and deal with edge case of signal interrupt as well.

```c
ssize_t n;
do {
	if ((n = read(...)) < 0 && errno != EINTR)
		break;
} while (n);
```

KlzXS: Is that what we are going with?

KlzXS: I mean this is a dumb question, but what do we do if we are given an endless input? Like `/dev/zero` or something simply huge. That shouldn't ever happen even on accident, but do we really just try to read it endlessly until the user aborts?

N-R-K: > Is that what we are going with?

Not sure. The upper `read` doesn't deal with `EINTR` either. So I guess we can leave it at `> 0`.

> I mean this is a dumb question, but what do we do if we are given an endless input?

Any specific reason we're draining the input to begin with? What happens if we don't drain it?

KlzXS: Also couldn't we just close the fd? We reopen `stdin` after the call to `load_input()` either way. Closing the fd would offload the task of dealing with the excess data to the kernel. And it can surely handle it much more efficiently.

src/nnn.c
```diff
@@ -7969,7 +7970,7 @@ static char *load_input(int fd, const char *path)
 	} else
 		xstrsncpy(cwd, path, PATH_MAX);
 
-	while ((chunk_count) < 512 && !msgnum) {
+	while (chunk_count < LIST_INPUT_MAX / chunk && !msgnum) {
 		input_read = read(fd, input + total_read, chunk);
 		if (input_read < 0) {
 			DPRINTF_S(strerror(errno));
```
N-R-K, line 7976: I suppose we can take this as an opportunity to make this more robust against signals.

```suggestion
		if (input_read < 0) {
			if (errno == EINTR)
				continue;
			DPRINTF_S(strerror(errno));
```

src/nnn.c
N-R-K: > Also couldn't we just close the fd? We reopen stdin after the call to load_input() either way

I'd go with that approach as well instead of trying to drain the input.

KlzXS: We are doing that. Kind of a weird decision to leave closing of the fd to the caller of the function, but it's called from `readpipe()` as well, which seems to also delegate the responsibility to its caller who finally is the one that opens and closes it. I'm not sure about all that. The other path through main is much simpler.

Also this just looks off to me. How is this even valid syntax.
```c
do
	rfd = open(g_pipepath, O_RDONLY);
while (rfd == -1 && errno == EINTR);
```

N-R-K: > Also this just looks wrong to me. How is this even valid syntax.

It's just omitting the  `{}` cus it's a single statement. I've seen (and done) worse stuff...

```c
	n[(char *)memcpy(d, s, n)] = '\0';
```

src/nnn.c
```diff
@@ -8001,29 +8004,23 @@ static char *load_input(int fd, const char *path)
...
+		if (chunk_count > LIST_INPUT_MAX / chunk) {
+			msgnum = MSG_SIZE_LIMIT;
+			break;
+		}
 
-		input = xrealloc(input, (chunk_count + 1) * chunk);
+		input = xrealloc(input, chunk_count * chunk);
 		if (!input)
-			return NULL;
+			goto malloc_1;
```
N-R-K, line 8020: I think `xrealloc` frees the old memory already, doesn't it?

I mean, given that `free(null)` is a no-op shouldn't be a problem, but just wanted to point it out.

src/nnn.c
```diff
@@ -190,6 +190,7 @@
 #define ASCII_MAX       128
 #define EXEC_ARGS_MAX   10
 #define LIST_FILES_MAX  (1 << 14) /* Support listing 16K files */
+#define LIST_INPUT_MAX  (1 << 28) /* Support reading in 256MiB */
```
N-R-K, line 193: Any specific reason for picking `256MiB` or just arbitrarily chosen limit?

N-R-K (approved):
Mostly seems okay to me.

src/nnn.c
KlzXS: > It's just omitting the {} cus it's a single statement.

Oh, I know why it is. But I don't really think it should.

I do like the `3[a]` way of indexing. And if it weren't hard to read that particular statement looks pretty nice. The worst I've done is something along the lines of
```c
ss = strstr(s, "...");
if (ss && ss != s && ss[-1] != '/') ...
```

src/nnn.c
KlzXS: Again you are right. I'm sure I saw it a few days ago when I was first looking into it. I say we leave it like that. It might be easier to follow if all paths lead to the same exit.

src/nnn.c
KlzXS: The original limit was 65536 paths each with `PATH_MAX == 4096` bytes, so 256 MiB in total. It seemed to me that @jarun wanted to keep it in case of invalid paths or something like that even though that's a bit overkill. I mean all of these numbers are arbitrary. Why did I originally pick 65k files? I don't know. It seemed like a nice number.

N-R-K: The manpage says the limit is 64MiB

https://github.com/jarun/nnn/blob/master/nnn.1#L315

So looks like this was just a mistake. I think we should change this to:

```suggestion
#define LIST_INPUT_MAX  (LIST_FILES_MAX * PATH_MAX)
```

Just like the manpage says.

KlzXS: He might have changed it afterwards. That's fine.

src/nnn.c
```diff
@@ -190,7 +190,7 @@
 #define ASCII_MAX       128
 #define EXEC_ARGS_MAX   10
 #define LIST_FILES_MAX  (1 << 14) /* Support listing 16K files */
-#define LIST_INPUT_MAX  (1 << 28) /* Support reading in 256MiB */
+#define LIST_INPUT_MAX  (LIST_FILES_MAX * PATH_MAX)
```
N-R-K, line 193: @KlzXS Cast this to `ssize_t` to make the CI happy, I guess.

```
 #define LIST_INPUT_MAX  ((ssize_t)LIST_FILES_MAX * PATH_MAX)
```

But now that I think about it, why do we even need this?? Cant we just do `chunk_count >= LIST_FILES_MAX` ? *Even if* we assume all of the paths are `== PATH_MAX` then this still should be enough.

N-R-K: Maybe we can just have a check to ensure all paths are `< PATH_MAX`

KlzXS: It would be enough, in fact it's way more than. Each chunk is 512 KiB. At 16384 files that's 8 GiB. We wouldn't be true to the limit to put it mildly. I'd much rather not do that.

Also someone might change `PATH_MAX` on their system to either be way smaller or even larger. I think it's better to keep it like this.

KlzXS: I added some extra parentheses. Probably not necessary, but I don't want to have to make another commit just for the CI. It won't hurt.

src/nnn.c
```diff
@@ -190,7 +190,7 @@
 #define ASCII_MAX       128
 #define EXEC_ARGS_MAX   10
 #define LIST_FILES_MAX  (1 << 14) /* Support listing 16K files */
-#define LIST_INPUT_MAX  (LIST_FILES_MAX * PATH_MAX)
+#define LIST_INPUT_MAX  ((ssize_t)(LIST_FILES_MAX * PATH_MAX))
```
N-R-K, line 193: The ci was complaining about the multiplication taking place at `int`. And I suppose that'll overflow on systems where int is 32 bits.

The additional braces make it so the mul still takes place in `int` but the result is just casted to ssize_t. Removing the inner braces should make the multiplication happen at `ssize_t`, which should more appropriate in theory.

N-R-K: But this is mostly pedantic, and I don't know weather we support 16bit int systems or not. So could just leave it be.

I have no strong opinion on this matter.

N-R-K: Ah just saw this https://github.com/jarun/nnn/pull/1451#discussion_r936073349

Yeah, i think we can leave it be. 16 bit int systems are rarely seen nowadays outside of embedded systems.

KlzXS: `ssize_t` is just too exotic of a type for me to reason through type coercion. As for the 16bit systems, I don't think we do. Don't worry too much about it. Lots of other stuff will probably break on 16bit systems. If you want you can spin up a VM and try it.

jarun:
Please squash the commits.

src/nnn.c
0xACE: On phone, but why `ssize_t` (`[-1, SSIZE_MAX]`)?

I'd say ˋ(size_t)LIST_FILES_MAX * PATH_MAXˋ is correct, though I haven't followed the entire context of the situation.

N-R-K: > On phone, but why `ssize_t` (`[-1, SSIZE_MAX]`)?

That's simply because `chunk_count` is `size_t`. 

> I'd say `(size_t)LIST_FILES_MAX * PATH_MAX` is correct, though I haven't followed the entire context of the situation.

That'd work, but then we'd need to cast the comparison to avoid comparing singed vs unsigned type: `(size_t)chunk_count >= LIST_INPUT_MAX`.

N-R-K: Now that I'm looking at it again, `chunk_count` has no reason to be `ssize_t`. I'll push a commit fixing it soon.

N-R-K: Pushed https://github.com/jarun/nnn/pull/1451/commits/ddf869675f2c5878f8aa0070fc5186d91bdd15db , see if you guys agree with it or not. @KlzXS @0xACE

0xACE: Looks good (presuming there's no need for `ssize_t` in this case)

I would prefer a easier to read `for`-loop, but I'm okay either way 🙂

N-R-K: > Looks good (presuming there's no need for `ssize_t` in this case)
> 
> I would prefer a easier to read `for`-loop, but I'm okay either way slightly_smiling_face

Hmm, I'm now also questioning why we are `free`-ing in reverse order instead of the good ol `for (i = 0; i < entries; ++i) free(paths[i]);`

N-R-K: Yeah, I'm not seeing why freeing in reverse would be needed. Just changed it to a regular loop at d208cffa

jarun:
Guys, please let me know if this is ready to merge. Also, I requested to squash all the changes.

jarun:
Sure! Thanks!

jarun:
Thank you!

PR #1456, opened 2022-08-07
Preview-tui fix image_preview stdout

### LINKED ISSUE #1455: preview-tui generate preview not working when not using uezerbug and kitty

- [x] Operating System: android, arch linux
- [ ] Desktop Environment: 
- [x] Terminal Emulator: termux, xfce4-terminal
- [x] Shell: bash
- [ ] Program options used:
- [x] Plugins are installed

In preview-tui, last few lines of `generate_preview()` 
```
399    if [ -f "$NNN_PREVIEWDIR/$3.jpg" ]; then
400        image_preview "$1" "$2" "$NNN_PREVIEWDIR/$3.jpg"
401    else
402       fifo_pager print_bin_info "$3"
403    fi
404 } >/dev/null 2>&1
```

the line 400 `image_preview` will  not working if not using ueberzug and kitty
the line 402 `fifo_pager` also not work

Because all the output are redirected to null by the line 404 `} >/dev/null 2>&1`.

Replacing the line 404 with `}` temporaily solve the problem for some system that do not have ueberzug, like termux.
`catimg` and `viu` can display the image properly after that.
But if ueberzug was installed back to system in arch linux, this will not work again. Need to replace `}` back to `} >/dev/null 2>&1`.
I am not clever enough to rewrite the script to solve the problem.

luukvbaal:
Fix #1455. @amalgame21 please confirm.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -379,7 +379,6 @@ generate_preview() {
                     printf "%s" "$!" > "$PREVIEWPID"
                     return
                  else
-                    exec >/dev/tty
                     image_preview "$1" "$2" "$3"
                     return
                  fi ;;
@@ -405,6 +404,7 @@ generate_preview() {
 
 image_preview() {
     clear
+    exec >/dev/tty
     if [ "$TERMINAL" = "kitty" ]; then
         # Kitty terminal users can use the native image preview method
         kitty +kitten icat --silent --scale-up --place "$1"x"$2"@0x0 --transfer-mode=stream --stdin=no "$3" &


amalgame21:
Seems fine to me now! Thanks!
I notice that the line `exec >/dev/tty` did the trick, but I dunno what it means actually.

Do you think the line `exec >/dev/tty` should also be added before line 401 `fifo_pager` inside the `generate_preview()` function for it to work properly? Without it seems not work if fall to that else statement.

jarun:
Thank you!

PR #1457, opened 2022-08-11
use standard C11 alignas

N-R-K:
nnn source code already uses C11, which introduced `_Alignas` [^0] as a
standardized alternative to `__attribute__((aligned))`.

there are no other usage of `__attribute__` in the nnn source code, so
in theory this should make things more portable across compilers.

also removes an unnecessary usage of `alloca()` with an aligned constant
sized buffer instead.

[^0]: https://en.cppreference.com/w/c/language/_Alignas

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -100,6 +100,8 @@
 #include <strings.h>
 #include <time.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

Issue #810, opened 2020-12-06
How to overwrite all files in one command?


midexe6380:
![16072943440](https://user-images.githubusercontent.com/65507149/101295280-b5d12b00-382d-11eb-9b3d-d4f305c31708.png)

0xACE:
before copying/moving/deleting the file it will be asking if it should use `force`, prompting yes there should delete all of the files

midexe6380:
@0xACE   I have no option `force`. How do I activate this option?

jarun:
In the following prompt, press `y`:

    rm -rf 4 file(s)? [Esc cancels]

midexe6380:
@jarun I don't understand you. I can't do that. Help me, please. Here's what I get
[https://youtu.be/eREn8KCeTR8](url)

jarun:
Ok. I got it. We always use interactive mode for cp and mv to be on the safer side. To overwrite without asking you'll have to change the code.

midexe6380:
LOL 😅 you use cp fork with a progress bar. It's not safe at all.

jarun:
It's optional.

0xACE:
@midexe6380 you are suppose to receive a prompt

`rm -rf current file? [Esc cancels]`

or similiar when you do writing operations onto disks... Are you using `master`?

midexe6380:
@0xACE 


> Are you using `master`?

Yes.



> you are suppose to receive a prompt

I get  [https://youtu.be/eREn8KCeTR8](https://youtu.be/eREn8KCeTR8l)

I thought what could be done ```ya``` (yes all) and replace all files.

midexe6380:
> ```
> rm -rf 4 file(s)? [Esc cancels]
> ```

where did you find this promt?
It feels like we are using different versions of NNN.

0xACE:
> I get [https://youtu.be/eREn8KCeTR8](https://youtu.be/eREn8KCeTR8l)

I'm not familiar with the `master` ui so I can't even tell what you are doing in that video

> where did you find this promt?
> It feels like we are using different versions of NNN.

That is correct I'm running an outdated personal build of `nnn`

I just tried it out, with copy operation and in my `nnn` it doesn't ask to use force... maybe:

> Ok. I got it. We always use interactive mode for cp and mv to be on the safer side. To overwrite without asking you'll have to change the code.

This is correct and my memory is serving me wrong... (maybe `nnn` asked to use force in the past? i can't remember)

[...]

PR #1458, opened 2022-08-11
Add cpoverwrite plugin

Darukutsu:
I stumbled across this closed [issue #810 ](https://github.com/jarun/nnn/issues/810) and tried to achieved something like that with plugin. Not sure if it's POSIX compliant since `cp -n`. I read somewhere it has some issues on mac or maybe not... I don't remember. I tested it, hopefully didn't forget something.

I was having some seizures while typing 'y' 'enter', 'y' 'enter'... The thing is if **nnn** has this feature built-in then I don't know about it.

Also what does "[✓]" mean? I didn't find reference.


jarun:
Why can't you have a local copy with yourself? It's pretty basic.

Darukutsu:
My bad bro. I thought someone was having same problem as me.

jarun:
No issues. Plugins are simple to write so we are a bit selective there.

PR #1459, opened 2022-08-12
Add plugin to open all selected files in tabs in vim

sumanchapai:
For vim users, this is a helpful feature if you are looking to edit
multiple files then this plugin will allow you to do exactly that.

### DIFF
--- a/plugins/vimopenall
+++ b/plugins/vimopenall
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Open selected files in vim
+# Modified from 
+# https://github.com/jarun/nnn/blob/v4.6/plugins/openall
+
+sel=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
+OPENER="vim"
+
+if [ -s "$sel" ]; then
+    targets=()
+    while IFS= read -r -d '' entry || [ -n "$entry" ]; do
+        # Only open files (not directories)
+        if [[ -f $entry ]]
+        then
+        targets+=( "$entry" )
+        fi
+    done < "$sel"
+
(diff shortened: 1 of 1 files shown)


jarun:
Please add the functionality to plugin `openall`.

sumanchapai:
I am not sure what we would like to do to the existing function of openall.

jarun:
Good. Raise the PR when you are sure.

sumanchapai:
Could you reopen the PR? I have integrated the functionality  in `openall`

N-R-K:
Why do we need to special case vim? AFIAK opening multiple files with vim would just open each on a new tab.

sumanchapai:
As of now, the plugin is using a hardcoded `OPENER` variable. One of the additions in this recent commit, 
https://github.com/sumanchapai/nnn/commit/3bfe5aaa4b02ad45ef58b724391304cb0e485ee4 is that it uses `EDITOR` env variable. Additionally, to open files in tabs one needs to supply `-p` flag. But users usually have vim aliased to `vim -p`.

N-R-K:
> One of the additions in this recent commit, [sumanchapai@3bfe5aa](https://github.com/sumanchapai/nnn/commit/3bfe5aaa4b02ad45ef58b724391304cb0e485ee4) is that it uses `EDITOR` env variable.

I don't think that's how the plugin is meant to be used. $OPENER is supposed to deal with the file-type and open it in a suitable application. The way you commit works, is it'll try an open *any type* of file with vim if `$EDITOR` is set to vim, which is certainly not a good behavior.

sumanchapai:
Thank you for elaborating. Looks like the nuke plugin handles my requested use case.

PR #1460, opened 2022-08-12
Add plugin to force delete selected files/directory

sumanchapai:
While there already exists a functionality in `nnn` to delete files or
directories, it requires many key presses. While this functionality is
dangerous, it's there as plugin for users who want it.


jarun:
We don't add plugins for basic functionality. Have your own local copy.

PR #1461, opened 2022-08-16
xstrdup: use memcpy

N-R-K:
the length is already known, so memcpy should be faster than xstrsncpy.


N-R-K:
The CI failure doesn't seem to be related to the PR. Not sure what it's about.

jarun:
Thank you!

Issue #1479, opened 2022-09-07
[Feature request] Keep selected count visible in status bar during action confirmation, e.g. deleting files.


amalgame21:
Very minor feature request here.

Since nnn allow selecting items in different directories, deleting files before confirming selected count is crucial to me to avoid deleting wrong items.

After selected files for deletion, pressing `x` `s` will go to `rm -rf selection?` dialog.
In this dialog, no selected number is shown.
Have to press `esc` to see the selected number to confirm correct number of file to be deleted, and then press `x` `s` `y` again to delete those file.

If selected count is shown in the `rm -rf selection` dialog. May prevent some extra keypresses. 
I tried changing my personal behavior to focus on the selection before pressing `x`, but failed. I found it not very intuitive.

luukvbaal:
I like it, implemented in #1480. Let's see what @jarun thinks.

amalgame21:
Thanks luukvbaal and jarun !!!!!

PR #1480, opened 2022-09-07
Show selected items in rm msg

luukvbaal:
Close https://github.com/jarun/nnn/issues/1479

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -634,7 +634,7 @@ static const char * const messages[] = {
 	"session name: ",
 	"'c'p/'m'v as?",
 	"'c'urrent/'s'el?",
-	"%s %s? [Esc cancels]",
+	"%s %s%s? [Esc cancels]",
 	"size limit exceeded",
 	"'f'ile/'d'ir/'s'ym/'h'ard?",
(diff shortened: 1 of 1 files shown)


N-R-K (approved):
I remember a couple times I trashed something else in another directory by mistake because they were still selected. So I think showing the selection count makes sense.

src/nnn.c
```diff
@@ -1429,7 +1429,8 @@ static char confirm_force(bool selection)
 
 	snprintf(str, 64, messages[MSG_FORCE_RM],
 		 g_state.trash ? utils[UTIL_GIO_TRASH] + 4 : utils[UTIL_RM_RF],
-		 (selection ? "selection" : "hovered"));
+		 (selection ? xitoa(nselected) : "hovered"),
```
jarun, line 1432: Why not something like:

- hovered
- selected 5

jarun (changes requested):

src/nnn.c
luukvbaal: Sure, done.

jarun:
Thank you!

PR #1485, opened 2022-09-12
Detect key collisions: make clear everywhere that it exits afterwards.

pataquets:
While trying command line switches, kept me scratching my head for a few minutes.
I wrongly understood that it would check for key collisions, then start. As if it was an optional check, skipped for fastest bootup. :facepalm:

Stating it explicitly, just in case.


N-R-K (approved):

jarun:
Thank you!

PR #1491, opened 2022-09-16
slightly improve n.fish

PatrickF1:
The expr is redundant because `test -ge` treats strings as numbers already. I combined the two tests into one line because it is slightly more idiomatic (adhering universal coding phrase of check if null, and if not null then check value equality).

Tested on fish by spinning up `n`, spawning a shell, and then executing `n` again.
Other tests
```
set NNNLVL 1
test -n "$NNNLVL" -a "$NNNLVL" -ge 1 # true
set NNNLVL 2
test -n "$NNNLVL" -a "$NNNLVL" -ge 1 # true
set NNNLVL 0
test -n "$NNNLVL" -a "$NNNLVL" -ge 1 # false
set --erase NNNLVL
test -n "$NNNLVL" -a "$NNNLVL" -ge 1 # false
```

### DIFF
--- a/misc/quitcd/quitcd.fish
+++ b/misc/quitcd/quitcd.fish
@@ -4,11 +4,9 @@
 
 function n --wraps nnn --description 'support nnn quit and change directory'
     # Block nesting of nnn in subshells
-    if test -n "$NNNLVL"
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1497, opened 2022-09-26
Check if selected and current is the same

IAmKapuze:
Do not display a "'c'urrent/'s'el?" message if selected and current is the same

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1403,6 +1403,13 @@ static int get_cur_or_sel(void)
 		if (cfg.prefersel && selbufpos)
 			return 's';
 
+		if (nselected == 1) {
+			char choicefullpath[PATH_MAX];
+			snprintf(choicefullpath, PATH_MAX, "%s/%s", g_ctx[cfg.curctx].c_path, pdents[cur].name);
+			if (!strcmp(choicefullpath, pselbuf))
+				return 'c';
+		}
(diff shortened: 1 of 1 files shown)


jarun:
By default, selection would move the cursor forward.
If we are talking of only 1 file in the current directory of the last file, it's a very easy case to identify.
Wouldn't add new logic for that.

Also, why not prioritize selection over current? In fact calling it current/selection can be confusing for the user who already knows he has selected one file explicitly and wants to delete it.

IAmKapuze:
I can select several files (from several directories) and later deselect part of them. I can remember, that I left selecting for only one file. But, when I move cursor to it and press 'x', I get question: "Delete current or selected?", but it's the same. It confus me, and I think "may be I not deselected ALL other files?". I press ESC and check selecting count, but no, selected only one file and it is current.

Thus I have to look at the number of selected files before deleting and I have to do extra (useless) action (choose current or selected). Why I need to choose between current and selected if it the same?

> Also, why not prioritize selection over current?

There is no practical difference, they are the same, but "selected" keep absolute path, "current" - relative path (it's shorter). To transmiting less data to command I chose "current".

jarun:
> but it's the same. It confus me, and I think "may be I not deselected ALL other files?"

Press `m` twice at this point.

PR #1501, opened 2022-10-01
Fix gitstatus patch git command

### LINKED ISSUE #1500: Entering directories with many directories inside is very slow with gitstatus patch
#### Environment details

- [x] Operating System: Void Linux
- [x] Desktop Environment: n/a
- [x] Terminal Emulator: Alacritty
- [x] Shell: Bash
- [x] Custom desktop opener (if applicable): n/a
- [x] Program options used: O_GITSTATUS
- [x] Configuration options set: none
- [ ] Plugins are installed

1. Build `nnn` with the `gitstatus` patch.
2. Clone a repository with lots of directories, e.g. [void-linux/void-packages](https://github.com/void-linux/void-packages).
3. Try to use `nnn` in the cloned repository and be annoyed that entering the `srcpkgs/` directory takes ~10 seconds.

---

The `srcpkgs/` directory has 10k+ directories inside it. Entering it without the `gitstatus` patch is instantaneous.

<strike>My C debugging skills are extremely limited, so I probably have no idea what I'm talking about, but it seems to me that the performance hit comes from doing a linear search for what the status of each directory is.</strike> I'm probably wrong here.

I know it says [here](https://github.com/void-linux/void-packages.git) that you don't "optimize handling 10k+ entries in a dir", but I thought that this performance hit should be considered a bug since entering the large directory is instantaneous without the patch.

Logs: [nnndbg](https://github.com/jarun/nnn/files/9690057/nnn.log) (it was too long to include here)

luukvbaal:
Fix #1500. No longer rely on the git status pathspec to filter out dotfiles for `!cfg.showhidden`. The pathspec we used expanded to all files/directories in the current directory, leading to poor performance in large directories. Now we filter it out ourselves by using the `-s` flag to `git status`, which shows relative paths, and skip entries that start with a dot.

Also added a missing semicolon for O_DEBUG.

### DIFF
--- a/patches/gitstatus/mainline.diff
+++ b/patches/gitstatus/mainline.diff
@@ -6,12 +6,12 @@
 # Authors: Luuk van Baal
 
 diff --git a/src/nnn.c b/src/nnn.c
-index f8a2c58..157cea5 100644
+index 83ecdb90..4397944a 100644
 --- a/src/nnn.c
 +++ b/src/nnn.c
-@@ -265,6 +265,25 @@
- #define FREE     0
- #define CAPACITY 1
+@@ -270,6 +270,25 @@
+ #define VFS_USED  1
+ #define VFS_SIZE  2
 
 +/* Git icons */
 +#ifdef NERD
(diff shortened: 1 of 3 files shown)


jarun:
Thank you!

PR #1504, opened 2022-10-06
Open in specified editor

kavulox:
This allows users to specify an editor that they would want to use in addition to $EDITOR

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -426,6 +426,7 @@ static uchar_t cfgsort[CTX_MAX + 1];
 static char *bmstr;
 static char *pluginstr;
 static char *orderstr;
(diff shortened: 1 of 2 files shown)


0xACE:
Doubt it will get merged to master, because it wouldn't serve anyone who hasn't compiled `nnn`, and whoever feels like compiling `nnn` can just hardcode `editor` with their own string...

Besides you have practically duplicated lines, there are multiple ways to refactor this...

N-R-K:
This can be easily achieved via a plugin. Make a plugin like this:

```sh
#!/bin/sh
editor "$1"
```

assign it to a whatever `key` you like via `$NNN_PLUG` then press `alt+key`.

PR #1510, opened 2022-10-24
Adding many new icons and colors

Babkock:
[Screenshot](https://babkock.github.io)

Wow people still write programs in C?? Hello I love this program. It has replaced Ranger as a file manager for me, I find it very efficient and fast. I hope this pull request finds you well. Here is a summary of the changes I am proposing:

* Added many new icons, file extensions, and file names
* Added new color for Emacs, lisp, and Org documents (`Magenta2`)
* Added new color for Rust source files (`IndianRed1`)
* C and C++ header files now have their own icon
* C++ is now a slightly different color than C file icon (`DarkSeaGreen2`)
* Added color for various "key" file types (`.pub`, `.key`, `.kbx`) (`Khaki3`)
* Added color for various "lock" file types (`.lock`, `.asc`, `.ssh`)
* Assembly source files now have a color (`Gold1`)
* HTML files now have a color (`LightSteelBlue`)
* Made a separate file icon for archives, to distinguish them from software packages
* Partially-downloaded files now have a clock icon and a color
* The standard home-directory folders "Music", "Downloads", "Pictures", and "Videos" now have the default colors for audio files, downloads, image files, and videos respectively

Please let me know if there are any changes I can make, or any questions you have. If you think something should be a different color, please feel free to suggest it. I tried not to change too much, or change the currently established icons and colors.

### DIFF
--- a/src/icons.h
+++ b/src/icons.h
@@ -66,6 +66,7 @@
 #define ICON_DOCUMENT      ICON_STR(FA_FILE_TEXT_O, "\uf718", "🗒 ")
 #define ICON_DOWNLOADS     ICON_STR(FA_DOWNLOAD, "\uf5d7", "📥")
 #define ICON_ELIXIR        ICON_STR(MFIZZ_ELIXIR, "\ue62d", "💧")
+#define ICON_EMACS         ICON_STR(DEV_GNU, "\ue779", "")
 #define ICON_ENCRYPT       ICON_STR("", "\uf805", "🔒")
 #define ICON_FSHARP        ICON_STR(DEV_FSHARP, "\ue7a7", "")
 #define ICON_GIT           ICON_STR(FA_GIT, "\ue5fb", "🌱")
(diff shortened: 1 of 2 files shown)


src/icons.h
```diff
@@ -53,56 +53,81 @@
 #define ICON_EXEC          ICON_STR(FA_COG, "\uf144", "⚙️ ")
 
 /* Top level and common icons */
-#define ICON_ARCHIVE       ICON_STR(FA_FILE_ARCHIVE_O, "\uf53b", "📦")
+#define ICON_ARCHIVE       ICON_STR("", "\uf410"," 📦")
```
N-R-K, line 56: Why remove `FA_FILE_ARCHIVE_O` ?

src/icons.h
```diff
@@ -178,19 +213,67 @@ static const struct icon file_icon = {ICON_FILE, 0};
...
+	{"Pictures",           ICON_PICTURES,      COLOR_IMAGE},
+	{"Public",             ICON_PUBLIC,        0},
+	{"Rakefile",           ICON_RUBY,          COLOR_RUBY},
+	{"rc.lua",             ICON_CONFIGURE,     COLOR_LUA},
+	{"sxhkdrc",            ICON_CONFIGURE,     COLOR_CONFIGURE},
+	{"Templates",          ICON_TEMPLATES,     0},
+	{"Vagrantfile",        ICON_CONFIGURE,     COLOR_CONFIGURE},
+	{"Videos",             ICON_VIDEOS,        COLOR_VIDEO},
+	{"webpack.config.js",  ICON_WEBPACK,       COLOR_JAVASCRIPT},
+	{"xmonad.hs",          ICON_CONFIGURE,     COLOR_CONFIGURE},
```
N-R-K, line 276: This part is probably not going to get added due to the reasons described [here](https://github.com/jarun/nnn/pull/1419#issuecomment-1179213035).

Could ofc turn this into a hash-table as well for faster lookups, but not sure if it's worth the effort.

N-R-K (changes requested):

jarun:
We are conservative in this area because adding every new icon increases memory usage and lookup time. And we want directories of the same color.

I understand you are from ranger. nnn follows different design principles.

jarun:
Also the image leads to your personal website. Why?

Babkock:
Okay I understand the reasons why although it seems a bit arbitrary. What if I made a patch for extra icons that people could choose to compile in if they wanted, like gitstatus and namefirst? So if you don't want that many icons you don't have to have them. I can remove directory colors, and use hash table idea if you want.

Would you be more open to merging a pull request from me if I took the diff from this, and submitted it as a patch (with directory colors reverted), and adjusted the `Makefile`, `check-patches.sh`, etc.? Or would it be not worth the effort?

jarun:
I would leave that to @luukvbaal and @N-R-K to decide.

N-R-K:
I don't think patches like that would be a good idea because it's going collide alot and cause significant maintenance overhead.

The goal with icons isn't to provide every icon under the sun, it's to provide a decent *but small* set of defaults. Users can then add more if they want to.

@Babkock I'll take a look here and see if there were anything that might be worth adding.

N-R-K:
> @Babkock I'll take a look here and see if there were anything that might be worth adding.

Added ksh, ttf and otf icon in #1511.

PR #1512, opened 2022-11-01
plugins/wallpaper: support wayland and fix indentation

Darukutsu:
nitrogen & pywal don't set background on wayland. Therefore add swww.

### DIFF
--- a/plugins/wallpaper
+++ b/plugins/wallpaper
@@ -8,29 +8,29 @@
 # Author: juacq97
 
 if [ -n "$1" ]; then
-	if [ "$(file --mime-type "$1" | awk '{print $NF}' | awk -F '/' '{print $1}')" = "image" ]; then
-	  if [ "$XDG_SESSION_TYPE" = "x11" ]; then
-	    if type nitrogen >/dev/null 2>&1; then
-		    nitrogen --set-zoom-fill --save "$1"
-	    elif type wal >/dev/null 2>&1; then
-		    wal -i "$1"
-	    else
-		printf "nitrogen or pywal missing"
-		read -r _
-	    fi
-	  else
-	    if type swww >/dev/null 2>&1; then
-		    swww img "$1"
-	    else
-		printf "swww missing"
-		read -r _
-	    fi
-	  fi
-
+    if [ "$(file --mime-type "$1" | awk '{print $NF}' | awk -F '/' '{print $1}')" = "image" ]; then
+        if [ "$XDG_SESSION_TYPE" = "x11" ]; then
+            if type nitrogen >/dev/null 2>&1; then
+                nitrogen --set-zoom-fill --save "$1"
+            elif type wal >/dev/null 2>&1; then
+                wal -i "$1"
+            else
+                printf "nitrogen or pywal missing"
+                read -r _
+            fi
+        else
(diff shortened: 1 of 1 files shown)


plugins/wallpaper
```diff
@@ -9,14 +9,24 @@
...
 	if [ "$(file --mime-type "$1" | awk '{print $NF}' | awk -F '/' '{print $1}')" = "image" ]; then
-        if type nitrogen >/dev/null 2>&1; then
-	        nitrogen --set-zoom-fill --save "$1"
-        elif type wal >/dev/null 2>&1; then
-	        wal -i "$1"
-        else
-            printf "nitrogen or pywal missing"
-            read -r _
-        fi
+	  if [ "$XDG_SESSION_TYPE" = "x11" ]; then
```
N-R-K, line 12: I don't think checking for this is necessary. If nitrogen and wal don't work on wayland then the user probably doesn't have it installed anyways.

plugins/wallpaper
```diff
@@ -9,14 +9,24 @@
...
+		    wal -i "$1"
+	    else
+		printf "nitrogen or pywal missing"
+		read -r _
+	    fi
+	  else
+	    if type swww >/dev/null 2>&1; then
+		    swww img "$1"
+	    else
+		printf "swww missing"
```
N-R-K, line 25: I'd drop the `XDG_SESSION_TYPE` check and make this error message more generic, something like: "couldn't find appropriate program to set the wallpaper"

N-R-K (changes requested):

plugins/wallpaper
Darukutsu: I think it good to have if you tend to switch between x11 and wayland. Bspwm to hyperland for example.

N-R-K: Ah, I see. Didn't consider people are using x11 and wayland both.

N-R-K (approved):

jarun:
Thank you!

jarun:
@Darukutsu if you can test Wayland, are there any other plugins which need changes to work on Wayland?

Darukutsu:
@jarun everything look ok, only plugin which behave incorectly is 'preview-tabbed'.
Universal solution for wayland would be using [wlrctl](https://git.sr.ht/~brocellous/wlrctl)
It only works with "urxvt" and need to add line in previewer_loop
```
...

**Comment by Darukutsu** [comment]
@jarun everything looks ok. The only plugin which behave incorrectly is 'preview-tabbed'.
Universal solution for wayland would be using [wlrctl](https://git.sr.ht/~brocellous/wlrctl)
It only works with 'urxvt' and need to add line
```
previewer_loop() {
...
    wlrctl window focus title:nnn
    
    # Bruteforce focus stealing prevention method,
    # works well in floating window managers like ...
...
}
```
It's not suitable since user may have open multiple instances with nnn, but other than that there is only compositor-specific solutions. I don't think it's needed if using something like 'preview-tui' with kitty.

Also today discovered that alacritty supports XEmbed so maybe add to preview-tabbed
```
elif type alacritty >/dev/null 2>&1 ; then
    TERMINAL="alacritty --embed"
```

PR #1531, opened 2022-11-25
Respect CFLAGS and LDFLAGS when building generator

sjmulder:


N-R-K:
What's the motivation for this? Does it make the generation process faster?

sjmulder:
These flags can be (and are) set in the build environment to configure things like include directories (e.g. `/usr/pkg/include`), extra compiler and linker flags for e.g. security, etc. These should be used when invoking the compiler or linker, and that's already the case in other places in this Makefile but not here.

N-R-K:
> that's already the case in other places in this Makefile but not here

That's because the generator is never installed or used by the user - it just generates the hash-table. I'm not against this change, but I'm curious what the motivation was because I don't think it brings any practical value.

sjmulder:
I know the generator isn't installed, however the flags are often necessary to make builds work on exotic platforms and we also use them for security hardening (enable fortify, relro). Those apply for intermediate tools as well. To be honest, I feel that the question should be about why such flags should be ignored.

jarun:
I'm fine with merging it.

jarun:
Thank you guys!

PR #1532, opened 2022-11-25
Add cppath plugin

raffaem:


jarun:
Please explain with path examples. Also, what's the use case?

raffaem:
You map the plugin to `;p`.
You select `/home/jarun/Documents/hello.txt`. You type `;p;p`. `/home/jarun/Documents/hello.txt` is copied in the clipboard.
You select `/home/jarun/Documents/hello.txt`. You type `;p`. You select `/home/jarun/Pictures/cat.jpg`. You type `;p`. `../Pictures/cat.jpg` is copied in the clipboard.

Use case:
You are editing `/home/jarun/mysoftware/src/main.cpp`.
You need data located in `/home/jarun/mysoftware/data/img/myimage.jpg`.
You can copy the relative path in the clipboard and then into the source file.

Another use case: you are writing in a forum to ask for help for the configuration file `/etc/systemd/systemd.d/myconf.f`. You copy the path in the clipboard and paste it into the forum.

jarun:
The `-x` program option already copies a selected file to clipboard.

raffaem:
I just had a look at it and it doesn't work for me because the choice between xclip and wl-copy should be based on XDG_SESSION_TYPE. I have both a Wayland desktop and a X11 desktop installed, but I'm running Wayland right now.

Anyway, after that is fixed, we can keep this for relative paths (first use case).

jarun:
> after that is fixed

You can add the same check in `.cbcp`.

jarun:
Also, your change is linux-only. Please use it locally. You can raise PR for the `XDG_SESSION_TYPE` check in `.cbcp`.

jarun:
Checks for `XDG_SESSION_TYP` added at commit 6bffaeeb445f10a88899626526316167aa4253cc.

raffaem:
I'm providing my plugin [here](https://github.com/raffaem/cppath) in case someone needs it. I think it is still useful for relative paths when you are programming and you need the relative paths to resources.

> Also, your change is linux-only.

It would be easy to add the checks for clip.exe, clipboard and the other tools the other plugins are using.

PR #1543, opened 2022-12-13
Improve kdeconnect plugin

raffaem:
- Work with multiple connected devices by either (1) sending files to the first device, if you only have sh, or (2) letting you choose which device to send to, if you have bash
- Support multi-files selection
- Support sending hovered file in case no file is selected

### DIFF
--- a/plugins/kdeconnect
+++ b/plugins/kdeconnect
@@ -1,24 +1,72 @@
 #!/usr/bin/env sh
 
-# Description: Send the selected files to your Android device using kdeconnect-cli.
+# Description: Send files or folders to your Android device using kdeconnect-cli.
 #              kdeconnect must be configured on the Android device and the PC.
 #
+# Usage:
+#   - Hover over a file or a folder and call the plugin.
+#   - Alternatively, select the files and folders you would like to send, and activate the plugin.
+#
 # Shell: POSIX compliant
-# Author: juacq97
+# Author: juacq97, raffaem
 
+# If you want system notification, put this equal to 1
+notify=1
+
+note() {
+	if [ $notify = 1 ]; then
+		notify-send -a "Kdeconnect" "$1"
+	else
+		echo "[Kdeconnect] $1"
+	fi
+}
+
+# If you have only sh, use this
+devname=$(kdeconnect-cli --list-available --name-only 2>/dev/null | awk NR==1)
+if [ -z "$devname" ]; then
+	note "No devices available"
+	exit
+fi
+
+# If you are privileged enough to have bash, use this instead
+# names=$(kdeconnect-cli --list-available --name-only 2>/dev/null)
+# IFS=$'\n' read -rd '' -a namesarr <<<"$names"
+# length=${#namesarr[@]}
+# if [ "$length" = 0 ]; then
+# 	note "No devices available"
+# 	exit
+# elif [ "$length" = 1 ]; then
+# 	devname="${namesarr[0]}"
+# else
+# 	for (( j=0; j<length; j++ ));
+# 	do
+# 	  printf "%d. %s\n" "$j" "${namesarr[$j]}"
+# 	done
+# 	read -r -p "Choice: " choice
+# 	devname="${namesarr[$choice]}"
+# fi
+
+# Get file that contains NULL-separated list of selected files
 selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
 
-id=$(kdeconnect-cli -a --id-only | awk '{print $1}')
-if [ -s "$selection" ]; then
-    kdeconnect-cli -d "$id" --share "$(cat "$selection")"
-
-    # If you want a system notification, uncomment the next 3 lines.
-    #notify-send -a "Kdeconnect" "Sending $(cat "$selection")"
-#else
-    #notify-send -a "Kdeconnect" "No file selected"
-
-    # Clear selection
-    if [ -p "$NNN_PIPE" ]; then
-        printf "-" > "$NNN_PIPE"
-    fi
+if [ ! -s "$selection" ]; then
+	if [ -n "$1" ]; then
+		selection=$(mktemp)
+		printf "%s" "$1" > "$selection"
+	else
+		note "No selected files, and no hovered file"
+		exit
+	fi
 fi
+
+# Send files over
+<"$selection" xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
+
+# Show notification
+note "Files sent"
+
+# Clear selection
+if [ -p "$NNN_PIPE" ]; then
+	printf "-" > "$NNN_PIPE"
+fi
+
--- a/plugins/kdeconnect
+++ b/plugins/kdeconnect
@@ -11,7 +11,7 @@
 # Author: juacq97, raffaem
 
 # If you want system notification, put this equal to 1
-notify=1
+notify=0
 
 note() {
 	if [ $notify = 1 ]; then
@@ -21,52 +21,23 @@ note() {
 	fi
 }
 
-# If you have only sh, use this
+send() {
+	xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
+	note "Files sent"
+}
+
 devname=$(kdeconnect-cli --list-available --name-only 2>/dev/null | awk NR==1)
 if [ -z "$devname" ]; then
 	note "No devices available"
 	exit
 fi
 
-# If you are privileged enough to have bash, use this instead
-# names=$(kdeconnect-cli --list-available --name-only 2>/dev/null)
-# IFS=$'\n' read -rd '' -a namesarr <<<"$names"
-# length=${#namesarr[@]}
-# if [ "$length" = 0 ]; then
-# 	note "No devices available"
-# 	exit
-# elif [ "$length" = 1 ]; then
-# 	devname="${namesarr[0]}"
-# else
-# 	for (( j=0; j<length; j++ ));
-# 	do
-# 	  printf "%d. %s\n" "$j" "${namesarr[$j]}"
-# 	done
-# 	read -r -p "Choice: " choice
-# 	devname="${namesarr[$choice]}"
-# fi
-
-# Get file that contains NULL-separated list of selected files
 selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
-
-if [ ! -s "$selection" ]; then
-	if [ -n "$1" ]; then
-		selection=$(mktemp)
-		printf "%s" "$1" > "$selection"
-	else
-		note "No selected files, and no hovered file"
-		exit
-	fi
+if [ -s "$selection" ]; then
+	send < "$selection"
+	[ -p "$NNN_PIPE" ] && printf "-" > "$NNN_PIPE" # clear selection
+elif [ -n "$1" ]; then
+	printf "$1" | send
+else
+	note "No selection and no hovered file"
 fi
-
-# Send files over
-<"$selection" xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
-
-# Show notification
-note "Files sent"
-
-# Clear selection
-if [ -p "$NNN_PIPE" ]; then
-	printf "-" > "$NNN_PIPE"
-fi
-


plugins/kdeconnect
```diff
@@ -4,21 +4,32 @@
...
 # Shell: POSIX compliant
-# Author: juacq97
+# Author: juacq97, raffaem
 
 selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
 
-id=$(kdeconnect-cli -a --id-only | awk '{print $1}')
-if [ -s "$selection" ]; then
-    kdeconnect-cli -d "$id" --share "$(cat "$selection")"
+if [[ ! -s "$selection" ]];
```
KlzXS, line 11: SC2039: In POSIX sh, [[ ]] is undefined.

plugins/kdeconnect
```diff
@@ -4,21 +4,32 @@
...
-    #notify-send -a "Kdeconnect" "No file selected"
+# Get ID of first available paired device
+id=$(kdeconnect-cli --list-available --id-only | awk '{print $1}')
 
-    # Clear selection
-    if [ -p "$NNN_PIPE" ]; then
-        printf "-" > "$NNN_PIPE"
-    fi
+# Read $selection as a list of NULL-separated strings
+while read -d $'\0' file ; do
```
KlzXS, line 25: SC2162: read without -r will mangle backslashes.
SC2039: In POSIX sh, read -d is undefined.
SC2039: In POSIX sh, $'..' is undefined.

plugins/kdeconnect
```diff
@@ -4,21 +4,32 @@
...
-    kdeconnect-cli -d "$id" --share "$(cat "$selection")"
+if [[ ! -s "$selection" ]];
+then
+	# If you want system notification, uncomment this line
+	#notify-send -a "Kdeconnect" "No file selected"
+	exit
+fi
+
+# Insert a NULL character at the end of $selection file
+echo -en '\0' >> "$selection"
```
KlzXS, line 19: SC2039: In POSIX sh, echo flags are undefined.

KlzXS (changes requested):
I added the complaints that shellcheck has. Please correct those before we go any further.

N-R-K:
What is the problem with the current script? Does `kdeconnect-cli` not accept multiple files as cli arg?

N-R-K:
But in any case, I'd use `xargs -0` instead of a loop. Try to see the existing plugins on how that works.

```sh
    xargs -0 -I{} < "$selection" kdeconnect-cli -d "$id" --share {}
```

Untested by this should probably work.

plugins/kdeconnect
```diff
@@ -4,21 +4,32 @@
 #              kdeconnect must be configured on the Android device and the PC.
 #
 # Shell: POSIX compliant
-# Author: juacq97
+# Author: juacq97, raffaem
```
jarun, line 7: Can you please explain what this change is about? You would also need to document the behaviour.

jarun (changes requested):

plugins/kdeconnect
raffaem: I'm not sure I understood your request.

You want me to write here in that part of the script what I did specifically?

raffaem:
> What is the problem with the current script? Does `kdeconnect-cli` not accept multiple files as cli arg?

It doesn't for me, or at least it doesn't with NULL-separated strings list.

It would be useful if someone can confirm that.

KlzXS:
> It doesn't for me, or at least it doesn't with NULL-separated strings list.

To avoid the need for appending a NULL you can use `xargs -0` as @N-R-K suggested.

If `kdeconnect-cli` really doesn't want to work with multiple files in one command (you can test that by hand) add a `-n 1` flag to `xargs` so it goes file by file.

Something like `xargs -0 -n 1 < "$selection" kdeconnect-cli -d "$id" --share `

plugins/kdeconnect
raffaem: The current documentation actually describes the new behavior, as it says "Send the selected file**s**"

raffaem:
> To avoid the need for appending a NULL you can use xargs -0 as @N-R-K suggested.

Yes, that works, thanks.

In fact the man page for `xargs` says, if I interpret it correctly:
>  As  of  GNU xargs version 4.2.9, the default behaviour of xargs is not to have a logical end-of-file marker.  POSIX
       (IEEE Std 1003.1, 2004 Edition) allows this.

> If kdeconnect-cli really doesn't want to work with multiple files in one command (you can test that by hand) add a -n 1 flag to xargs so it goes file by file.

My understanding is that @N-R-K suggested -I implies -L 1, which passes file one by one

raffaem:
> What is the problem with the current script? Does `kdeconnect-cli` not accept multiple files as cli arg?

The answer seems to be no.

The man page for `kdeconnect-cli` says:

> --share <path or URL>         Share **a** file/URL to a said device

plugins/kdeconnect
jarun: together or one by one? what difference does your change make?

raffaem: > together or one by one? what difference does your change make?

Select multiple files with Space and activate the plugin.

Current version: only the first file is sent to the smartphone

This version: All files are sent to the smarphone

KlzXS:
> My understanding is that @N-R-K suggested -I implies -L 1, which passes file one by one

That's correct, but I prefer using `-n 1` when the argument should get tacked on the end.

I think that snippet from GNU xargs is referencing the `-E` option. Tbh today is the first time I read up on what that does. Apparently you can make `xargs` go until it hits a certain argument. By default either "_" or non-existant. GNU picked one.

plugins/kdeconnect
jarun: OK. Then no further documentation is necessary.

jarun:
@KlzXS @N-R-K I think this is a good improvement. Please let me know when it's ready to merge.

raffaem:
I would like to implement also the following:

1. Send the hovered file if no selection is given
2. Support for multiple paired devices (ask the user which one to send to)

I don't know if that would overblow the addon.

Also, what are the requisites to have an addon under bash? Can we assume that, if the user has kdeconnect-cli, then it also has bash?

I'm asking because the shellcheck is remarking bash-like syntax which is not POSIX sh compatible.

plugins/kdeconnect
```diff
@@ -4,21 +4,37 @@
...
+
+# Get ID of first available paired device
+id=$(kdeconnect-cli --list-available --id-only | awk '{print $1}')
 
-    # If you want a system notification, uncomment the next 3 lines.
-    #notify-send -a "Kdeconnect" "Sending $(cat "$selection")"
-#else
-    #notify-send -a "Kdeconnect" "No file selected"
+# Build command to share file
+cmd="kdeconnect-cli -d '$id' --share '{}'"
```
N-R-K, line 28: Can you check if this works if the filename has `'` in it?

plugins/kdeconnect
```diff
@@ -4,21 +4,37 @@
...
+# If you want system notification, put this equal to 1
+notify=0
 
 selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
 
-id=$(kdeconnect-cli -a --id-only | awk '{print $1}')
-if [ -s "$selection" ]; then
-    kdeconnect-cli -d "$id" --share "$(cat "$selection")"
+if [ ! -s "$selection" ];
+then
```
N-R-K, line 15: try to put the `then` on the same line as the `if`

jarun:
1. Send the hovered file if no selection is given

is OK. Other one we can skip.

N-R-K (changes requested):

KlzXS:
> Also, what are the requisites to have an addon under bash? Can we assume that, if the user has kdeconnect-cli, then it also has bash?

In general we try to keep it as POSIX compliant as possible. The main exception is if it would make the plugin absurdly complex or perform weakly.  One such example would be arrays. POSIX shell doesn't support arrays per se and getting around that is just not worth it.

I wouldn't make any assumptions on what shell someone uses. They might use classic sh for whatever reason or they may use something like fish, ksh, or whatever else. Not all of them are compatible with bash, but they all follow POSIX wherever possible, so it's a safe move.

Other than that we try to keep non-plugin-specific utilities conforming to POSIX. The notable exception being `-0` for `xargs` which is a GNU extension.

raffaem:
> 1. Send the hovered file if no selection is given
> 
> 
> is OK. Other one we can skip.

Errata corrige: The current plugin does not work in case multiple devices are paired and reachable. The IDs of all devices are sent to kdeconnect-cli, which consider it as the ID of one device only and don't fine it.

I've temporarily fixed it by sending to the first device in the list.

What should we do in case multiple devices are paired and reachable?

plugins/kdeconnect
raffaem: no it doesn't work. What do you propose?

N-R-K: Just get rid of the notification. I'd rather have a script that works correctly than a flashy one that fails on certain input.

plugins/kdeconnect
```diff
@@ -1,24 +1,49 @@
...
+if [ ! -s "$selection" ]; then
+	if [ -n "$1" ]; then
+		printf "$1" > "$selection"
+	else
+		if [ $notify = 1 ]; then
+			notify-send -a "Kdeconnect" "No selected files, and no hovered file"
+		else
+			echo "[Kdeconnect] No selected files, and no hovered file"
+		fi
+		exit
```
N-R-K, line 32: Pretty sure this exit should be in the outer branch, no?

plugins/kdeconnect
```diff
@@ -1,24 +1,49 @@
...
-        printf "-" > "$NNN_PIPE"
-    fi
+# Build command to share file
+cmd="kdeconnect-cli -d '$id' --share '{}'"
+if [ $notify = 1 ]; then
+	cmd="notify-send -a 'Kdeconnect' 'Sending {}';$cmd;"
 fi
+
+# Send files over
+xargs -0 < "$selection" -I{} sh -c "$cmd"
```
N-R-K, line 43: Try this:

```suggestion
<"$selection" xargs -0 -I{} kdeconnect-cli -d "$id" --share {}
```

Should fix the quoting issue.

N-R-K (changes requested):

plugins/kdeconnect
raffaem: I think it's in the right spot.
If there is no selection but there is an hovered fine, we use the hovered fine.
If there is no selection and there is no hovered file (is that possible?) we exit.
So we don't exit in all cases if there is no selection.

N-R-K: If there was no selection then we shouldn't be executing the `xargs` command below.

875d: @N-R-K I don't understand. If there is no selection, **but** there IS a hovered file, then we execute the xargs below to send the hovered file to the Android device.

If there is no selection, and there is no hovered file, then we don't execute the xargs below: we print an error message and we exit.

That's why the exit at line 40 it's in the "no selection" AND "no hovered file" branch.

Maybe I'm missing something.

You would like the exit on the outer branch, so you would like this:
```
# In case of no selection, use hovered file
if [ ! -s "$selection" ]; then
	if [ -n "$1" ]; then
		printf "$1" > "$selection"
	else
		if [ $notify = 1 ]; then
			notify-send -a "Kdeconnect" "No selected files, and no hovered file"
		else
			echo "[Kdeconnect] No selected files, and no hovered file"
		fi
	fi
	exit
fi
```

But this means that we exit if there is no selection, regardless on whether there is a hovered file to send or not.

Maybe if you could write the code you would like I can understand better :)

raffaem: @N-R-K I don't understand. 

If there is no selection and there is a hovered file, then we execute the xargs below to send the hovered file to the Android device.

If there is no selection and there is no hovered file, then we don't execute the xargs below: we print an error message and we exit.

That's why the exit at line 40 it's in the "no selection" AND "no hovered file" branch.

Maybe I'm missing something.

If I understand you correctly, you would like the exit on the outer branch, so you would like this:

```
# In case of no selection, use hovered file
if [ ! -s "$selection" ]; then
	if [ -n "$1" ]; then
		printf "$1" > "$selection"
	else
		if [ $notify = 1 ]; then
			notify-send -a "Kdeconnect" "No selected files, and no hovered file"
		else
			echo "[Kdeconnect] No selected files, and no hovered file"
		fi
	fi
	exit
fi
```

But this means that we exit if there is no selection, regardless on whether there is a hovered file to send or not.

Then why the `printf "$1" > "$selection"` at all?

Maybe if you could write the code you would like I can understand better :)

N-R-K: > Then why the `printf "$1" > "$selection"` at all?

AH, I didn't even notice that. It's not a good idea to mess with the selection file like that. Just run the kdeconnect command there directly instead.

plugins/kdeconnect
```diff
@@ -1,24 +1,56 @@
...
+# Get file that contains NULL-separated list of selected files
 selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
 
-id=$(kdeconnect-cli -a --id-only | awk '{print $1}')
-if [ -s "$selection" ]; then
-    kdeconnect-cli -d "$id" --share "$(cat "$selection")"
+# In case of no selection, use hovered file
+if [ ! -s "$selection" ]; then
+	if [ -n "$1" ]; then
+		printf "$1" > "$selection"
```
N-R-K, line 33:
```suggestion
		kdeconnect-cli -d "$id" --share "$1"
```

plugins/kdeconnect
raffaem: I'm asking for a temporary file. I would like to mantain the connection code at the end since I would like notifications too

raffaem:
@N-R-K can you check the new implementation?

I also have bash code to manage multiple devices. I don't know if (1) it can be rewritten in POSIX sh (2) you would like to split it into another extension (3) you would like to maintain it like that, commented out

plugins/kdeconnect
KlzXS: If you're using `mktemp` add it as a dependency in `README.md`. Though you could also just make two commands for sending and resume on a common path to get your notifications. I'm inclined to leave it up to you.

KlzXS:
> I don't know if (1) it can be rewritten in POSIX sh

You just opened a deranged can of worms my friend. It _can_ be done, but it's not pretty.
What you are looking for is something like this

```
getname() {
    j=0
    for name in "$@"; do
        printf "%d. %s\n" "$j" "$(eval printf "\${$j})"
        j=$((j+1))
    done

    read -r -p "Choice: " choice
    devname="$(eval printf "\${$choice})"
    printf "%s" "$devname"
}

devname=$(getname $(kdeconnect-cli --list-available --name-only 2>/dev/null))
```

might need slight corrections and I have no idea what the output of `kdeconnect` actually looks like so you might even need to transform that a bit. But there you go, that's how you do arrays POSIX style, by manipulating the arguments "_array_".

N-R-K:
@raffaem 

1. Commented out code is almost always a bad sign, so I like the new `note` function.
2. We can add multi-device support - but let's do that on another PR. For now just focus on selection support.
3. I don't think there's any need for abusing the selection file and don't think there was any need for a tempfile either.

I've pushed some changes. Untested, please let me know if it works fine or not.

If it works, then I think we can merge this (and then figure out multi-device support; which I think can be done without bash).

Also in the future - please use a separate git branch when opening pull requests.

raffaem:
Yes it works.

```
send < "$selection"
```

I don't understand why this line works

```
[ -p "$NNN_PIPE" ] && printf "-" > "$NNN_PIPE" # clear selection
```
Does this work because the interpreter, when it meets the and operator, evaluates the second term only if the first term is true?

Is this guaranteed by POSIX? The evaluation of the second term if and only if the first is true?

So it's a bit of a hack to write it in one line?

```
printf "$1" | send
```

This works because `xargs` read from stdin?

I didn't know you could pipe into a function.

> Also in the future - please use a separate git branch when opening pull requests.

I didn't know the PR would give you write permission on my repo.

Does it give you write permission only on the branch on which the PR is based?

N-R-K:
> I don't understand why this line works
> 
> ```
> [ -p "$NNN_PIPE" ] && printf "-" > "$NNN_PIPE" # clear selection
> ```

It's because of how "logical-operators" work. Look up "short-circuiting", I bet someone one the internet has probably done a better job at explaining it more than I could.

> ```
> printf "$1" | send
> ```
> 
> This works because `xargs` read from stdin?
> 
> I didn't know you could pipe into a function.
Yes. And the `<` operator redirects a processes' stdin.

> Does it give you write permission only on the branch on which the PR is based?

Only the branch that you PR-ed from. (And I'll lose it when this PR is closed).

N-R-K:
Thanks, merged the PR. Let's discuss multi-device support on another PR.

I think you should be able to do with without bash, here's some (completely untested) snippet to get to started @raffaem:

```
names=$(kdeconnect-cli --list-available --name-only 2>/dev/null)
printf "$names" | while IFS=$(printf "\n") read -r name; do
	i=0
	printf "%d. %s\n" "$i" "$name"
	i=$(( i + 1 ))
done
printf "pick a device: "
read -r pick
if [ -n "$pick" ] && [ "$pick" -eq "$pick" ]; then
	# use sed/awk to grab the nth 
fi
```

jarun:
Thank guys!

raffaem:
> Thanks, merged the PR. Let's discuss multi-device support on another PR.
> 
> I think you should be able to do with without bash, here's some (completely untested) snippet to get to started @raffaem:
> 
> ```shell
> names=$(kdeconnect-cli --list-available --name-only 2>/dev/null)
> printf "$names" | while IFS=$(printf "\n") read -r name; do
> 	i=0
> 	printf "%d. %s\n" "$i" "$name"
> 	i=$(( i + 1 ))
> done
> printf "pick a device: "
> read -r pick
> if [ -n "$pick" ] && [ "$pick" -eq "$pick" ]; then
> 	# use sed/awk to grab the nth line from $names
> fi
> ```

isn't `"$pick" -eq "$pick"` always TRUE?

N-R-K:
> isn't `"$pick" -eq "$pick"` always TRUE?

No, because `$pick` might not be an integer. `-eq` tests for numeric equality, so if `$pick` happened to be a non numeric string, then it'll error.

```sh
[~]~> test 5 -eq five
test: integer expression expected: five
```

It's basically an idiomatic way to check weather a variable is an integer or not.

PR #1547, opened 2022-12-20
Miscellaneous improvements to nmount plugin

8B411:
- Prevent `sync` from syncing all devices, instead of syncing just the single one chosen by user;
- Fix interactive authentication request by `udiskctl` when trying to power-off HDD;
- Fix instances when `lsblk` fails to provide mountpoint;
- Check if user-provided device exists and is a block device, otherwise fallback gracefully.
- Make some stylistic changes to the code for overall consistency.


jarun:
Do you think you need to list down the improvements in the PR?

jarun:
Thank you!

PR #1551, opened 2022-12-27
[kdeconnect] Improvements

raffaem:
- Support multiple devices
- Filter out non-regular files

### DIFF
--- a/plugins/kdeconnect
+++ b/plugins/kdeconnect
@@ -8,7 +8,7 @@
 #   - Alternatively, select the files and folders you would like to send, and activate the plugin.
 #
 # Shell: POSIX compliant
-# Author: juacq97, raffaem
+# Author: juacq97, raffaem, N-R-K
 
 # If you want system notification, put this equal to 1
 notify=0
@@ -21,15 +21,38 @@ note() {
 	fi
 }
 
+# kdeconnect doesn't cope with non-files
+filter_files() {
+	xargs -0 -I{} sh -c '
+		if [ -f "{}" ]; then
+			printf "%s\0" "{}";
+		else
+			printf "Error: not a regular file: %s\n" "{}" >&2;
+		fi;'
+}
+
 send() {
-	xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
+	filter_files | xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
 	note "Files sent"
 }
 
-devname=$(kdeconnect-cli --list-available --name-only 2>/dev/null | awk NR==1)
-if [ -z "$devname" ]; then
-	note "No devices available"
+# Select paired device
+names=$(kdeconnect-cli --list-available --name-only 2>/dev/null)
+if [ -z "$names" ]; then
+	note "No devices paired and available"
 	exit
+fi
+
+ndevs=$(printf "%s" "$names" | awk 'END{print NR}')
+if [ "$ndevs" -eq 1 ]; then
+	devname="$names"
+else
+	printf "%s" "$names" | awk '{ print NR ". " $0 }'
+	printf "Pick a device: "
+	read -r pick
+	if [ -n "$pick" ] && [ "$pick" -eq "$pick" ]; then
+		devname=$(printf '%s' "$names" | awk "NR==$pick")
+	fi
 fi
 
 selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}


plugins/kdeconnect
```diff
@@ -26,12 +26,36 @@ send() {
...
+# Select paired device
+names=$(kdeconnect-cli --list-available --name-only 2>/dev/null)
+if [ -z "$names" ]; then
+	note "No devices paired and available"
 	exit
 fi
 
+ndevs=$(printf "%s" "$names" | awk 'END{print NR}')
+if [ "$ndevs" -eq 1 ]; then
+	devname=$(printf "%s" "$names" | awk 'NR==1')
```
N-R-K, line 38: If there's just one name, why call awk on it?

plugins/kdeconnect
```diff
@@ -26,12 +26,36 @@ send() {
...
+else
+	# Append a newline at the end of `names` to read the last device
+	# Note that command substitution removes trailing newlines
+	# see: https://stackoverflow.com/a/15184414/1719931
+	names=$(printf "%s\nx" "$names");
+	names=${names%x}
+	i=0
+	printf "%s" "$names" | while IFS=$(printf "\n") read -r name; do
```
N-R-K, line 46: This loop can probably be replaced by something like this ` printf "%s" "$names" | awk '{ print NR ". " $0 }'`.

plugins/kdeconnect
```diff
@@ -26,12 +26,36 @@ send() {
...
+		printf "%d. %s\n" "$i" "$name"
+		i=$((i+1))
+	done
+	printf "Pick a device: "
+	read -r pick
+	if [ -n "$pick" ] && [ "$pick" -eq "$pick" ]; then
+		pick=$((pick+1))
+		devname=$(printf '%s' "$names" | awk "NR==$pick")
+		printf "You have selected: %s\n" "$devname"
```
N-R-K, line 55: Don't think this is necessary. The user can figure out what he selected by looking at the list.

N-R-K (changes requested):

plugins/kdeconnect
```diff
@@ -22,16 +22,29 @@ note() {
 }
 
 send() {
-	xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
+	xargs -0 -I{} sh -c 'if [ ! -d "{}" ]; then printf "%s\0" "{}"; fi;' |  xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
```
N-R-K, line 25: What's this about?

raffaem: if you pass a directory, kndeconnect creates an empty file on the device with the name of the directory. Basically it doesn't support directory transfers.

N-R-K: As far as I see, this simply filters out directories, right? It doesn't recursively send all the files inside the dir?

If that's the case, then just leave it out. The user can simply not select a directory. And if he does select one, it's better to see the empty file than to silently discard his selection.

raffaem: Yes it just filters out directory.

1. But I think we will need some kind of checks. What happens if you send over a block device?
2. In general, how can I have xargs call a user-defined function? On SO [I found I can export the function](https://stackoverflow.com/a/11003457/1719931), but `shellcheck` complains that is not POSIX sh-compatible.

raffaem: it hangs

N-R-K: In that case, filtering out only the files seems like a good idea. Does this work?

```diff
diff --git a/plugins/kdeconnect b/plugins/kdeconnect
index 0d3ef8b7..209bf137 100755
--- a/plugins/kdeconnect
+++ b/plugins/kdeconnect
@@ -21,8 +21,18 @@ note() {
 	fi
 }
 
+# kdeconnect doesn't cope with non-files
+filter_files() {
+	xargs -0 -I{} sh -c \
+		'if [ -f "{}" ]; then'\
+		'	printf "%s\0" "{}";'\
+		'else'\
+		'	printf "Error: not a file: %s\n" "{}" >&2;'\
+		'fi;'
+}
+
 send() {
-	xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
+	filter_files | xargs -0 -I{} kdeconnect-cli --name "$devname" --share {}
 	note "Files sent"
 }
 
```

raffaem: No it says:
```
Pick a device: 1
	printf "%s\0" "src";: -c: line 2: syntax error: unexpected end of file
	printf "%s\0" "CHANGELOG";: -c: line 2: syntax error: unexpected end of file
	printf "%s\0" "LICENSE";: -c: line 2: syntax error: unexpected end of file
```

but I made it to work by removing the `\` characters :)

raffaem: How can we call:

```
note "Error: not a file:
```

instead of:

```
printf "Error: not a file:
```
?

N-R-K: Don't think that's possible without doing something dirty. So I'd leave it be.

N-R-K (approved):
@jarun This should be good to merge.

jarun:
Thanks guys!

PR #1553, opened 2022-12-31
icons-hash: misc improvements

N-R-K:
* Use a cheaper hash function
* Use a better prng
* Fix compiler warnings


jarun:
Thank you!

PR #1555, opened 2023-01-02
icons-hash: fix bitwise rotation

N-R-K:
in case the rotation is 0, `v >> (32 - r)` would end up doing a 32 right shift which is equal to the width of `v` and thus undefined behavior.

ref: https://blog.regehr.org/archives/1063


jarun:
Thank you!

PR #1566, opened 2023-01-14
General simplification

KlzXS:
`get_output()`:
- The function now uses `UTIL_SH_EXEC` to execute the command.
- The command writes directly to a tmp file where applicable.
- Got rid of parameter `multi` as it's no longer needed.
- A bad use case is now blocked.
- Got rid of `cmd` and `fd` as they are no longer needed. Correct me if I'm wrong, but `fd` wasn't ever used and there is a path to call `free(cmd)` while it's `NULL` (`multi == FALSE` and `pipe()` failing).
- The comment above wasn't accurate, now it is.
- `g_buf` is now clobbered only in some cases. Not that it would matter much, I imagine.

@jarun I still need to look into using `UTIL_SH_EXEC` for running GUI apps as plugins as you suggested, but I thought I should share this regardless.

I only tried the simple case of setting `NNN_HELP` and opening the help screen. I might do more extensive testing later, or if someone wants to give it a shot as well that would be great.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4490,12 +4490,29 @@ static bool get_output(char *file, char *arg1, char *arg2, int fdout, bool page)
 	pid = fork();
 	if (pid == 0) {
 		/* In child */
+		char *bufptr = file;
+
 		close(cmd_in_fd);
 		dup2(cmd_out_fd, STDOUT_FILENO);
 		dup2(cmd_out_fd, STDERR_FILENO);
 		close(cmd_out_fd);
 
-		spawn(utils[UTIL_SH_EXEC], file, arg1, arg2, F_MULTI);
+		if (bufptr && arg1) {
+			char argbuf[CMD_LEN_MAX];
+
+			len = xstrsncpy(argbuf, file, xstrlen(file) + 1);
+			argbuf[len - 1] = ' ';
+			bufptr = argbuf + len;
+			len = xstrsncpy(bufptr, arg1, xstrlen(arg1) + 1);
+			if (arg2) {
+				bufptr[len - 1] = ' ';
+				xstrsncpy(bufptr + len, arg2, xstrlen(arg2) + 1);
+			}
+
+			bufptr = argbuf;
+		}
+
+		spawn(utils[UTIL_SH_EXEC], bufptr, NULL, NULL, F_MULTI);
 		_exit(EXIT_SUCCESS);
 	}
 
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4511,7 +4511,9 @@ static bool get_output(char *file, char *arg1, char *arg2, int fdout, bool page)
 
 		unlink(g_tmpfpath);
 		return TRUE;
-	} else if (have_file)
+	}
+
+	if (have_file)
 		// Case 3
 		return TRUE;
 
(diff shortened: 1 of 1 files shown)


jarun:
I see the following issue with `nnn -i` (to show file info):

Constant message in the info bar: `Usage: file [-bcCdEhikLlNnprsSvzZ0] [--apple] [--extension] [--mime-encoding]`

jarun:
I think all internal commands are failing e.g. listing a zip file results in:

```
atool: no command specified
Try `atool --help' for more information.
```

src/nnn.c
```diff
@@ -4425,92 +4425,106 @@ static void set_smart_ctx(int ctx, char *nextpath, char **path, char *file, char
...
-		dup2(pipefd[1], STDOUT_FILENO);
-		dup2(pipefd[1], STDERR_FILENO);
-		close(pipefd[1]);
-		execvp(*argv, argv);
+		close(cmd_in_fd);
+		dup2(cmd_out_fd, STDOUT_FILENO);
+		dup2(cmd_out_fd, STDERR_FILENO);
+		close(cmd_out_fd);
+
+		spawn(utils[UTIL_SH_EXEC], file, arg1, arg2, F_MULTI);
```
jarun, line 4498: Maybe we are seeing the missing argument issue because `sh` expects a single argument string... like:

    sh "file arg1 arg2"
But concatenating would require an internal temporary buffer.

jarun (changes requested):

jarun (approved):

jarun:
Thanks @KlzXS!

jarun:
@KlzXS I reverted this change and the changes I made on top of it. It appears concatenating the arguments don't work when the filename has spaces in it due to the usual shell quirks.

jarun:
Please re-raise if you find a way around it.

KlzXS:
I guess it's best to forego using `UTIL_SH_EXEC` and go the good old way it was while using the rest of the simplification.

I also had thoughts of implementing a new variant of `spawn()` that would allow us to pass in file descriptors to be mapped to stdin, stdout and stderr. I think a few places in the code might benefit from that. Say we pipe the output directly into `nnn` instead of writing to a file and reading from it. What do you think?

jarun:
> I guess it's best to forego using UTIL_SH_EXEC and go the good old way it was while using the rest of the simplification.

I agree!

> I also had thoughts of implementing a new variant of spawn() that would...

We had problems and breaks with such experiments earlier. I think spawn() is good for now.

KlzXS:
> I think spawn() is good for now.

Sure. I'll do it the old way. I might submit my experiment at a later date. Might.

jarun:
👍

N-R-K:
> I also had thoughts of implementing a new variant of `spawn()` that would allow us to pass in file descriptors to be mapped to stdin, stdout and stderr. I think a few places in the code might benefit from that. Say we pipe the output directly into `nnn` instead of writing to a file and reading from it. What do you think?

@KlzXS You may want to take a look at [`posix_spawn`](https://linux.die.net/man/3/posix_spawn) - it's much faster than `fork + dup + exec` and I think it should be capable of doing what you want. You can take a look at nsxiv for some reference (the error handling is quite ugly, but otherwise it works quite well): [util.c#L251-L294](https://codeberg.org/nsxiv/nsxiv/src/commit/75849adb88187ab93ab2224706693676a3ea0fb3/util.c#L251-L294)

PR #1570, opened 2023-01-25
Second attempt at #1566

KlzXS:
`get_output()` no longer uses `UTIL_SH_EXEC` since that presented the usual shell expansion problems. Now it directly invoked the command.
Tried it out with the examples of failing commands in #1566 and those seem fine now.

BTW sorry for the terrible commit message. Didn't realize it wouldn't allow me to merge all the commits again. Besides a bunch of "Revert 'Revert ...'" is not that pretty.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -837,7 +837,7 @@ static void move_cursor(int target, int ignore_scrolloff);
 static char *load_input(int fd, const char *path);
 static int set_sort_flags(int r);
 static void statusbar(char *path);
-static bool get_output(char *file, char *arg1, char *arg2, int fdout, bool multi, bool page);
+static bool get_output(char *file, char *arg1, char *arg2, int fdout, bool page);
 #ifndef NOFIFO
 static void notify_fifo(bool force);
 #endif
@@ -3188,7 +3188,7 @@ static void showfilterinfo(void)
 
 	i = getorderstr(info);
 
-	if (cfg.fileinfo && ndents && get_output("file", "-b", pdents[cur].name, -1, FALSE, FALSE))
+	if (cfg.fileinfo && ndents && get_output("file", "-b", pdents[cur].name, -1, FALSE))
 		mvaddstr(xlines - 2, 2, g_buf);
 	else {
 		snprintf(info + i, REGEX_MAX - i - 1, "  %s [/], %4s [:]",
@@ -4550,7 +4550,7 @@ static bool show_stats(char *fpath)
 		return FALSE;
 
 	while (r)
-		get_output(cmds[--r], fpath, NULL, fd, TRUE, FALSE);
+		get_output(cmds[--r], fpath, NULL, fd, FALSE);
 
 	close(fd);
 
@@ -4697,7 +4697,7 @@ static bool handle_archive(char *fpath /* in-out param */, char op)
 	if (op == 'x') /* extract */
 		spawn(util, arg, fpath, NULL, F_NORMAL | F_MULTI);
 	else /* list */
-		get_output(util, arg, fpath, -1, TRUE, TRUE);
+		get_output(util, arg, fpath, -1, TRUE);
 
 	if (x_to) {
 		if (chdir(xdirname(fpath)) == -1) {
@@ -5097,7 +5097,7 @@ static void show_help(const char *path)
 
 	char *prog = xgetenv(env_cfg[NNN_HELP], NULL);
 	if (prog)
-		get_output(prog, NULL, NULL, fd, TRUE, FALSE);
+		get_output(prog, NULL, NULL, fd, FALSE);
 
 	start = end = helpstr;
 	while (*end) {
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -4425,92 +4425,106 @@ static void set_smart_ctx(int ctx, char *nextpath, char **path, char *file, char
...
-	argv[++index] = arg2;
+		cmd_in_fd = STDIN_FILENO;
+		cmd_out_fd = fdout;
+	} else if (have_file) {
+		// Case 3
+		cmd_in_fd = STDIN_FILENO;
+		cmd_out_fd = fdout;
+	} else {
+		// Case 1
+		if (pipe(pipefd) == -1)
```
N-R-K, line 4472: `pipe2` would've been nice. Unfortunately it's not POSIX. So :shrug:

src/nnn.c
```diff
@@ -4425,92 +4425,106 @@ static void set_smart_ctx(int ctx, char *nextpath, char **path, char *file, char
...
 
-	if (tmpfile) {
+	/*
+	 * In this case the logic of the function dictates that we should write the output of the command
+	 * to `fd` and show it in the pager. But since we didn't open the file descriptor we have no right
+	 * to close it, the caller must do it. We don't even know the path to pass to the pager and
+	 * it's a real hassle to get it. In general this just invites problems so we are blocking it.
+	 */
+	if (have_file && page)
+		return FALSE;
```
N-R-K, line 4455: Is there any actual cases of this function being called like this? I'd personally use `assert` but `nnn` codebase doesn't seem to use it. A debug log would be nice I think.

KlzXS: No, there are currently no code paths that do this, but I thought I'd still prevent it from going through. An `assert` is a bit aggressive. We don't need to crash the whole program, just prevent shenanigans from happening. I'll add a debug log, good idea.

N-R-K: > An `assert` is a bit aggressive. We don't need to crash the whole program, just prevent shenanigans from happening.

The idea is that if you compile with `-N DEBUG` then the "release build" won't suffer from it and in "debug builds" it will get caught easily and instantly (instead of potentially silently ignoring the bogus argument and having the program march on forward as if everything is ok).

But a debug log is fine by me too, especially since `nnn` doesn't use any assert currently.

jarun:
I see the following problem with this:

I have an entry in `NNN_PLUG` as: `c:!|cp -rv "$nnn" "$nnn".cp*`

I have a file: `IEEE. SpectrumCollections.1.HandsOn.072221.pdf`

When I press `Alt-c` on the above file I get the following error:

    cp: cannot stat '"$nnn"': No such file or directory

I have also modified `run_cmd_as_plugin()` as:

```lang=C
static void run_cmd_as_plugin(const char *file, char *runfile, uchar_t flags)
{
        size_t len; 

        xstrsncpy(g_buf, file, PATH_MAX);

        len = xstrlen(g_buf);
        if (len > 1 && g_buf[len - 1] == '*') {
                flags &= ~F_CONFIRM; /* Skip user confirmation */
                g_buf[len - 1] = '\0'; /* Get rid of trailing no confirmation symbol */
                --len;
        }

        if (flags & F_PAGE) {
                get_output(g_buf, NULL, NULL, -1, TRUE);                                                              
#if 0
        if ((flags & F_PAGE) || (flags & F_NOTRACE)) {
                if (is_suffix(g_buf, " $nnn"))
                        g_buf[len - 5] = '\0';
                else
                        runfile = NULL;

                if (flags & F_PAGE)
                        get_output(g_buf, runfile, NULL, -1, TRUE);
                else // F_NOTRACE
                        spawn(g_buf, runfile, NULL, NULL, flags);
#endif
        } else 
                spawn(utils[UTIL_SH_EXEC], g_buf, NULL, NULL, flags);
}
```

KlzXS:
You used `|`, it's going the `F_PAGE` path, but you didn't set it to use `UTIL_SH_EXEC`. The env variable is not getting expanded.

jarun:
I don't get it. You mentioned:

> No longer uses UTIL_SH_EXEC since that presented the usual shell expansion problems.

Can you share the steps you meant? My `UTIL_SH_EXEC` is `sh`.

jarun:
In this case I wanted to page the output as well as complete the copy operation.

jarun:
OK. I changed:

    get_output(g_buf, NULL, NULL, -1, TRUE);
to
    get_output(utils[UTIL_SH_EXEC], g_buf, NULL, -1, TRUE);

and it worked. Is this what you meant?

jarun:
Let me use this for some more time and if there are no issues, I'll merge.

KlzXS:
> Is this what you meant?

That's exactly it. I updated the main comment to better reflect what the change compared to the previous PR is.

jarun:
Thank you!

jarun:
Thanks @KlzXS! This is a great improvement.
Not only `$nnn`, run commands as plugin now supports all exported variables. Something I wanted for very long...

PR #1571, opened 2023-01-25
Simplify the command as plugin codepath

KlzXS:
Invoking commands as plugins now uses `UTIL_SH_EXEC` which should allow for lifting the restrictions about the placement of `$nnn` it some commands. The new behavior also requires the user to quote `$nnn` so that the shell can properly expand it, since we no longer pass it as a separate argument to `exec()`. An exception is made is ` $nnn` is detected to be a suffix. In that case the old behavior is preserved as to not break users' configs. We can slate this for removal at some point if you wish, but for now I think it's best to keep it.

As always, my testing is rather light, so you should probably try it out for yourself before merging. Have you considered developing an environment for automatic testing of regressions or is that just not worth it at this point?

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -119,7 +119,7 @@ If the plugins list gets too long, try breaking them up into sections:
 ```
 NNN_PLUG_PERSONAL='g:personal/convert2zoom;p:personal/echo'
 NNN_PLUG_WORK='j:work/prettyjson;d:work/foobar'
-NNN_PLUG_INLINE='e:!go run $nnn*'
+NNN_PLUG_INLINE='e:!go run "$nnn"*'
 NNN_PLUG_DEFAULT='1:ipinfo;p:preview-tui;o:fzz;b:nbak'
 NNN_PLUG="$NNN_PLUG_PERSONAL;$NNN_PLUG_WORK;$NNN_PLUG_DEFAULT;$NNN_PLUG_INLINE"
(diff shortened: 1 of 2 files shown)


KlzXS:
The CI failing is the result of my poor git branch management in this case. If the other PR is merged it should go away on its own. I made the PR under the assumption that the other changes were already merged.

plugins/README.md
```diff
@@ -154,56 +154,52 @@ Now <kbd>;x</kbd> can be used to make a file executable, <kbd>;g</kbd> can be us
...
+1. Use single quotes for `$NNN_PLUG` so `"$nnn"` is not interpreted
+2. (_Again_) add `!` before the command
+3. To disable directory refresh after running a _command as plugin_, prefix with `-!`
 
 #### Some useful key-command examples
 
 | Key:Command | Description |
 |---|---|
-| `c:!convert $nnn png:- \| xclip -sel clipboard -t image/png*` | Copy image to clipboard |
+| `c:!convert "$nnn" png:- \| xclip -sel clipboard -t image/png*` | Copy image to clipboard |
```
jarun, line 192: Please update the man page as well.

jarun (changes requested):

jarun:
Thank you!

plugins/README.md
jarun: I'll take care of this.

Issue #1572, opened 2023-01-25
chksum incorrect return


875d:
When trying to verify a `md5` file with `chksum`:

```
/home/[USERNAME]/.config/nnn/plugins/chksum: line 64: return: can only `return' from a function or sourced script
```

We probably need `exit` instead

N-R-K:
What's your `sh` shell? I'm using `dash` and it seems to work fine. Not saying we shouldn't change to `exit`, but I'm just curious.

875d:
```
❯ /usr/bin/sh --version
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
```

N-R-K:
Hmm, yeah. I can reproduce it on bash:

```
[nnn nrk]~> bash --version
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
```

875d:
> Hmm, yeah. I can reproduce it on bash:
> 
> ```
> [nnn nrk]~> bash --version
> GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
> ```

You know POSIX sh.

So either fix it or report to GNU bash.

N-R-K:
> You know POSIX sh.

I don't. Only bits and pieces, so I don't know if it's a valid shell code or not.

KlzXS:
From what I can see `return` is part of POSIX. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#return

To me `checksum_type()` definitely looks like a function. I don't know why it would complain about this.

N-R-K:
That return isn't part of `checksum_type()`.

875d:
> checksum_type()



> From what I can see `return` is part of POSIX. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#return
> 
> To me `checksum_type()` definitely looks like a function. I don't know why it would complain about this.

The `return` it complains about is not inside `checksum_type()`

875d:
@N-R-K don't you have write access?

KlzXS:
Oh, my bad. Then that's why. It must be inside of a function or "dot script".

N-R-K:
Jarun typically does the merging. I don't know what the policy is towards collaborators directly pushing into `master` but I'd rather not do that unless I've been given explicit permissions.

KlzXS:
[...]

PR #1573, opened 2023-01-25
plugins/chksum: use exit instead of return

N-R-K:
From https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_24:

| The return utility shall cause the shell to stop executing the current
| function or dot script. If the shell is not currently executing a
| function or dot script, the results are unspecified.

Closes: https://github.com/jarun/nnn/issues/1572


jarun:
Thank you!

PR #1576, opened 2023-01-29
preview-tui: add chafa and PREFERRED_IMAGE_VIEWER env variable

greenfoo:
This PR does two things:
* Adds "chafa" as another option to display images in the terminal (in addition to the already present "ueberzug", "catimg" and "viu"), which seems to generate better results.
* Adds a new environment variable (PREFERRED_IMAGE_VIEWER) that can be used to select which of these image viewers to use when more than one are installed (before, the first one found using a predefined order was always used)

Please feel free to discard this PR if you don't like the idea of further increasing the complexity of these user scripts. Thanks! : )

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -20,7 +20,7 @@
 #   - man
 #   - optional: bsdtar or atool for additional archive preview
 #   - optional: bat for code syntax highlighting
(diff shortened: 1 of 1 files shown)


luukvbaal (changes requested):
I'm fine with adding this

luukvbaal (approved):

Issue #1069, opened 2021-06-12
Preview-tui not showing directories and files correctly


etrigan63:
#### Environment details (Put `x` in the checkbox along with the information)

- [X ] Operating System: ArcoLinux
- [ X Desktop Environment: bspwm
- [X ] Terminal Emulator: kitty
- [X ] Shell: zsh
- [ ] Custom desktop opener (if applicable):
- [ ] Program options used:
- [ X] Configuration options set: 
export NNN_PLUG='[:preview-tui;'
export NNN_OPTS="eaEox"

- [ ] Issue exists on `nnn` master

#### Exact steps to reproduce the issue
Open preview-tui using Alt-[ or ;[ while hovering over a file or directory
Files look like this:
![image](https://user-images.githubusercontent.com/7347264/121765314-ec0ea080-cb17-11eb-8479-3c65d4b71475.png)
Directories look like this:
![image](https://user-images.githubusercontent.com/7347264/121765324-0779ab80-cb18-11eb-8c2a-d9e526dac566.png)
Font is Fira Code Nerd Font.
Images work correctly:
![image](https://user-images.githubusercontent.com/7347264/121765346-3abc3a80-cb18-11eb-83a8-0dac1c0b95d4.png)

jarun:
PAGER is not interpreting ASCII color codes. Here's what I have:

```
$ echo $PAGER
less -Ri
```

Look up the options in the man page.

larrybotha:
For anyone else experiencing this, and to add some keywords for issue search (`ESC`, `ESC[`), my `$PAGER` was just set to `less`.

I'm not sure exactly where the `$PAGER` variable is set, but setting it to `less -R`, as @jarun mentioned resolves the rendering due to escape ANSI color escape sequences:

```
#!/usr/bin/env sh

# custom nnn configs, e.g. in ~/.config/nnn/my-custom-config.sh

# fix preview-tui not rendering correctly
export PAGER="less -R"

# other NNN env vars
```

PR #1580, opened 2023-02-08
Preview-tui!: refactor environment variables

luukvbaal:
Problem:    List of environment variables is copied for each previewer.
            Environment variable names are too general.
Solution:   Parameterize environment variable flags with static list of
            variables. Prepend user-facing variables with `NNN_`.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -11,8 +11,8 @@
 #       - wezterm (https://wezfurlong.org/wezterm), or
 #       - QuickLook on WSL (https://github.com/QL-Win/QuickLook), or
 #       - Windows Terminal (https://github.com/Microsoft/Terminal | https://aka.ms/terminal) with WSL, or
-#       - $TERMINAL set to a terminal (it's xterm by default).
-#   - less or $PAGER
+#       - $NNN_TERMINAL set to a terminal (it's xterm by default).
+#   - less or $NNN_PAGER
 #   - tree or exa or ls
 #   - mediainfo or file
 #   - mktemp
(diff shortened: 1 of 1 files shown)


plugins/preview-tui
```diff
@@ -11,8 +11,8 @@
 #       - wezterm (https://wezfurlong.org/wezterm), or
 #       - QuickLook on WSL (https://github.com/QL-Win/QuickLook), or
 #       - Windows Terminal (https://github.com/Microsoft/Terminal | https://aka.ms/terminal) with WSL, or
-#       - $TERMINAL set to a terminal (it's xterm by default).
-#   - less or $PAGER
+#       - $NNN_TERMINAL set to a terminal (it's xterm by default).
+#   - less or $NNN_PAGER
```
jarun, line 15: What is the problem with using the default pager?

jarun (changes requested):

plugins/preview-tui
jarun: The problem is - it's difficult for users to decide which one to setup which one not to.

Maybe we can list and document all of them in one place.

jarun (changes requested):

plugins/preview-tui
luukvbaal: I think there is merit to not using the default shell `$PAGER` to avoid issue like this: https://github.com/jarun/nnn/issues/1069.
This way we enforce a value that works by default.

jarun: I think we have the PAGER thing documented somewhere in the wiki. Even if that's not the case, users have to know that it should be `less -Ri`. Changing the variable name does not help.

luukvbaal: It helps in the sense that new users that have defined `$PAGER` in their shell profile will no longer run into problems...

luukvbaal: To reiterate: I think it's bad practice to blindly import default shell variables that we know can cause issues with `preview-tui`.

Why not set default values that we know works but still allow modifying them through shell variables, prefixed with `_NNN`. That why I made that change but if you disagree I can omit that. The main point of this PR was refactoring/de-duplicating the list of shell variables to be passed to the previewer.

jarun: Makes sense.

jarun:
LGTM. @N-R-K please have a look.

N-R-K:
Seems to work OK, tested with xterm.

jarun:
Thanks guys!

PR #1585, opened 2023-02-17
Preview-tui hide wezterm split-pane output

luukvbaal:
Avoids wezterm's split-pane command from writing it's output over nnn (note the `18`):
image


jarun:
Thank you!

PR #1586, opened 2023-02-17
Preview-tui restore current working directory

luukvbaal:
This could result in writing the wrong selection to `$CURSEL` here:
https://github.com/jarun/nnn/blob/ade477e705630660baf9ee2345001233047741b5/plugins/preview-tui#L469

I think @michaelzusev described a problem as a result from this on gitter(which I checked out after you mentioned it on the dev-list @jarun, I hadn't before).


jarun:
Thank you!

Issue #1584, opened 2023-02-16
Colemak users should compile with colemak patch, not use stale Gitlab fork


jacmoe:
#### Environment details (Put `x` in the checkbox along with the information)

- [x ] Keyboard Variant: Colemak
- [x ] Issue exists on `nnn` master

#### Exact steps to reproduce the issue
1. Discover *nnn*
2. Navigate to Github repository
3. Read the Wiki, and see that Colemak users should use Queer Bird's `nnn-colemak` on Gitlab

The problem, of course, is that we should instead compile *nnn* from source, with the `colemak` patch. Much better!

I guess not every colemak using *nnn* convert is as observant as I was, so perhaps the wiki could be fixed with a note that colemak users might want to compile from source with the colemak patch.

Thanks for *nnn*! :heart: 

EDIT: I would normally just have issued a PR, but I am not sure what to edit.
Maybe saying something like we should do a `make O_COLEMAK=1`

jacmoe:
That sounds great, thanks!
I applied the patch and I find it very useful, but it would of course be better if I didn't have to reach down for M because H and M is swapped. That's truly minor!
I would be even happier if @luukvbaal would be so kind to update the patch for people who don't use colemak-dh.

luukvbaal:
I really would rather someone invested in the patch would step up to submitting/maintaining a patch they're interested in, that was the idea of the user-patch framework. I can't speak to the quality of these keymap patches(the current colemak patch was not written by me nor do I use it, I just added it to the framework).

But yeah as I understand it there are two colemak variants. I guess since we already have a keymap patch in the framework others are welcome too. 

> EDIT: I would normally just have issued a PR, but I am not sure what to edit.
Maybe saying something like we should do a make O_COLEMAK=1

I think in this case a PR would indeed replace the current colemak patch with a (regular?) colemak patch and move the current patch to `patches/colemak-dh` and add a `O_COLEMAK-DH` make variable.

jacmoe:
OK, here's how I suggest that this is done:

1. Someone else renames the patch and update the wiki/makefile with new O_COLEMAK_DH flag
2. I learn how to export a diff.
3. I experiment with a good colemak patch.
4. After having used `nnn` for a month with the patch applied, and the kinks ironed out, I will add the new O_COLEMAK patch.

How does that sound?

KlzXS:
@luukvbaal Oh, I thought you were invested in it since you recently updated it. My bad, I didn't look into the origin of that patch.

[...]

PR #1587, opened 2023-02-17
Added patch for colemak keyboard layout. Existing renamed to colemak-dh

jacmoe:
A PR as discussed in #1584

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ O_NOSORT := 0  # disable sorting entries on dir load
 
 # User patches
 O_COLEMAK := 0 # change key bindings to colemak compatible layout
+O_COLEMAK-DH := 0 # change key bindings to colemak-dh compatible layout
 O_GITSTATUS := 0 # add git status to detail view
 O_NAMEFIRST := 0 # print file name first, add uid and guid to detail view
 O_RESTOREPREVIEW := 0 # add preview pipe to close and restore preview pane
@@ -168,6 +169,7 @@ LOGOSVG = misc/logo/logo.svg
 LOGO64X64 = misc/logo/logo-64x64.png
 
 COLEMAK = patches/colemak
+COLEMAK-DH = patches/colemak-dh
 GITSTATUS = patches/gitstatus
 NAMEFIRST = patches/namefirst
 RESTOREPREVIEW = patches/restorepreview
@@ -339,6 +341,9 @@ ifeq ($(strip $(O_RESTOREPREVIEW)),1)
 endif
 ifeq ($(strip $(O_COLEMAK)),1)
 	patch --forward $(PATCH_OPTS) --strip=1 --input=$(COLEMAK)/mainline.diff
+endif
+ifeq ($(strip $(O_COLEMAK-DH)),1)
+	patch --forward $(PATCH_OPTS) --strip=1 --input=$(COLEMAK-DH)/mainline.diff
 endif
 
 postpatch:
--- a/Makefile
+++ b/Makefile
@@ -361,6 +361,9 @@ endif
 ifeq ($(strip $(O_COLEMAK)),1)
 	patch --reverse $(PATCH_OPTS) --strip=1 --input=$(COLEMAK)/mainline.diff
 endif
+ifeq ($(strip $(O_COLEMAK-DH)),1)
+	patch --reverse $(PATCH_OPTS) --strip=1 --input=$(COLEMAK-DH)/mainline.diff
+endif
 
 skip: ;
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -340,6 +342,9 @@ endif
 ifeq ($(strip $(O_COLEMAK)),1)
 	patch --forward $(PATCH_OPTS) --strip=1 --input=$(COLEMAK)/mainline.diff
 endif
+ifeq ($(strip $(O_COLEMAK-DH)),1)
+	patch --forward $(PATCH_OPTS) --strip=1 --input=$(COLEMAK-DH)/mainline.diff
```
luukvbaal, line 346: Please also add the reverse patch in the Makefile below. We do this to keep the repository in a clean state:
```diff
+++ b/Makefile
@@ -361,6 +361,9 @@ endif
 ifeq ($(strip $(O_COLEMAK)),1)
        patch --reverse $(PATCH_OPTS) --strip=1 --input=$(COLEMAK)/mainline.diff
 endif
+ifeq ($(strip $(O_COLEMAK-DH)),1)
+       patch --reverse $(PATCH_OPTS) --strip=1 --input=$(COLEMAK-DH)/mainline.diff
+endif

 skip: ;
```

We normally also add a new patch to the array in the `patches/checkpatches.sh` script but in this case I think we are good with just a single colemak patch in there.

luukvbaal (changes requested):

jarun:
@luukvbaal please merge when this is ready.

Makefile
jacmoe: Will do!

jacmoe: Done!
But, there is a slight issue: O_COLEMAK and O_COLEMAK-DH are mutually exclusive. Every test in the check_patches script fails when it tries to apply both of those.
Since I am not familiar with the patch system, I can't do much about that.

jacmoe:
Updated the PR with suggested changes, but `check_patches` fails whenever it tries to apply both `O_COLEMAK` and `O_COLEMAK-DH`. Those two are mutually exclusive.

N-R-K:
> Updated the PR with suggested changes, but `check_patches` fails whenever it tries to apply both `O_COLEMAK` and `O_COLEMAK-DH`. Those two are mutually exclusive.

Prefer `O_COLEMAK` over `O_COLEMAK-DH` when both are set. You can do something like this:

https://github.com/jarun/nnn/blob/b49cc28d80bc2236c33954b625245d9fbb180279/Makefile#L69-L79

- - -

But honestly, I feel like the current patching system is not very scalable. It used to work fine when there was only 2~3 patches, but every new patch will end up increasing the number of possible combination by an exponential factor (2<sup>n</sup>).

Though I don't exactly have a better solution either.

N-R-K:
Also the difference between colemak and colemak-dh seems to be really low (looks like just 3~4 changes).

So I'd rather just rm `colemak-dh` if @jarun agrees.

jacmoe:
I tried to remedy the Makefile, but it errors out. Individually, both O_COLEMAK and O_COLEMAK-DH applies succesfully, but not together.
Would be better if both could coexist because I suspect there's quite a few colemak-dh users among the potential audience of `nnn`. I am fine with ordinary colemak because I don't have an ortholinear / split keyboard.

luukvbaal:
> Also the difference between colemak and colemak-dh seems to be really low (looks like just 3~4 changes).

> So I'd rather just rm colemak-dh if @jarun agrees.

I'm not familiar with these keyboard layouts to judge whether they both warrant their own keyboard layouts for nnn. If they differ enough where it matters, I would be fine with providing a patch for both, given that we allow keymap patches...

> Updated the PR with suggested changes, but check_patches fails whenever it tries to apply both O_COLEMAK and O_COLEMAK-DH. Those two are mutually exclusive.

I suggested to omit adding an additional entry to the checkpatches array.

> But honestly, I feel like the current patching system is not very scalable. It used to work fine when there was only 2~3 patches, but every new patch will end up increasing the number of possible combination by an exponential factor (2n).

The patch system itself is scalable enough I think, but indeed the checkpatches script isn't really. However it it just to catch merge conflicts. So like I said I don't think we need multiple entries for keymap patches that edit the same lines.

jacmoe:
That makes sense!
If the `COLEMAK-DH` is removed from the check script, all is fine. Yes, it will fail at compile time if both are set, but at least it's a logical thing to do.

The difference between dh and standard colemak is significant enough to warrant a patch for each.

N-R-K (approved):
Looks OK, I think. Though, I would've preferred to just rm the `dh` variant.

PR #1588, opened 2023-02-17
nmv: prefer selection if -u is active

N-R-K:
the rename plugin always asks for "selection vs current" even when -u flag is active. pass it to the plugin via `NNN_PREFER_SELECTION` so that there's less distracting prompts.

### DIFF
--- a/plugins/.nmv
+++ b/plugins/.nmv
@@ -17,10 +17,12 @@
 # Shell: bash
 # Author: KlzXS
 
+# shellcheck disable=SC1090,SC1091
+. "$(dirname "$0")"/.nnn-plugin-helper
+
 EDITOR="${EDITOR:-vi}"
 TMPDIR="${TMPDIR:-/tmp}"
 NNN_INCLUDE_HIDDEN="${NNN_INCLUDE_HIDDEN:-0}"
-NNN_PREFER_SELECTION="${NNN_PREFER_SELECTION:-0}"
 VERBOSE="${VERBOSE:-0}"
 RECURSIVE="${RECURSIVE:-0}"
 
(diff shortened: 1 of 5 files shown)


N-R-K:
This also looks like something a lot of other plugin can benefit from as well. Should we export it for all plugins ?

jarun:
Yes, please export for all plugins.

plugins/.nmv
```diff
@@ -38,12 +39,17 @@ exit_status=0
 dst_file=$(mktemp "$TMPDIR/.nnnXXXXXX")
 
 if [ -s "$selection" ]; then
-	printf "Rename 'c'urrent / 's'election? "
-	read -r resp
+	printf "%s\n" "$NNN_PREFER_SELECTION" >&2
```
jarun, line 42: Why do we need this print?
Also, can we make this a common function in `.nnn-plugin-helper`?

jarun (changes requested):

plugins/.nmv
N-R-K: > Why do we need this print?

Was a debug print. I'll remove. :upside_down_face: 

> Also, can we make this a common function in `.nnn-plugin-helper`?

I'll take a look, and see what other plugins could use this.

jarun:
Thank you!

jarun:
Please check the build break:

```

In plugins/.nmv line 43:
	arr=$(tr '\0' '\n' < "$selection")
                              ^--------^ SC2154: selection is referenced but not assigned.


In plugins/.nnn-plugin-helper line 51:
        [ -n "$1" ] && printf "$1 "
                              ^---^ SC2059: Don't use variables in the printf format string. Use printf "..%s.." "$foo".


In plugins/gpge line 24:
		files=$(tr '\0' '\n' < "$selection")
                                        ^--------^ SC2154: selection is referenced but not assigned.

For more information:
  https://www.shellcheck.net/wiki/SC2154 -- selection is referenced but not a...
  https://www.shellcheck.net/wiki/SC2059 -- Don't use variables in the printf...

```

PR #1589, opened 2023-02-17
fix shellcheck issues

N-R-K:


jarun:
Thank you!

Is shellcheck not running on PRs? The earlier PR passed.

KlzXS:
I think it sometimes asks for approval. I don't know what that's about

jarun:
okk

N-R-K:
> I think it sometimes asks for approval. I don't know what that's about

I think that's only for "first time contributors" ? I don't think it matters for us since we don't have any "secret" in our CI.

Issue #1583, opened 2023-02-15
[preview-tui]: some pager pid error


kuntau:
#### Environment details (Put `x` in the checkbox along with the information)

- [x] Operating System: MacOS
- [ ] Desktop Environment:
- [x] Terminal Emulator: Wezterm
- [x] Shell: zsh
- [ ] Custom desktop opener (if applicable):
- [ ] Program options used:
- [ ] Configuration options set:
- [x] Plugins are installed
- [ ] Issue exists on `nnn` master

#### Exact steps to reproduce the issue

I always got this error but not sure what the root cause.

```sh
preview-tui: line 305: 81359 Terminated: 15          $NNN_PAGER < "$FIFOPATH"
```

I guess it's had something to do with `pid` that was killed and this error should be redirected to `/dev/null`

The previewer works fine except those error appear on top of preview window

luukvbaal:
Haven't been able to reproduce this, I did notice another issue with wezterm that should be fixed by #1585.

Could you elaborate how/when you see the error?

kuntau:
<img width="732" alt="Screenshot 2023-02-16 at 2 34 00 AM" src="https://user-images.githubusercontent.com/327489/219710828-4d56f340-d74b-4d7d-9db0-31f7ca6f023f.png">

> Could you elaborate how/when you see the error?

Nothing special. Just open `preview-tui` and browser around. What I've noticed the error didn't appear when the pager is `bat` or images

kuntau:
https://user-images.githubusercontent.com/327489/219715689-a336503f-5597-4aef-9dd2-26208f8c9800.mov

luukvbaal:
Not sure what could cause this. I do wonder why the error states `line 305...`. Are you using the latest `preview-tui`? Line 305 is a function definition on master.

kuntau:
> Not sure what could cause this. I do wonder why the error states `line 305...`. Are you using the latest `preview-tui`? Line 305 is a function definition on master.

Yes, latest `preview-tui`. Anyway, I've found the offending line. Can you verify @luukvbaal ?

PR #1591, opened 2023-02-19
fix(preview-tui): fix message leakage cause by kill signal

kuntau:
Honestly, I'm not sure what this `printf` does, but it fixes https://github.com/jarun/nnn/issues/1583

So far, I haven't seen any different behaviour other than no more that flashing error on preview pane.

I have test on following terminal

1. ✅ Wezterm
2. ✅ Kitty
3. ✅ iTerm2
4. ✅ Tmux

All work fine.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -148,7 +148,7 @@ done; unset IFS
 
 exists() { type "$1" >/dev/null 2>&1 ;}
 pkill() { command pkill "$@" >/dev/null 2>&1 ;}
-pidkill() { [ -f "$1" ] && kill "$(cat "$1" 2>/dev/null)" >/dev/null 2>&1 ;}
+pidkill() { [ -f "$1" ] && kill -2 "$(cat "$1" 2>/dev/null)" >/dev/null 2>&1 ;}
 prompt() { printf "%b" "$@"; cfg=$(stty -g); stty raw -echo; head -c 1; stty "$cfg" ;}
 
 start_preview() {
@@ -215,6 +215,7 @@ fifo_pager() {
     mkfifo "$FIFOPATH" || return
 
     $NNN_PAGER < "$FIFOPATH" &
+    printf "%s" "$!" > "$PREVIEWPID"
 
     (
         exec > "$FIFOPATH"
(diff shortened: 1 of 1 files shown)


luukvbaal:
It is not unnecessary, removing the line will result in the pager not being killed:
image

Can you check if redirecting stderr will hide the message?
```diff
+++ b/plugins/preview-tui
@@ -215,7 +215,7 @@ fifo_pager() {
     mkfifo "$FIFOPATH" || return

     $NNN_PAGER < "$FIFOPATH" &
-    printf "%s" "$!" > "$PREVIEWPID"
+    printf "%s" "$!" 2>/dev/null > "$PREVIEWPID"

     (
         exec > "$FIFOPATH"
```

kuntau:
> Can you check if redirecting stderr will hide the message?

Nope. Same error appear.

kuntau:
This should properly fix the problem. Can you verify @luukvbaal

luukvbaal:
I still haven't been able to reproduce the message from showing up so I can't really verify whether it resolves the issue. 

But no, `SIGINT` is not enough to terminate the pager on my system:
```
pgrep less
22612
22626
22640
22656
22671
22686
22700
22715
22730
22747
22761
22775
22789
22803
22818
22835
22849
22864
22879
22893
```
`preview-tui`'s FIFO loop survives the signal as well so toggling the plugin through <kbd>;p</kbd> stops working with this change.

Knowing where the message originates from helps though, perhaps you can figure out something else?

kuntau:
What system are you on? Another method is to redirect all `stderr` with `exec 2> /dev/null`.

I have test that, but it _might_ also hide other error message. What do you think?

luukvbaal:
I am on linux. I would rather avoid redirecting all stderr.

Yes, pagers are killed on current `preview-tui` on my system.

What shell does `/usr/bin/env sh` link to for you?

kuntau:
> What shell does `/usr/bin/env sh` link to for you?

```
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)
```

> I would rather avoid redirecting all stderr.

Here's a couple of solutions I found on SO [Q1](https://stackoverflow.com/questions/81520/how-to-suppress-terminated-message-after-killing-in-bash) [Q2](https://stackoverflow.com/questions/5719030/bash-silently-kill-background-function-process)

I can't get `wait` to work. It broke more things than it fixes. Haven't tried `disown` though.

luukvbaal:
@kuntau please try #1593. Still can't reproduce the terminated messages in preview-tui but going by those SO threads and testing locally in my shell this seems to hide the messages.

PR #1592, opened 2023-02-19
nmv: ensure the tmpfile is cleaned up in all cases

N-R-K:
currently, there's a couple exit paths where the tmpfile doesn't get cleaned up.

### DIFF
--- a/plugins/.nmv
+++ b/plugins/.nmv
@@ -37,8 +37,6 @@ esac
 
 exit_status=0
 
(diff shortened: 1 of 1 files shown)


KlzXS (approved):
Looks good. Are there any other plugins where this should be done as well? I know at least a couple others use `mktemp`.

N-R-K:
> Looks good. Are there any other plugins where this should be done as well? I know at least a couple others use `mktemp`.

Haven't checked. This one stood out since I modified it recently.

jarun:
Thanks guys!

PR #1597, opened 2023-02-25
Add old 1997-2003 .doc and .xls preview support

amalgame21:
Add old 1997-2003 .doc and .xls preview support


luukvbaal (approved):

luukvbaal:
Untested on my part but seems to match a quick look here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types, thanks.

amalgame21:
> Untested on my part but seems to match a quick look here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types, thanks.

wow nice webpage!
That exactly matches the output of `file -bL --mime-type` on my computer.

jarun:
Thanks guys!

PR #1598, opened 2023-02-25
More correct tilde handling

N-R-K:


jarun:
While testing this I noticed I can't create a file `~` within any directory. Can that be handled?

N-R-K:
> While testing this I noticed I can't create a file `~` within any directory. Can that be handled?

Sure, I'll take a look in a while.

N-R-K:
OK, I think `mkpath` needs to be updated to support this. But `mkpath` is called in so many places, that I'm not very confident about changing it's behavior.

jarun:
I tried a quick path and was able to create it. But then I inadvertently did `^x` on the file and lost my home. ;)
So please be extremely careful about this.

N-R-K:
> But then I inadvertently did `^x` on the file and lost my home. ;)

image

Were you able to restore it?

N-R-K:
> But then I inadvertently did `^x` on the file and lost my home. ;)

OK, I'm now testing this in a chroot. It looks like both this PR *and* current nnn `master` will end up deleting $HOME if you try to delete a file named `~`.

This needs to be fixed regardless of weather we support creating `~` file or not.

jarun:
I wasn't able to restore but I generally take backups. So not much is lost. But I do need to configure the env... :(

Yes, this needs to be fixed.

N-R-K:
OK, I've traced down why this is happening. It's here:

https://github.com/jarun/nnn/blob/2a38132012583c60b1a1ec0ed8bf48e402ee116e/src/nnn.c#L7716-L7718

If the user inputs `~` (or if a file name is just `~`), we call `abspath()` on it - which obviously ends up expanding `~` to `$HOME`.

I think we'll need to track down every single `abspath` calls where the 2nd argument (cwd) is NULL.

N-R-K:
OK, I've made `mkpath` and `abspath` more "strict" about tilde handling. This seems to be working fine. Try it out (in a chroot or some sort of jail, of course).

jarun:
Request other devs and users to try this out.
It would take some time for me to get back to my comfort zone.

jarun:
Did someone get a chance to test it?

N-R-K:
I tested it on a chroot, but it'd be nice if someone else also takes a look at the code and can test it (under some sort of chroot) as well. @KlzXS

src/nnn.c
```diff
@@ -3837,8 +3840,8 @@ static char *get_kv_val(kv *kvarr, char *buf, int key, uchar_t max, uchar_t id)
 				return pluginstr + kvarr[r].off;
 
 			val = bmstr + kvarr[r].off;
-			convert_tilde(val, g_buf);
-			return abspath(((val[0] == '~') ? g_buf : val), NULL, buf);
+			bool tilde = convert_tilde(val, g_buf);
```
KlzXS, line 3843: Maybe add the declaration at the top of the function to be more in line with the rest of the code.

src/nnn.c
```diff
@@ -1137,16 +1147,8 @@ static size_t mkpath(const char *dir, const char *name, char *out)
 {
 	size_t len = 0;
 
-	if (name[0] == '~') { //NOLINT
+	if (tilde_is_home_strict(name)) { //NOLINT
```
KlzXS, line 1150: Do you think a comment here explaining the rationale would be a good addition?

KlzXS (approved):
I ran it raw, no need for `chroot`s of any kind.

It no longer deletes the users' `$HOME` when deleting `~`. You can actually rename, copy, move and remove files that start with `~` without any issues. Plugins also seem to handle it properly getting passed the proper path.

I'd say this is good to go. If any extra issues arise (which I don't expect we'll even see reported) we can handle them as they come. As long as we don't nuke someone's system it's fine.

src/nnn.c
N-R-K: Absolutely. Added in 003228a.

src/nnn.c
N-R-K: The declaration style seems to be very mixed in the codebase. On some functions, it's strictly on top while on others there are mid declaration.

In this case, the result is used just once, so I think it's best to declare it close to usage instead.

KlzXS: I know not every function is strictly ANSI C declarations, but they're also no "mid-sentence" if that makes sense. They are usually above a block that deals with a particular thing. It's just aesthetics you can leave it like that for all I care.

src/nnn.c
KlzXS: Great!

jarun:
Thanks guys!

PR #1602, opened 2023-02-28
Create quitcd script for nushell

mmai:
This adds the "cd on quit" functionality for Nushell https://www.nushell.sh/


jarun:
Thank you!

PR #1606, opened 2023-03-07
Fix paths in nushell quitcd script

mmai:
This set the correct path to the temp nnn file when XDG_CONFIG_HOME is set. It fixes https://github.com/jarun/nnn/discussions/1605


jarun:
Thank you!

Issue #1615, opened 2023-03-21
`quitcd` not running properly in folders with name containing `\\` or ending with `\`


amalgame21:
To reproduce:
1. Use this script https://github.com/jarun/nnn/blob/master/misc/quitcd/quitcd.bash_zsh 
2. Create folder `tes\\t` and `test\`
3. Running `n` in terminal, in folder `tes\\t`, press `^G` with output `/home/amalgame/.config/nnn/.lastd:cd:1: no such file or directory: /home/amalgame/tes\t`
4. Running `n` in terminal, in folder `test\`, press `^G` with output `/home/amalgame/.config/nnn/.lastd:1: unmatched "`

N-R-K:
It's due to this:

https://github.com/jarun/nnn/blob/c4678e311641d328cd127a5f83f7f566aeda09d2/src/nnn.c#L2783

Just wrapping something in quotes isn't enough, we'll have to either:

1. Do proper shell escaping
2. Change it to `dprintf(fd, "%s", curpath);` and then adjust the shell quit-cd functions to do the `cd` manually.

2 seems like a better idea.

N-R-K:
Very lightly tested, but I think this should probably work:

```diff
diff --git a/src/nnn.c b/src/nnn.c
index 0e41cd1f..a4caf91a 100644
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2761,6 +2761,39 @@ static void archive_selection(const char *cmd, const char *archive)
 	free(buf);
 }
 
+static char *shell_escape(const char *s)
+{
+	size_t n = strlen(s), w = 0;
+
+	if (s == g_buf) {
+		DPRINTF_S("s == g_buf");
+		return NULL;
+	}
+
+	g_buf[w++] = '\''; /* begin single quote */
+	for (size_t r = 0; r < n; ++r) {
+		/* potentially too big: 4 for the single quote case, 2 from
+		 * outside the loop */
+		if (w + 6 >= sizeof g_buf)
+			return NULL;
+
+		switch (s[r]) {
+		case '\'':
+			g_buf[w++] = '\''; /* end single quote */
+			g_buf[w++] = '\\'; /* put \' so it's treated as literal single quote */
+			g_buf[w++] = '\'';
+			g_buf[w++] = '\''; /* start single quoting again */
+			break;
+		default:
+			g_buf[w++] = s[r];
+			break;
+		}
+	}
+	g_buf[w++] = '\''; /* end single quote */
[...]

PR #1616, opened 2023-03-22
add shell_escape() to properly escape filenames fed to shell

N-R-K:
Fixes: https://github.com/jarun/nnn/issues/1615

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1299,6 +1299,42 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
 	return resolved_path;
 }
 
+/* wraps the argument in single quotes so it can be safely fed to shell */
+static bool shell_escape(char *output, size_t outlen, const char *s)
+{
+	size_t n = xstrlen(s), w = 0;
+
+	if (s == output) {
+		DPRINTF_S("s == output");
+		return FALSE;
+	}
+
+	output[w++] = '\''; /* begin single quote */
+	for (size_t r = 0; r < n; ++r) {
+		/* potentially too big: 4 for the single quote case, 2 from
+		 * outside the loop */
+		if (w + 6 >= outlen)
+			return FALSE;
+
+		switch (s[r]) {
+		/* the only thing that has special meaning inside single
+		 * quotes are single quotes themselves. */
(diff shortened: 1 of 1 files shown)


KlzXS (approved):
The changes here look good. Did you want to replace all potentially problematic paths in shell commands we invoke in this PR or did you want to leave that for another one when all such places are identified.

In either case commands in `patterns` could use this treatment, though `shell_escape()` won't suffice as is. In some of them paths are passed inside bigger string in `''`, see `P_CPMVRNM`.

Escaping with `''` should work for all potentially problematic paths except maybe those that contain certain non-printable characters. I think those need some different type of escaping, but I'm sure we don't need to worry about that<sup>TM</sup>.

src/nnn.c
```diff
@@ -1299,6 +1299,42 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
 	return resolved_path;
 }
 
+/* wraps the argument in single quotes so it can be safely fed to shell */
+static bool shell_escape(char *output, size_t outlen, const char *s)
```
jarun, line 1303: Would this handle unicode characters as well?

jarun (changes requested):

src/nnn.c
KlzXS: I had no problem both with and without escaping `t̶̩͇̜̔͒́̈́̃ē̸̪̽s̶̨͛t̴͍̙̳̳͙̅f̷͎̐͒ỉ̴̧̦͙̯̞̇̿l̵̡̯̝̞̑͌̄͊e̶͉̟̙̫͑͂̉` while working with it in the shell. If it can handle Zalgo it can handle sane everyday Unicode, I'd say.

PS. Sorry if this breaks your emails.

jarun: Cool!

N-R-K:
> Did you want to replace all potentially problematic paths in shell commands we invoke in this PR or did you want to leave that for another one when all such places are identified.

Ideally, yes. But doesn't necessarily need to be done now. We can fix the other cases up in a later PR.

jarun:
Thank you!

Issue #1614, opened 2023-03-19
Plugin `preview-tui` failed to launch in `tmux` while hovering on files with special characters in the filename.


amalgame21:
Since commit 80c0774, then preview cannot be launched while hovering on files with `space` , open or close bracket `(` `)` in the filename.

amanjitsk:
Just wanted to add that this only happens when starting `preview-tui` when hovered on such a file. If it starts as normal, then hovering over such a file still works as expected. With `kitty`, it does not even work if the hovered file is in a directory with spaces in its name (even if the hovered file does not) - this second variant of this bug does not happen in `tmux` for example. The first variant can be fixed on line 532 by
```
toggle_preview "'$1'" &
```
but this doesn't explain why the bug exists in the first place, and seems like it should be unnecessary because everything is double quoted anyway. Similarly, for the kitty bug, on line 173
```
--cwd "'$PWD'" ...
```
but once again, this shouldn't be needed with the double quotes.

jarun:
@luukvbaal can you please have a look?

amalgame21:
Sorry I did not mentioned that I was using `nnn` inside `tmux`, since I alias it long time ago and treat it as standard ,I just forgot I was using `tmux`.
What I described above only for `tmux` case. Outside `tmux`, the `preview-tui` works perfectly.

Now also tested with no other NNN env (only `NNN_PLUG='p:preview-tui'` and `export NNN_PLUG`) and launching `nnn` with `-a`, also not work inside `tmux`. 

I will also look into it later.

amalgame21:
This is the output when I launch the `preview-tui` plugin for file `~/test/a b c (.mp4`

```bash
/home/amalgame/.config/nnn/plugins/preview-tui: eval: line 170: syntax error near unexpected token `('
[...]

PR #1617, opened 2023-03-22
preview-tui: shell escape filenames being fed to eval

N-R-K:
Fixes: https://github.com/jarun/nnn/issues/1614

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -131,6 +131,10 @@ elif [ "$NNN_SPLIT" != 'h' ]; then
     NNN_SPLIT='v'
 fi
 
+sh_escape() {
+    printf "%s" "$1" | sed "s|'|'\\\''|g;s|^|'|;s|$|'|"
+}
+
 ENVVARS="$ENVVARS
 NNN_SPLIT=$NNN_SPLIT
 NNN_TERMINAL=$NNN_TERMINAL"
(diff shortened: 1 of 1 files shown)


N-R-K:
Tested on filenames with spaces and parenthesis.

@amalgame21 please confirm if this fixes all your test cases or not.

amalgame21:
For filename only, seems ok, will test it when I back home.
For folder name (I mean current directory), probably not because the `$PWD` and `$PREVIEW_MODE` will contain current directory name which is a part of `$ENVSTRING`.

N-R-K:
> For folder name (I mean current directory), probably not because the `$PWD` and `$PREVIEW_MODE` will contain current directory name which is a part of `$ENVSTRING`.

Hmm, good call. I'll test it out soon (and fix if needed).

luukvbaal:
Do we think https://github.com/jarun/nnn/pull/1580 was worth it? It seems figuring out workarounds for that change and the way to properly pass the environment variables to a newly support preview may be more work than simply copying the list of variables again.

If `preview-tui` was written in bash we could have used a proper array of variables but this POSIX sh array I used seems to have downsides. Not sure if worth it...

luukvbaal:
I'm not to keen on creating a subshell for each envvar we have to escape. It noticeably increases the time for the first preview to show up.

If we can't figure out a way that avoids that and preferably even the for loop to begin with, I think I'd rather revert the refactor part of https://github.com/jarun/nnn/pull/1580.

amalgame21:
Tested, working fine, thanks for you work! But it seems slower than before.
Understand it is hard to do in POSIX, hope it is easier using `bash` array.
The `finder` plugin also require bash ,which is the first example in the nnn [plugin](https://github.com/jarun/nnn/tree/master/plugins) page.

N-R-K:
> But it seems slower than before.

Spawning a subshell for each env var will do that :)

I think we should either just use bash (it has builtin substitution with `${var/s/r}` and has real arrays) or revert the refactor part of the other commit.

luukvbaal:
Yeah I started converting to bash arrays a while ago but got sidetracked on other OS work :)

PR #1624, opened 2023-04-09
plocate plugin

pielgrzym:
Hi,

Thanks for this great tool! I created a simple plugin that uses locate/plocate command to search for file/path name and pipes the result to fzf. User can select desired path and NNN will cd into given path. Plugin has fallback in case user uses locate and detects if path is a file or directory.

### DIFF
--- a/plugins/plocate
+++ b/plugins/plocate
@@ -14,13 +14,15 @@
 printf "find: "
 read -r pattern
 
+printf "%s" "+l" > "$NNN_PIPE"
+
 if type plocate >/dev/null 2>&1; then
-    dest_path=$(plocate "$pattern" | fzf)
+    dest_path=$(plocate -0 "$pattern" > "$NNN_PIPE")
 else
-    dest_path=$(locate "$pattern" | fzf)
+    dest_path=$(locate -0 "$pattern" > "$NNN_PIPE")
 fi
 
-if [[ -f "$dest_path" ]]; then
+if [ -f "$dest_path" ]; then
     dest_dir=$(dirname "$dest_path")
 else
     dest_dir="$dest_path"
--- a/plugins/plocate
+++ b/plugins/plocate
@@ -1,0 +1,24 @@
+#!/usr/bin/env sh
+
+# Description: Find files using locate/plocate and cd to their directory
+#
+# Dependencies: fzf, locate/plocate
+#
+# Note: Plugin automatically fallbacks to locate in absence of plocate
+#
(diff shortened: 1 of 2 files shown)


plugins/plocate
```diff
@@ -0,0 +1,28 @@
...
+printf "find: "
+read -r pattern
+
+if type plocate >/dev/null 2>&1; then
+    dest_path=$(plocate "$pattern" | fzf)
+else
+    dest_path=$(locate "$pattern" | fzf)
+fi
+
+if [[ -f "$dest_path" ]]; then
```
N-R-K, line 23: `[[ ]]` is a bash feature. Change it to `[ ]` if you want it to be posix compliant.

plugins/plocate
```diff
@@ -0,0 +1,28 @@
...
+# Shell: POSIX compliant
+# Author: Jakub Nawalaniec
+
+. "$(dirname "$0")"/.nnn-plugin-helper
+
+printf "find: "
+read -r pattern
+
+if type plocate >/dev/null 2>&1; then
+    dest_path=$(plocate "$pattern" | fzf)
```
N-R-K, line 18: I'd avoid depending on fzf. `nnn` has a listing feature alreay, see [quick grep](https://github.com/jarun/nnn/tree/master/plugins#quick-grep-using-rg) example (although I'm wondering why the example is using `eval`).

N-R-K (changes requested):

jarun:
Yes, we have too many plugins already to find, cd etc.

plugins/plocate
pielgrzym: Done :heavy_check_mark:

plugins/plocate
pielgrzym: I tried the code above but the result is a root level directory list not a list of all paths I can easily filter to see what has been matched. I passed "+l" as in grep plugin. Also it seems length of results is heavily limited this way. When I search for a very common dir name (like: "git") I get an error "file limit exceeded" (unlike fzf, which just flys with whatever I toss at it). Any way around this?

pielgrzym:
Ah, I see - no problem, shall I close the PR? :)

jarun:
Yes.

jarun:
I closed it.

jarun:
You may want to try the fzcd and fzopen plugins.

pielgrzym:
Thanks for tips - those plugins just search a subpath, not mlocate/slocate/plocate file db :)

PR #1625, opened 2023-04-09
refactor tables to use x-macros

N-R-K:
a lot of the look up tables we have declares the `#define` and the array contents in different places. which is error prone and harder to change. for example adding something in the middle requires manually updating all the later `#defines`.

using [x-macros](https://www.drdobbs.com/the-new-c-x-macros/184401387) they can be grouped together much more nicely.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -707,19 +707,20 @@ static const char * const env_cfg[] = {
 };
 
 /* Required environment variables */
-#define ENV_SHELL  0
-#define ENV_VISUAL 1
-#define ENV_EDITOR 2
-#define ENV_PAGER  3
-#define ENV_NCUR   4
+#define ENVS \
+	X(ENV_SHELL, "SHELL") \
+	X(ENV_VISUAL, "VISUAL") \
+	X(ENV_EDITOR, "EDITOR") \
+	X(ENV_PAGER, "PAGER") \
+	X(ENV_NCUR, "nnn")
 
-static const char * const envs[] = {
-	"SHELL",
-	"VISUAL",
-	"EDITOR",
-	"PAGER",
-	"nnn",
-};
+#define X(ENUM, STRING) ENUM,
+enum { ENVS };
+#undef X
+
+#define X(ENUM, STRING) STRING,
+static const char * const envs[] = { ENVS };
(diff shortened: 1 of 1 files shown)


N-R-K:
Just a work in progress for now. What do you guys think, is it worth it or should I drop it? @KlzXS

jarun:
The current arrangement is fine and straightforward. :)

KlzXS:
I like the idea. It can sometimes be bothersome to try match a name to a string.

But I don't like these x-macros. If we were to refactor these things why not just go for:
```c
#define ENV_SHELL  "SHELL"
#define ENV_VISUAL "VISUAL"
#define ENV_EDITOR "EDITOR"
#define ENV_PAGER  "PAGER"
#define ENV_NCUR   "nnn"
```
I think that's much cleaner and we never use `envs` (or other such arrays for that matter) except with the corresponding index variables/defines. Any decent compiler should optimize it properly, so there wouldn't even be any extra overhead.

I'll leave it to @jarun to give the final word.

N-R-K:
> It can sometimes be bothersome to try match a name to a string.

Could also just use `enum` instead of `#define` (which gets incremented automatically) and then use designated initializer for the tables:

```c
enum {
	ENV_SHELL,
	ENV_VISUAL,
	ENV_EDITOR,
	ENV_PAGER,
	ENV_NCUR,
};
 
static const char * const envs[] = {
	[ENV_SHELL] = "SHELL",
	[ENV_VISUAL] = "VISUAL",
	[ENV_EDITOR] = "EDITOR",
	[ENV_PAGER] = "PAGER",
	[ENV_NCUR] = "nnn",
};
```

This doesn't involve any macros but still makes adding/moving things around easier.

N-R-K:
> we never use `envs` (or other such arrays for that matter) except with the corresponding index variables/defines.

Hmm, I didn't thoroughly check the usage sites. But if the table itself wasn't necessary, then this seems like a good idea.

0xACE:
I feel like we could benefit for this to be implemented for all the messages we use...

jarun:
Yes, we can take the enum approach. BTW I am planning to make a release tomorrow.

PR #1630, opened 2023-04-19
feat(preview-tui): handle quoting in `start_preview` more robustly

### LINKED ISSUE #1614: Plugin `preview-tui` failed to launch in `tmux` while hovering on files with special characters in the filename.

musjj:
This PR makes the script more resistant to naughty filenames.
The script now depends on bash for the following features:
- Arrays
Correctly creating and passing argument lists is now simple
- Parameter transformation
`${parameter@Q}` makes it easy to correctly quote a string so that it can be safely re-evaluated by the interpreter later.

On iTerm, the shell command used to render the preview is now passed to osascript via a named pipe: `$FIFO_OSASCRIPT`.
By not embedding the shell command directly, we now no longer need to worry about osascript's quoting rules. It's not perfect, because `$SHELL` and `$TMPDIR` might contain naughty characters, but it's quite unlikely to happen.

Fix #1614

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 # Description: Terminal based file previewer
 #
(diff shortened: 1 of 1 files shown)


luukvbaal:
Thanks! I had a similar commit laying around but there were problems and hadn't gotten around to fixing them yet.

I can check windows terminal tomorrow and iterm through a VM but other testers are welcome.

N-R-K:
Looks OK to me (lightly tested). A couple comments:

* The `$FIFO_OSASCRIPT` thing is new. I can guess what it's doing and why but the commit body should explain it too.
* Additionally the PR description should be included in the commit body. It's much more convenient to have the rationale behind a change documented in the commit message so that it can be viewed directly in git instead of hunting down PR/issues in some web ui.

plugins/preview-tui
```diff
@@ -119,6 +121,7 @@ elif [ -n "$WEZTERM_PANE" ]; then
     NNN_TERMINAL=wezterm
 elif [ -z "$NNN_TERMINAL" ] && [ "$TERM_PROGRAM" = "iTerm.app" ]; then
     NNN_TERMINAL=iterm
+    mkfifo "$FIFO_OSASCRIPT"
```
N-R-K, line 124:
```suggestion
    mkfifo "$FIFO_OSASCRIPT" || exit 1
```

Should error check, just in case.

luukvbaal:
I pushed a commit that is noticeably faster when toggling `preview-tui` (comparable to env refactor). I confirmed that it works with `xterm`, `tmux`, `kitty` and `wezterm` as previewer.

I have yet to test `iTerm` and `Windows Terminal`.

musjj:
Nice, that's probably faster than a manual loop. It works pretty well for me.

luukvbaal:
I must say I am not a big fan of writing the osascript command to a file. We already had a working solution by properly escaping that can now be replaced by two simple bash parameter expansions. The latest commit does that, what do you guys think? 
Besides, it is not really a FIFO, just a file and another (unnecessary IMO) one that we must clean up after ourselves at that.

It works on `iTerm` (with the shipped bash version) and `Windows Terminal`, although I made some minor changes after testing on windows so I will test once more tomorrow.

jarun:
Thank you!

Issue #1632, opened 2023-04-22
Please don't force enable assertions.


patlefort:
Typically release builds are built with `NDEBUG` as assertions with `assert()` are meant for debugging. It should be left to the user compiling the code to decide if they want to live dangerously. If some of these checks are that important (or worse, expected), it shouldn't use `assert()`. For me it's a pain to have to make another exception when compiling code from the AUR (Arch User Repository).

#### Environment details (Put `x` in the checkbox along with the information)

Irrelevant as it affect any environment.

#### Exact steps to reproduce the issue

Compile with `-DNDEBUG`.

N-R-K:
nnn itself doesn't use any asserts. It's only used in the hashtable generator. Originally the generator didn't even use the environment flags, but it does since 2fc9d51.

Honestly didn't think there was any need for the generator to use environmental flags since it won't ever be used by the user. But it's whatever, I'll remove the asserts and replace it with something handmade sometime soon.

patlefort:
Are you saying it is only used during the build process? In any case, simply removing the lines at https://github.com/jarun/nnn/blob/6a8d74a43a2135a186dc59c5a1f561444ca098e4/src/icons-hash.c#L31 would solve the issue or undefining NDEBUG before including any files.

N-R-K:
> Are you saying it is only used during the build process?

Yes. And now that I think about it, using CC for the generator isn't right, it won't work for cross compile situation since the generator needs to be built and run on the *host machine* not the target.

On gentoo the convention seems to be use `BUILD_{CC,CFLAGS,LDFLAGS}` for situations like this. I'll probably take a look into what other distros do and fix accordingly.

> simply removing the lines at

That would just disable the checks, which is not a proper solution.

patlefort:
It could simply be a separate project/makefile that's compiled and installed on the host machine. On Arch, we have make dependencies, it would be very simple to manage.

jarun:
@N-R-K is it OK to remove those?

N-R-K:
> is it OK to remove those?

Nah the checks shouldn't be removed. I have a commit ready fixing the issue. Will open a PR sometime today hopefully (been pretty sick lately).

jarun:
Take your time! Get well soon!

PR #1633, opened 2023-04-23
icons-hash: replace assert with handmade version

N-R-K:
since 2fc9d51, the hash-table generator inherits environmental CFLAGS and so we shouldn't disallow setting -DNDEBUG.

fixes: https://github.com/jarun/nnn/issues/1632


jarun:
Thank you!

PR #1638, opened 2023-05-04
Update nerd-font icons to v3

luukvbaal:
* Replace all nerd-font code points with their unicode icon in `icons.h`. (`for i in "$(cat icons.h)"; do echo "$i"; done`)
* Update all deprecated nerd-font code points in `icons.h` and `.iconlookup` with their v3 variants. (`nerdfix check icons.h`)
* Add some "recent" `icons.h` additions to `.iconlookup` (should make it completely in sync again with the new top level icons from `icons.h`...)


plugins/.iconlookup
```diff
@@ -140,6 +145,9 @@ awk 'BEGIN {
     icons["elf"][1] = icons["linux"][1]; icons["elf"][2] = icons["linux"][2]
     icons["epub"][1] = icons["manual"][1]; icons["epub"][2] = icons["manual"][2]
     icons["exe"][1] = icons["exec"][1]; icons["exe"][2] = icons["exec"][2]
+    icons["ex"][1] = icons["elixir"][1]; icons["ex"][2] = icons["elixir"][2]
```
jarun, line 148: Have you added new icons? If yes, why should that be part of this PR?

jarun (changes requested):

plugins/.iconlookup
N-R-K: I think it's just syncing iconlookup with existing icons header.

luukvbaal: > > Add some "recent" icons.h additions to .iconlookup (should make it completely in sync again with the new top level icons from icons.h...)

Yes I added some recent additions to `icons.h` that were missing from `.iconlookup`...

jarun: OK.

src/icons.h
```diff
@@ -43,88 +43,88 @@
 #define ICON_PADDING_RIGHT_LEN (sizeof ICON_PADDING_RIGHT - 1)
 
 /* ARROWS */
-#define ICON_ARROW_UP      ICON_STR(MD_ARROW_UPWARD, "\uf55c", "⬆")
-#define ICON_ARROW_FORWARD ICON_STR(MD_ARROW_FORWARD, "\uf553", "➡")
-#define ICON_ARROW_DOWN    ICON_STR(MD_ARROW_DOWNWARD, "\uf544", "⬇")
+#define ICON_ARROW_UP      ICON_STR(MD_ARROW_UPWARD, "󰁝", "⬆")
```
N-R-K, line 46: Any specific reason for switching to "raw icons" from the escaped version?

(TBH I prefer it this way, so that the icon is visible in my editor compared to having to lookup the code point. But curious if there's any other reason.)

luukvbaal: I needed to convert it to raw icons to be able to run the `nerdfix` program on it. Saw no reason to convert it back as indeed it seems more convenient than having the code points in `icons.h`.

N-R-K:
> @N-R-K can you please test and confirm this one?

Will do in a bit.

plugins/.iconlookup
```diff
@@ -29,6 +29,7 @@ awk 'BEGIN {
     color_docs=202    #color_docs=173      #color_docs="208;135;112"
     color_archive=209 #color_archive=179   #color_archive="235;203;139"
     color_c=81        #color_c=150         #color_c="163;190;140"
+    color_elixir=104  #color_elixir=109    #color_elixir="143;188;187"
```
jarun, line 32: We need to document the change in support model for Nerd Fonts somewhere. The best place is:

https://github.com/jarun/nnn/wiki/Advanced-use-cases#file-icons

jarun (approved):

N-R-K (approved):
Just tried it out with a v3 nerd font and everything seems to be working OK.

jarun:
Thanks guys! Please update the wiki!

PR #1639, opened 2023-05-05
quitcd.bash_zsh: make POSIX compliant; minor fixes and style changes

atomcult:
This commit makes the following changes to quitcd.bash_zsh:
- POSIX compliance and an according rename of the script
- Enforces consistent if-then statements and indentation
- Minor comment fixes

### DIFF
--- a/misc/quitcd/quitcd.bash_sh_zsh
+++ b/misc/quitcd/quitcd.bash_sh_zsh
@@ -1,16 +1,16 @@
 n ()
 {
     # Block nesting of nnn in subshells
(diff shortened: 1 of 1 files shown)


misc/quitcd/quitcd.sh
```diff
@@ -1,16 +1,16 @@
 n ()
```
jarun, line 1: Please change the name to quitcd_bash_sh_zsh.

jarun (changes requested):

jarun:
Thank you!

PR #1644, opened 2023-05-12
Update Haskell icon colour

jaspwr:
Using the purple used for Elixir makes a little more sense than green. I can change this to add a new color to avoid mixing names if you want but I don't see the need to.


jarun:
Thank you!

PR #1646, opened 2023-05-13
Fix some broken table links in docs

Delgan:
Hi.

While reading the docs I noticed that some links were broken (at least on Firefox). Can you reproduce?

I'm referring to the table of content here: https://github.com/jarun/nnn/tree/master/plugins#table-of-contents


N-R-K:
> Can you reproduce?

Yes, it's broken for me too.

Perhaps we can drop the table of content all together because github automatically generates them already (it's the top right icon right beside the pen icon):

image

However, with their recent UI update they seem to have broken the auto-TOC thingy. Because clicking it doesn't do anything for me anymore.

jarun:
Yes, I think it's a good idea to drop the ToC. Just confirm the links are exactly the same.

Is something like that available for the wiki too?

Delgan:
The documentation can be read locally and the project migrated somewhere other than Github, so the ToC may makes sense, but it is a bit redundant indeed.

The auto-ToC is broken for me as well when browsing `plugins` folder, but works fine when cliking the `README.md` file.

Anyway if you want me to delete the ToC or close/update the PR to address the problem more globally, let me know.

N-R-K (approved):
Think we can merge this as is since it fixes the broken links.

Since the auto-TOC on github is broken, we can keep the manual TOC until it's fixed later.

jarun:
Thank you!

PR #1647, opened 2023-05-14
added option "-z" to force yes in cp/mv prompts

mortang2410:
Hi, 

I understand that `cp` and `mv` can either have `-i` for interactive prompts (for every file) or `-f` for saying yes to all overwrites. It was suggested in issue #810 that people should hard-code this replacement (trading some flexibility), while in issue #1458 someone tried to make a plugin (over-engineering?). But, why not handle it the unix way like `cp` and `mv` with an option? By default it will be interactive, but with the  option "-z" nnn will say yes to `cp`/`mv` prompts for overwriting. 

I hope you can consider this simple change. It's a very common usage pattern (overwrite all) for users, many of whom prefer not to hard-code it. I have also updated the help string. Please also excuse any possible error I made. And finally thanks for creating the software.


src/nnn.c
```diff
@@ -8510,7 +8512,15 @@ int main(int argc, char *argv[])
 			g_state.oldcolor = 1;
 			break;
 		case 'd':
-			cfg.showdetail = 1;
+			cfg.showdetail = 1; 
+			break;
+		case 'z':
+			cfg.forceyesprompt = 1; /* unused var, just for easy search */
+#ifdef __linux__
+		cp[6]=mv[6]='f';
```
N-R-K, line 8520: This type of thing is extremely dangerous. If the `cp`/`mv` string changes in the future then this will be silently broken.

I know there's one instance of doing this in the current code, but we ought to fix that instead of adding more dangerous code.

N-R-K (changes requested):

mortang2410:
I now use strcpy to modify the strings (and changed some sizeof to strlen). It should be more explicit and future-proof now. What do you think?

N-R-K:
> I now use strcpy to modify the strings (and changed some sizeof to strlen). It should be more explicit and future-proof now. What do you think?

Oh sorry, my comment was in general, I wasn't asking you to fix it. I already had a patch fixing it locally, didn't open the PR yet.

As for your patch, that strlen turns that buffer into a Variable Length Array (compile with -Wvla and you'll get a warning), which is not acceptable.

mortang2410:
Oh sorry. But yeah, if you have a better solution we'll go with that.

N-R-K:
Opened https://github.com/jarun/nnn/pull/1651 fixing the finicky cp/mv code.

jarun:
> I hope you can consider this simple change.

The reason  this is not implemented is to protect users from overwriting files silently. It is not possible that one would want to keep overwriting files for every use-case over years.

Please write a simple plugin of your own to implement the overwrite feature.

mortang2410:
Sorry if it was not clear but the option `-z` is not turned on by default. It is precisely how `cp` and `mv` also work, with an explicit option to force overwrite, and is not meant for every use-case, but to turn on when needed.
And someone else already tried writing a plug-in. If you think it's basic enough that anyone could just modify the program, I have nothing more to say. I am curious what you use personally when you need to overwite a large number of files yourself. Whether you step out of nnn or there's a way we were not aware of.

N-R-K:
> Whether you step out of nnn or there's a way we were not aware of.

Although it is not for overwriting files, I often need to copy files to root owned dirs. The way I typically accomplish it is with nnn's built-in prompt, which has access to the selection. E.g press <kbd>]</kbd> to get into the builtin prompt and use [special variable %j or %J](https://github.com/jarun/nnn/wiki/concepts#special-variables), e.g `doas cp "%j" ./`.

Note: I don't think we do shell escape on the filenames (yet). So you'd want to be careful if you have some files with characters that might cause issues (such as `$`).

jarun:
> I am curious what you use personally when you need to overwite a large number of files yourself. Whether you step out of nnn or there's a way we were not aware of.

I never overwrite a large number of files in a batch mode just by name without a precise comparison.
The problem with the option is once it is used, the user will never remember to turn it off till see finds one day that he has silently overwritten something of importance.

PR #1648, opened 2023-05-14
Makefile: use -g3 for debug builds

N-R-K:
-g3 builds additional information such as macro definition and so on.


jarun:
Thank you!

PR #1650, opened 2023-05-14
Fix crash when PWD is set to empty string

leo-arch:
nnn crashes when PWD is set to empty string: `PWD="" nnn`

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -8716,7 +8716,7 @@ int main(int argc, char *argv[])
 			/* Start in the current directory */
 			char *startpath = getenv("PWD");
 
-			initpath = startpath ? xstrdup(startpath) : getcwd(NULL, 0);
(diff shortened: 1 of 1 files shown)


N-R-K (approved):
Good catch.

Did you verify if this is the only occurrence of such issue or might there be more?

leo-arch:
Hey @N-R-K. Thanks! There _might_ be more issues like this, but I cannot tell for sure if that's actually the case. As a rule of thumb, I would just check all string pointer checks and make sure they all check the pointer address AND the first byte:

```c
if (ptr && *ptr)
...
```

to avoid this kind of issues. 

Just a suggestion. Maybe it's not even necessary.

jarun:
Thank you!

leo-arch:
You're welcome @jarun!

PR #1651, opened 2023-05-15
make the cp/mv modification more robust

N-R-K:
this makes it so that if the cp/mv commands are changed the in the future, it will continue to work reliably instead of having hardcoded indexes to modify the array.

the `#ifdef __linux__` is also removed, compilers should be smart enough to see that PROGRESS_{CP,MV} are unused (on non-linux systems) and optimize it out.


jarun:
Thank you!

PR #1666, opened 2023-06-06
added NNN_SHELL environment variable [experimental]

israel24961:
dash/fish usage.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -8796,7 +8796,7 @@ int main(int argc, char *argv[])
 	DPRINTF_S(pager);
 
 	/* Get SHELL */
-	shell = xgetenv(envs[ENV_SHELL], utils[UTIL_SH]);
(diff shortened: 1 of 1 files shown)


israel24961:
Using a different shell, `!` keybinding, while using `nnn`.

i.e. Having [fish shell](https://fishshell.com) __Non posix compliant shell__   as the shell used with `!`.
Otherwise the user would need to set  __fish__ shell as default which would make most of the scripts uncompatible.

Unless already exists an option to map `!`.

N-R-K:
> Otherwise the user would need to set **fish** shell as default which would make most of the scripts incompatible.

AFAIK the `SHELL` environment variable should be set to the interactive shell. It has nothing to do with the default `/bin/sh`. Here's what I get on my system (using zsh as interactive shell and dash as `/bin/sh`):

```console
$ echo "$SHELL"
/bin/zsh
$ ls -lah /bin/sh
lrwxrwxrwx 1 root root 4 Feb 10 14:45 /bin/sh -> dash
```

PR #1672, opened 2023-06-14
Preview-tui add -C to NNN_PAGER

luukvbaal:
This makes it so that when the paged text is fewer lines than the preview pane, it is placed at the top instead of at the bottom (#1665).

Not sure when or if the default for this in `less` changed but aligning at the top does seem better as `nnn` does so as well.


jarun:
Thank you!

PR #1679, opened 2023-07-01
Remove some unnecessary ifdefs

N-R-K:
### define _FILE_OFFSET_BITS 64 unconditionally

according to the manpage, it won't have any effect on 64bit system
anyways. and musl always uses 64bit so this macro doesn't have any
effect there either.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -28,13 +28,12 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#define _FILE_OFFSET_BITS 64 /* Support large files on 32-bit glibc */
+
 #if defined(__linux__) || defined(MINGW) || defined(__MINGW32__) \
 	|| defined(__MINGW64__) || defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


N-R-K:
Okay, I give up. No clue why macos is failing, it is completely evil.

src/nnn.c
```diff
@@ -28,14 +28,11 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#define _GNU_SOURCE
```
jarun, line 31: We treat warnings as errors in compilation options. Is there are chance of this failing on any platform due to double definition?

jarun (changes requested):

src/nnn.c
N-R-K: The CI seems to be happy, but I've dropped the _GNU_SOURCE commit, just in case.

My main goal was to remove using internal libc macros but macos kinda ruined that.

As for `_FILE_OFFSET_BITS`, glibc recommends defining it to 64 unconditionally AFAIK. So it should be okay.

jarun:
Thank you!

PR #1686, opened 2023-07-08
nmount: make `pmount` optional

Anomalocaridid:
Edit: Rather than entirely replacing pmount, this PR will now make pmount optional and use `udisksctl` if it is not installed.

Currently, `nmount` uses `pmount` for mounting and unmounting drives and `udisksctl` for powering them off. However, `udisksctl` can handle mounting and unmounting drives as well, so a dependency on `pmount` seems somewhat redundant, barring a reason to use `pmount` specifically.

In case it matters, this has the side effect of changing the default mount location from `/media/<device>` to udisks2's mount directory (which defaults to `/run/media/<username>/<volume label>`), although one can set a udev rule to change it back if desired: https://wiki.archlinux.org/title/Udisks#Mount_to_/media.

However, this side effect also makes `nmount` work better with other `udisks2`-based programs such as [`udiskie`](https://github.com/coldfix/udiskie), which uses `udisks2` to automatically mount removable drives. Currently, `nmount` does not play well with drives mounted by `udiskie`; attempting to unmount them causes an error.

### DIFF
--- a/plugins/nmount
+++ b/plugins/nmount
@@ -32,14 +32,24 @@ while [ -n "$dev" ]; do
     elif [ "$dev" = "q" ]; then
         exit
     else
-        if grep -qs "$dev " /proc/mounts; then
+        # LUKS volumes mounted with udisksctl appear differently than with pmount
+        if grep -qs "$dev " /proc/mounts || [ -n "$(lsblk -n "/dev/$dev" -o MOUNTPOINT)" ]; then
             sync "$(lsblk -n "/dev/$dev" -o MOUNTPOINT | sed "/^$/d")"
             if type pumount >/dev/null 2>&1; then
                 pumount "/dev/$dev"
                 exit_code="$?"
             else
-                udisksctl unmount -b "/dev/$dev" --no-user-interaction
-                exit_code="$?"
+                # Unlike pmount, udisksctl does not transparently handle LUKS volumes
+                # We need to manually get the unlocked device, unmount it, and then lock the volume
+                if lsblk -n "/dev/$dev" -o FSTYPE | grep "crypto_LUKS" >/dev/null; then
+                    dev_map="$(udisksctl info -b /dev/"$dev" | grep "CleartextDevice" | grep -o "dm_2d[[:digit:]]*" | sed "s/_2d/-/")"
+                    udisksctl unmount -b "/dev/$dev_map" --no-user-interaction >/dev/null
+                    exit_code="$?"
+                    udisksctl lock -b "/dev/$dev" --no-user-interaction >/dev/null
+                else
+                    udisksctl unmount -b "/dev/$dev" --no-user-interaction >/dev/null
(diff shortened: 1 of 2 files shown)


N-R-K:
See this https://github.com/jarun/nnn/discussions/1667#discussioncomment-6137278

Anomalocaridid:
I see. I apologize for neglecting to check the discussions for similar topics.

In that case, how about I change my PR to instead make `pmount` optional and have `nmount` fall back to `udisksctl` if it is not installed?

jarun:
> In that case, how about I change my PR to instead make pmount optional and have nmount fall back to udisksctl if it is not installed?

I'm fine with that. It seems you are using 8 spaces instead of 4.
Also, does `udisksctl` work with LUKS encrypted volumes?

Anomalocaridid:
> I'm fine with that. It seems you are using 8 spaces instead of 4.

Sorry about that. I have fixed the formatting.

> Also, does `udisksctl` work with LUKS encrypted volumes?

Yes, however unlike `pmount`, it requires running separate commands for (un)mounting and (un)locking. I have addressed this by running those commands for all drives mounted using `udisksctl` and silencing the error output for the (un)locking commands. As far as I was able to determine, all the (un)locking commands do for non-LUKS volumes is exit with an error message.

I am unsure if this is the best approach for handling LUKS and non-LUKS volumes, so I added it in a separate commit so it can be easily removed or replaced if necessary.

plugins/nmount
```diff
@@ -34,14 +34,30 @@ while [ -n "$dev" ]; do
...
         if grep -qs "$dev " /proc/mounts; then
             sync "$(lsblk -n "/dev/$dev" -o MOUNTPOINT | sed "/^$/d")"
-            if pumount "/dev/$dev"; then
+            if type pumount >/dev/null 2>&1; then
+                pumount "/dev/$dev"
+                exit_code="$?"
+            else
+                udisksctl unmount -b "/dev/$dev" --no-user-interaction
+                exit_code="$?"
+                udisksctl lock -b "/dev/$dev" --no-user-interaction 2>/dev/null
```
jarun, line 43: We can't attempt lock/unlock on volumes without identifying if they are LUKS encrypted or not.
Please check the following link for methods: https://www.baeldung.com/linux/check-disk-encryption

jarun (changes requested):

plugins/nmount
Anomalocaridid: I have re-implemented the locking and unlocking of LUKS volume with udiskctl so that it works properly. This time, I made sure to test my implementation with both un-encrypted and encrypted flash drives.

jarun:
Can someone else also test and confirm this change with udiskctl and pmount?

jarun:
Thank you!

Anomalocaridid:
Thank you as well

jarun:
I found it quite useful!

Issue #1692, opened 2023-07-14
O_NERD and O_EMOJI creates duplicate letters in filename


FelixKratz:
#### Environment details (Put `x` in the checkbox along with the information)

- [x] Operating System: macOS Ventura
- [x] Desktop Environment: Aqua
- [x] Terminal Emulator: kitty, Alacritty
- [x] Shell: zsh
- [ ] Custom desktop opener (if applicable):
- [x] Program options used: none
- [x] Configuration options set: O_NERD
- [ ] Plugins are installed
- [x] Issue exists on `nnn` master

#### Exact steps to reproduce the issue
Clone the repo, enable nerdfonts in the `makefile`
```makefile
O_NERD := 1  # support icons-nerdfont
```
and compile with `LDLIBS="-L/opt/homebrew/opt/ncurses/lib/" make`.

Some files (with certain icons it appears) have duplicate initial letters as seen in the screenshot below:
![Screenshot 2023-07-14 at 11 48 58](https://github.com/jarun/nnn/assets/22680421/a01ccdcf-55b0-4a7e-bfb2-a0152226fc9e)

as soon as I select the makefile it will also have the duplicate initial letter:
![Screenshot 2023-07-14 at 11 54 34](https://github.com/jarun/nnn/assets/22680421/9801c5e9-2fe1-4a66-b458-145b471e69ba)

I have done some testing, this seems to also happen with `O_EMOJI` is set to enabled:
![Screenshot 2023-07-14 at 11 56 51](https://github.com/jarun/nnn/assets/22680421/ce931411-2488-4bfb-ba9a-46b0bd2f7396)

 but not with `O_ICONS` enabled:
![Screenshot 2023-07-14 at 11 57 37](https://github.com/jarun/nnn/assets/22680421/765c2c14-f100-4e8d-b1e3-6e65be515474)

I have peeked into the source code to find what is different between those options, but did not find a convincing clue.

I have also checked with kitty and alacritty, they both face the same problem.

These are the dynamic libs used by nnn:
```bash
 x otool -L ./nnn
./nnn:
	/usr/lib/libedit.3.dylib (compatibility version 2.0.0, current version 3.0.0)
	/opt/homebrew/opt/ncurses/lib/libncursesw.6.dylib (compatibility version 6.0.0, current version 6.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)
```

N-R-K:
[...]

PR #1694, opened 2023-07-18
fix O_NERD and O_EMOJI on macOS

FelixKratz:
This fixes the problem discussed in #1692 

I had to remove the "fast redraw" path on macOS, I don't know if this is a problem.
The important thing seems to be the `clear()` call. I have also tried to keep the "fast redraw" path by utilizing `clrtoeol` but that did not work.

I think this is more of a mitigation than a fix and shows that it may be a problem in ncurses on macOS.

Edit: I have updated the PR to only affect the code generated for macOS

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -6493,6 +6493,7 @@ static int adjust_cols(int n)
 	return (n - 2);
 }
 
+#if !defined(ICONS_ENABLED) || !defined(__APPLE__)
 static void draw_line(int ncols)
 {
 	bool dir = FALSE;
@@ -6526,6 +6527,7 @@ static void draw_line(int ncols)
 
 	markhovered();
 }
+#endif
 
 static void redraw(char *path)
 {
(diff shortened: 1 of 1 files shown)


N-R-K (changes requested):
Thanks for your effort, but frankly I'm not very happy with this patch because `ifdef` mess like this gets out of hand really quickly (and `nnn` has too much ifdef already IMO).

> The important thing seems to be the clear() call.

If I change `erase()` to `clear()` and then hold down `j`, I see severe flickering whenever it scrolls close to the bottom of the screen. Do you not see it on your end?

FelixKratz:
I understand that this is not a very elegant patch but it was the only patch that seemed to work for me.

My personal patch for this problem will be to revert the nerdfont update.

N-R-K:
> I have also tried to keep the "fast redraw" path by utilizing `clrtoeol` but that did not work.

Btw, what exactly did you try? Does the following make any difference?

```diff
@@ -4192,6 +4192,8 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
        char ind = '\0';
        int attrs;
 
+       clrtoeol();
+
        if (cfg.showdetail) {
                int type = ent->mode & S_IFMT;
```

FelixKratz:
I have tried adding the `clrtoeol` call just after https://github.com/jarun/nnn/blob/b835cfcc2abf0503966dec1b9226b3a7c34d371a/src/nnn.c#L6507

and
https://github.com/jarun/nnn/blob/b835cfcc2abf0503966dec1b9226b3a7c34d371a/src/nnn.c#L6520

Didnt help at all with the problem.

PR #1695, opened 2023-07-20
preview-tui SSH compatibility

amalgame21:
When accessing the remote machine through SSH, skip ueberzug preview method which does not support preview in SSH, even though ueberzug in installed on the remote machine.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -418,7 +418,7 @@ image_preview() {
         kitty +kitten icat --silent --scale-up --place "$1"x"$2"@0x0 --transfer-mode=stream --stdin=no "$3" &
     elif [ "$NNN_TERMINAL" = "wezterm" ] && [ -z "$NNN_PREVIEWIMGPROG" ]; then
         wezterm imgcat "$3" &
-    elif exists ueberzug && { [ -z "$NNN_PREVIEWIMGPROG" ] || [ "$NNN_PREVIEWIMGPROG" = "ueberzug" ] ;} && [ -z "$SSH_CONNECTION" ] ; then⏎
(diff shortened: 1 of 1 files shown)


jarun:
@luukvbaal @N-R-K please review and approve if looks good.

plugins/preview-tui
```diff
@@ -418,7 +418,7 @@ image_preview() {
         kitty +kitten icat --silent --scale-up --place "$1"x"$2"@0x0 --transfer-mode=stream --stdin=no "$3" &
     elif [ "$NNN_TERMINAL" = "wezterm" ] && [ -z "$NNN_PREVIEWIMGPROG" ]; then
         wezterm imgcat "$3" &
-    elif exists ueberzug && { [ -z "$NNN_PREVIEWIMGPROG" ] || [ "$NNN_PREVIEWIMGPROG" = "ueberzug" ] ;}; then
+    elif exists ueberzug && { [ -z "$NNN_PREVIEWIMGPROG" ] || [ "$NNN_PREVIEWIMGPROG" = "ueberzug" ] ;} && [ -z "$SSH_CONNECTION" ] ; then⏎
```
N-R-K, line 421: What's this?

image

N-R-K (changes requested):

plugins/preview-tui
amalgame21: vim eol character also copied in tmux... sorry about that.

amalgame21:
Sorry, seems it is still not working properly in my test. 
The preview can be generated by catimg/viu while ssh, but it cannot generate preview of next items after pressing up/down. Have to toggle preview-tui on and off manually for another preview.

PR #1698, opened 2023-07-22
Jump to the next young file

ANtlord:
Forward jumping to the next young file is implemented.  ~The starting position from which the next young file is searched is the current cursor position or the next position if the current position of the cursor points to a young file.~  The starting position from which the next young file is searched is the next position.  If there no young file has been found, the search starts from the beginning.  It stops at the initial position where the search has started.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5107,6 +5107,7 @@ static void show_help(const char *path)
 		"a1-4  Context%11(Sh)Tab  Cycle/new context\n"
 	    "62Esc ^Q  Quit%20q  Quit context\n"
 		 "b^G  QuitCD%18Q  Pick/err, quit\n"
+		 "b^y  Next young file\n"
 	"0\n"
 	"1FILTER & PROMPT\n"
 		  "c/  Filter%17^N  Toggle type-to-nav\n"
@@ -6115,6 +6116,20 @@ static void handle_screen_move(enum action sel)
 	case SEL_END:
 		move_cursor(ndents - 1, 1);
 		break;
+	case SEL_YOUNG:
+	{
+		for (int r = cur;;) {
+			if (++r >= ndents)
+				r = 0;
+			if (r == cur)
+				break;
+			if (pdents[r].flags & FILE_YOUNG) {
+				move_cursor(r, 0);
+				break;
+			}
+		}
+		break;
+	}
 	default: /* case SEL_FIRST */
 	{
 		int c = get_input(messages[MSG_FIRST]);
@@ -7196,7 +7211,8 @@ nochange:
 		case SEL_HOME: // fallthrough
 		case SEL_END: // fallthrough
 		case SEL_FIRST: // fallthrough
-		case SEL_JUMP:
+		case SEL_JUMP: // fallthrough
+		case SEL_YOUNG:
 			if (ndents) {
 				g_state.move = 1;
 				handle_screen_move(sel);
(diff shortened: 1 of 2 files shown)


src/nnn.c
```diff
@@ -5107,6 +5107,7 @@ static void show_help(const char *path)
 		"a1-4  Context%11(Sh)Tab  Cycle/new context\n"
 	    "62Esc ^Q  Quit%20q  Quit context\n"
 		 "b^G  QuitCD%18Q  Pick/err, quit\n"
+	       "cy  Next young file\n"
```
N-R-K, line 5110: Try to keep this aligned with the rest:

```diff
                 "b^G  QuitCD%18Q  Pick/err, quit\n"
-              "cy  Next young file\n"
+                 "cy  Next young file\n"
```

src/nnn.c
```diff
@@ -6115,6 +6116,30 @@ static void handle_screen_move(enum action sel)
...
+		int is_found = false;
+		for (; r < ndents; ++r) {
+			if (pdents[r].flags & FILE_YOUNG) {
+				move_cursor(r, 0);
+				is_found = true;
+				break;
+			}
+		}
+
+		if (!is_found) {
```
N-R-K, line 6132: No need for 2 loops doing the same thing.

```c
	case SEL_YOUNG:
	{
		for (int r = cur;;) {
			if (++r == ndents)
				r = 0;
			if (r == cur)
				break;
			if (pdents[r].flags & FILE_YOUNG) {
				move_cursor(r, 0);
				break;
			}
		}
		break;
	}
```

src/nnn.h
```diff
@@ -256,6 +257,7 @@ static struct key bindings[] = {
 	{ CONTROL('J'),   SEL_AUTONEXT },
 	/* Edit in EDITOR */
 	{ 'e',            SEL_EDIT },
+	{ 'y',            SEL_YOUNG },
```
N-R-K, line 260: `y` seems like a bad key for this. Usually `y` means yank/copy in programs that have vim-like bindings.

N-R-K (changes requested):

src/nnn.h
ANtlord: I agree.  Any suggestions?

src/nnn.c
```diff
@@ -5107,6 +5107,7 @@ static void show_help(const char *path)
 		"a1-4  Context%11(Sh)Tab  Cycle/new context\n"
 	    "62Esc ^Q  Quit%20q  Quit context\n"
 		 "b^G  QuitCD%18Q  Pick/err, quit\n"
+	      "cy  Next young file\n"
```
jarun, line 5110: The alignment is still wrong. Please check on github diff view.

src/nnn.h
jarun: <kbd>^y</kbd>

Would be handy for users in nav-to-type mode also.

jarun (changes requested):

N-R-K (commented):
Looks okay. But please remove this line from the commit message:

> The starting position from which the next young file is searched is the current cursor position or the next position if the current position of the cursor points to a young file.

This is not the case anymore with the new loop. And even in the old loop, this was dead code:

* If initial file was young, it'd get skipped (before the loop).
* If initial file wasn't young, it'd get skipped as well (but this time inside the loop).

So the initial file was always being skipped anyways.

ANtlord:
Looks like I should correct the patches also.  Is that correct?

N-R-K:
Weird, looks like the `--merge` option in `check-patches.sh` is what's causing the patches to fail... It works fine otherwise. I'll look into it later.

Issue #157, opened 2018-11-30
Nothing happens when pressing help key (big terminal, less pager with -F option)


matclab:
Pressing `?`  (`SHIFT+,`) or `D` does nothing.

jarun:
fr == french?

jarun:
And which shortcut is `SHIFT+,`?

jarun:
I don't know French so you'll have to help yourself here. Compile `nnn` in debug mode (`make debug`) and the keycode will be printed here: github.com/jarun/nnn/blob/master/src/nnn.c#L1051

The ascii code of `?` shouldn't change but see what you find for these keys.

matclab:
Ok. In fact it works when my terminal (tilix) is not full screen.
So it is not a problem of the key but a problem of integration between nnn, the terminal and the pager…

I tried with xterm and I have the same problem.

Pressing `?` I sometime see the help flashing very briefly.

jarun:
Awesome! Many thanks!

jarun:
@matclab I have added support for 1 env var combination for PAGER, LESS and SHELL on latest master.

So now (on fish shell) I can run:

    env PAGER='less -FSRXIj.2' nnn

ahmedelgabri:
I have the same issue, I tried both [`kitty`](https://github.com/kovidgoyal/kitty) & [`alacritty`](https://github.com/jwilm/alacritty) & my `$PAGER` is set to `less` only. I'm running macOS Mojave 10.14.2

jarun:
Please try master.
Also, if it doesn't work please try to debug yourself. Debugging local env specific issues is too taxing for the bandwidth we have. Thanks for your understanding.

jarun:
Use the debug build 'make debug'

add logs. the debug log filr would be in /tmp/nnndbg

ahmedelgabri:
One thing I forgot to mention, that when I quit `nnn` after trying `?` multiple times, is that the help is already printed `n` times in the commandline. Where `n` is the number of times I clicked `?` inside `nnn`.

Anyway that's what I did to debug, not sure if I missed anything.

### Steps:

- `make debug`
- Open `nnn` 
- type `?` to get `help` _(2-3 times)_
- quit

#### logfile

```
dents=0x7fb94c95a200
pnamebuf=0x7fb94d000000
ts2.tv_nsec - ts1.tv_nsec=421000
fname=
cur=0
path=/private/tmp/nnn
c=-1
cur=0
[...]

PR #1701, opened 2023-07-25
Reset less --quit-if-one-screen for NNN_PAGER default

czak:
This is a proposal to update the `NNN_PAGER` default command.

## Context

I've  run into the [Help not showing](https://github.com/jarun/nnn/wiki/Troubleshooting#help-not-showing) issue as I was playing with `nnn` for the first time.

My `$LESS` includes `-F`/`--quit-if-one-screen` flag, so (after finally finding #157 and the Wiki entry) I understand why it happens and how to work around it.

However, I'd like to keep the flag in my global `$LESS`.

## Proposal

What do you think about enforcing the behavior for the `NNN_PAGER` default?
According to `man less`, this can be done using `-+F`:

> If an option appears in the LESS variable, it can be reset to its default value on the command line by beginning the command line option with "-+".
> -- https://man.archlinux.org/man/less.1#OPTIONS


jarun:
@N-R-K @luukvbaal any comments on this?

N-R-K:
> I've run into the [Help not showing](https://github.com/jarun/nnn/wiki/Troubleshooting#help-not-showing) issue as I was playing with `nnn` for the first time.
> 
> My `$LESS` includes `-F`/`--quit-if-one-screen` flag, so (after finally finding #157 and the Wiki entry) I understand why it happens and how to work around it.
> 
> However, I'd like to keep the flag in my global `$LESS`.

This all seems fine in theory. However, this patch is changing the default for `preview-tui` plugin and shouldn't have any effect on the help screen (shown via <kbd>?</kbd>) - so I'm a bit confused here.

czak:
> this patch is changing the default for `preview-tui` plugin and shouldn't have any effect on the help screen (shown via `?`)

Ayayay that's my bad :sweat: I must have tested it with a smaller window and thought it worked. Sorry about that.

I updated the fix in [42a3154](https://github.com/jarun/nnn/pull/1701/commits/42a315482d0c3cbe696ffc94e772c03a5a8dd933). It works, but I don't really like this change anymore. The other `utils[]` don't set flags (except for `rm -rf` and `sh -c`), so it feels a bit out of place here.

Please feel free to decide what is best here.
Thank you for taking a look.

N-R-K:
`UTIL_LESS` affects a lot more than just the help screen. I don't think we can merge this without exhaustive testing to make sure it doesn't cause regression elsewhere.

czak:
I'll just go ahead and close it. Maybe it will be useful as reference for others who run into the issue. Thanks!

PR #1711, opened 2023-08-09
Double-width icons can cause duplicated first characters in the filename on macOS

### LINKED ISSUE #1692: O_NERD and O_EMOJI creates duplicate letters in filename

quantonganh:
Fix #1692

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -131,6 +131,18 @@
 #include "icons.h"
 #endif
 
+#if defined(ICONS_ENABLED) && defined(__APPLE__)
+/*
+ * For some reason, wcswidth returns 2 for certain icons on macOS
+ * leading to duplicated first characters in filenames when navigating.
+ * https://github.com/jarun/nnn/issues/1692
+ * There might be a better way to fix it without requiring a refresh.
+ */
+#define macos_icons_hack() do { clrtoeol(); refresh(); } while(0)
+#else
+#define macos_icons_hack()
+#endif
+
 #ifdef TOURBIN_QSORT
 #include "qsort.h"
 #endif
@@ -6520,6 +6532,7 @@ static void draw_line(int ncols)
 	}
 
 	move(2 + last - curscroll, 0);
+	macos_icons_hack();
 	printent(&pdents[last], ncols, FALSE);
 
 	if (g_state.oldcolor && (pdents[cur].flags & DIR_OR_DIRLNK)) {
@@ -6533,6 +6546,7 @@ static void draw_line(int ncols)
 	}
 
 	move(2 + cur - curscroll, 0);
+	macos_icons_hack();
 	printent(&pdents[cur], ncols, TRUE);
 
 	/* Must reset e.g. no files in dir */


quantonganh:
Why this failed?

```
Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.
Cloning git repository
Enumerating objects: 17975, done.
Counting objects: 100% (164/164), done.
Compressing objects: 100% (80/80), done.
Total 17975 (delta 88), reused 135 (delta 81), pack-reused 17811

error fetching: couldn't find remote ref "refs/pull/1711/head"
```

jarun:
Can you please make the change macOS-specific?

src/nnn.c
```diff
@@ -6520,6 +6520,10 @@ static void draw_line(int ncols)
 	}
 
 	move(2 + last - curscroll, 0);
+#if defined(ICONS_ENABLED) && defined(__APPLE__)
+	clrtoeol();
+	refresh();
```
N-R-K, line 6525: Is the `refresh()` also needed? What happens without it?

quantonganh: The issue still persists if we do not refresh the screen:

Screen Shot 2023-08-10 at 06 15 45

N-R-K: I was thinking we can avoid unnecessary flushes. But the performance impact here doesn't seem too noticeable (unlike https://github.com/jarun/nnn/pull/1694). So we can probably merge this in as a hack until someone comes up with something better.

Could you please add a comment under the `#ifdef` referencing the original issue url? And also leave a comment stating that there might be a better way to fix it without a flush (so that it's easily visible to anyone else interested in this) @quantonganh

src/nnn.c
```diff
@@ -6520,6 +6520,10 @@ static void draw_line(int ncols)
 	}
 
 	move(2 + last - curscroll, 0);
+#if defined(ICONS_ENABLED) && defined(__APPLE__)
```
N-R-K, line 6523: Instead of duplicating this, you can put this in a single place on top of the file:

```c
#if defined(ICONS_ENABLED) && defined(__APPLE__)
#define macos_icon_hack() do { clrtoeol(); refresh(); } while(0)
#else
#define macos_icon_hack()
#endif
```

Then just use `macos_icon_fix()` in those places.

quantonganh: I have made the updates as you suggested. Kindly review them.

N-R-K (approved):

src/nnn.c
```diff
@@ -131,6 +131,18 @@
...
 #endif
 
+#if defined(ICONS_ENABLED) && defined(__APPLE__)
+/*
+ * For some reason, wcswidth returns 2 for certain icons on macOS
+ * leading to duplicated first characters in filenames when navigating.
+ * https://github.com/jarun/nnn/issues/1692
+ * There might be a better way to fix it without requiring a refresh.
+ */
+#define macos_icons_hack() do { clrtoeol(); refresh(); } while(0)
```
jarun, line 141: Why do we need a do while loop for this?

jarun (changes requested):

src/nnn.c
N-R-K: It's a common idiom to avoid surprises when used inside if statements. E.g

```c
if (c)
    MACRO();
```

Will not work without the `do {} while(0)` trick because it'll expand to this:

```c
if (c)
    statement_1;
statement_2;
```

jarun: Can't we just have the braces:

#define macos_icons_hack() { clrtoeol(); refresh(); }

N-R-K: Then you won't be able to put the semi-colon when invoking the macro. From [c-faq 10.4](https://c-faq.com/cpp/multistmt.html):

> If the macro expanded to a simple compound statement, the final, caller-supplied semicolon would be a syntax error:

```c
if (cond)
	{ stmt1; stmt2; };
else	/* some other code */
```

jarun:
Thank you!

PR #1712, opened 2023-08-12
Fix bug in fzopen when selection is canceled

Rahlir:
Empty selection is not handled in the current version of fzopen. This means that error is printed when nuke is used and, even worse, on macOS the current directory is opened in Finder. This is a tiny commit to fix that.


jarun:
Thank you!

PR #1717, opened 2023-08-22
Add Image2PDF Conversion Plugin

karimhussein1:
Enables easy conversion of image files to PDF using ImageMagick.


jarun:
Thank you for the PR!

Since there are many many file formats and requirements can change from user to user we refrain from adding file conversion plugins. Can you please share the sample plugin in the discussions sections instead? That way anyone looking for something like this can easily find it.

PR #1719, opened 2023-08-27
Update finder plugin find command with default path #1718

### LINKED ISSUE #1718: 'finder' plugin doesn't work on MacOS. Error message: 'find: illegal option --m'
`finder` plugin doesn't work on MacOS. I get the following error message: `find: illegal option --m`.

Cause: `find` expects the first parameter to be the path where it should start looking. In `gnu find`, it is defaulted to `.` but it MacOS which uses `bsd find`, it is not defaulted.

Fix: The suggested fix is to add `.` to the find command in the `finder` plugin so that it works across all platforms.

- [x] Operating System: MacOS
- [ ] Desktop Environment:
- [ ] Terminal Emulator:
- [ ] Shell:
- [ ] Program options used:
- [x] Plugins are installed: finder

umaranis:
Fixes #1718

### DIFF
--- a/plugins/finder
+++ b/plugins/finder
@@ -55,7 +55,7 @@ readexpr() {
              [ -n "$fexpr" ] && readexpr ;;
         \$*) cmd="${fexpr:1}" ;;
         *)   mapexpr && readexpr
-             cmd="find $fexpr -print0" ;;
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1722, opened 2023-09-04
fzhist plugin: add support for zsh history #1721

### LINKED ISSUE #1721: fzhist plugin: add support for zsh history
Currently, `fzhist` plugin only supports bash and fish history. Enhance it to also support `zsh`.

umaranis:
resolves #1721

### DIFF
--- a/plugins/fzhist
+++ b/plugins/fzhist
@@ -5,6 +5,8 @@
 #
 # Note: Supports only bash, zsh and fish history
 #
+# TODO: For zsh there's also $ZDOTDIR which might need to be checked as well for the -z $HISTFILE && -n $ZDOTDIR case.
+#
 # Shell: POSIX compliant
 # Author: Arun Prakash Jana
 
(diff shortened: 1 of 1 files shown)


plugins/fzhist
```diff
@@ -19,6 +19,9 @@ shellname="$(basename "$SHELL")"
 if [ "$shellname" = "bash" ]; then
     hist_file="$HOME/.bash_history"
     entry="$("$fuzzy" < "$hist_file")"
+elif [ "$shellname" = "zsh" ]; then
+    hist_file="$HOME/.zsh_history"
```
N-R-K, line 23: `$HISTFILE` should be checked for both zsh and bash: https://www.gnu.org/software/bash/manual/html_node/Bash-History-Facilities.html.

(For zsh there's also `$ZDOTDIR` which *might* need to be checked as well for the `-z $HISTFILE && -n $ZDOTDIR` case)

N-R-K (changes requested):

plugins/fzhist
umaranis: Pushed a new commit. I am checking if $HISTFILE available, otherwise going to the default location.

plugins/fzhist
```diff
@@ -17,7 +19,18 @@ fi
...
 
 if [ "$shellname" = "bash" ]; then
-    hist_file="$HOME/.bash_history"
+    if [ -f "$HISTFILE" ]; then
+        hist_file="$HISTFILE"
+    else
+        hist_file="$HOME/.bash_history"
+    fi
+    entry="$("$fuzzy" < "$hist_file")"
+elif [ "$shellname" = "zsh" ]; then
```
jarun, line 28: Please fix alignment.

jarun (changes requested):

plugins/fzhist
umaranis: Not sure what's wrong with alignment... looks good to me.

image

N-R-K: I see that there's some trailing spaces after the `fi`. 

image

Though I don't know if that's what jarun is actually talking about or not.

jarun:
Thank you!
Can you please squash all changes into one?

jarun:
I took care of it. Thanks!

PR #1724, opened 2023-09-07
plugins for integration with MacOS clipboard

### LINKED ISSUE #1723: Plugins for integration with MacOS clipboard
Implement plugins to support file copy and paste operation from MacOS clipboard. This will help sharing files between `nnn` and other programs like Email, Slack, Finder etc. 

There is a drag and drop plugin which caters for this requirement on linux but nothing is available for MacOS that allows the use of `nnn` with other GUI applications (as far as I can see).

umaranis:
resolves #1723

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
             echo "########## clang-tidy-15 ##########"
             clang-tidy-15 src/* -- -I/usr/include -I/usr/include/ncursesw
             echo "########## shellcheck ##########"
-            find plugins/ -type f -not -name "*.md" -exec shellcheck {} +
+            find plugins/ -type f ! \( -name "*.md" -o -name "*-mac" \) -exec shellcheck {} +
 
   package-and-publish:
     machine: true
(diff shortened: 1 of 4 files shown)


jarun:
Why can;t this be integrated in existing plugin `.cbcp`?

Also, CI fails:

```

In plugins/cbcopy-mac line 1:
#!/usr/bin/osascript
^-- SC1008 (error): This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh. Add a 'shell' directive to specify.


In plugins/cbcopy-mac line 12:
  set filePath to (second item of args & "/" & first item of args)
      ^------^ SC2121 (warning): To assign a variable, use just 'var=value', no 'set ..'.
                  ^-- SC1036 (error): '(' is invalid here. Did you forget to escape it?
                  ^-- SC1088 (error): Parsing stopped here. Invalid use of parentheses?


In plugins/cbpaste-mac line 12:
fs=($( osascript -e "use framework \"Foundation\"
   ^-- SC3030 (warning): In POSIX sh, arrays are undefined.
    ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In plugins/cbpaste-mac line 23:
cp -R "${fs[@]}" "$2/"
       ^------^ SC3054 (warning): In POSIX sh, array references are undefined.

For more information:
  https://www.shellcheck.net/wiki/SC1008 -- This shebang was unrecognized. Sh...
  https://www.shellcheck.net/wiki/SC2121 -- To assign a variable, use just 'v...
  https://www.shellcheck.net/wiki/SC2207 -- Prefer mapfile or read -a to spli...

```

umaranis:
If you understand if correctly, the `.cbcp` plugin copies the paths of selected files into the clipboard whereas my plugins copy/paste actual files. Not sure how the two can be integrated.

I think the pipeline is failing because these script can only run on macos (not on linux).

jarun:
I see. In that case, fine.

In that case, please ignore shellcheck for these files.

umaranis:
@jarun shellcheck is ignored now for these plugins

jarun:
Thank you!

PR #1729, opened 2023-09-20
Add support for eza as a replacement for exa

azuline:
exa is unmaintained (see https://github.com/ogham/exa), and https://github.com/eza-community/eza is the community-maintained successor.

This PR adds support for eza as a replacement to exa.


N-R-K (approved):
Seems fine but I'm confused as to why they made a new fork with a different name instead of continuing on the `exa` repository? It seems like other people have commit access to it, so what was the problem?

jarun:
Maybe for more control on the development cycle...

jarun:
Thank you!

PR #1732, opened 2023-09-23
Fix CI

N-R-K:
Also splits the list of checks into multiple lines. IMO makes it more easily readable and editable. And produces better diff when editing.


jarun:
Thank you!

PR #1737, opened 2023-09-26
fix nu quitcd: switched quotes

### LINKED ISSUE #1736: nushell `cd` on quit broken
Commit 57882ffab713e66690d29714a7fbea2cb2e61a5f broke the recommended nushell `cd` on quit

Quitting Nnn now outputs the following error
```nushell
Error: nu::shell::access_beyond_end

  × Row number too large (empty content).
    ╭─[/home/jkappes/.config/nushell/config.nu:26:1]
 26 │     if ($env.NNN_TMPFILE | path exists) {
 27 │       let newpath = (open $env.NNN_TMPFILE | parse 'cd "{nnnpath}"').0.nnnpath
    ·                                                                      ┬
    ·                                                                      ╰── index too large (empty content)
 28 │       ^rm -f $env.NNN_TMPFILE
    ╰────
```

J-Kappes:
- quitcd target directory has been in single quotes since 57882ff
- I updated the nushell snippet to correctly parse it. This should close #1736

### DIFF
--- a/misc/quitcd/quitcd.nu
+++ b/misc/quitcd/quitcd.nu
@@ -7,6 +7,7 @@ def-env n [...x] {
   ^nnn ($x | str join)
   let newpath = (
     if ($env.NNN_TMPFILE | path exists) {
+      # FIXME: fails if path contains single-quote
       let newpath = (open $env.NNN_TMPFILE | parse "cd '{nnnpath}'").0.nnnpath
       ^rm -f $env.NNN_TMPFILE
       echo $newpath
--- a/misc/quitcd/quitcd.nu
+++ b/misc/quitcd/quitcd.nu
@@ -7,7 +7,7 @@ def-env n [...x] {
   ^nnn ($x | str join)
   let newpath = (
     if ($env.NNN_TMPFILE | path exists) {
-      let newpath = (open $env.NNN_TMPFILE | parse 'cd "{nnnpath}"').0.nnnpath
+      let newpath = (open $env.NNN_TMPFILE | parse "cd '{nnnpath}'").0.nnnpath
       ^rm -f $env.NNN_TMPFILE
       echo $newpath
     } else {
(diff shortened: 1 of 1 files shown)


N-R-K:
Are you sure this will work on path which contain `'` ?

J-Kappes:
It doesn't work and I haven't found a fix yet.

N-R-K:
> It doesn't work and I haven't found a fix yet.

If you can't figure out a way to fix it then leave a FIXME comment explaining the edge case.

J-Kappes:
This is pretty hacky, so leaving the FIXME might be preferable.

misc/quitcd/quitcd.nu
```diff
@@ -7,6 +7,7 @@ def-env n [...x] {
   ^nnn ($x | str join)
   let newpath = (
     if ($env.NNN_TMPFILE | path exists) {
+      #FIXME fails on dir names containing single-quotes
```
N-R-K:
```suggestion
      # FIXME: fails if path contains single-quote
```

N-R-K (changes requested):
Small reword.

misc/quitcd/quitcd.nu
```diff
@@ -7,7 +7,8 @@ def-env n [...x] {
   ^nnn ($x | str join)
   let newpath = (
     if ($env.NNN_TMPFILE | path exists) {
-      let newpath = (open $env.NNN_TMPFILE | parse 'cd "{nnnpath}"').0.nnnpath
+      let rawpath = (open $env.NNN_TMPFILE | parse --regex 'cd (?P<dir>.+)').0.dir
+      let newpath = ($rawpath | $"echo ($in)" | /bin/bash -s)
```
N-R-K, line 11: Adding bash dependency on another shell's config doesn't seem like a good idea.

N-R-K:
Feel free to open another PR if you manage to fix it without requiring `bash` or other non-posix dependency.

Issue #1735, opened 2023-09-25
Remove colemak-dh patch ?


jacmoe:
In PR #1587 a patch for colemak was added, effectively replacing the existing colemak-dh patch, but colemak-dh was not removed.

Maybe it is time to actually remove it?
I wouldn't want support for colemak being dropped because of the overhead of having to maintain two separate patches for colemak when only one of them is actually used.

N-R-K:
I support removing the colemak-dh patch. If needed, we can link to it in the wiki or something.

jarun:
Please raise the PR.

PR #1741, opened 2023-09-30
rm colemak-dh

N-R-K:
one colemak patch ought to be enough.

Closes: https://github.com/jarun/nnn/issues/1735


N-R-K:
I've also added a [colemak section](https://github.com/jarun/nnn/wiki/Advanced-use-cases#colemak-keybinds) on the wiki which links to colemak-dh and other colemak variants.

jarun:
Thank you!

PR #1747, opened 2023-10-10
Preview-tui avoid early exit after SIGWINCH

### LINKED ISSUE #1731: preview-tui plugin in kitty crashes after second window resize

- [x] Operating System: Linux
- [x] Desktop Environment: none+i3
- [x] Terminal Emulator: kitty
- [x] Shell: fish
- [x] Program options used: adERx
- [x] Configuration options set: unsure which are relevant, my configuration is here https://github.com/azuline/nixos/blob/master/home/nnn/default.nix
- [x] Plugins are installed: preview-tui, fzopen, imgview, playdir (custom plugin)
- [ ] Issue exists on `nnn` master: Unsure, I'm on 4.8 but have the latest plugins

1. Execute preview-tui plugin in kitty.
2. Resize terminal once. Everything should be fine.
3. Resize terminal again. The preview window exits.

I've poked through the plugin code, but I'm not sure what's causing this issue. That this only happens on the second resize, and not on the first, breaking my intuition a little bit. I'm not familiar with the actions kitty takes when resizing the window (does it send signals to running processes?).

Recording:

https://github.com/jarun/nnn/assets/51880422/c574bdc6-6eda-48f8-a0bd-3480ed75d4e6

I'm happy to put up a pull request if I can understand what the cause is!

luukvbaal:
Fix #1731

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -475,12 +475,14 @@ if [ "$PREVIEW_MODE" -eq 1 ] 2>/dev/null; then
     fi
 
     preview_file "$PWD/$1"
-    preview_fifo &
+    preview_fifo & WAITPID=$!
     printf "%s" "$!" > "$FIFOPID"
     printf "%s" "$PWD/$1" > "$CURSEL"
-    trap 'winch_handler; wait' WINCH
+    trap 'winch_handler' WINCH
     trap 'rm "$PREVIEWPID" "$CURSEL" "$FIFO_UEBERZUG" "$FIFOPID" "$POSOFFSET" 2>/dev/null' INT HUP EXIT
-    wait "$!" 2>/dev/null
+    while kill -s 0 $WAITPID; do
+      wait $WAITPID 2>/dev/null
+    done
     exit 0
 else
     if [ ! -r "$NNN_FIFO" ]; then


luukvbaal:
Tried fixing this in other ways for a bit but this is the only way that I could find that works. Haven't looked into the reason behind the behavior discrepancy here between POSIX sh and bash 🤷🏻‍♂️ 

Alternative suggestions welcome.

plugins/preview-tui
```diff
@@ -478,9 +479,17 @@ if [ "$PREVIEW_MODE" -eq 1 ] 2>/dev/null; then
     preview_fifo &
     printf "%s" "$!" > "$FIFOPID"
     printf "%s" "$PWD/$1" > "$CURSEL"
-    trap 'winch_handler; wait' WINCH
+    trap 'winch_handler' WINCH
     trap 'rm "$PREVIEWPID" "$CURSEL" "$FIFO_UEBERZUG" "$FIFOPID" "$POSOFFSET" 2>/dev/null' INT HUP EXIT
     wait "$!" 2>/dev/null
```
N-R-K, line 484: I haven't run it, but maybe "$!" is the problem here? What happens when this is put into an explicit variable. E.g `cmd & cmd_pid=$!` and then using `$cmd_pid` later instead?

luukvbaal: Nah don't think so, the first `wait` works as expected. Any trapped signal stops executing of the `wait` command, this is according to spec. In posix sh simply adding a `wait` to the `WINCH` trap worked, in bash it no longer does apparently. Seems to suggest that the trap handler/child proccess we actually want to wait for run in different subshells now.

luukvbaal:
Updated to something that appears to work as well and seems somewhat cleaner.

jarun:
Thank you!

Issue #1590, opened 2023-02-17
mpv sixel/kitty support for preview


mia-0:
So I don’t use nnn but I’ve added kitty support to mpv a while ago and I’m using it in a [TUI moderation app for my file hoster 0x0.st](https://git.0x0.st/mia/0x0/src/branch/master#moderation-ui), so I thought you might be interested in this as an alternative to XEmbed.

In short, with the “buffered” options of its sixel/kitty video outputs, mpv can run “embedded” in other TUI applications to provide video previews. In my app, I’m also using it for images, whether from files or in-memory streams (such as PDF previews from MuPDF). It works well over SSH (provided there is enough network bandwidth—`ssh -C` helps) and does not require synchronization of terminal output (since its writes are atomic).

N-R-K:
As far as I see, [preview-tui](https://github.com/jarun/nnn/blob/master/plugins/preview-tui#L10) already supports kitty (but I don't see mpv mentioned anywhere). So my understanding is that you're asking to embed an mpv instance inside kitty to do the previews, is that correct?

In any case, this is going to be @luukvbaal's call.

mia-0:
Yeah, somewhat analogous to the XEmbed approach in preview-tabbed. So mpv does the rendering and you just tell it where, relative to the terminal window. It could replace image viewers and e.g. calls to ffmpegthumbnailer.

luukvbaal:
I won't be looking into this as I'm not a kitty user but a PR is welcome AFAIC. I guess https://github.com/mpv-player/mpv/commit/874e28f4a41a916bb567a882063dd2589e9234e1 can be used for reference.

mia-0:
It also supports sixels and there’s almost no difference between the two as far as usage goes (see e.g. [here](https://git.0x0.st/mia/0x0/src/branch/master/modui/mpvwidget.py)).

kuntau:
Ah... this requires `mpv v0.36`, no wonder I got error no kitty vo. Latest release still v0.35

mia-0:
Right, I forgot that wasn’t actually in any release yet. I’m on a distro that packages git snapshot (as per upstream recommendation)…

jarun:
I have added this to the tracker as no one is free to pick this up at the moment.
Closing accordingly.

PR #1749, opened 2023-10-15
preview-tui: mpv kitty and sixel support for img, gif & vid preview

abhinav3398:
adds sixel image, video and gif preview support when tmux is buit with `./configure --enable-sixel`
requirements for preview apart from sixel supported tmux:
- mpv
- img2preview
2023-10-15T00:31:13,007901883-06:00
2023-10-15T00:31:44,573211029-06:00

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -21,8 +21,12 @@
 #   - man
 #   - optional: bsdtar or atool for additional archive preview
 #   - optional: bat for code syntax highlighting
-#   - optional: ueberzug, kitty terminal, wezterm terminal, viu, catimg or chafa for images
-#   - optional: convert(ImageMagick) for playing gif preview (required for kitty image previews)
+#   - optional: ueberzug, kitty terminal, wezterm terminal, img2sixel, viu, catimg or chafa for images
+#   - optional: convert(ImageMagick) for playing gif preview (mandatory for kitty image previews)
+#   - optional: mpv for gif and video
+#       Also requires a terminal supporting the sixel (https://www.arewesixelyet.com/)
+#       or kitty (https://sw.kovidgoyal.net/kitty/graphics-protocol) video_output backends.
+#       Requires tmux compiled with `./configure --enable-sixel` if used.
 #   - optional: ffmpegthumbnailer for video thumbnails (https://github.com/dirkvdb/ffmpegthumbnailer)
 #   - optional: ffmpeg for audio thumbnails
 #   - optional: libreoffce for opendocument/officedocument preview
@@ -40,11 +44,6 @@
 #   - optional: pistol file viewer (https://github.com/doronbehar/pistol).
 #       1. install pistol
 #       2. set/export $NNN_PISTOL as 1
-#   - optional: sixel
-#       - terminal with sixel support. ex. foot, wezterm, iterm2, macterm, mintty, etc. (check https://www.arewesixelyet.com/)
-#       - tmux with sixel support (should be compiled with `./configure --enable-sixel`)
-#       dependencies: mpv, img2sixel
-#
 #
 # Usage:
 #   You need to set a NNN_FIFO path and a key for the plugin with NNN_PLUG,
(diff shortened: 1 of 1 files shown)


luukvbaal:
Please check `shellcheck` errors and style/documentation format. Also, which terminals does this target?

abhinav3398:
yes you are right. This is targeted towards foot, putty, wezterm, alacritty and other terminals wherein which the user wants to use tmux as multiplexer for nnn window split.

jarun:
@luukvbaal or @N-R-K please merge when this is ready.

luukvbaal:
Sure, I haven't had the chance to test this yet but it looks better now.

plugins/preview-tui
```diff
@@ -428,12 +434,27 @@ image_preview() {
...
         fifo_pager print_bin_info "$3" && return
     fi
     printf "%s" "$!" > "$PREVIEWPID"
 }
 
+video_preview() {
+    clear
+    exec >/dev/tty
+    vid_end=$(( RANDOM % 100 ))
+    vid_start=$(( RANDOM % vid_end ))
```
N-R-K, line 449: What happens if both of these are beyond the file duration? `--start` in mpv accepts a percent, see `man mpv`. I'd use it instead of random stuff. E.g `--start=50% --length=4` to play 4 seconds from the middle.

plugins/preview-tui
```diff
@@ -428,12 +434,27 @@ image_preview() {
...
+    if type mpv >/dev/null 2>&1 ; then
+        mpv --no-config --msg-level=all=warn,ao/alsa=error --vo=sixel --profile=sw-fast --loop-file --mute=yes --start="$vid_start"% --end="$vid_end"% "$3" &
```
N-R-K, line 451: I don't get the msg-level flag. Shouldn't we just disable terminal logging entirely? See the `--no-terminal` flag in mpv.

N-R-K (changes requested):

plugins/preview-tui
luukvbaal: mpv recommends the `--really-quiet` flag in the sixel section of their man page.

plugins/preview-tui
N-R-K: > Why cant it just start at the beginning?

I presume this is because a lot of videos might have some common opening part. E.g some opening of some series etc.

But thinking about movies and series, starting at 50% might not be a great idea then, since it might end up spoiling the story. Playing 4 seconds starting at 10% probably is a good default.

jarun:
@N-R-K please merge when ready.

plugins/preview-tui
```diff
@@ -40,6 +40,10 @@
 #   - optional: pistol file viewer (https://github.com/doronbehar/pistol).
 #       1. install pistol
 #       2. set/export $NNN_PISTOL as 1
+#   - optional: sixel
+#       - tmux sixel support (should be compiled with `./configure --enable-sixel`)
```
N-R-K, line 44: It should be obvious, but the terminal emulator needs to support sixel too. Should probably be mentioned.

plugins/preview-tui
```diff
@@ -428,12 +435,30 @@ image_preview() {
...
     printf "%s" "$!" > "$PREVIEWPID"
 }
 
+video_preview() {
+    clear
+    exec >/dev/tty
+    vid_start=10
+    if type mpv >/dev/null 2>&1 ; then
+        if [ "$4" -eq 1 ]; then # not gif
+            mpv --no-config --really-quiet --vo=sixel --profile=sw-fast --loop-file --mute=yes --start="$vid_start"% --length=4 "$3" &
```
N-R-K, line 452: No point in making `vid_start` a variable when it's used only once. I'd hardcode `--start="10%"`

N-R-K (commented):
The code looks okay, but I don't have my mpv built with sixel support so I didn't test.

luukvbaal:
Sorry for the delay, finally got around to looking into this. Looks alright but I think some more work is needed to implement this properly. 

We need to make sure we don't try to use the sixel backend on terminals that do not support it. I actually havent gotten the sixel backend to mpv work yet on my machine. It prints the raw sixel data when running tmux with sixel support, in a terminal that doesn't support sixel.
In terminals that supposedly do support sixel, I'm getting `[vo] Video output sixel not found!` when using the mpv sixel backend, even though `img2sixel` works fine. Anything else needed to make the mpv sixel backend work @abhinav3398?

#1590 also mentions the kitty backend, I tried adding support for that and that seems to work fine. I'll try to add a commit and merge this PR this weekend if I can get sixel to work...

luukvbaal:
OK mpv in the arch repo is not built with the sixel backend, installing mpv-full did the trick. What terminal are you using this on @abhinav3398?

I'm currently testing on WezTerm which supports both the sixel and kitty backends. Kitty backend works all right but the sixel backend is horribly slow. Is performance alright for you?

luukvbaal:
Added a commit to also support the mpv kitty backend. Also try to detect sixel support and better honor the `NNN_PREVIEWIMGPROG` and newly added `NNN_PREVIEWVIDEO` veriables.

abhinav3398:
> installing mpv-full did the trick. 

 That's exactly what I used.

> What terminal are you using this on @abhinav3398 

> I'm currently testing on WezTerm which supports both the sixel and kitty backends. Kitty backend works all right but the sixel backend is horribly slow. Is performance alright for you?

I'm using foot. works perfectly fine for me. Mpv used to display videos fine about 2 weeks ago. It just recently started lagging for me. But the lag is fine for prewing purposes. Haven't tried on other terminals yet but I'll try it out on other terminals and try to add a check for sixel support for the terminal after the weekend. 

Thanks for the help @luukvbaal as I'm busy during the weekend.

luukvbaal:
> It just recently started lagging for me. But the lag is fine for prewing purposes.

Yeah I suppose so. I doubt there is much we can do about performance anyways besides hoping it gets improved upstream ^^ Thanks for kicking off the mpv support, it's cool :) Maybe it's time to consider switching from st again, although I never end up doing so.

plugins/preview-tui
```diff
@@ -416,24 +433,37 @@ generate_preview() {
...
         fifo_pager print_bin_info "$3" && return
     fi
     printf "%s" "$!" > "$PREVIEWPID"
 }
 
+video_preview() {
+    clear
+    exec >/dev/tty
+    if [ -n "$NNN_PREVIEWVIDEO" ]; then
+        mpv --no-config --really-quiet --vo="$NNN_PREVIEWVIDEO" --profile=sw-fast --loop-file --mute=yes "$4" "$3" &
```
N-R-K, line 460: You can use `--ao=null` instead of `--mute=yes`. A null output should skip some unnecessary work compared to using an audio output and muting it.

N-R-K:
> I doubt there is much we can do about performance anyways besides hoping it gets improved upstream

Doesn't seem like there's any performance issues on sixel output being reported upstream (that's probably because no one uses it). Pinging @mia-0.

plugins/preview-tui
mia-0: Or just `--no-audio` to also stop mpv from using any audio tracks in the first place.

mia-0:
Performance is mostly bottlenecked by libsixel (the quantization step in particular; dithering options may affect it) and the terminal emulator. So there’s not much we can do about it on mpv’s side without replacing libsixel I’m afraid.

N-R-K:
> I'm fine to merge this now, but I'll test a bit more and wait for reviews one more time.

Assuming this is good to go then?

jarun:
Thanks guys!

PR #1755, opened 2023-11-03
preview-tabbed: Add alacritty as xembed client for text-based preview

cronyakatsuki:
I added alacritty as an option for xembed client for text-based preview.


jarun:
Thank you!

PR #1762, opened 2023-11-17
restore sort function pointers when restoring cfg

### LINKED ISSUE #1757: sorting state is shared between contexts
#### Environment details

- [x] Operating System: NixOS
- [x] Desktop Environment: none+i3
- [x] Terminal Emulator: kitty
- [x] Shell: fish
- [x] Program options used: aAERx
- [x] Configuration options set: see https://github.com/azuline/nixos/blob/master/home/nnn/default.nix#L24
- [x] Plugins are installed
- [ ] Issue exists on `nnn` master: pretty sure, i might be a few commits behind the latest commit

When I change contexts, the sort order of the current context is preserved in the second context. This preservation occurs when I set the sort order by hand (video 1), but also when the sort order is part of the `NNN_ORDER` envvar (video 2). Even if case 1 is intentional, I do not believe case 2 is, as leaving the `NNN_ORDER`-ed directory in the same context unsets the sort and does not preserve it.

There is another strange behavior: If i unset the sorting state in the 2nd context, go back to the first context, and then attempt to re-set the sort, nothing happens. It is as if my first attempt to re-set the sort actually performed the unset operation in the first context, and the second attempt actually did re-set the sort (though I'm not too sure what's really going on in the code).

Video 1: https://github.com/jarun/nnn/assets/51880422/e56ac3ac-94cd-41ef-89cf-31c56798c202
Video 2: https://github.com/jarun/nnn/assets/51880422/50536d6c-3f52-4618-8c0b-bd3c3b66a4eb

What is the desired behavior? I'm happy to open a pull request, but I don't want to implement the wrong behavior ^_^

azuline:
closes #1757 

This PR fixes a sorting synchronization issue.

The root cause of the bug in #1757 is that the sorting function `entrycmpfn` is not restored when changing contexts, causing `reverse` (and `version`) to be "global" in that setting them on one context caused them to be set on all contexts.

This PR fixes the bug by setting `entrycmpfn` and `namecmpfn` to the correct pointers whenever `cfg` is set.

---

simulation of state variables in the bugged scenario:

1. starting state
   - `cfg0.reverse = 0`, `entrycmpfn = &entrycmp`
2. set context 1 to `r`
   - `cfg0.reverse = 1`, `entrycmpfn = &reventrycmp`
3. create context 2. context 2 should default to `r`.
   - `cfg0.reverse = 1`, `cfg1.reverse = 1`, `entrycmpfn = &reventrycmp`
4. unset `r` in context 2.
   - `cfg0.reverse = 1`, `cfg1.reverse = 0`, `entrycmpfn = &entrycmp`
5. switch back to context 1. context 1 should not appear to be `r`.
   - `cfg0.reverse = 1`, `cfg1.reverse = 0`, `entrycmpfn = &entrycmp`
6. try to set `r` again. no-ops when it should do something.
   - `cfg0.reverse = 0`, `cfg1.reverse = 0`, `entrycmpfn = &entrycmp`

the "double unset" behavior is because the statusbar uses a function pointer comparison to print `R` and `V`. it does not use `cfg.reverse` or `cfg.version`. if the statusbar used `cfg.reverse`, we would see `R` in the statusbar, yet no reverse sort in the entrylist.

video :)

https://github.com/jarun/nnn/assets/51880422/5832c1c4-91d4-4460-842d-dad59c54a186

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4308,6 +4308,9 @@ static void savecurctx(char *path, char *curname, int nextctx)
 
 	tmpcfg.curctx = nextctx;
 	cfg = tmpcfg;
+	/* Restore the global function pointers alongside the cfg. */
+	entrycmpfn = cfg.reverse ? &reventrycmp : &entrycmp;
+	namecmpfn = cfg.version ? &xstrverscasecmp : &xstricmp;
 }
 
 #ifndef NOSSN
@@ -6248,11 +6251,8 @@ static int set_sort_flags(int r)
 			r = 'd';
 		}
 
-		if (cfg.version)
-			namecmpfn = &xstrverscasecmp;
-
-		if (cfg.reverse)
-			entrycmpfn = &reventrycmp;
+		entrycmpfn = cfg.reverse ? &reventrycmp : &entrycmp;
+		namecmpfn = cfg.version ? &xstrverscasecmp : &xstricmp;
 	} else if (r == CONTROL('T')) {
 		/* Cycling order: clear -> size -> time -> clear */
 		if (cfg.timeorder)
@@ -8018,6 +8018,9 @@ nochange:
 					lastname = g_ctx[r].c_name;
 
 					cfg = g_ctx[r].c_cfg;
+					/* Restore the global function pointers alongside the cfg. */
+					entrycmpfn = cfg.reverse ? &reventrycmp : &entrycmp;
+					namecmpfn = cfg.version ? &xstrverscasecmp : &xstricmp;
 
 					cfg.curctx = r;
 					setdirwatch();
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4276,6 +4276,18 @@ static void printent(const struct entry *ent, uint_t namecols, bool sel)
 		addch(ind);
 }
 
+/**
+ * Sets the global cfg variable and restores related state to match the new
+ * cfg.
+ */
+static void setcfg(settings newcfg)
+{
+	cfg = newcfg;
+	/* Synchronize the global function pointers to match the new cfg. */
+	entrycmpfn = cfg.reverse ? &reventrycmp : &entrycmp;
+	namecmpfn = cfg.version ? &xstrverscasecmp : &xstricmp;
+}
+
 static void savecurctx(char *path, char *curname, int nextctx)
 {
 	settings tmpcfg = cfg;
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -6248,11 +6251,8 @@ static int set_sort_flags(int r)
 			r = 'd';
 		}
 
-		if (cfg.version)
-			namecmpfn = &xstrverscasecmp;
-
-		if (cfg.reverse)
-			entrycmpfn = &reventrycmp;
+		entrycmpfn = cfg.reverse ? &reventrycmp : &entrycmp;
+		namecmpfn = cfg.version ? &xstrverscasecmp : &xstricmp;
```
azuline, line 6255: i'm not sure if this is necessary, since a new session is spawned with the last session's active directory and sorts (?).

jarun: It would be great if you can test and confirm whether it works without this change or not.

src/nnn.c
```diff
@@ -8018,6 +8018,9 @@ static bool browse(char *ipath, const char *session, int pkey)
 					lastname = g_ctx[r].c_name;
 
 					cfg = g_ctx[r].c_cfg;
+					/* Restore the global function pointers alongside the cfg. */
+					entrycmpfn = cfg.reverse ? &reventrycmp : &entrycmp;
```
jarun, line 8022: Since we are doing it in multiple places, can we have a function?

jarun (changes requested):

src/nnn.c
azuline: the behavior i observed should be fixed without this change, but i am not sure if there is a scenario where a new session results in an inconsistent state between cfg and the function pointers. i didnt fully model the session mechanics when debugging this.

it shouldn't cause any problems, since this logic is a NoOp if everything is in sync, but we can remove this change and wait to see if any bugged behavior is observed.

jarun: OK. In that case please have have the code (call tot he funciton) commented and add sufficient documentation.

azuline: the function i factored out does more than set the function pointers, it encapsulates the switching of the `cfg` global, as whenever `cfg` is switched, the function pointers should be synchronized to match the new `cfg`. i left a comment above these two lines expressing intent.

src/nnn.c
azuline: added a function to encapsulate setting `cfg =`, since we should synchronize global function pointers whenever we set `cfg =`

jarun:
Thank you!

PR #1770, opened 2023-11-21
When handling SEL_NEW pass `path` to `abspath` call

7ocb:
If `path` is not provided to `abspath`, later will do `getcwd`, and it's result will differ from `path`. This causes problem that when creating directory inside path reached with symlink, subsequent call to get_cwd_entry does not recognize newly created path as subpath of current path, thus not selecting newly created element.


jarun:
@N-R-K is this required?

PR #1771, opened 2023-11-21
When handling bookmark, use readlink, not realpath

7ocb:
This is to cd to path as it pointed by symlink, not to it's real path. Bookmarked directory may itself contain symlinks in path, which should be respected.

For example: if directory is physically in `/mnt/storage/some` and it's symlinked to `~/some` and directory `~/some/dir` added to bookmarks, it's expected that when following bookmark directory will be changed to `~/some/dir` (as in bookmark's link) not to `/mnt/storage/some/dir` (as dir real path).

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7053,7 +7053,7 @@ nochange:
 
 			pent = &pdents[cur];
 			if (!g_state.selbm || !(S_ISLNK(pent->mode) &&
-			                        realpath(pent->name, newpath) &&
+			                        readlink(pent->name, newpath, PATH_MAX) &&
 			                        xstrsncpy(path, lastdir, PATH_MAX)))
 				mkpath(path, pent->name, newpath);
 			g_state.selbm = 0;


src/nnn.c
```diff
@@ -7053,7 +7053,7 @@ static bool browse(char *ipath, const char *session, int pkey)
 
 			pent = &pdents[cur];
 			if (!g_state.selbm || !(S_ISLNK(pent->mode) &&
-			                        realpath(pent->name, newpath) &&
+			                        readlink(pent->name, newpath, PATH_MAX) &&
```
KlzXS, line 7056: Try our custom function `abspath()`. It was made for this exact reason.

7ocb: @KlzXS thanks! I already found that function, but it does not do resolving symlink's path. 

So I'm thinking about something like this:

```c
/* finds abspath of link pointed by filepath, taking cwd into account */
static char *bmtarget(const char *filepath, char *cwd, char *buf) {
   char target[PATH_MAX + 1];

   const int n = readlink(filepath, target, PATH_MAX);
   target[n] = 0;

   return abspath(target, cwd, buf);
}
```

7ocb:
If using readlink, the user may create direct links (i.e. links that point
to resolved physical path directly) and get desired behaviour. I.e. in this
case both behaviours achievable (i.e. by how link is created user may
decide short or long path to be followed).

But if using realpath, opposite behaviour os not possible, as even if link
points to "short" path, it still will follow long (real) path.

7ocb:
Maybe make this configurable via option?

PR #1773, opened 2023-11-22
When handling bookmark, use readlink, not realpath

7ocb:
This is to cd to path as it pointed by symlink, not to it's real path. Bookmarked directory may itself contain symlinks in path, which should be respected.

For example: if directory is physically in `/mnt/storage/some` and it's symlinked to `~/some` and directory `~/some/dir` added to bookmarks, it's reasonable to expect that when following bookmark directory will be changed to `~/some/dir` (as in bookmark's link) not to `/mnt/storage/some/dir` (as dir real path).

It's reasonable to expect this behavior because this follows the `cd` behavior (both shell `cd` and `nnn` changing directory) when changing directory into symlink.

Also, with behavior as in this PR user can have both bookmarks to "short" path (i.e. `~/some/dir` in example above) and long path (i.e. `/mnt/storage/some/dir`) - which path is actually bookmarked, that path will be followed. Original behavior disallows using bookmarks to go to "short" path.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1310,6 +1310,18 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
 	return resolved_path;
 }
 
+/* finds abspath of link pointed by filepath, taking cwd into account */
+static char *bmtarget(const char *filepath, char *cwd, char *buf) 
+{
+	char target[PATH_MAX + 1];
+	ssize_t n = readlink(filepath, target, PATH_MAX);
+	if (n != -1) {
+		target[n] = '\0';
+		return abspath(target, cwd, buf);
+	} 
+	return NULL;
+}
+
 /* wraps the argument in single quotes so it can be safely fed to shell */
 static bool shell_escape(char *output, size_t outlen, const char *s)
 {
@@ -7053,7 +7065,7 @@ nochange:
 
 			pent = &pdents[cur];
 			if (!g_state.selbm || !(S_ISLNK(pent->mode) &&
-			                        realpath(pent->name, newpath) &&
+			                        bmtarget(pent->name, path, newpath) &&
 			                        xstrsncpy(path, lastdir, PATH_MAX)))
 				mkpath(path, pent->name, newpath);
 			g_state.selbm = 0;


src/nnn.c
```diff
@@ -1310,6 +1310,16 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
 	return resolved_path;
 }
 
+/* finds abspath of link pointed by filepath, taking cwd into account */
+static char *bmtarget(const char *filepath, char *cwd, char *buf) {
```
N-R-K, line 1314: `{` goes to the next line.

src/nnn.c
```diff
@@ -1310,6 +1310,16 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
...
+   char target[PATH_MAX + 1];
```
N-R-K, line 1315: nnn uses TABs for indent.

src/nnn.c
```diff
@@ -1310,6 +1310,16 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
...
+
+   const int n = readlink(filepath, target, PATH_MAX);
```
N-R-K, line 1317: Drop the `const`, it's useless noise.

src/nnn.c
```diff
@@ -1310,6 +1310,16 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
...
+   target[n] = 0;
```
N-R-K, line 1318: Use `'\0'`.

And `readlink` isn't guaranteed to succeed and can return -1. That needs to be checked properly.

src/nnn.c
```diff
@@ -7053,7 +7063,7 @@ static bool browse(char *ipath, const char *session, int pkey)
 
 			pent = &pdents[cur];
 			if (!g_state.selbm || !(S_ISLNK(pent->mode) &&
-			                        realpath(pent->name, newpath) &&
+			                        bmtarget(pent->name, path, newpath) &&
```
N-R-K, line 7066: The function name says `bmtarget` but this will affect all navigation logic. Not just bookmarks.

N-R-K (changes requested):

src/nnn.c
7ocb: @N-R-K in this case only bookmarks as far as I can see (as per `g_state.selbm` check). Am I missing something?

src/nnn.c
7ocb: Done.

src/nnn.c
7ocb: Done.

src/nnn.c
7ocb: Done.

src/nnn.c
7ocb: Done.

src/nnn.c
```diff
@@ -1310,6 +1310,21 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
...
+static char *bmtarget(const char *filepath, char *cwd, char *buf) 
+{
+	char target[PATH_MAX + 1];
+
+	int n = readlink(filepath, target, PATH_MAX);
+	if (n != -1) {
+		target[n] = '\0';
+
+		return abspath(target, cwd, buf);
+	} else {
```
KlzXS, line 1323: Don't use else after return as per CI.

KlzXS (changes requested):

src/nnn.c
N-R-K: You're right. I didn't notice the `selbm` check.

src/nnn.c
7ocb: Done.

src/nnn.c
```diff
@@ -1310,6 +1310,21 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
...
+{
+	char target[PATH_MAX + 1];
+
+	int n = readlink(filepath, target, PATH_MAX);
+	if (n != -1) {
+		target[n] = '\0';
+
+		return abspath(target, cwd, buf);
+	} 
+
```
N-R-K, line 1324: This is a really small function. There's no real need for all these blank lines. I'd drop them all.

src/nnn.c
```diff
@@ -1310,6 +1310,21 @@ static char *abspath(const char *filepath, char *cwd, char *buf)
 	return resolved_path;
 }
 
+/* finds abspath of link pointed by filepath, taking cwd into account */
+static char *bmtarget(const char *filepath, char *cwd, char *buf) 
+{
+	char target[PATH_MAX + 1];
+
+	int n = readlink(filepath, target, PATH_MAX);
```
N-R-K, line 1318: Forgot to mention this earlier, but it's better to use `ssize_t` since that's what `readlink` actually returns (even though it doesn't really matter here).

N-R-K (commented):
Seems to work fine. A couple style nits.

src/nnn.c
N-R-K: TBH, I was never a fan of that warning. To me it seems perfectly fine to have `else` after `return` if all the branches end up returning.

(To be clear, the above is just a small rant, not a suggestion to change it back. Keep the PR as it currently is to keep the CI happy).

src/nnn.c
7ocb: Done.

src/nnn.c
7ocb: Done.

N-R-K (approved):

jarun:
Thank guys!

PR #1775, opened 2023-12-02
feat(preview-tui): allow image previews inside tmux on kitty terminal

JefeDavis:
Image previews can work correctly inside tmux while using kitty terminal by using `--transfer-mode=memory` and adding the following to your tmux config:
`set -g allow-passthrough on`

requires tmux 3.3a

i have added a clause to check if terminal is tmux and $KITTY_LISTEN_ON is defined, if so use the icat kitten inside tmux


jarun:
I don't see a chance of any regression since the code is isolated. Hence, merged.
Thank you!

PR #1782, opened 2023-12-14
fix nushell quitcd script: remove deprecated keywords

BeyondMagic:
`def-env` was removed, correct use now is `def --env`.
`let-env` was deprecated, correct use now is `$env.`.


jarun:
Thank you!

Issue #1765, opened 2023-11-19
NNN command line forward slash filter command won't allow other commands


slferris:
#### Environment details (Put `x` in the checkbox along with the information)

- [ x] Operating System: macOS Sonoma 14.1.1
- [x ] Desktop Environment: Mac mini M2 Pro, Apple Studio Display, Retina LCD
- [ x] Terminal Emulator: Studio Display, 
- [zsh ] Shell:
- [ ] Custom desktop opener (if applicable):
- [ x] Program options used: nnn plugins
- [ ] Configuration options set: nnn plugins installed along with nnn using homebrew 
- [x ] Plugins are installed in both nnn and oh-my-zsh
- [ x] Issue exists on `nnn` master

#### Exact steps to reproduce the issue
When I click on the command line in nnn it displays only a forward slash filter character to the left and on the line above to the right the characters str [/}, ic [:] or noic [:]. It does not allow me to use any other character besides / including the colon : to do other non-filter commands.

jarun:
You can't run commands in the filter prompt.
To get the commands prompt, press `]`. It's there in the help page.

N-R-K:
I think the issue he's describing is different. If you click on the statusbar **with the mouse** then pretty much all other keybindings stop working.

I was able to reproduce it on my end. But haven't yet looked into what's causing it.

jarun:
I see. It's been ages I have used the mouse with nnn. Reopened.

0xACE:
> then pretty much all other keybindings stop working.

even `ESC` doesnt exit filter mode? iirc it happened in my build too, it was related to the filtermode thing in that case...

N-R-K:
> even `ESC` doesnt exit filter mode?

It does, but the visual doesn't update. And sometimes clicking on the statusbar doesn't even enable filter mode. And sometimes it shows filter mode being activated but it actually isn't. Something seems to be going pretty wrong somewhere.

fre3nando:
By trying to reproduce the issue, I came across the following behaviour:
1. When activating filter mode using the mouse, and then disabling it by pressing the escape key, filter mode is from then on automatically activated when changing directories.
2. When activating filter mode using the forward-slash ('/') key, clicking on the prompt once doesn't de-activate it, rather requiring two successive clicks in order to turn filter mode off. In fact, if the escape key is pressed after clicking on the prompt once, the aforementioned problem arises again: changing directories triggers filter mode.
[...]

PR #1783, opened 2023-12-14
Setting PWD: set on use case, not on browse

7ocb:
This moves setting PWD environment variable closer to the places where child process (that needs PWD correctly set) is started instead of set it on browse.

@jarun this is alternative for https://github.com/jarun/nnn/pull/1778. This moves setting of `PWD` to "setexports" function which is being called before calling subshell and before spawning plugin.

I'm not 100% sure that this will cause `PWD` properly set for _every_ case (i.e. there are cases when "spawn" called without "setexports" being called before - for example when plugin is called as command handling, like `.cbcp`, or when running edit of selection), but this covers all use cases I initially wanted to fix. 

I also tried to put setting of `PWD` directly into `spawn` but that caused a lot of changes, as this function used in a lot of places, and in most of them this variable actually was not needed (like with `.cbcp`).

What do you think?

### DIFF
--- a/patches/restorepreview/mainline.diff
+++ b/patches/restorepreview/mainline.diff
@@ -238,7 +238,7 @@ index 0388b23c..66d3316a 100644
  		case SEL_PROMPT:
 +			if (g_state.previewer)
 +				notify_fifo(FALSE, TRUE);
- 			r = handle_cmd(sel, newpath);
+ 			r = handle_cmd(sel, path, newpath);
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

jarun:
We are good with this.

jarun:
Could you take a look at https://github.com/jarun/nnn/issues/1765?

7ocb:
I will.

PR #1803, opened 2024-01-23
Changed exa from grid to tree view (preview-tui)

manceras:
As explained in discussion #1801, changing exa view from grid (-G) to tree (-T).

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -368,7 +368,7 @@ preview_file() {
         elif exists tree; then
             fifo_pager tree --filelimit "$(find . -maxdepth 1 | wc -l)" -L 3 -C -F --dirsfirst --noreport
         elif exists exa; then
-            exa -G --group-directories-first --colour=always
+            exa -T --group-directories-first --colour=always
         elif exists eza; then # eza is a community fork of exa (exa is unmaintained)
             eza -G --group-directories-first --colour=always
         else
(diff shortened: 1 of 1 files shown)


plugins/preview-tui
```diff
@@ -368,7 +368,7 @@ preview_file() {
         elif exists tree; then
             fifo_pager tree --filelimit "$(find . -maxdepth 1 | wc -l)" -L 3 -C -F --dirsfirst --noreport
         elif exists exa; then
-            exa -G --group-directories-first --colour=always
+            exa -T --group-directories-first --colour=always
         elif exists eza; then # eza is a community fork of exa (exa is unmaintained)
             eza -G --group-directories-first --colour=always
```
azuline, line 373: can you also update eza so that it is consistent with exa?

```suggestion
            eza -T --group-directories-first --colour=always
```

azuline (commented):
to match `tree`, we'd also want the `-L 3` flag. `eza -T` hangs in my home directory unless i limit its printing depth. there is no substitution for `--filelimit` though, which would cause a very large subdirectory to prevent previewing the highlighted directory

manceras:
I added fifo_pager so we start viewing the top of the tree instead of the bottom. Is that better?

luukvbaal (approved):
> As exa has a feature to list a directory in a tree, wouldn't be more consistant showing it that way?

All the alternatives we have in `preview-tui` by now were just added as users have suggested them, consistency was never really taken into account. If we have multiple users who prefer the tree view, I'm not against changing the default. But keep in mind it always possible to just change the plugin locally to your desire.

> I added fifo_pager so we start previewing the top of the tree instead of the bottom. Is that better?

I think so yes.

jarun:
Thank you!

PR #1804, opened 2024-01-27
show linenumber when in SEL_JUMP

Darukutsu:
Related [discussion](https://github.com/jarun/nnn/discussions/1708).
I reworked my previous answer a bit. I think this should be part of nnn, since it's (IMO) hard to manually count down. Or at least include as native patch.

We could also show relative/absolute linenumber always(with minor tweaks), but that doesn't seem to be useful in any way yet.

linenumber

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -306,6 +306,7 @@ typedef struct entry {
 	uid_t uid; /* 4 bytes */
 	gid_t gid; /* 4 bytes */
 #endif
+        int ncount;
 } *pEntry;
 
 /* Selection marker */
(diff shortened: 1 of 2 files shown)


jarun:
Can you test this will very long filenames?

Also, what is the keybind to show the relative numbers?

src/nnn.c
```diff
@@ -306,6 +306,7 @@ typedef struct entry {
 	uid_t uid; /* 4 bytes */
 	gid_t gid; /* 4 bytes */
 #endif
+        int ncount;
```
jarun, line 309: This increases memory usage per file. Why do we need to store this and not calculate on the fly?

jarun (changes requested):

src/nnn.c
N-R-K: Simpler approach in https://github.com/jarun/nnn/pull/1808 which doesn't require storing counts.

jarun: Already merged.

Issue #1105, opened 2021-07-11
zstd


syntezoid:
**Is your feature request related to a problem? Please describe.**
I can't open .zst .tar.zst archives.
I am currently using Zstandard for archiving files, but the distribution I am using, Arch, provides applications in the .tar.zst archive format and I need to check their contents or extract them.

**Describe the solution you'd like**
Add somehow Zstandard, but I'm not sure how. As I see atool and bsdtar doesn't support .zst. I found patool as a more recent incarnation of atool, but it doesn't support Zstandard either.

**Describe alternatives you've considered**
If not via atool, patool, bsdtar is some nnn plugin that already does this?

**Additional context**

**Consider contributing**

Please consider contributing the feature back to `nnn`.

Feel free to discuss in the ToDo list thread. We are more than happy to help.

jarun:
> Add somehow Zstandard, but I'm not sure how. As I see atool and bsdtar doesn't support .zst. I found patool as a more recent incarnation of atool, but it doesn't support Zstandard either.

Doesn't look like Zstandard is widely adopted yet. `nnn` uses `atool`/`bsdtar` internally so we can't do much natively. However, you can always write 2 plugins - one to archive and one to extract zst and assign 2 keybinds to them if that format is so important to you. I think you can also merge them into 1 i.e. if the file extension has .zst extract instead of archiving,

eugene55:
I don't like to comment on closed issues but Arch's bsdtar in fact supports zst format. To support it from nnn side though, you just have to configure NNN_ARCHIVE environment variable to include zst extension to the list. After that zst will be treat as any other archive.

jarun:
Link please.

eugene55:
https://man.archlinux.org/man/bsdtar.1

jarun:
It appears the FreeBSD version does too - https://www.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&format=html

Stig124:
Debian does too  (but Ubuntu doesn't):

https://manpages.debian.org/buster/libarchive-tools/bsdtar.1.en.html

PR #1805, opened 2024-01-27
add zip plugin

Darukutsu:
I don't like that built-in zip includes absolute paths by default. I also miss zstd as option. This could do probably any 
compression. You can also password protect archive.

image

Related issues [#1105](https://github.com/jarun/nnn/issues/1105).

### DIFF
--- a/plugins/zip
+++ b/plugins/zip
@@ -128,7 +128,6 @@ zip_selection_individualy(){
       xargs -0 -I{} tar -cvf "{}.tar" "{}" < "$selection"
       compress_tar 'printf "%s" "{}"' "$selection"
     fi
-    password_protect 'basename "{%}"' "$selection"
   else
     if [ "$alg" = "zip" ]; then
       xargs -0 -I{} sh -c '
@@ -145,7 +144,6 @@ zip_selection_individualy(){
       ' < "$selection"
       compress_tar 'basename "{}"' "$selection"
     fi
-    password_protect 'basename "{%}"' "$selection"
   fi
   password_protect 'basename "{%}"' "$selection"
 }
--- a/plugins/zip
+++ b/plugins/zip
@@ -93,7 +93,7 @@ password_protect(){ # filename
 }
 
 compress_tar(){ # filename
-  if [ $alg != tar ] && [ $alg != zip ]; then
+  if [ "$alg" != tar ] && [ "$alg" != zip ]; then
     cd "$working_dir" || exit # some needs this
     # 'keep' is due to zstd not following standard
     if $alg -"$compression" --keep "$1.tar"; then
@@ -107,7 +107,7 @@ zip_selection_individualy(){
     for abspath in $(tr '\0' '\n' < "$selection"); do
       if [ "$alg" = "zip" ]; then
         # dont need .zip
-        zip -r -$compression "$abspath.zip" "$abspath"
+        zip -r -"$compression" "$abspath.zip" "$abspath"
       else
         tar -cvf "$abspath.tar" "$abspath"
         compress_tar "$abspath"
@@ -129,7 +129,7 @@ zip_selection_individualy(){
       file=$(basename "$abspath")
 
       if [ "$alg" = "zip" ]; then
-        cd "$ppath" && zip -r -$compression "$working_dir/$file.zip" "$file"
+        cd "$ppath" && zip -r -"$compression" "$working_dir/$file.zip" "$file"
       else
         tar -C "$ppath" -cvf "$file.tar" "$file"
         compress_tar "$file"
@@ -142,7 +142,7 @@ zip_selection_individualy(){
 zip_selection(){
   if [ "$include_abs" = "y" ] || [ "$include_abs" = "Y" ]; then
     if [ "$alg" = "zip" ]; then
-      xargs -0 zip -r -$compression "$zipname.zip" < "$selection"
+      xargs -0 zip -r -"$compression" "$zipname.zip" < "$selection"
     else
       xargs -0 tar -cvf "$zipname.tar" < "$selection"
     fi
@@ -152,7 +152,7 @@ zip_selection(){
       file=$(basename "$abspath")
 
       if [ "$alg" = "zip" ]; then
-        cd "$ppath" && zip -r -$compression "$working_dir/$zipname.zip" "$file"
+        cd "$ppath" && zip -r -"$compression" "$working_dir/$zipname.zip" "$file"
       else
         tar -C "$ppath" -rvf "$zipname.tar" "$file"
       fi
@@ -168,7 +168,7 @@ zip_hovered(){
   zipname="$1"
 
   if [ "$alg" = "zip" ]; then
-    zip -r -$compression "$zipname.zip" "$1"
+    zip -r -"$compression" "$zipname.zip" "$1"
   else
     tar -cvf "$zipname.tar" "$1"
     compress_tar "$zipname"
--- a/plugins/zip
+++ b/plugins/zip
@@ -1,6 +1,6 @@
 #!/usr/bin/env sh
 
-#set -x
+# set -x
 # Better zip solution, compress selection - archive or batch(deterministic way '$filename.zip')
 # or hover individual compression
 #
@@ -13,11 +13,26 @@
 default_alg=gzip
 default_compression=5
 # this will be determined from filename if not provided
-default_archivename="archive$(date +%Y%m%d_%H%M%S)"
+default_archivename="archive_$(date +%Y%m%d_%H%M%S)"
 
 working_dir="$PWD"
+selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
 
-choose_alg(){
+choose_alg() {
+  cat << EOF
+# Reference: https://linuxreviews.org/Comparison_of_Compression_Algorithms
+1. bzip2
+2. gzip
+3. lzip
+4. lzma
+5. tar (archive only, no compression)
+6. zip
+7. zstd
+8. xz
+EOF
+
+  printf "Pick algorithm [default - %s]: " "$default_alg"
+  read -r alg
   case "$alg" in
     "1"|"bzip2") alg=bzip2
       ;;
@@ -39,53 +54,14 @@ choose_alg(){
       ;;
     *)
       alg="$default_alg"
-      printf "picking default - '%s'\n" "$alg"
+      printf "unrecognized input, picking default - '%s'\n" "$alg" >&2
       ;;
   esac
 
   command -v "$alg" >/dev/null 2>&1 || { printf "'%s' not installed, cancelling...\n" "$alg" >&2; exit 1; }
 }
 
-print_alg(){
-  printf "
-#https://linuxreviews.org/Comparison_of_Compression_Algorithms
-
-1. bzip2
-2. gzip
-3. lzip
-4. lzma
-5. tar (archive only, no compression)
-6. zip (windows friend :D)
-7. zstd
-8. xz
-  \n\n"
-}
-
-print_alg
-printf "Pick algorithm[%s - default]: " "$default_alg"
-read -r alg
-choose_alg
-
-if ! [ "$alg" = "tar" ]; then
-  # TODO: zstd has more levels
-  printf "Choose compression level(except tar)[1.3.5.7.9, %s - default]: " "$default_compression"
-  read -r compression
-  case "$compression" in
-    [0123456789])
-      #do nothing
-      ;;
-    *)
-      compression="$default_compression"
-      ;;
-  esac
-fi
-
-
-
-
-selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
-
-password_protect(){
+password_protect() {
   if [ "$encrypt" = "y" ] || [ "$encrypt" = "Y" ]; then
     cd "$working_dir" || exit # some needs this
     if [ "$2" ]; then
@@ -100,7 +76,7 @@ password_protect(){
   fi
 }
 
-compress_tar(){
+compress_tar() {
   if [ "$alg" != tar ] && [ "$alg" != zip ]; then
     cd "$working_dir" || exit # some needs this
 
@@ -120,7 +96,7 @@ compress_tar(){
   fi
 }
 
-zip_selection_individualy(){
+zip_selection_individualy() {
   if [ "$include_abs" = "y" ] || [ "$include_abs" = "Y" ]; then
     if [ "$alg" = "zip" ]; then
       xargs -0 -I{} sh -c '
@@ -154,7 +130,7 @@ zip_selection_individualy(){
   password_protect 'basename "{%}"' "$selection"
 }
 
-zip_selection(){
+zip_selection() {
   if [ "$include_abs" = "y" ] || [ "$include_abs" = "Y" ]; then
     if [ "$alg" = "zip" ]; then
       xargs -0 zip -r -"$compression" "$zipname.zip" < "$selection"
@@ -181,7 +157,7 @@ zip_selection(){
   password_protect "$zipname"
 }
 
-zip_hovered(){
+zip_hovered() {
   #zipname=${default_archivename:-"$1"}
   zipname="$1"
 
@@ -194,6 +170,28 @@ zip_hovered(){
   password_protect "$zipname"
 }
 
+### main
+
+choose_alg
+
+if ! [ "$alg" = "tar" ]; then
+  compression_limit=9
+  [ "$alg" = "zstd" ] && compression_limit=19
+  printf "Choose compression level 1-%d [default - %s]: " "$compression_limit" "$default_compression"
+  read -r compression
+  case "$compression" in
+    [0123456789]*)
+      #do nothing
+      ;;
+    "")
+      compression="$default_compression"
+      ;;
+    *)
+      compression="$default_compression"
+      printf "unrecognized input, picking default - %s\n" "$default_compression" >&2
+      ;;
+  esac
+fi
 
 printf "Encrypt[y/N]: "
 read -r encrypt
@@ -211,7 +209,7 @@ if [ -s "$selection" ]; then
   case "$individual" in
     "n"|"N")
       individual=""
-      printf "Archive name['%s' - default]: " "$default_archivename"
+      printf "Archive name [default - '%s']: " "$default_archivename"
       read -r zipname
       zipname=${zipname:-$default_archivename}
       ;;
--- a/plugins/zip
+++ b/plugins/zip
@@ -147,6 +147,7 @@ zip_selection_individualy(){
     fi
     password_protect 'basename "{%}"' "$selection"
   fi
+  password_protect 'basename "{%}"' "$selection"
 }
 
 zip_selection(){
--- a/plugins/zip
+++ b/plugins/zip
@@ -15,7 +15,7 @@ default_compression=5
 # this will be determined from filename if not provided
 default_archivename="archive$(date +%Y%m%d_%H%M%S)"
 
-working_dir="$2"
+working_dir="$PWD"
 
 choose_alg(){
   case "$alg" in
@@ -70,12 +70,12 @@ if ! [ "$alg" = "tar" ]; then
   # TODO: zstd has more levels
   printf "Choose compression level(except tar)[1.3.5.7.9, %s - default]: " "$default_compression"
   read -r compression
-  case $compression in
+  case "$compression" in
     [0123456789])
       #do nothing
       ;;
     *)
-      compression=$default_compression
+      compression="$default_compression"
       ;;
   esac
 fi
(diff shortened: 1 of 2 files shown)


plugins/zip
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env sh
```
jarun, line 1: Please add the details in the plugin readme table.

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+      ;;
+    "3"|"lzip") alg=lzip
+      ;;
+    "4"|"lzma") alg=lzma
+      ;;
+    "5"|"tar") alg=tar
+      ;;
+    "6"|"zip") alg=zip
+      ;;
+    "7"|"zstd") alg=zstd
```
jarun, line 34: What happens when the user chooses a compression which is not available in the system?

jarun (changes requested):

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+
+zip_selection_individualy(){
+  if [ "$include_abs" = "y" ] || [ "$include_abs" = "Y" ]; then
+    for abspath in $(sed 's#\x0# #g' < "$selection"); do
+      if [ "$alg" = "zip" ]; then
+        # dont need .zip
+        zip -r -$compression "$abspath.zip" "$abspath"
+      else
+        tar -cvf "$abspath.tar" "$abspath"
+        compress_tar "$abspath"
```
N-R-K, line 111: You can generally just pipe the output of `tar` over to the compressor program without needing to create any intermediate file.

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+# Shell: POSIX compliant
+# Author: Darukutsu
+
+# Change these for better suited defaults
+default_alg=zstd
+default_compression=5
+# this will be determined from filename if not provided
+#default_archivename=archive
+
+working_dir=$(pwd)
```
N-R-K, line 18: Use either `$PWD` or `$2` (see https://github.com/jarun/nnn/tree/master/plugins#access-level-of-plugins)

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+
+
+
+
+selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
+
+password_protect(){ # filename
+  if [ "$encrypt" = "y" ] || [ "$encrypt" = "Y" ]; then
+    cd "$working_dir" || exit # some needs this
+    find . -maxdepth 1 -type f \( -name "$1.zip" -o -name "$1.tar*" \) \! -name "*.gpg" -exec gpg -c "{}" \; -exec rm "{}" \;
```
N-R-K, line 89: Why not just pass the full filename instead of doing this. Also `gpg` might accept `stdin` as well. So you might be able to just pipe it in and avoid the intermediate file once again.

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+    #if [ "$alg" = "zip" ]; then
+    #  xargs -0 -I{} zip -r -$compression "{}.zip" "{}" < "$selection"
+    #elif [ "$alg" = "tar" ]; then
+    #  xargs -0 -I{} tar -cvf "{}.tar" "{}" < "$selection"
+    #else
+    #  xargs -0 -I{} tar -I"$alg -$compression" -cvf "{}.tar.$suffix" "{}" < "$selection"
+    #fi
+  else
+    for abspath in $(sed 's#\x0# #g' < "$selection"); do
+      ppath=$(echo "$abspath" | sed 's#\(/.*/\).*#\1#')
```
N-R-K, line 126: It's not generally safe to echo a filepath since it might contain special chars. Use `printf` instead.

Also what exactly is this sed doing?

N-R-K (changes requested):

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+
+zip_selection(){
+  if [ "$include_abs" = "y" ] || [ "$include_abs" = "Y" ]; then
+    if [ "$alg" = "zip" ]; then
+      xargs -0 zip -r -$compression "$zipname.zip" < "$selection"
+    else
+      xargs -0 tar -cvf "$zipname.tar" < "$selection"
+    fi
+  else
+    for abspath in $(sed 's#\x0# #g' < "$selection"); do
```
KlzXS, line 148: All other plugins use `tr '\0' '\n'` to get rid of '\0'. Let's keep it consistent.

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+      ppath=$(echo "$abspath" | sed 's#\(/.*/\).*#\1#')
```
KlzXS, line 149: Use `dirname` to get the containing directory of a file. Keep in mind that it doesn't include a trailing `/` if that matters to you.

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+      file=$(echo "$abspath" | sed 's#/.*/\(.*\)#\1#')
```
KlzXS, line 150: Use `basename` to get the filename.

plugins/zip
```diff
@@ -0,0 +1,212 @@
...
+      printf "Archive name['%s' - default]: " "$default_archivename"
+      read -r zipname
+      zipname=${zipname:-$default_archivename}
+      ;;
+    "y"|"Y")
+      #leave this empty
+      ;;
+  esac
+
+  if test "$individual"; then
```
KlzXS, line 203: `[` is an alias for `test`. Keep it consistent and use `if [ "$individual" ]; then`.

KlzXS (changes requested):
I only marked a single occurrence of each issue. Please change all of them.

plugins/zip
KlzXS: My comments below address this in entirety.

Darukutsu: You can also zip files individualy with absolute path or not. Let's say for some reason you have 2 different directories which you want compress separately or maybe 2 big files and you want to have A.zip B.zip. **I needed to pass this in loop instead of `xargs` for comment mentioned above and when you don't want absolute path you have to be in that path... `zip -j` is not very good, breaks structure, when you want compress dirs within dirs.** Because we want selection separated `sed` removes `\0` character since `read -0` is not POSIX. Anyway changed it for `tr`. I didn't want to repeat although I have. Because of highlighted text I didn't used `tar -I` and piping gpg. Don't know any better

plugins/zip
Darukutsu: Yeah I know. For some reason shellcheck recommended this but not now... weird :D.

plugins/zip
```diff
@@ -0,0 +1,214 @@
...
+    find . -maxdepth 1 -type f \( -name "$1.zip" -o -name "$1.tar*" \) \! -name "*.gpg" -exec gpg -c "{}" \; -exec rm "{}" \;
+  fi
+}
+
+compress_tar(){ # filename
+  if [ $alg != tar ] && [ $alg != zip ]; then
+    cd "$working_dir" || exit # some needs this
+    # 'keep' is due to zstd not following standard
+    if $alg -"$compression" --keep "$1.tar"; then
+      rm "$1.tar"
```
KlzXS, line 100: CI is flagging this. Use `"$alg"`.

KlzXS (changes requested):

plugins/zip
Darukutsu: There are more variables should I put `""` for all of them?
Also is it ok to amend commit? Don't know what's proper way when fixing issues like that.

KlzXS: Please do, I missed those. It'll keep complaining until all of them are quoted. In general, if it has `$` it goes between `""`.

You can add more commits, amend, force push, whatever you like. PRs usually just get squashed anyway.

KlzXS: You know how every rule has an exception? This is it for the quoting rule: `$(tr '\0' '\n' < "$selection")` shouldn't be quoted. We want that to split at whitespace.

Darukutsu: what if filename has white space?

Darukutsu: anyway `selection` should only contain selected file nothing else right?

Darukutsu: Ah ok... for loop will break

Darukutsu: Ah, for loop isn't great for this since it will break if filename contains `\n` or white space.

Darukutsu: well I will have to come with other solution or maybe I'm not right and already talking bullshits :D.

plugins/zip
```diff
@@ -0,0 +1,214 @@
...
+      ;;
+    "") alg="$default_alg"
+      ;;
+    *)
+      alg="$default_alg"
+      printf "picking default - '%s'" "$alg"
+      ;;
+  esac
+
+  which "$alg" >/dev/null 2>&1 || (printf "'%s' not installed, cancelling..." "$alg" && exit 1)
```
N-R-K, line 46: 0. Use `command -v` instead of `which`. It should be more portable.
0. Switch from `(..)` to `{..}` to group commands.
0. It should exit regardless of whether `printf` failed or not. So change `&&` to `;`
0. printf doesn't add newline unlike echo. Use `printf "...\n" ...` instead. Also I think this should go to standard error instead. So add a `>&2` redirect.

plugins/zip
N-R-K: Just do it with `xargs -0` and add a `alg_to_extension()` function or something to add the extension.

KlzXS: Yeah, it would be best if you can get it to work with `xargs -0`. Just keep in mind that you can't run functions inside `xargs`. That's a bash extension.

Darukutsu:
Well boys, will this work in other shells? I tried to preserve original behavior, but it's kinda weird. I mean it works for me :D. Maybe I should include some tests.

I could possibly rewrite it in way in which I specify file extensions in `choose_alg()` and then use native `tar --use-compress-program=COMMAND`.
Or let user manually everytime specify extension in input like `myarchive.tar.gz`.

It works now in a way that users need to specify only name without extensions.

plugins/zip
```diff
@@ -0,0 +1,226 @@
...
+
+selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
+
+password_protect(){
+  if [ "$encrypt" = "y" ] || [ "$encrypt" = "Y" ]; then
+    cd "$working_dir" || exit # some needs this
+    if [ "$2" ]; then
+      xargs -0 -I{} sh -c '
+      filename=$('"$1"') &&
+      find . -maxdepth 1 -type f \( -name "$filename.zip" -o -name "$filename.tar*" \) \! -name "*.gpg" -exec gpg -c "{}" \; -exec rm "{}" \;
```
N-R-K, line 94: Won't this just end up encrypting unrelated files if it happened to start with the same `$filename`?

N-R-K (changes requested):

plugins/zip
Darukutsu: oh yeah it will. I forgot to add encryption to individual selection so it wasn't encrypting at all. My bad.

jarun:
@N-R-K and @KlzXS is this ready to merge?

KlzXS:
@jarun There are still some outstanding comments from me that should be addressed. CI is currently failing; I'd say we probably just silence it in this case. I myself haven't tried the plugin out; will probably do that at some point.

When those issues are address and if we confirm that it does work, we'll be ready to merge. Personally, I'd like to see the script simplified a bit, there's a lot of commands nested inside of `xargs ... sh -c '' ` constructs, but I haven't had time to look into it and see if something can be done about that.

Darukutsu:
There are some issues. I finally made somewhat testing script, so now it will be easier to fix them. I will provide script with fix tomorrow.

Darukutsu:
When testing using script comment line `229`. You still need to make sure output is ok. I tested using this.
```
  # Clear selection
  printf "-" > "$NNN_PIPE"
```
<details>
  <summary>test script</summary>
  
  ```
  #!/usr/bin/env sh

g=$(tput setaf 2)
n=$(tput sgr0)

command=~/.config/nnn/plugins/zip

#tar
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rY\rY\rY\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 1${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rY\rY\rN\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 2${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rY\rN\rN\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 3${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rY\rN\rY\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 4${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rN\rY\rN\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 5${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rN\rY\rY\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 6${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rN\rN\rY\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 7${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "tar\rN\rN\rN\r\r"))& $command && printf "\n%s\n\n" "${g}tar passed 8${n}"

#tar.gz
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rY\rY\rY\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 1${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rY\rY\rN\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 2${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rY\rN\rN\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 3${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rY\rN\rY\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 4${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rN\rY\rN\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 5${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rN\rY\rY\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 6${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rN\rN\rY\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 7${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "gzip\r\rN\rN\rN\r\r"))& $command && printf "\n%s\n\n" "${g}gzip passed 8${n}"

#zip
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rY\rY\rY\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 1${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rY\rY\rN\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 2${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rY\rN\rN\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 3${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rY\rN\rY\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 4${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rN\rY\rN\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 5${n}" ; read
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rN\rY\rY\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 6${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rN\rN\rY\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 7${n}" ; read; read
(sleep 1 ; xdotool type --delay 100 $(printf "zip\r\rN\rN\rN\r\r"))& $command && printf "\n%s\n\n" "${g}zip passed 8${n}"
```
<\details>

**Comment by Darukutsu** [comment]
Should I fix anything else?

**Comment by N-R-K** [comment]
Did some basic tests (didn't test encryption). Couple remarks:

0. Why does compressing a single file produce a tar? I'd expect it to be compressed directly without the tar archive.
0. Should we really hardcode 5 as default compression level? Each tool comes with it's own default and IMO we should let the tool decide for itself what the default should be instead of hardcoding 5.

**Comment by N-R-K** [comment]
Also, I'm really not a fan of this type of thing that's going on: `-o -name "$filename.tar*"`. All this would go away if we just used stdin/out instead of making intermediate files. Here's an example of how to do stdin/out pipe to do `tar.gz` on selection:

```console
$ xargs -0 tar -cvf - < "$selection" | gzip -c -9 > archive.tar.gz
```

Or if you want to remove the $pwd from the paths:

```console
sed -z "s|^$PWD/||g" < "$selection" | xargs -0 tar -cvf - | zstd -c -19 > archive.tar.zst
```

Darukutsu:
> ```
> sed -z "s|^$PWD/||g" < "$selection" | xargs -0 tar -cvf - | zstd -c -19 > archive.tar.zst
> ```

does this work because after trying it i got something like this which is not what we want.

```
# run under .config/nnn
" tar.vim version v32
" Browsing tarfile /home/daru/.config/nnn/archive.tar.zst
" Select a file with cursor and press ENTER

plugins/x2sel
plugins/xdgdefault
plugins/zfs-snapdir
sessions/
sessions/@                              # my implementation does this
home/daru/.config/pgcli/                #pgcli/
home/daru/.config/pgcli/config          #pgcli/config
home/daru/.config/pgcli/log             #pgcli/log
```
also doing this piping to gpg and thus creating another `elseif` entry isn't very something... or maybe gpg should be separate script then, or don't include it at all.

> Should we really hardcode 5 as default compression level? Each tool comes with it's own default and IMO we should let the tool decide for itself what the default should be instead of hardcoding 5.

As a workaround to prevent including more if-statements we could do `"default_compression=-verbose` since expanding it will make `--verbose` and it's unharmful behavior and should be same across all compression algorithms.

jarun:
@Darukutsu CI is failing, can you check?

Also, this is open for a while, if this doesn't move, I'll close it.

Darukutsu:
```
 ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.
 ```
  just like @KlzXS said
 >CI is currently failing; I'd say we probably just silence it in this case.
 
 I couldn't come up with better solution. It's pretty much working for me(haven't spotted any issue, when using/[testing](https://github.com/jarun/nnn/pull/1805#issuecomment-1937017504)), but that's not what means "production ready". Feel free to close. In either case thank you guys for perfectioning this script.

KlzXS:
The main issue I have with the script is that it's a nightmare to maintain with all those nested substitutions, extracting and constructing names and a lot of `if`s that run in parallel across multiple functions.  Unfortunately, I don't have a solution to those issues.

If you're fine with such code merge it in.

@Darukutsu Thank you for giving it a shot. Personally, I'd have written this in something like Python as bash isn't really great at dealing with complex logic. You get ugly things like that `find` for removing files after encrypting them. Ideally you should know the name of the temporary file you just wrote to, but alas, you don't.

We don't really have examples of non-shell scripts, but a plugin can be any old executable file, we just happen to mostly do it in `sh` as that's available everywhere.

KlzXS:
I think it's best that we close this for now. I've added it as a task in the Tracker. Someone can give it a shot in the future.

My suggestions for those attempting this task:
- Don't use POSIX `sh`. This is a complicated task and it's very limited in its functionality. At the very least use `bash` so that you can execute a function inside `xargs`. That should simplify the script significantly and make it easier to modify and maintain. Functions are your friends.
- You can try another scripting language but try to use something widespread and popular. I've mentioned Python above.
- If you try to support encryption, see if you can support both symmetric and asymmetric `gpg`. See `gpge` and `gpgd` for an example.
- If using `gpg`, see if you can remove the intermediate files directly, without having to resort to `find`ing them after `gpg`'s done its job.
- In general, try to make the script as performant as possible. It should spend most of its runtime doing useful work and as little as possible executing helper functions.
- `nnn`'s core archiving solution has support for `atool`. See if you can use that to simplify the task.
- Consider if you want to support actions for extracting (Extract Here, Extract to..., Extract to `archive_name/`) available in many archiving solutions. Splitting that off into a separate plugin would be a good choice.

Now that I've written this it sounds more like a whole separate tool instead of a "simple" plugin. Make of that what you will.

All in all, this is functionality that would be nice to have in `nnn` in some form. Thanks to @Darukutsu for bringing it up.

PR #1806, opened 2024-02-01
quitcd: fix old bug and feat. for modular export for nushell

BeyondMagic:
+ now users can import as a module instead of just copying and paste with `use nnn.nu`
+ also refactor everything with internal commands, and add commentary, and remove unnecessary `--force` in `rm`.
+ remove having to import `nnn` environment variable `NNN_TMPFILE`, now it only launches for the command.

image

### DIFF
--- a/misc/quitcd/quitcd.nu
+++ b/misc/quitcd/quitcd.nu
@@ -1,19 +1,37 @@
-# The behaviour is set to cd on quit (nnn checks if NNN_TMPFILE is set)
-let cfgHome = ($env | default $"($env.HOME)/.config" XDG_CONFIG_HOME | get XDG_CONFIG_HOME)
-$env.NNN_TMPFILE = $"($cfgHome)/nnn/.lastd"
+# Run nnn with dynamic changing directory to the environment.
+#
+# $env.XDG_CONFIG_HOME sets the home folder for `nnn` folder and its $env.NNN_TMPFILE variable.
+# See manual NNN(1) for more information.
+#
+# Import module using `use quitcd.nu n` to have `n` command in your context.
+export def --env n [
+	...args : string # Extra flags to launch nnn with.
+	--selective = false # Change directory only when exiting via ^G.
+] -> nothing {
 
-def --env n [...x] {
-  # Launch nnn. Add desired flags after `^nnn`, ex: `^nnn -eda ($x | str join)`
-  ^nnn ($x | str join)
-  let newpath = (
-    if ($env.NNN_TMPFILE | path exists) {
-      # FIXME: fails if path contains single-quote
-      let newpath = (open $env.NNN_TMPFILE | parse "cd '{nnnpath}'").0.nnnpath
-      ^rm -f $env.NNN_TMPFILE
-      echo $newpath
-    } else {
-      pwd
-    }
-  )
-  cd $newpath
+	# The behaviour is set to cd on quit (nnn checks if $env.NNN_TMPFILE is set).
+	# Hard-coded to its respective behaviour in `nnn` source-code.
+	let nnn_tmpfile = $env
+		| default '~/.config/' 'XDG_CONFIG_HOME'
+		| get 'XDG_CONFIG_HOME'
+		| path join 'nnn/.lastd'
+		| path expand
+
+	# Launch nnn. Add desired flags after `^nnn`, ex: `^nnn -eda ...$args`,
+	# or make an alias `alias n = n -eda`.
+	if $selective {
+		^nnn ...$args
+	} else {
+		NNN_TMPFILE=$nnn_tmpfile ^nnn ...$args
+	}
+
+	if ($nnn_tmpfile | path exists) {
+		# Remove <cd '> from the first part of the string and the last single quote <'>.
+		# Fix post-processing of nnn's given path that escapes its single quotes with POSIX syntax.
+		let path = open $nnn_tmpfile | str substring 4..-1 | str replace --all `'\''` `'`
+
+		^rm $nnn_tmpfile
+
+		cd $path
+	}
 }


misc/quitcd/quitcd.nu
```diff
@@ -1,19 +1,28 @@
...
+export def --env n [
+	...flags : string # Extra flags to launch nnn with.
+] -> nothing {
+
+	# Launch nnn. Add desired flags after `^nnn`, ex: `^nnn -eda ($x | str join)`
+	^nnn ...$flags
+
+	if ($env.NNN_TMPFILE | path exists) {
+		# Remove <cd '> from the first part of the string and the last single quote <'>.
+		let path = (open $env.NNN_TMPFILE | str substring 4..-1)
```
N-R-K, line 22: Did you actually test this on a buggy path? e.g one that contains `'`? https://github.com/jarun/nnn/pull/1737#issuecomment-1735351494

N-R-K (changes requested):

N-R-K:
> instead of just copying and paste

There was no need for copy paste. You can just `source` the file (I'm assuming nu shell has `source`).

BeyondMagic:
> There was no need for copy paste. You can just `source` the file (I'm assuming nu shell has `source`).

Nushell aims to focus on modularity, by replacing `source` alike-files with modules we can hide and use whenever we want.

misc/quitcd/quitcd.nu
BeyondMagic: Fixed in 43a44d50c57c36e9428e4ad295db5610482de38f.

N-R-K:
> with modules we can hide and use whenever we want.

Not sure if this sense or not since this isn't some "library" meant to be used wherever. But then again, I don't use nushell, so I could be wrong.

- - -

@J-Kappes @mmai could you guys review and/or test this pull request?

BeyondMagic:
> > with modules we can hide and use whenever we want.
> 
> Not sure if this makes sense or not since this isn't some "library" meant to be used wherever.

You can start by reading about [modules here](www.nushell.sh/book/modules.html).

And you can test by downloading nushell and then opening the terminal, typing out the following commands step by step:
1. `nu`
2. `use <this repo>/misc/quitcd/quitcd.nu`
3. `n`

Then go to another directory and exit. There we go.

Think of this by also enabling users to build upon this simple function.

N-R-K:
> @N-R-K is this good to go?

Waiting for other nushell users to test/review.

> And you can test by downloading nushell [...]

Sorry, not interested in downloading/running pre-built binaries and also not interested in compiling [thousands of dependencies](https://github.com/nushell/nushell/blob/main/Cargo.lock) just to test something I have no intention to use.

jarun:
@BeyondMagic please squash to a single commit.

@mmai can you please review?

BeyondMagic:
@jarun done.

misc/quitcd/quitcd.nu
```diff
@@ -1,19 +1,27 @@
...
-      # FIXME: fails if path contains single-quote
-      let newpath = (open $env.NNN_TMPFILE | parse "cd '{nnnpath}'").0.nnnpath
-      ^rm -f $env.NNN_TMPFILE
-      echo $newpath
-    } else {
-      pwd
-    }
-  )
-  cd $newpath
+	let config_home = if ($env.XDG_CONFIG_HOME | is-empty) {
```
mmai, line 6: There is an error when XDG_CONFIG_HOME is not defined :
```
Error: nu::shell::name_not_found

  × Name not found
    ╭─[/home/henri/.config/nushell/nnn.nu:35:1]
 35 │
 36 │     let config_home = if ($env.XDG_CONFIG_HOME | is-empty) {
    ·                                ───────┬───────
    ·                                       ╰── did you mean 'XDG_CONFIG_DIRS'?
 37 │         $'($env.HOME)/.config/'
    ╰────
```
this is why there is a default value in the original code : ` | default $"($env.HOME)/.config" XDG_CONFIG_HOME`

misc/quitcd/quitcd.nu
```diff
@@ -1,19 +1,27 @@
...
+
+	# Launch nnn. Add desired flags after `^nnn`, ex: `^nnn -eda ($x | str join)`
+	NNN_TMPFILE=$nnn_tmpfile ^nnn ...$flags
+
+	if ($nnn_tmpfile | path exists) {
+		# Remove <cd '> from the first part of the string and the last single quote <'>.
+		# Fix post-processing of nnn with path, escaping single quotes with POSIX syntax.
+		let path = (open $nnn_tmpfile | str substring 4..-1 | str replace --all `'\''` `'`)
+
+		rm $nnn_tmpfile
```
mmai, line 23: without the --force option in  `rm $nnn_tmpfile` I have now a confirmation prompt when I exit nnn : `rm: remove '/home/henri/.config/nnn/.lastd'?  [Y/N]:`

mmai (commented):
Hello. Thank you for the modular export. I saw two problems with the config_home setting and the temp file removal.

misc/quitcd/quitcd.nu
BeyondMagic: I see, although the default environment value without any info at all is in my opinion bad, can we resolve it by adding an optional flag and then using the default value there?

misc/quitcd/quitcd.nu
BeyondMagic: Can't replicate, which version are you running? If the same, seems like a **nushell** platform's incompatibility issue assuming that you don't use Linux, given `rm` information, the force flag should only regard itself to ignore errors, not make it non-interactive like the **GNU coreutils** `rm`:

```
rm (Nushell 0.89.0)
...
  -f, --force - suppress error when no file
...
  Delete a file, ignoring 'file not found' errors
  > rm --force file.txt
```

But assuming also that that file can be deleted before it runs that line, adding the force flag may not cause any harm. If we're on the same version, I can add the flag and, hopefully, review it later.

mmai: I use nushell 0.87 on NixOS.

The confirmation prompt occurs when you have an alias  `alias rm = rm -i`  in your nushell config (which is common).
You are right the "--force" option on the nushell `rm` command don't change anything, you have to use the external `rm`  with the caret prefix :   `^rm -f`.

misc/quitcd/quitcd.nu
mmai: Yes that should work.

misc/quitcd/quitcd.nu
BeyondMagic: Try `^rm` without the force flag.

misc/quitcd/quitcd.nu
BeyondMagic: Updated the branch, you can try it now.

N-R-K: > can we resolve it by adding an optional flag and then using the default value there?

The scripts need to follow `nnn`'s logic on how to determine the config dir. Making it a user facing flag makes no sense because if the user provides some path then the script will use one directory whereas `nnn` will use another.

Please try to keep the logic same as other cd-quit scripts.

misc/quitcd/quitcd.nu
mmai: Yes. `^rm $nnn_tmpfile` works.

misc/quitcd/quitcd.nu
mmai: @N-R-K @BeyondMagic in this case we should stick to the original code in order to have HOME be the fallback to XDG_CONFIG_HOME as in the other quitcd scripts.

BeyondMagic: Please, read the comments and logic of the code, the logic itself has not changed. The flag is aliased for setting its respective environment variable; meaning, using the same default value whereas if given other value, it will use that instead.

In simpler terms, hard-coded string constant can be set to other value, and will raise readable and understandable errors when fail to use any value.

N-R-K: > whereas if given other value, it will use that instead

This is precisely what I'm saying should not be done.

As far as I understand, if the user passes, let's say `/tmp/` as `--config_home` then the script will try to use that dir. But `nnn` will not be using it.

BeyondMagic: As far as I'm aware, `quitcd` scripts use only the `nnn`'s environment variable `NNN_TMPFILE`, and as the manual says:

```
       NNN_TMPFILE: always cd on quit and write the  command  in
       the file specified.

           export NNN_TMPFILE='/tmp/.lastd'
```

This script then builds that environment variable from the value of the flag, then calls externally `nnn` with the same environment variable, which means, `nnn` can only use the value of that environment variable.

Any other environment variables remain unchanged.

If it'd better, the flag instead could be `--tmp-file` and would specify the value of `NNN_TMPFILE` directly with the default value being `~/.config/nnn/.lastd`. What are your thoughts?

N-R-K: If you look at the other scripts, you'll notice that they allow for cd-on-quit selectively on `^G` via removing `export`:

https://github.com/jarun/nnn/blob/5b05c8b9b1e6fe5549680465e05aa89f45d265df/misc/quitcd/quitcd.bash_sh_zsh#L9-L14

This works because if `NNN_TMPFILE` is not set, pressing `^G` inside `nnn` will write the location to the "hardcoded" `.lastd` file inside the config dir. And thus the script's config dir and nnn's config dirs must match.

If you *don't* allow for selective cd-on-quit (the current state of the PR) then you can add `--tmp-file` flag to set `NNN_TMPFILE` directly. If you allow selective cd-on-quit then the config dir logic needs to 1:1 follow `nnn`'s logic.

Which one to do is up to you. But I'd advise keeping things similar to other scripts because not everyone might want to always cd-on-quit. Some people might prefer to cd-on-quit explicitly via `^G`.

BeyondMagic: > This works because if NNN_TMPFILE is not set, pressing ^G inside nnn will write the location to the "hardcoded" .lastd file inside the config dir

Ok, I don't think the Nushell's quitcd script ever supported picker mode then.

I fixed it now, and instead of a `tmp-file` flag, it uses again the previous hard-coded path builder, like the other scripts. However, I added a `selective` flag so that users don't have to edit the code of the module itself to enable this behaviour.

```nu
-> quitcd n --help
Run nnn with dynamic changing directory to the environment.

$env.XDG_CONFIG_HOME sets the home folder for `nnn` folder and its $env.NNN_TMPFILE variable.
See manual NNN(1) for more information.

Import module using `use quitcd.nu n` to have `n` command in your context.

Usage:
  > n {flags} ...(args) 

Flags:
  --selective <Boolean> - Whether only change directory upon exit of ^G. (default: false)
  -h, --help - Display the help message for this command

Parameters:
  ...args <string>: Extra flags to launch nnn with.
```

BeyondMagic: @N-R-K, I believe this resolves the issue completely. E.j. of what it is now doing:
```nu
$ use quitcd.nu n
$ alias n = n --selective true
$ n
# go to other folder and exit with ^G
# context is now in that folder.
$ n
# go to new folder and exit with q.
# context is not in this new folder.
```

```nu
$ use quitcd.nu n
$ n
# go to other folder and exit with ^G
# context is now in that folder.
$ n
# go to new folder and exit with q.
# context is now in this new folder.
```

misc/quitcd/quitcd.nu
```diff
@@ -1,19 +1,37 @@
...
-$env.NNN_TMPFILE = $"($cfgHome)/nnn/.lastd"
+# Run nnn with dynamic changing directory to the environment.
+#
+# $env.XDG_CONFIG_HOME sets the home folder for `nnn` folder and its $env.NNN_TMPFILE variable.
+# See manual NNN(1) for more information.
+#
+# Import module using `use quitcd.nu n` to have `n` command in your context.
+export def --env n [
+	...args : string # Extra flags to launch nnn with.
+	--selective = false # Whether only change directory upon exit of ^G.
```
N-R-K, line 9:
```suggestion
	--selective = false # Change directory only when exiting via ^G.
```

N-R-K (approved):
This looks better. One small nit on the wording of `--selective` flag.

BeyondMagic:
> This looks better. One small nit on the wording of `--selective` flag.

Fixed, had a few problems pulling from the main branch, but now the log history is fine.

jarun:
Thanks guys!

PR #1811, opened 2024-02-03
Add option to `rm -rf` irrespective of trash

62832:
Heya. I was slightly annoyed by the fact that using `NNN_TRASH` completely disabled the ability to `rm -rf` within nnn, so I wanted to add a separate option for `^X` that would just use `rm -rf` regardless of the trash backend being used.

Hope this is alright.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -1551,12 +1551,12 @@ static void xdelay(useconds_t delay)
 	usleep(delay);
 }
 
-static char confirm_force(bool selection)
+static char confirm_force(bool selection, bool use_trash)
 {
 	char str[64];
 
 	snprintf(str, 64, messages[MSG_FORCE_RM],
-		 g_state.trash ? utils[UTIL_GIO_TRASH] + 4 : utils[UTIL_RM_RF],
+		 use_trash ? utils[UTIL_GIO_TRASH] + 4 : utils[UTIL_RM_RF],
 		 (selection ? "selected" : "hovered"));
 
 	int r = get_input(str);
@@ -1565,7 +1565,7 @@ static char confirm_force(bool selection)
 		return '\0'; /* cancel */
 	if (r == 'y' || r == 'Y')
 		return 'f'; /* forceful for rm */
-	return (g_state.trash ? '\0' : 'i'); /* interactive for rm */
+	return (use_trash ? '\0' : 'i'); /* interactive for rm */
 }
 
 /* Writes buflen char(s) from buf to a file */
@@ -2548,13 +2548,13 @@ static void opstr(char *buf, char *op)
 	snprintf(buf, CMD_LEN_MAX, "xargs -0 sh -c '%s \"$0\" \"$@\" . < /dev/tty' < %s", op, selpath);
 }
 
-static bool rmmulstr(char *buf)
+static bool rmmulstr(char *buf, bool use_trash)
 {
-	char r = confirm_force(TRUE);
+	char r = confirm_force(TRUE, use_trash);
 	if (!r)
 		return FALSE;
 
-	if (!g_state.trash)
+	if (!use_trash)
 		snprintf(buf, CMD_LEN_MAX, "xargs -0 sh -c 'rm -%cr \"$0\" \"$@\" < /dev/tty' < %s",
 			 r, selpath);
 	else
@@ -2565,13 +2565,13 @@ static bool rmmulstr(char *buf)
 }
 
 /* Returns TRUE if file is removed, else FALSE */
-static bool xrm(char * const fpath)
+static bool xrm(char * const fpath, bool use_trash)
 {
-	char r = confirm_force(FALSE);
+	char r = confirm_force(FALSE, use_trash);
 	if (!r)
 		return FALSE;
 
-	if (!g_state.trash) {
+	if (!use_trash) {
 		char rm_opts[] = "-ir";
 
 		rm_opts[1] = r;
@@ -2702,8 +2702,8 @@ static bool cpmvrm_selection(enum action sel, char *path)
 			return FALSE;
 		}
 		break;
-	default: /* SEL_RM */
-		if (!rmmulstr(g_buf)) {
+	default: /* SEL_TRASH, SEL_RM_ONLY */
+		if (!rmmulstr(g_buf, g_state.trash && sel == SEL_TRASH)) {
 			printmsg(messages[MSG_CANCEL]);
 			return FALSE;
 		}
@@ -5175,8 +5175,8 @@ static void show_help(const char *path)
 	          "ca  Select all%14A  Invert sel\n"
 	       "9p ^P  Copy here%12w ^W  Cp/mv sel as\n"
 	       "9v ^V  Move here%15E  Edit sel list\n"
-	       "9x ^X  Delete%18S  Listed sel size\n"
-		"aEsc  Send to FIFO\n"
+	       "9x ^X  Delete or trash%09S  Listed sel size\n"
+		  "cX  Delete (rm -rf)%07Esc  Send to FIFO\n"
 	"0\n"
 	"1MISC\n"
 	      "8Alt ;  Select plugin%11=  Launch app\n"
@@ -7651,9 +7651,10 @@ nochange:
 		case SEL_CP: // fallthrough
 		case SEL_MV: // fallthrough
 		case SEL_CPMVAS: // fallthrough
-		case SEL_RM:
+		case SEL_TRASH: // fallthrough
+		case SEL_RM_ONLY:
 		{
-			if (sel == SEL_RM) {
+			if (sel == SEL_TRASH || sel == SEL_RM_ONLY) {
 				r = get_cur_or_sel();
 				if (!r) {
 					statusbar(path);
@@ -7664,7 +7665,7 @@ nochange:
 					tmp = (listpath && xstrcmp(path, listpath) == 0)
 					      ? listroot : path;
 					mkpath(tmp, pdents[cur].name, newpath);
-					if (!xrm(newpath))
+					if (!xrm(newpath, g_state.trash && sel == SEL_TRASH))
 						continue;
 
 					xrmfromsel(tmp, newpath);
--- a/src/nnn.h
+++ b/src/nnn.h
@@ -96,7 +96,8 @@ enum action {
 	SEL_CP,
 	SEL_MV,
 	SEL_CPMVAS,
-	SEL_RM,
+	SEL_TRASH,
+	SEL_RM_ONLY,
 	SEL_OPENWITH,
 	SEL_NEW,
 	SEL_RENAME,
@@ -239,8 +240,9 @@ static struct key bindings[] = {
 	{ 'w',            SEL_CPMVAS },
 	{ CONTROL('W'),   SEL_CPMVAS },
 	/* Delete from selection buffer */
-	{ 'x',            SEL_RM },
-	{ CONTROL('X'),   SEL_RM },
+	{ 'x',            SEL_TRASH },
+	{ CONTROL('X'),   SEL_TRASH },
+	{ 'X',            SEL_RM_ONLY },
 	/* Open in a custom application */
 	{ 'o',            SEL_OPENWITH },
 	{ CONTROL('O'),   SEL_OPENWITH },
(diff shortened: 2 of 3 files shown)


KlzXS (changes requested):
I think this is a fine addition. It always seemed redundant to me that there were two keybinds for the exact same thing bound to the exact same key.

Though it is a bit dangerous for the people who've grown accustomed to using `^X` for deleting files with trash enabled. The last *change* (there have been additions since, but those wouldn't affect existing workflows) in keybinds was in April 2022 and that was much less impactful than this would be.

src/nnn.h
```diff
@@ -240,7 +241,7 @@ static struct key bindings[] = {
 	{ CONTROL('W'),   SEL_CPMVAS },
 	/* Delete from selection buffer */
 	{ 'x',            SEL_RM },
-	{ CONTROL('X'),   SEL_RM },
+	{ CONTROL('X'),   SEL_RMONLY },
```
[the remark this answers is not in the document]
jarun: I don't understand what is "only". Not here, not in the help.

src/nnn.c
```diff
@@ -2815,7 +2815,7 @@ static char *get_archive_cmd(const char *archive)
 static void archive_selection(const char *cmd, const char *archive)
 {
 	char *buf = malloc((xstrlen(patterns[P_ARCHIVE_CMD]) + xstrlen(cmd) + xstrlen(archive)
-	                   + xstrlen(selpath)) * sizeof(char));
```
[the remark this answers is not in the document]
62832: I believe I *am* using tabs there, it's just that those were previously spaces. I use an alias to tab-indent anything in this case since I typically use 4 spaces to indent in my own projects.

src/nnn.h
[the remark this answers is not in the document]
62832: In this case "only" is supposed to mean "only use `rm -rf`, ignoring `NNN_TRASH`". I can see why it would be confusing, however.

src/nnn.c
```diff
@@ -1551,12 +1551,12 @@ static void xdelay(useconds_t delay)
 	usleep(delay);
 }
 
-static char confirm_force(bool selection)
+static char confirm_force(bool selection, enum action rmopt)
```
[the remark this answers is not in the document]
62832: By that point, `sel` can only be either `SEL_RM` or `SEL_RMONLY`, so I assumed this would be okay.

src/nnn.c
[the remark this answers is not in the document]
KlzXS: Oh, you're right. I swapped the colors around. In any case, revert it back to what it was. That's how it's done everywhere else.

src/nnn.c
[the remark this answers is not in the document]
KlzXS: I was expecting something along the lines of `bool use_trash`. It's kind of weird passing down the action to this mundane function this deep in the call stack, after we've pretty much used it up for its intended purpose.

It's just personal preference, we can keep it like this. It's just that I don't remember something like this ever coming up, so it got my attention.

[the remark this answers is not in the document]
62832: Yeah, that makes more sense, actually. I'll go ahead and change it as such.

src/nnn.h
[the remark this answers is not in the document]
KlzXS: Maybe `SEL_RMNOTRASH`, implying that `SEL_RM` does or can use trash like it was up until now? I've been trying to come up with something, but everything is either somewhat vague or too verbose.

N-R-K:
You can accomplish the same thing with a simple plugin. Bind it to `x` and then call it with <kbd>alt+x</kbd>. Don't think we need to have this builtin, and I *especially* don't think we should bind it to `^x` and change existing bind.

KlzXS:
On the other hand, I think this is a useful bultin. It's a bit odd that you can't specify to permanently delete a file if you configure trash. Maybe even going as far as to say that most file managers have such a feature. So, delegating it to a one-liner plugin seems a bit off.

I do agree about the keybind tough. I don't expect it to affect anybody realistically, but who knows. This would have been perfect back before keybinds were stabilized, but apparently no one thought of that at the time.

KlzXS:
...unless we do something like with batch rename where we call a plugin if it's found and otherwise do the builtin thing.

It's a bit convoluted, but would at the same time keep the existing behavior for most users while allowing something extra for those willing to opt in. Also, there was some talk recently about supporting more trash managers. I didn't follow that through, but I imagine giving them a script they can edit would allow them better compatibility than just building all those options directly in.

What do you think?

N-R-K:
> do something like with batch rename where we call a plugin if it's found and otherwise do the builtin thing.

That might be a good idea. Would allow people to use other trash utilities or add special logic to skip trash and rm directly and what not.

KlzXS:
@62832 the plugin in question is `.nmv` and the corresponding symbol in code is `UTIL_NMV`. In case you want to take a look at that.

src/nnn.c
```diff
@@ -5175,8 +5175,8 @@ static void show_help(const char *path)
 	          "ca  Select all%14A  Invert sel\n"
 	       "9p ^P  Copy here%12w ^W  Cp/mv sel as\n"
 	       "9v ^V  Move here%15E  Edit sel list\n"
-	       "9x ^X  Delete%18S  Listed sel size\n"
-		"aEsc  Send to FIFO\n"
+	          "cx  Delete%18S  Listed sel size\n"
+		 "b^X  Delete (rm-only)%06Esc  Send to FIFO\n"
```
jarun, line 5179: Does ^X result in `rm` or `rm -rf`?

jarun (changes requested):

src/nnn.c
jarun: Also, when the user presses `^X` does the next prompt explicitly state that `rm` is going to be invoked?

62832: Yes to both of those.

62832: `^X` results in `rm -rf` as it would if there was no trash backend being used. The prompt does also explicitly state that `rm -rf` will be invoked.

62832: The `(rm-only)` clause could probably be amended to just say `Delete (rm -rf)`.

jarun: Yes, that should be the correct help documentation.

jarun:
@KlzXS there is one problem. For users who use the `type-to-nav` mode, `^X` is the natural way to trash. In my case, I still use `rm -rf` but someone used to trash files would inadvertently remove the files.

So I am inclined to agree with @N-R-K on this aspect.

However, I think we can have a best of both the worlds. We can add a program option `-X` (capital-X) which would force `rm -rf` with `^X` when a different trash is configured.

What do you think?

@62832 hold on till we conclude this discussion.

src/nnn.c
```diff
@@ -1565,7 +1565,7 @@ static char confirm_force(bool selection)
 		return '\0'; /* cancel */
 	if (r == 'y' || r == 'Y')
 		return 'f'; /* forceful for rm */
-	return (g_state.trash ? '\0' : 'i'); /* interactive for rm */
+	return (g_state.trash && use_trash ? '\0' : 'i'); /* interactive for rm */
```
N-R-K, line 1568: I don't like how you need to check the same thing everywhere in multiple places. Just make the caller check for `g_state.trash` and make these functions only use the `use_trash` argument.

src/nnn.h
```diff
@@ -240,7 +241,7 @@ static struct key bindings[] = {
 	{ CONTROL('W'),   SEL_CPMVAS },
 	/* Delete from selection buffer */
 	{ 'x',            SEL_RM },
-	{ CONTROL('X'),   SEL_RM },
+	{ CONTROL('X'),   SEL_RMNOTRASH },
```
N-R-K, line 244: Capital x is free. You can use that instead of changing `ctrl+x`.

N-R-K (changes requested):

src/nnn.h
jarun: @N-R-K yes, that is also a good idea. That way we don't need to change the behaviour of the current keybinds.

jarun:
@62832 I am fine with a new keybind `X` for `rm -rf` always.

Also, please squash all the commits into one for easy maintenance.

62832:
Should be done now, having addressed @N-R-K's concerns and switched to using SHIFT+X as the bind for `rm -rf`. Ready when you are.

src/nnn.h
```diff
@@ -241,6 +242,7 @@ static struct key bindings[] = {
 	/* Delete from selection buffer */
 	{ 'x',            SEL_RM },
 	{ CONTROL('X'),   SEL_RM },
+	{ 'X',            SEL_RMNOTRASH },
```
jarun, line 245: Can you please rename the above 2 to SEL_TRASH and this one to SEL_RM_FIXED?

jarun (changes requested):

src/nnn.h
N-R-K: `SEL_RM_ONLY` sounds better to me. And for the existing one, I think `SEL_RM` is fine. But `SEL_TRASH` also works if you prefer that.

src/nnn.c
```diff
@@ -5176,7 +5176,7 @@ static void show_help(const char *path)
 	       "9p ^P  Copy here%12w ^W  Cp/mv sel as\n"
 	       "9v ^V  Move here%15E  Edit sel list\n"
 	       "9x ^X  Delete%18S  Listed sel size\n"
```
N-R-K, line 5178: Maybe this should be changed to "Delete or trash"? A bit weird seeing both the bindings with "Delete" description.

N-R-K (approved):
Small documentation nit. Code seems ok to me.

KlzXS (approved):
If we are happy with the new names of the actions and the description in the help this should be good to go.

Also a reminder that we should update the wiki after this is merged.

jarun:
We are good! I'll take care of the Wiki update. Thanks guys!

62832:
Is that failing `ubuntu-patches` test going to be a problem?

jarun:
./patches/check-patches.sh fails in CI. @62832 please have a look.

```
./patches/check-patches.sh
7
O_COLEMAK=1 O_GITSTATUS=0 O_NAMEFIRST=0 O_RESTOREPREVIEW=0 [FAILED]
8
O_COLEMAK=0 O_GITSTATUS=1 O_NAMEFIRST=0 O_RESTOREPREVIEW=0 [SUCCESS]
9
O_COLEMAK=1 O_GITSTATUS=1 O_NAMEFIRST=0 O_RESTOREPREVIEW=0 [FAILED]
10
O_COLEMAK=0 O_GITSTATUS=0 O_NAMEFIRST=1 O_RESTOREPREVIEW=0 [SUCCESS]
11
O_COLEMAK=1 O_GITSTATUS=0 O_NAMEFIRST=1 O_RESTOREPREVIEW=0 [FAILED]
12
O_COLEMAK=0 O_GITSTATUS=1 O_NAMEFIRST=1 O_RESTOREPREVIEW=0 [SUCCESS]
13
O_COLEMAK=1 O_GITSTATUS=1 O_NAMEFIRST=1 O_RESTOREPREVIEW=0 [FAILED]
14
O_COLEMAK=0 O_GITSTATUS=0 O_NAMEFIRST=0 O_RESTOREPREVIEW=1 [SUCCESS]
15
O_COLEMAK=1 O_GITSTATUS=0 O_NAMEFIRST=0 O_RESTOREPREVIEW=1 [FAILED]
16
O_COLEMAK=0 O_GITSTATUS=1 O_NAMEFIRST=0 O_RESTOREPREVIEW=1 [SUCCESS]
17
O_COLEMAK=1 O_GITSTATUS=1 O_NAMEFIRST=0 O_RESTOREPREVIEW=1 [FAILED]
18
O_COLEMAK=0 O_GITSTATUS=0 O_NAMEFIRST=1 O_RESTOREPREVIEW=1 [SUCCESS]
19
O_COLEMAK=1 O_GITSTATUS=0 O_NAMEFIRST=1 O_RESTOREPREVIEW=1 [FAILED]
20
O_COLEMAK=0 O_GITSTATUS=1 O_NAMEFIRST=1 O_RESTOREPREVIEW=1 [SUCCESS]
21
O_COLEMAK=1 O_GITSTATUS=1 O_NAMEFIRST=1 O_RESTOREPREVIEW=1 [FAILED]
22
make: *** [Makefile:326: checkpatches] Error 1
23
Error: Process completed with exit code 2.
```

62832:
@jarun Should be sorted now, conflicts have been resolved with the Colemak patch.

jarun:
Re-ran CI.

jarun (approved):

jarun:
Thank you!

PR #1816, opened 2024-02-09
fix: use a more generic way to print NUL with awk

mdhooge:
As described here: https://github.com/jarun/nnn/discussions/1815

### DIFF
--- a/plugins/mimelist
+++ b/plugins/mimelist
@@ -12,4 +12,4 @@ printf "mime (e.g., video/audio/image): "
 read -r mime
 
 printf "%s" "+l" > "$NNN_PIPE"
-find . | file -if- | grep "$mime" | awk -F: '{printf "%s\0", $1}' > "$NNN_PIPE"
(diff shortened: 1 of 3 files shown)


plugins/mimelist
```diff
@@ -12,4 +12,8 @@ printf "mime (e.g., video/audio/image): "
 read -r mime
 
 printf "%s" "+l" > "$NNN_PIPE"
-find . | file -if- | grep "$mime" | awk -F: '{printf "%s\0", $1}' > "$NNN_PIPE"
+if type mimetype >/dev/null 2>&1; then
```
jarun, line 15: Please add yourself as the author, all file and mimelist as dependencies (refer to other plugins).

Also add the deps in the readme.

jarun (changes requested):

plugins/mimelist
mdhooge: I checked package name for `mimetype`, it is different for debian & ArchLinux so I opted for a reference to the PERL module…

jarun:
Fixes #1816.

PR #1821, opened 2024-02-13
Added dashes to progress versions of cp and mv

KlzXS:
I did a grep for cp and mv, but that didn't pick up these two which messed up the size of the strings below.


N-R-K (approved):

jarun:
Thank you!

PR #1831, opened 2024-02-25
Support more trash utility

rennsax:
This PR add more options for the environment variable `NNN_TRASH`.

Currently, `NNN_TRASH` only support two external utilities: `1` for `trash-put` in [andreafrancia/trash-cli](https://github.com/andreafrancia/trash-cli), `2` for `gio trash`. These two doesn't seem to act perfectly in macOS.

In macOS, there are trash utilities like [sindresorhus/macos-trash](https://github.com/sindresorhus/macos-trash) and [trash](https://hasseg.org/trash/), which just move files/folders to the native trash bin. The two utilities can be installed with:
```sh
brew install macos-trash
brew install trash
```
These two both provide the command interface `trash`. They are useful when an user prefer to use the native trash folder.

Thanks for your review.

### DIFF
--- a/nnn.1
+++ b/nnn.1
@@ -562,7 +562,7 @@ separated by \fI;\fR:
 \fBNNN_TRASH:\fR trash (instead of \fIrm -rf\fR) files to desktop Trash.
 .Bd -literal
     export NNN_TRASH=n
-    # n=1: trash-cli, n=2: gio trash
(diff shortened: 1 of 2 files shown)


N-R-K (changes requested):
I don't think we want to keep hardcoding more and more trash utils. Instead making `NNN_TRASH=3` call out to some specific plugin is more flexible (see https://github.com/jarun/nnn/pull/1811#issuecomment-1925903856).

PR #1834, opened 2024-02-29
preview-tabbed: show sxiv/nsxiv in thumbnail mode when watching Pictures folder

TheUtopian:
It'll use `xdg-user-dir PICTURES` to get Pictures folder location and if user enters the folder it'll  use sxiv/nsxiv to show thumbnails instead of nnn.

### DIFF
--- a/plugins/preview-tabbed
+++ b/plugins/preview-tabbed
@@ -69,6 +69,9 @@ else
     echo "No xembed term found" >&2
 fi
 
+if type xdg-user-dir >/dev/null 2>&1 ; then
+    PICTURES_DIR=$(xdg-user-dir PICTURES)
+fi
 
 term_nuke () {
     # $1 -> $XID, $2 -> $FILE
@@ -177,7 +180,17 @@ previewer_loop () {
                 fi
                 ;;
             inode/directory)
-                $TERMINAL "$XID" -e nnn "$FILE" &
+                if [[ -n $PICTURES_DIR && "$FILE" == "$PICTURES_DIR"* ]] ; then
+                    if type sxiv >/dev/null 2>&1 ; then
+                        sxiv -te "$XID" "$FILE" &
+                    elif type nsxiv >/dev/null 2>&1 ; then
+                        nsxiv -te "$XID" "$FILE" &
+                    else
+                        $TERMINAL "$XID" -e nnn "$FILE" &
+                    fi
+                else
+                    $TERMINAL "$XID" -e nnn "$FILE" &
+                fi
                 ;;
             text/*)
                 if [ -x "$NUKE" ] ; then


TheUtopian:
I think it's very convenient to show thumbnails in Pictures instead of just folders and I think this feature will only improve experience for most users. Users who save images in different folders also will be able to easily modify the code to include their favorite folders. For others who don't use sxiv or nsxiv it'll act as before and if they wouldn't like this behavior they could just comment 3 lines, so I don't see any downside to not enabling this by default.

KlzXS:
@TheUtopian Do you plan on continuing to work on this?

TheUtopian:
@KlzXS, Yes, I think I found a solution, it's not as elegant as I initially wanted, but it seems to work pretty good.
I'll update the PR. I'd be glad to receive any advice on improving the code

plugins/preview-tabbed
```diff
@@ -177,7 +180,21 @@ previewer_loop () {
                 fi
                 ;;
             inode/directory)
-                $TERMINAL "$XID" -e nnn "$FILE" &
+                if [[ -n $PICTURES_DIR && "$FILE" == *"$PICTURES_DIR"* ]] ; then
+                    if find "$FILE" -maxdepth 1 | file -bLf- --mime-type | grep -q image ; then
```
N-R-K, line 184: These are some pretty expensive operation.

plugins/preview-tabbed
```diff
@@ -177,7 +180,21 @@ previewer_loop () {
                 fi
                 ;;
             inode/directory)
-                $TERMINAL "$XID" -e nnn "$FILE" &
+                if [[ -n $PICTURES_DIR && "$FILE" == *"$PICTURES_DIR"* ]] ; then
```
N-R-K, line 183: Why is there a glob before as well? If the picture dir is "/picture" why should "/some/random/dir/picture" match it?

N-R-K (changes requested):

plugins/preview-tabbed
KlzXS: If I'm reading this correctly:

_IF_ we are currently hovering over the "Pictures" directory _AND_ it directly contains at least one file whose MIME is an image _THEN_ we preview the directory in a specified program if available _ELSE_ we do what we were doing previously.

What do `sxiv` and `nsxiv` do if you open a directory that contains no images? Is it any different to what happens if it contains an image? I can't try this out at the moment. If it's not a significant difference just drop this check.

N-R-K: > What do sxiv and nsxiv do if you open a directory that contains no images?

It exits due to not having any images to show.

> Is it any different to what happens if it contains an image?

If it contains images then those images will be opened in thumbnail view (since `-t` flag is specified).

KlzXS: Thanks for clarifying that .

I assume it exits with a non-zero error code. Just trying it and falling back if it fails should be less expensive, right?

N-R-K: > I assume it exits with a non-zero error code

It exits with `1`, yes.

> Just trying it and falling back if it fails should be less expensive

Yeah that's probably better. But how would you do that? The (n)sxiv process gets backgrounded.

KlzXS: Yeah, that's a slight inconvenience.

I'm fairly sure you can do something like `{ command1 || command2 } &`.

A bit uglier than I hoped for the final result, but it should do the trick.

N-R-K: > A bit uglier than I hoped for the final result, but it should do the trick.

Well, `nsxiv` might exit with non-zero for other reasons too. So IDK about that.

Also don't we send signals to close/stop the backgrounded process? That's almost certainly going to result in a non-zero exit code as well and trigger `command2`.

KlzXS: > Well, nsxiv might exit with non-zero for other reasons too

If it fails for any reason I think it's still better to fall back. A preview > no preview.

> Also don't we send signals to close/stop the backgrounded process?

To be honest, I was away when the whole preview thing was being implemented and never got into it, so I have no idea how it works.

Would the signal be sent to `nsxiv` or the subshell? If it's the subshell I think we're fine. It makes more sense that it should be the subshell. How would we even know the PID of a process it spawns?

KlzXS:
@TheUtopian please address the issues we raised.

TheUtopian:
@KlzXS, @N-R-K, thanks for your comments.

> I'm fairly sure you can do something like `{ command1 || command2 } &`.

If you'll take a look at the last commit you'll see I was using something similar:
`( nsxiv -te "$XID" "$FILE" 2>/dev/null || $TERMINAL "$XID" -e nnn "$FILE" )`, but that didn't work, it was opening (n)sxiv tabs and not closing them.
I've tried to use curly brackets as well, but it was just returning an error.

**But I think I found a better solution:**
I've noticed that if we use `if` statement that always run, it will open (n)sxiv and if (n)sxiv will abort, script will run the `else` statement with nnn.

```shell
if true; then
    nsxiv -te "$XID" "$FILE" &
else
    $TERMINAL "$XID" -e nnn "$FILE" &
fi
``` 
What do you think?

> Why is there a glob before as well? If the picture dir is "/picture" why should "/some/random/dir/picture" match it?

You're right, I'll remove the left match, I was also considering regex but as I read, the glob match should be faster.

N-R-K:
> and if (n)sxiv will abort, script will run the else statement with nnn

But that's not what the code is doing, though. 

```sh
    sxiv -te "$XID" "$FILE" &
```

If the above fails for example, then it's not going to run anything else. Or am I missing something?

I'm assuming you meant to do something like this:

```sh
  if sxiv -te "$XID" "$FILE" 2>/dev/null; then
     :  # no-op
  elif nsxiv -te "$XID" "$FILE" 2>/dev/null; then
     :  # no-op
  else
     $TERMINAL "$XID" -e nnn "$FILE"
  fi &
```

This would "work" but consider what I said about (n)sxiv exiting with non-zero for other reasons too.

TheUtopian:
> If the above fails for example, then it's not going to run anything else. Or am I missing something?

You're totally right, It doesn't work that way unfortunately, I was a little hasty.

I've also tired this but it still opens new tabs without closing them:
```shell
{
    nsxiv -te "$XID" "$FILE"
    if [ $? -ne 0 ]; then
	$TERMINAL "$XID" -e nnn "$FILE"
    fi
} &
```

> This would "work" but consider what I said about (n)sxiv exiting with non-zero for other reasons too, what happens in those cases?

If I understood correctly, even if nsxiv returns non-zero value e.g. when we close it by switching the preview file, nnn will be opened and closed immediately, so it shouldn't be noticable.

KlzXS:
> If you'll take a look at the last commit you'll see I was using something similar

I missed that. I've since tried my suggestion on an actual machine and confirmed that you indeed can't terminate the subshell directly. Try what @N-R-K suggested.

> what happens in those cases?

Do we actually care about differentiating those? As @TheUtopian said, at worst we just spawn another preview process and it fails if, for example, it doesn't have permission.

TheUtopian:
@KlzXS, I have tried to do what **N-R-K** suggested, but unfortunately it also creates a subshell, so the tabs don't close.
I found another solution which is much faster, I took every image extension supported by imlib2 which is used by sxiv and nsixv.
The only disadvantage I see is that it looks ugly.
```shell
if \ls -f "$FILE" | grep -qm 1 "jpg$\|jpeg$\|png$\|svg$\|gif$\|jxl$\|webp$\|ico$\|bmp$\|ani$\|argb$\|ff$\|heif$\|heifs$\|heic$\|heics$\|avci$\|avcs$\|avif$\|avifs$\|jfif$\|jfi$\|jp2$\|j2k$\|iff$\|ilbm$\|lbm$\|ppm$\|pgm$\|pbm$\|qoi$\|ps$\|raw$\|arw$\|cr2$\|dcr$\|dng$\|nef$\|orf$\|raf$\|rw2$\|rwl$\|srw$\|tga$\|tiff$\|xbm$\|xpm$"; then
    if type sxiv >/dev/null 2>&1 ; then
        sxiv -te "$XID" "$FILE" &
    elif type nsxiv >/dev/null 2>&1 ; then
        nsxiv -te "$XID" "$FILE" &
    else
        $TERMINAL "$XID" -e nnn "$FILE" &
    fi
else
    $TERMINAL "$XID" -e nnn "$FILE" &
fi
```

N-R-K:
> I took every image extension supported by imlib2

Imlib2 allows for disabling many of the loader support. On my system many of those formats are not supported because I've disabled them.

Just assume that if the user has picture directory enabled, that it contains pictures.

TheUtopian:
@N-R-K, I squashed everything into one commit

N-R-K (approved):

jarun:
Thank you!

Issue #1860, opened 2024-03-28
Iconlookup doesn't work in nnn because it prints the wrong codepoints


aakhavanQC:
#### Environment details (Put `x` in the checkbox along with the information)

- [x] Operating System: SLES 12 (over ssh)
- [x] Desktop Environment: Windows 11
- [x] Terminal Emulator: Windows Terminal
- [x] Shell: fish
- [x] Custom desktop opener (if applicable): N/A
- [x] Program options used: none
- [x] Configuration options set: none
- [x] Plugins are installed: no plugins
- [x] Issue exists on `nnn` master

#### Exact steps to reproduce the issue

- Compile nnn from master with `make CC=gcc O_ICONS=1 O_NERD=1`.
- Run `./nnn`
- See this:
![image](https://github.com/jarun/nnn/assets/161640887/a534ec7d-86df-467c-89ca-04a3318c9e33)

Some of the icons seem to work, but most don't. preview-tui with .iconlookup works just fine, so I don't think it's related to shell/terminal settings, but rather, that they're printing different codepoints. As an example, .iconlookup is outputting codepoint `U+E5FF` for the directory icon, but nnn is outputting `U+E155`. I thought maybe it's locale-related, but playing around with that didn't fix it. It's a pretty old OS, but unfortunately I can't change it.

![image](https://github.com/jarun/nnn/assets/161640887/9fb94f92-3664-42e7-8c4f-38697461fd45)

aakhavanQC:
Re-read the dev docs and realized `O_ICONS` is **not** needed with `O_NERD`. Removing the former fixed it.

PR #1861, opened 2024-03-28
Prevent multiple icons options being selected at the same time

KlzXS:
It's a bit weird that we continue the build process if multiple options for icons are set. I wouldn't expect it to work and yet it does. Especially with one of the flags being named the most generic thing (`O_ICONS`) it's bound to happen on accident.

We've never had conditions for failing builds before, so I took the liberty of naming it `T_ICONS` for "test icons".

Inspired by #1860.


jarun:
Thank you!

PR #1864, opened 2024-04-04
Fix file creation on OpenBSD

horrad:
On OpenBSD at least one of O_RDONLY, O_WRONLY or O_RDWR is needed to open a file.

In creating a new file none of those is set, which leads to an EINVAL error ("invalid argument").

Since the new file is only created and never read, I chose to use O_WRONLY.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4732,7 +4732,7 @@ next:
 			return FALSE;
 		}
 	} else {
-		int fd = open(path, O_CREAT | O_TRUNC, S_IWUSR | S_IRUSR); /* Forced create mode for files */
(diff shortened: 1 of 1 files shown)


N-R-K (approved):

horrad:
> @horrad Please put what you wrote in the PR description in the commit message body as well.
Done.

jarun:
Can you please make the patch BSD-specific?

KlzXS:
I think it's best we avoid adding unnecessary `#ifdef`s. [`open(2)`](https://www.man7.org/linux/man-pages/man2/open.2.html) says:

> The argument flags must include one of the following access
       modes: O_RDONLY, O_WRONLY, or O_RDWR.

We have none of those.

jarun:
Makes sense. Merged. Thanks!

PR #1868, opened 2024-04-16
remove common prefix before comparing numerical filenames 

azuline:
Currently, nnn sorts files like this:

image

After code change:

image


N-R-K:
Isn't this what version sort already does? E.g press `t` and then `v` (or use `-Tv` cli option).

azuline:
just tried it; yep! going to close this now

i think what i did there also doesn't create a total ordering, so it wouldn't be a good change anyways without some more analysis.

PR #1869, opened 2024-04-16
avoid cursor jump when file under cursor is renamed by plugin/`r`

azuline:
This PR prevents the cursor from jumping back to the top of the directory when the active file (file underneath the cursor) is renamed by the `r` renamer tool or an external plugin.

https://github.com/jarun/nnn/assets/51880422/a98d56c8-68ec-4570-80d3-e7d74c15860a

My use case:

I use `nnn` to navigate my music library, with a custom plugin to operate on music directories. Several of these commands rename the directory. Currently, once the directory is renamed (often the change of a single character), the cursor jumps back to the top of the working directory, requiring me to navigate back to where I was in a fairly large directory. After this change, the cursor no longer jumps.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5394,7 +5394,7 @@ static char *readpipe(int fd, char *ctxnum, char **path)
 	return nextpath;
 }
 
-static bool run_plugin(char **path, const char *file, char *runfile, char **lastname, char **lastdir)
+static bool run_plugin(char **path, const char *file, char *runfile, char **lastname, char **lastdir, bool *cd)
 {
 	pid_t p;
 	char ctx = 0;
@@ -5477,6 +5477,8 @@ static bool run_plugin(char **path, const char *file, char *runfile, char **last
 	nextpath = readpipe(rfd, &ctx, path);
 	if (nextpath)
 		set_smart_ctx(ctx, nextpath, path, runfile, lastname, lastdir);
+	else
+		*cd = FALSE;
 
 	close(rfd);
 
@@ -5991,7 +5993,7 @@ exit:
 	return ndents;
 }
 
-static void populate(char *path, char *lastname)
+static void populate(char *path, char *lastname, bool cd)
 {
 #ifdef DEBUG
 	struct timespec ts1, ts2;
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -5475,8 +5475,11 @@ static bool run_plugin(char **path, const char *file, char *runfile, char **last
 	while (rfd == -1 && errno == EINTR);
 
 	nextpath = readpipe(rfd, &ctx, path);
-	if (nextpath)
+	if (nextpath) {
 		set_smart_ctx(ctx, nextpath, path, runfile, lastname, lastdir);
+	} else {
+		*cd = FALSE;
+	}
```
azuline, line 5482: Prior, there was an assumption that every plugin execution could change the directory, and so `cd = TRUE` for every plugin run. We now accurately set `cd` based on whether the plugin changed the directory or not.

I believe this is the only way for a plugin to change the directory, would like confirmation.

src/nnn.c
```diff
@@ -5477,6 +5477,8 @@ static bool run_plugin(char **path, const char *file, char *runfile, char **last
 	nextpath = readpipe(rfd, &ctx, path);
 	if (nextpath)
 		set_smart_ctx(ctx, nextpath, path, runfile, lastname, lastdir);
+	else
+		*cd = FALSE;
```
azuline, line 5481: Prior, there was an assumption that every plugin execution could change the directory, and so `cd = TRUE` for every plugin run. We now accurately set `cd` based on whether the plugin changed the directory or not.

I believe this is the only way for a plugin to change the directory, would like confirmation.

src/nnn.c
```diff
@@ -6014,7 +6016,18 @@ static void populate(char *path, char *lastname)
...
+		target = dentfind(lastname, ndents);
+		/* If we failed to find the previous filename, and we didn't change directories, */
+		/* default to the previous cursor position. This way, when a file is renamed by */
+		/* a plugin, the cursor does not jump to the top of the directory. */
+		if (target == 0 && !cd)
+			target = cur;
+	} else
+		target = 0;
+
+	move_cursor(target, 0);
```
KlzXS, line 6030: What if now `lastname` _is_ on position 0 and we didn't `cd`? You set `target = cur` which is probably not what we want.

Also, we generally don't know if `cur` will be a valid position after running external actions. Any number of entries may be added or removed or renamed, in which case you might as well place the cursor on a random line.

jarun:
> Currently, once the directory is renamed (often the change of a single character)

From the `nnn` help page:

`^R  Rename/dup              r  Batch rename`

If you are changing only 1 file/dir, use `^R`.

Alternatively, use filters to reach a file quickly. Since you have just renamed it, you should be knowing the name.

azuline:
I am neither using `^R` or `r`, but a custom plugin in a very large directory. I will keep this behavior in a fork since it is only important to my use case. 👍🏼

src/nnn.c
azuline: Thanks, I didn't consider these cases. I will fix them in my fork.

PR #1888, opened 2024-05-21
fix: Expand on preview-tui docs

jasper-at-windswept:
Parts of preview-tui docs were vague on steps for settings up kitty, this pull requests tries to make it clearer.


luukvbaal:
There are 3 empty lines added for no reason. I also don't think this actually clarifies anything that is specific to `preview-tui`.
`$TMPDIR` is a common environment variable. Moreover, it doesn't need to be substituted with the absolute path, kitty will expand the variable so the example in the doc can be used as is...

PR #1890, opened 2024-05-27
quitcd: fix range limits for new nushell version

BeyondMagic:
[Last version of nushell](https://www.nushell.sh/blog/2024-04-30-nushell_0_93_0.html) changed the way ranges work, so the command `str substring 4..-1` isn't behaving the way it was anymore.

### DIFF
--- a/misc/quitcd/quitcd.nu
+++ b/misc/quitcd/quitcd.nu
@@ -29,8 +29,7 @@ export def --env n [
 		# Remove <cd '> from the first part of the string and the last single quote <'>.
 		# Fix post-processing of nnn's given path that escapes its single quotes with POSIX syntax.
 		let path = open $nnn_tmpfile
-			| str substring 3..
-			| str trim --char "'"
+			| str replace --all --regex `^cd '|'$` ``
 			| str replace --all `'\''` `'`
 
 		^rm -- $nnn_tmpfile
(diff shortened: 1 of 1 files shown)


N-R-K:
Won't this break for people who have older nushell versions or is the new patch backwards compatible? If it's not backwards compatible, then please document the minimum required version in the comment at the top of the file.

misc/quitcd/quitcd.nu
```diff
@@ -28,7 +28,10 @@ export def --env n [
 	if ($nnn_tmpfile | path exists) {
 		# Remove <cd '> from the first part of the string and the last single quote <'>.
 		# Fix post-processing of nnn's given path that escapes its single quotes with POSIX syntax.
-		let path = open $nnn_tmpfile | str substring 4..-1 | str replace --all `'\''` `'`
+		let path = open $nnn_tmpfile
+			| str substring 3..
+			| str trim --char "'"
```
N-R-K, line 33: What happens if a path ends in `'`? Won't strim end up removing the `'` as well? https://www.nushell.sh/commands/docs/str_trim.html

I think you should only remove the last char, not "trim" every `'`.

N-R-K (changes requested):

misc/quitcd/quitcd.nu
BeyondMagic: True, I hoped NNN'd send a trailing slash at the end of the path before the single quotes; but that's no the case.

BeyondMagic:
> Won't this break for people who have older nushell versions or is the new patch backwards compatible?

No; not in the way it is coded right now.

BeyondMagic:
Resolved with backwards compatibility and fixed trimming issue.

jarun:
Thank you!

PR #1894, opened 2024-06-11
Draft: Add mark plugin

TheKhanj:
It can be useful to mark certain files, such as when watching a TV show or a course, to keep track of where you left off. I couldn't find any other plugin that does this.

However, it's a bit annoying that each time this script changes a file name, the cursor jumps back to the first file. If there's a way to fix that, I'd be happy to know about it.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -41,6 +41,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [ipinfo](ipinfo) | Fetch external IP address and whois information | sh | curl, whois |
 | [kdeconnect](kdeconnect) | Send selected files to an Android device [✓] | sh | kdeconnect-cli |
 | [launch](launch) | GUI application launcher | sh | fzf |
+| [mark](mark) | Mark and unmark files | sh | - |
 | [mimelist](mimelist) | List files by mime in subtree | sh | file/mimetype |
 | [moclyrics](moclyrics) | Show lyrics of the track playing in moc | sh | [ddgr](https://github.com/jarun/ddgr), [moc](http://moc.daper.net/) |
 | [mocq](mocq) | Queue/play selection/dir/file in moc [✓] | sh | [moc](http://moc.daper.net/) |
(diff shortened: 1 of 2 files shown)


N-R-K:
> However, it's a bit annoying that each time this script changes a file name, the cursor jumps back to the first file. If there's a way to fix that, I'd be happy to know about it.

There isn't currently. Patches welcome. See also: https://github.com/jarun/nnn/discussions/1889

> It can be useful to mark certain files, such as when watching a TV show or a course, to keep track of where you left off.

The plugin basically just prepend/removes `* ` to a filename, is that correct? Seems like a niche thing tbh.

TheKhanj:
Yes, it essentially adds an '*' to the filename. It might be better to implement this as a built-in feature in nnn, though it could potentially make nnn bloated. Your call :).

Ranger has a similar feature, and in my opinion, it's very handy.

jarun:
shellcheck in CI fails:

```

In plugins/mark line 9:
. "$(dirname "$0")"/.nnn-plugin-helper
  ^-- SC1091 (info): Not following: ./.nnn-plugin-helper was not specified as input (see shellcheck -x).


In plugins/mark line 15:
        files="$(echo "$PWD/$1")"
               ^---------------^ SC2116 (style): Useless echo? Instead of 'cmd $(echo foo)', just use 'cmd foo'.


In plugins/mark line 18:
is_marked() { echo "$1" | grep '*$' >/dev/null 2>&1; }
                               ^--^ SC2063 (warning): Grep uses regex, but this looks like a glob.

For more information:
  https://www.shellcheck.net/wiki/SC2063 -- Grep uses regex, but this looks l...
  https://www.shellcheck.net/wiki/SC1091 -- Not following: ./.nnn-plugin-help...
  https://www.shellcheck.net/wiki/SC2116 -- Useless echo? Instead of 'cmd $(e...

```

@N-R-K do you find it useful?

N-R-K:
> @N-R-K do you find it useful?

Personally, no. We have batch renamer which (paired with a good editor) can do this already.

TheKhanj:
Yeah you're right, I agree, but implementing it as a builtin feature "might" be useful:)

jarun:
In that case I am closing this feature.

We have a built-in batch renamer too. However, it has some limitations:
https://github.com/jarun/nnn/wiki/Troubleshooting#default-batch-rename-limits

PR #1898, opened 2024-06-22
dirs in context color fix

a-gruzdev:
When you launch `nnn -D` , directories use some default color instead of context color

I made some investigation and found out that this check `if (color_pair && fcolors[color_pair])` is always false for context colors, as it's values in `fcolors` are always `0`.
dbg

I added it's initialization in `initcurses` to fix this. Now it looks like this
col1
col2

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2284,8 +2284,10 @@ static bool initcurses(void *oldmask)
 						msg(env_cfg[NNN_COLORS]);
 						return FALSE;
 					}
-				} else
+				} else {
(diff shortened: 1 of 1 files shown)


jarun:
Did you test without `-D` to confirm there are no side effects?

N-R-K (approved):
Looks okay to me. Did some light testing and everything seems to work as expected.

a-gruzdev:
yes, tested without `-D`, dirs use their color from `fcolors` (blue by default), and everything else works as expected

jarun:
Thank you!

Issue #1678, opened 2023-06-28
Missing large file support for older toolchains


hgy59:
nnn requires the option `_FILE_OFFSET_BITS==64` on 32 bit archs to handle Large files (> 4GB).

This is forced for 32-bit arm and x86 archs in nnn.c by 
```
#if defined(__arm__) || defined(__i386__)
#define _FILE_OFFSET_BITS 64 /* Support large files on 32-bit */
#endif
```

This was introduced in #42 and fixed for 32-bit arm with #135

Since `nnn.c` depends on included `fts.h` I get a comiler error for the current version (4.8) with older toolchains:
```
<fts.h> cannot be used with -D_FILE_OFFSET_BITS==64
```

I can remove the the definition of `_FILE_OFFSET_BITS 64`  to build nnn successfully with these toolchains, but than it will lack to display large files correctly.

Is it possible to build nnn without the dependency of `fts.h` to fix this?
i.e. I am searching for a solution to build nnn without multithreading support that was introduced with nnn v4.1.

N-R-K:
This is a bug in glibc, from `man 3 fts`:

```
BUGS
       Before  glibc  2.23,  all  of  the  APIs described in this man page are not safe when compiling a program using the LFS APIs
       (e.g., when compiling with -D_FILE_OFFSET_BITS=64).
```

I don't think we have any plans to workaround libc bugs which have long been fixed (glibc 2.23 was released in 2016). But if you really want to use such old toolchains, maybe projects like [fts-standalone](https://github.com/pullmoll/musl-fts) interests you.

- - -

Also, I think we should remove those ifdefs and define `_FILE_OFFSET_BITS 64` unconditionally.

PR #1910, opened 2024-07-15
Disable large file support on old glibc

sjmulder:
On glibc <2.23 fts.h doesn't support the feature:

https://sourceware.org/bugzilla/show_bug.cgi?id=11460

```
In file included from src/nnn.c:78:0:
/usr/include/fts.h:41:3: error: #error "<fts.h> cannot be used with -D_FILE_OFFSET_BITS==64"
 # error "<fts.h> cannot be used with -D_FILE_OFFSET_BITS==64"
```

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -35,7 +35,7 @@
 #endif
 #include <features.h>		/* __GLIBC__ etc */
 /* large file support on 32-bit glibc >= 2.23 where fts.h supports it */
-#if !defined(__GLIBC__) || __GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 23)
+#if __GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 23)
 #define _FILE_OFFSET_BITS 64
 #endif
 #if defined(__linux__)
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -28,12 +28,15 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#define _FILE_OFFSET_BITS 64 /* Support large files on 32-bit glibc */
-
 #if defined(__linux__) || defined(MINGW) || defined(__MINGW32__) \
 	|| defined(__MINGW64__) || defined(__CYGWIN__)
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
+#include <features.h>		/* __GLIBC__ etc */
+/* large file support on 32-bit glibc >= 2.23 where fts.h supports it */
+#if !defined(__GLIBC__) || __GLIBC__ > 2 || __GLIBC_MINOR__ >= 23
+#define _FILE_OFFSET_BITS 64
 #endif
 #if defined(__linux__)
 #include <sys/inotify.h>
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -35,7 +35,7 @@
 #endif
 #include <features.h>		/* __GLIBC__ etc */
 /* large file support on 32-bit glibc >= 2.23 where fts.h supports it */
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -28,13 +28,16 @@
...
  */
 
-#define _FILE_OFFSET_BITS 64 /* Support large files on 32-bit glibc */
-
 #if defined(__linux__) || defined(MINGW) || defined(__MINGW32__) \
 	|| defined(__MINGW64__) || defined(__CYGWIN__)
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
+#include <features.h>		/* __GLIBC__ etc */
```
N-R-K, line 36: I don't think this is a public header for us to be including.

N-R-K (changes requested):
This was discussed here: https://github.com/jarun/nnn/issues/1678#issuecomment-1615220416

I don't think we want to keep adding workarounds for buggy libc behavior which have been fixed for almost a decade now.

sjmulder:
It's your call to make - the advantage is that nnn can still run older targets such as CentOS 7, at the cost of a bit of preprocessor crud. Functionally there's no degradation on platforms where nnn already works.

src/nnn.c
sjmulder: I wasn't aware of this but indeed feature_test_macros(7) documents this. I can shuffle includes around a bit to avoid this if you're not against this PR a priori.

N-R-K: It's upto @jarun to make the call. I personally don't think trying to work around a decade old glibc bug is worth it, especially since glibc maintainers now recommend defining it unconditionally.

sjmulder: I understand the perspective but the one extra ifdef keeps nnn working on legacy platforms where system utilities like this remain useful. Note that RHEL 7 is still under extended lifecycle support. I'll keep it working in pkgsrc while I can but it would be nice if vanilla nnn kept working too.

jarun: I am fine with this. We should try to keep `nnn` working on older systems.

jarun (changes requested):

src/nnn.c
N-R-K: > I am fine with this. We should try to keep nnn working on older systems.

In that case, @sjmulder please try to see if you can do this without including private/non-posix libc headers.

N-R-K: > then unconditionally include one of the base headers like stddef.h to get the effects of features.h

This sounds okay.

> That would assume it's okay to delay _FILE_OFFSET_BITS until after stddef.h

Unfortunately, I don't think this will work. From the `feature_test_macros` manpage:

"NOTE: In order to be effective, a feature test macro must be  defined  before  including  any header files"

sjmulder: `_FILE_OFFSET_BITS` is indeed checked by features.h, so a bit of a chicken and egg problem here. Bit of a chicken and egg problem.

sjmulder: I can't think of a way to do this correctly without a configure script. Unless you do I propose we close this PR and I think of something better (or leave it as is).

N-R-K: We have a bit of pseudo configure script going on in the Makefile over here

https://github.com/jarun/nnn/blob/86d883eee4986c7125bd77415e1a9b6b94e6bb83/Makefile#L189-L190

If you can do something like this without too much baggage (we don't want to slow down the build step noticeably) then that seems acceptable to me.

N-R-K: Here's a command that you can use to test glibc version in the makefile if you're still interested in pursuing this:

```sh
printf "#include <stdlib.h>\n#if __GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 23)\nYES\n#else\nNO\n#endif\n" | cc -E - | tail -n1
```

jarun:
Closing this since this is working for a very long time.

PR #1911, opened 2024-07-15
dprintf() polyfill for Illumos

sjmulder:
I realise it's a bit of a long shot, submitting a PR to fix a legacy platform, but I thought I might at least try upstreaming this patch we've been using for a while.

Edit: this is also for Illumos, which is actively maintained.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -121,6 +121,10 @@
 #define alloca(size) __builtin_alloca(size)
 #endif
 
+#ifdef __sun
+#define NEED_DPRINTF
+#endif
+
 #include "nnn.h"
 #include "dbg.h"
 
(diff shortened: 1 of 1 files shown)


src/nnn.c
```diff
@@ -856,6 +860,27 @@ static void notify_fifo(bool force);
...
 
+#ifdef NEED_DPRINTF
+static int dprintf(int fd, const char *format, ...)
+{
+	va_list ap;
+	char *s;
+	int len, nwritten;
+
+	va_start(ap, format);
+	len = vasprintf(&s, format, ap);
```
N-R-K, line 871: Although posix doesn't require it, the general expectation is for `dprintf` to not do any memory allocation so that it can be used in signal handler context. Not doing allocation would also remove allocation related failure point.

src/nnn.c
```diff
@@ -121,6 +121,10 @@
 #define alloca(size) __builtin_alloca(size)
 #endif
 
+#ifdef __sun
```
N-R-K, line 124: If Illumos is maintained, why don't they add `dprintf`? It's more economical for them to add it instead of N projects to carry their own implementation.

src/nnn.c
sjmulder: I agree it would be better but I couldn't find a better alternative with the available primitives. One could use a static buffer but then you're length limited.

src/nnn.c
sjmulder: Would be better if they did indeed. But that does leave legacy systems, and even there nnn can be very useful.

jarun:
@N-R-K please merge when ready.

sjmulder:
Turns out there aren't that many dprintf() so perhaps this alternative is better: https://github.com/jarun/nnn/pull/1912

jarun:
Closed with #1912.

PR #1900, opened 2024-06-24
Added New Plugin

Zazucki:
Added upl0ad: Upload files to 0x0.st

I'm open to alternative names, or integrating this into the upload plugin.

PR #1917, opened 2024-07-25
plugins/upload: remove inactive services

N-R-K:
this removes ix.io and ffsend since they are no longer active.

binary files are uploaded to file.io as before. selections are now tar-balled and uploaded to file.io.

text files are uploaded to 0x0.st

Closes: https://github.com/jarun/nnn/pull/1900

### DIFF
--- a/plugins/upload
+++ b/plugins/upload
@@ -1,12 +1,11 @@
 #!/usr/bin/env sh
 
-# Description: Selections are uploaded using Firefox Send
+# Description: Selections are archived into a tar file (uncompressed) and uploaded to file.io
 #              For single files:
-#              Upload to Firefox Send if ffsend is found, else
-#              Paste contents of a text a file http://ix.io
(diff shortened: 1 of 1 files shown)


N-R-K:
@KlzXS @zazucki Please test this out and see if it works as expected or not.

Zazucki:
I might be able to test tonight or this weekend.
Is taring and uploading the selection correct or should it upload each individually? I think the user should compress/zip/tar/whatever before (possibly using a different nnn plugin) and have the upload plugin only handle uploading.

N-R-K:
> Is taring and uploading the selection correct or should it upload each individually?

I looked at the [ffsend readme](https://github.com/timvisee/ffsend) and the preview gif there shows that when doing multiple files it archives the file before uploading. So that's the behavior I went with.

jarun:
Thank you!

PR #1928, opened 2024-09-09
Fix implicit function declaration strcasecmp

s-hamann:
`icons-hash.c` uses `strcasecmp` from `strings.h`, but does not include it. At least when building against musl with gcc 15, this is a fatal error.

See https://bugs.gentoo.org/939164


N-R-K (approved):

jarun:
Thank you!

PR #1930, opened 2024-09-13
Have NNN_PREVIEWIMGPROG accept a generic command for preview-tui

danielb2:
Make NNN_PREVIEWIMGPROG flexible enough to accept any program, including necessary flags.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -460,6 +460,8 @@ image_preview() {
         chafa "$3" &
     elif exists img2sixel && [[ "$NNN_PREVIEWIMGPROG" == +(|img2sixel) ]]; then
         img2sixel -g "$3" &
+    elif exists "$NNN_PREVIEWIMGPROG"; then
+        "$NNN_PREVIEWIMGPROG" "$3" &
     else
         fifo_pager print_bin_info "$3" && return
     fi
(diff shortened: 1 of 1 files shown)


plugins/preview-tui
```diff
@@ -460,6 +460,8 @@ image_preview() {
         chafa "$3" &
     elif exists img2sixel && [[ "$NNN_PREVIEWIMGPROG" == +(|img2sixel) ]]; then
         img2sixel -g "$3" &
+    elif exists "$NNN_PREVIEWIMGPROG"; then
```
luukvbaal, line 463: > it seems odd to add every single program out there

Some of the provided options require a bunch of flags, which this script provides for the user. But yes, adding a more generic else clause would be good. In that case I would change the `exists` check to check the first word in the variable like so:

```suggestion
    elif exists "${NNN_PREVIEWIMGPROG%% *}"; then
```

That way the user can add any necessary flags to the command as well. Feel free to also remove the current else clauses that do not contain any flags and would be covered by this generic one.

danielb2:
I modified exists instead since it will never check more than one file `exists() { type "${1%% *}" >/dev/null 2>&1 ;}`

plugins/preview-tui
danielb2: I modified exists instead since it will never check more than one file exists() { type "${1%% *}" >/dev/null 2>&1 ;}

luukvbaal: Not sure I prefer that, that obfuscates what we are doing. Would be better to properly document `NNN_PREVIEWIMGPROG` in that it can be a full command including flags, rather than just a program path. And then keep the variable expansion at the callsite like I suggested. I have pushed a commit that does that.

luukvbaal (approved):

jarun:
Thank you!

plugins/preview-tui
danielb2: It still can. I'm just moving the logic over to exists to check the first part of the argument since exists may be used in other places and it's cleaner this way because then the argument doesn't need to be changed everytime before passing

PR #1935, opened 2024-09-20
Do not default to `xterm`

kovmir:
It makes more sense to default to the terminals people are less likely to have. If a user has `xterm` installed, then that is, most probably, because it ships with X11. Whereas if a user has `st` or `alacritty` instlled, then that is because he/she chose to install it.


jarun:
@N-R-K or @luukvbaal can one of you please address the CI failures?

jarun:
Thank you!

PR #1937, opened 2024-09-22
Indicate type-to-nav mode with symbol F

amalgame21:
Notify the user the current context menu is in type-to-nav mode with symbol N after pressing `ESC` in type-to-nav mode.
After pressing `^N`, the symbol N disappear, indicate it is no longer in type-to-nav mode.

Feel free to reject it if it does not match nnn's design.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -3238,6 +3238,9 @@ static int getorderstr(char *sort)
 {
 	int i = 0;
 
+	if (cfg.filtermode)
+		sort[i++] = 'F';
+
 	if (cfg.showhidden)
 		sort[i++] = 'H';
 


src/nnn.c
```diff
@@ -3238,6 +3238,9 @@ static int getorderstr(char *sort)
 {
 	int i = 0;
 
+	if (cfg.filtermode)
```
jarun, line 3241: Why do we need this? The prompt should show `/` when the mode is enabled.

jarun (changes requested):

src/nnn.c
```diff
@@ -3238,6 +3238,9 @@ static int getorderstr(char *sort)
...
+		sort[i++] = 'N';
```
jarun, line 3242: In case you really want it (appears to be useful in some cases), please use `F`.

jarun (changes requested):

src/nnn.c
jarun: Got it.

jarun:
Thank you!

src/nnn.c
```diff
@@ -3238,6 +3238,9 @@ static int getorderstr(char *sort)
 {
 	int i = 0;
 
+	if (cfg.filtermode)
+		sort[i++] = 'F';
```
N-R-K, line 3242: Is this the right place to add this though? This string is related to sorting, but type to nav mode isn't really related to sorting order.

jarun: True but it's the smallest patch. :)

amalgame21: Sorry I don't understand the code design very much and I just hacked it with my very very limited knowledge in C. If it is not the proper place to add this code and mess thing up please reverse it. Local patch is easy anyway.

jarun: I have taken care of it. The code is better paled now.

amalgame21: Thanks! Really appreciated!!!

PR #1938, opened 2024-09-22
Toggle type-to-nav (^N) like global switch

amalgame21:
This commit bring the current type-to-nav mode state to next context while changing to the next context, making toggling the type-to-nav (^N) like a global switch, instead of the original behavior which store each type-to-nav mode state per context.

I think it is much easier to control the workflow because I no longer need to remember which context is in type-to-nav mode or not.

Feel free to reject it if it does not match nnn's design.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4330,6 +4330,7 @@ static void savecurctx(char *path, char *curname, int nextctx)
 {
 	settings tmpcfg = cfg;
 	context *ctxr = &g_ctx[nextctx];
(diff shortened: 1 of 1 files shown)


jarun:
It's not about design, a use case would be broken.

When we create the context, the type-to-nav feature is also enabled. So if it's disabled for a context, it's because the user has decided to disable it explicitly. We can't override that every time there is a context switch because in that case there would be now simple way to switch contexts without disabling type-to-nav.

PR #1948, opened 2024-11-05
restore sigpipe handler in plugin child processes

### LINKED ISSUE #1947: find doesn't exit early in fzopen if NNN_FIFO exists
When nnn is run with the fifo enabled and I try to use fzopen in a directory with a large number of children, fzopen will hang for a while until `find` finishes listing every file. If I run the same command that fzopen runs in a normal shell the whole command stops as soon as I press enter. This problem doesn't seem to be specific to `find` or `fzf` and the same problem seems to happen if anything that takes a while to pipe a bunch of data is piped into a command that quickly exits (even tested with `true`).

- [x] Operating System: NixOS unstable (24.11.20241102)
- [x] Desktop Environment: KDE
- [x] Terminal Emulator: Kitty
- [x] Shell: zsh
- [ ] Program options used: none
- [x] Configuration options set: NNN_FIFO=/tmp/nnn.fifo
- [x] Plugins are installed
- [ ] Issue exists on `nnn` master

- run nnn with NNN_FIFO set or with -a
- go to `/`
- run fzopen
- immediately press enter

babbaj:
fixes #1947

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -5529,6 +5529,9 @@ static bool run_plugin(char **path, const char *file, char *runfile, char **last
 		if (wfd == -1)
 			_exit(EXIT_FAILURE);
 
+		/* Restore sigpipe handler to default */
+		sigaction(SIGPIPE, &(struct sigaction){.sa_handler = SIG_DFL}, NULL);
+
 		if (!cmd_as_plugin) {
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

Issue #1931, opened 2024-09-17
Crash When Attempting to Archive


levone1:
#### Environment details (Put `x` in the checkbox along with the information)

- [ x] Operating System: Linux Mint 22
- [ x] Desktop Environment: KDE Plasma 5.27
- [ x] Terminal Emulator: Kitty and others
- [ x] Shell: Bash
- [ ] Custom desktop opener (if applicable):
- [ x] Program options used: None, and any tried
- [ x] Configuration options set: NNN_ARCHIVE="\\.(7z|a|ace|alz|arc|arj|bz|bz2|cab|cpio|deb|gz|jar|lha|lz|lzh|lzma|lzo|rar|rpm|rz|t7z|tar|tbz|tbz2|tgz|tlz|txz|tZ|tzo|war|xpi|xz|Z|zip)$"
- [ x] Plugins are installed
- [ x] Issue exists on `nnn` master

#### Exact steps to reproduce the issue

- select multiple files
- type 'z' then 's' for "selected"
- enter archive name, (example.zip; example.tar.gz)
- press enter

Any attempt to archive ends in a crash, with "*** buffer overflow detected ***:  terminated" message.

Tried with several different variables and configurations.  Also built master and tested with O_DEBUG=1.  Here is nnndbg contents:
```
ln 8751: VERSION=5.0
ln 8531: g_tmpfpath=/tmp
ln 8532: tmpfplen=5
ln 8784: home=/home/me
ln 8479: cfgpath=/home/me/.config
ln 8485: cfgpath=/home/me/.config/nnn
ln 8513: selpath=/home/me/.config/nnn/.selection
ln 8792: opener=xdg-open
ln 8902: getenv(envs[ENV_VISUAL])=hx
ln 8903: getenv(envs[ENV_EDITOR])=emacs
ln 8904: editor=hx
ln 8908: pager=supercat
ln 8912: shell=/bin/bash
ln 8914: getenv("PWD")=/home/me/Desktop/tmp/nnn
ln 2254: COLORS=256
ln 2255: COLOR_PAIRS=32767
ln 5761: __func__=dentfill
ln 6028: ts2.tv_nsec - ts1.tv_nsec=52311
ln 6632: __func__=redraw
ln 6647: path=/home/me/Desktop/tmp/nnn
ln 6632: __func__=redraw
ln 6647: path=/home/me/Desktop/tmp/nnn
ln 6632: __func__=redraw
[...]

PR #1953, opened 2024-11-11
fix crash under _FORTIFY_SOURCE

N-R-K:
when built with _FORTIFY_SOURCE it will check whether the buffer has as much space as the argument passed to snprintf: https://github.com/bminor/glibc/blob/7b544224f82d20019f9b28522ebf8114a372d1a2/debug/snprintf_chk.c#L28-L29

this results in some false positives when the snprintf provided len argument is bigger than the buffer size (but the result would have fit into the buffer anyways).

fix this by passing the proper size to snprintf as argument. (the +1 len isn't necessary, but add it just in case.)

Fixes: https://github.com/jarun/nnn/issues/1931

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2834,15 +2834,16 @@ static char *get_archive_cmd(const char *archive)
 
 static void archive_selection(const char *cmd, const char *archive)
 {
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #1956, opened 2024-11-14
Use consistent language for disk mode in manual

roygbyte:
Was trying to see folder size and had difficulty moving from the mention of "du mode" to sorting by disk use. My suggestion makes a soft reference to sorting by disk use. Hopefully clearer for peeps.


jarun:
Thank you!

PR #1961, opened 2024-11-25
bugfix/directory-preview-doesnt-work-on-mac-os

moonvent:
On macOS, without additional utilities, the directory preview feature was not working due to the --group-directories-first flag, which is not supported by the BSD version of ls.

To resolve this, an alternative method was implemented and tested.

### DIFF
--- a/plugins/preview-tui
+++ b/plugins/preview-tui
@@ -373,7 +373,7 @@ preview_file() {
         elif exists eza; then # eza is a community fork of exa (exa is unmaintained)
             fifo_pager eza -T --group-directories-first --colour=always -L 3
         else
-            fifo_pager ls -F --group-directories-first --color=always
+            fifo_pager ls -1 --color=always | sort
         fi
         cd ..
     elif [ "${encoding#*)}" = "binary" ]; then


plugins/preview-tui
```diff
@@ -373,7 +373,7 @@ preview_file() {
         elif exists eza; then # eza is a community fork of exa (exa is unmaintained)
             fifo_pager eza -T --group-directories-first --colour=always -L 3
         else
-            fifo_pager ls -F --group-directories-first --color=always
+            fifo_pager ls -1 --color=always | sort
```
N-R-K, line 376: These are not the same thing. You should either install GNU ls, or any of the other alternatives listed as a dependency at the top of the plugin:

```
#   - tree or exa/eza or (GNU) ls
```

N-R-K (changes requested):

plugins/preview-tui
moonvent: Alright the request is closed

N-R-K: The plugins typically state any non-portable dependencies (if any) on top. If you find anything that doesn't work despite having the proper dependencies installed then feel free to report those.

PR #1967, opened 2024-12-11
Support macOS trash (`export NNN_TRASH=3`)

p-mng:
Most of the details for this PR can be found in #1963, I've been using this patch since then without any problems. Since I haven't received a response yet, I thought I would just open a PR. In the meantime, I was also able to verify that the `trash` command exists on all new installations of macOS using a Sonoma VM.


N-R-K:
I don't think adding another hardcoded util is the way to go. There's a lot of different trashing utils nowadays and we can't keep hardcoding more and more stuff.

Issue #1968, opened 2024-12-12
session restore (s r) crash


doremiyeon:
https://github.com/jarun/nnn/blob/5e5a17f9548f3267b94263c6ac6466456d6c3d06/src/nnn.c#L4546

when attempting to restore a session by hitting 's r', _sname_ is **NULL** here. should probably check for that and write @\0 to _curssn_.

jarun:
Raise a PR please.

doremiyeon:
#1969 here you go, sorry.

PR #1969, opened 2024-12-13
fix 's r' session restore #1968

doremiyeon:


src/nnn.c
```diff
@@ -4543,7 +4543,7 @@ static bool load_session(const char *sname, char **path, char **lastdir, char **
 	*lastdir = g_ctx[cfg.curctx].c_last;
 	*lastname = g_ctx[cfg.curctx].c_name;
 	set_sort_flags('\0'); /* Set correct sort options */
-	xstrsncpy(curssn, sname, NAME_MAX);
+	xstrsncpy(curssn, sname ? sname : "@", NAME_MAX);
```
N-R-K, line 4546: how can `sname` be null here? Can you provide a stack trace of the crash?

N-R-K (changes requested):

src/nnn.c
N-R-K: Ok, I see. It's coming from here:

    8122                                    if (load_session(NULL, &path, &lastdir, &lastname, r == 'r')) {

jarun:
Thank you!

PR #1970, opened 2024-12-15
filter function restore on session load

doremiyeon:
loading a session or switching to a different context with a different filter mode setting (regex vs string) will not update the filter function and can lead to crashes
- **on session load, restore correct filter function**
- **also switch filterfn on ctx changes**

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4385,6 +4385,7 @@ static void setcfg(settings newcfg)
 	/* Synchronize the global function pointers to match the new cfg. */
 	entrycmpfn = cfg.reverse ? &reventrycmp : &entrycmp;
 	namecmpfn = cfg.version ? &xstrverscasecmp : &xstricmp;
+	filterfn = cfg.regex ? &visible_re : &visible_str;
 }
 
 static void savecurctx(char *path, char *curname, int nextctx)
@@ -4542,10 +4543,9 @@ static bool load_session(const char *sname, char **path, char **lastdir, char **
 	*path = g_ctx[cfg.curctx].c_path;
 	*lastdir = g_ctx[cfg.curctx].c_last;
 	*lastname = g_ctx[cfg.curctx].c_name;
+	setcfg(cfg);
 	set_sort_flags('\0'); /* Set correct sort options */
 	xstrsncpy(curssn, sname ? sname : "@", NAME_MAX);
-	/* restore correct filter function */
-	filterfn = cfg.regex ? &visible_re : &visible_str;
 	status = TRUE;
 
 END:
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -4544,6 +4544,8 @@ static bool load_session(const char *sname, char **path, char **lastdir, char **
 	*lastname = g_ctx[cfg.curctx].c_name;
 	set_sort_flags('\0'); /* Set correct sort options */
 	xstrsncpy(curssn, sname ? sname : "@", NAME_MAX);
+	/* restore correct filter function */
+	filterfn = cfg.regex ? &visible_re : &visible_str;
 	status = TRUE;
 
 END:


src/nnn.c
```diff
@@ -4542,6 +4543,7 @@ static bool load_session(const char *sname, char **path, char **lastdir, char **
 	*path = g_ctx[cfg.curctx].c_path;
 	*lastdir = g_ctx[cfg.curctx].c_last;
 	*lastname = g_ctx[cfg.curctx].c_name;
+	setcfg(cfg);
```
jarun, line 4546: How are the other 2 working without calling `setcfg()?` Is `savecurtx()` being called in some way?

jarun (changes requested):

src/nnn.c
doremiyeon: the other 2 contexts? `savecurtx()` should be called on any context switch, I think. I don't really like calling `setcfg` here, because it doesn't change the `cfg` assignment and `set_sort_flags()` already sets the other two function pointers. the important point is that the filter function pointer is set for the current context after the session is loaded. i think one of the individual commits has some more information on that

jarun: No, `entrycmpfn` and `namecmpfn`.

jarun: If I am not clear, I wanted to understand how `entrycmpfn` and `namecmpfn` are being set without `setcfg` not being called in the current code?

N-R-K:
> this should fix more crashes related to switching between contexts with
different types of filters active. particularly when switching from a regex
filtered context to a string filtered one.

I can reproduce this by compiling with ASan:

```console
$ make CFLAGS_OPTIMIZATION='-g3 -fsanitize=address,undefined'
```

And then switching between string and regex filtered context.

Commit https://github.com/jarun/nnn/pull/1970/commits/0beb2b8b6dc80827a27d235f26d2f0c489fb21fd fixes the issue and looks good to me. I don't know about the other commit since I don't use sessions and haven't tested it.

N-R-K:
> Commit [0beb2b8](https://github.com/jarun/nnn/commit/0beb2b8b6dc80827a27d235f26d2f0c489fb21fd) fixes the issue and looks good to me. I don't know about the other commit since I don't use sessions and haven't tested it.

Actually, nevermind. I only tested the 

```c
filterfn = cfg.regex ? &visible_re : &visible_str;
```

Part of that commit. I haven't tested the changes in `load_session()`.

src/nnn.c
N-R-K: > If I am not clear, I wanted to understand how `entrycmpfn` and `namecmpfn` are being set without `setcfg` not being called in the current code?

Okay, just tested it out and the answer is that they're not being set. To reproduce:

1. Open nnn
2. Switch to regex filtering
3. Save session (e.g name it "regex")
4. Close nnn
5. Open nnn again (with string filter being default)
6. Load "regex" session
7. Try to search with regex, e.g `^s` and notice that there's no match even though it should match `src/` folder.

image

With the current PR, if I repeat steps 5-7 then the regex works properly.

image

I think the patch looks fine.

N-R-K (approved):

src/nnn.c
N-R-K: > I wanted to understand how `entrycmpfn` and `namecmpfn` are being set without `setcfg` not being called in the current code?

Ah, wait. I didn't read carefully. `entrycmpfn` and `namecmpfn` are being set inside the `set_sort_flags('\0');`.

> I don't really like calling setcfg here, because it doesn't change the cfg assignment and set_sort_flags() already sets the other two function pointers.

Me neither after looking a bit more. I've changed it back to only doing `filterfn` now.

N-R-K (approved):

jarun:
Thank you!

PR #1971, opened 2024-12-15
xreadline by word movement

doremiyeon:
is this useful? added by-word navigation and deletion bindings for xreadline()
- **extend xreadline() with some by-word movement**
- **xreadline ctrl-w memmove less**

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -180,6 +180,7 @@
 #define MAX(x, y)       ((x) > (y) ? (x) : (y))
 #define ISODD(x)        ((x) & 1)
 #define ISBLANK(x)      ((x) == ' ' || (x) == '\t')
+#define ISSPACE(x)      (ISBLANK(x) || (x) == '\n' || (x) == '\r' || (x) == '\f' || (x) == '\v')
 #define TOUPPER(ch)     (((ch) >= 'a' && (ch) <= 'z') ? ((ch) - 'a' + 'A') : (ch))
 #define TOLOWER(ch)     (((ch) >= 'A' && (ch) <= 'Z') ? ((ch) - 'A' + 'a') : (ch))
 #define ISUPPER_(ch)    ((ch) >= 'A' && (ch) <= 'Z')
@@ -3717,13 +3718,13 @@ static char *xreadline(const char *prefill, const char *prompt)
 				continue;
 			case CONTROL('W'):
 				printmsg(prompt);
-				do {
-					if (pos == 0)
-						break;
-					memmove(buf + pos - 1, buf + pos,
-						(len - pos) * WCHAR_T_WIDTH);
-					--pos, --len;
-				} while (buf[pos - 1] != ' ' && buf[pos - 1] != '/'); // NOLINT
+				lpos = pos;
+				while (pos > 0 && ISSPACE(buf[pos-1]))
+					--pos;
+				while (pos > 0 && !ISSPACE(buf[pos-1]))
+					--pos;
+				memmove(buf + pos, buf + lpos, (len - lpos) * WCHAR_T_WIDTH);
+				len -= lpos - pos;
 				continue;
 			case CONTROL('K'):
 				printmsg(prompt);
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -100,6 +100,7 @@
 #include <time.h>
 #include <unistd.h>
 #include <stddef.h>
+#include <wctype.h>
 #include <stdalign.h>
 #ifndef __USE_XOPEN_EXTENDED
 #define __USE_XOPEN_EXTENDED 1
@@ -3637,7 +3638,7 @@ static void addcmdtohist(char *cmd)
 /* Show a prompt with input string and return the changes */
 static char *xreadline(const char *prefill, const char *prompt)
 {
-	size_t len, pos;
+	size_t len, pos, lpos;
 	int x, r;
 	const int WCHAR_T_WIDTH = sizeof(wchar_t);
 	wint_t ch[1];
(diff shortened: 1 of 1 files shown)


jarun:
Can you explain each of the behavioural changes for me to take a call?

src/nnn.c
```diff
@@ -3716,13 +3716,17 @@ static char *xreadline(const char *prefill, const char *prompt)
 				continue;
 			case CONTROL('W'):
 				printmsg(prompt);
+				lpos = pos;
 				do {
-					if (pos == 0)
+					if (pos <= 0)
```
N-R-K, line 3721: `pos` can't be less than 0 since it's unsigned.

src/nnn.c
```diff
@@ -3716,13 +3716,17 @@ static char *xreadline(const char *prefill, const char *prompt)
...
-					memmove(buf + pos - 1, buf + pos,
-						(len - pos) * WCHAR_T_WIDTH);
-					--pos, --len;
+					--pos;
 				} while (buf[pos - 1] != ' ' && buf[pos - 1] != '/'); // NOLINT
+				if (pos < lpos) {
+					if (len > lpos)
+						memmove(buf + pos, buf + lpos, (len - lpos) * WCHAR_T_WIDTH);
+					len -= lpos - pos;
+				}
```
N-R-K, line 3729: I think the previous code was fine. `prompt` should never get large enough for multiple memmoves to become noticeable.

Also, I don't like the `len > lpos` special casing. `len` can't be bigger than `lpos`, only equal. And if `len == lpos` then the memmove with 0 length argument would've gracefully handled that as a no-op anyways.

src/nnn.c
```diff
@@ -3745,8 +3749,50 @@ static char *xreadline(const char *prefill, const char *prompt)
 				pos = 0;
 				continue;
 			case ESC: /* Exit prompt on Esc, but just filter out Alt+key */
-				if (handle_alt_key(ch) != ERR)
+				if (handle_alt_key(ch) != ERR) {
+					switch (*ch) {
+					case 'd':
+						printmsg(prompt);
```
N-R-K, line 3755: Is this supposed to be here?

N-R-K (changes requested):
Untested, but some cursory review.

src/nnn.c
N-R-K: Okay, nevermind. I see what's going on. I thought it was some debug print message mistakenly left behind.

N-R-K:
I've pushed some changes:

1. Removed the superfluous checks. Just make the code gracefully work with 0.
2. `isalnum` (and all the other `ctype.h` functions) only accept argument in the rage `[0, 255]`, we can't feed it `wchar_t` arguments. Added custom macro for it instead.
3. We might be dealing with unicode chars, so instead of `isalnum` I changed the checks to `isspace` which would work on unicode chars instead of just ascii.
4. I notice a bug in the current `Ctrl+W` handling, we only delete a single space if there's multiple before the cursor while readline will remove all. I've fixed it, but it also no longer treats `/` as a boundary (neither does readline).

@jarun do you want `xreadline` to follow readline's behavior or do you want to keep the custom `/` as word boundary behavior? If you want the custom behavior, do you want it just for `Ctrl+w` or the rest as well?

src/nnn.c
doremiyeon: sure, it's just a habit. i prefer specifying stricter conditions to safeguard against type changes or the like.

src/nnn.c
doremiyeon: len is always bigger than lpos unless cursor is at final position. it's true that memmove is called with 0 at len == lpos. semantics don't change, just saves the call.

doremiyeon:
> Can you explain each of the behavioural changes for me to take a call?

I added Meta-b, Meta-f conforming to https://www.gnu.org/software/bash/manual/bash.html#index-forward_002dword-_0028M_002df_0029 as well as the corresponding deletion commands, M-d and M-bspc.

They move cursor by word boundary, where a word is defined to be composed of letters and digits.

doremiyeon:
@N-R-K 

> I've pushed some changes:
>     3. We might be dealing with unicode chars, so instead of `isalnum` I changed the checks to `isspace` which would work on unicode chars instead of just ascii.

I hadn't considered unicode, but could isspace be turned into something to include all the other non-word (non alphanum) characters within ascii range as well?

N-R-K:
> but could isspace be turned into something to include all the other non-word (non alphanum) characters within ascii range as well?

What meaningful difference does it make?

doremiyeon:
> > but could isspace be turned into something to include all the other non-word (non alphanum) characters within ascii range as well?
> 
> What meaningful difference does it make?

the word boundaries differ, do they not? many filenames have underscores in  them, or dashes. navigating between those is the whole point of these movements. paths contain slashes, same thing.

also: strictly speaking, Ctrl-w explicitly differs from Meta-bspc by what it considers a word boundary. whitespace vs non-alphanumerical (including unicode, not sure if it distinguishes between unicode "special characters" and the rest off the top of my head).

N-R-K:
> the word boundaries differ, do they not? many filenames have underscores in them, or dashes. navigating between those is the whole point of these movements.

Ok, I see. I've switched to `iswalnum`. Hopefully it doesn't contain any weird pitfalls like the `ctype.h` counterparts do.

> strictly speaking, Ctrl-w explicitly differs from Meta-bspc by what it considers a word boundary.

I've left out `ctrl-w` change for now, I'll look into it later. But first I'm going to wait for @jarun to reply on whether he wants to keep the custom `/` behavior.

> edit: there's also iswalnum() that "should just work."

Yes, I thought about it just now as well.

jarun:
> do you want xreadline to follow readline's behavior or do you want to keep the custom / as word boundary behavior? If you want the custom behavior, do you want it just for Ctrl+w or the rest as well?

We can keep it the same as readline.

N-R-K:
> We can keep it the same as readline.

Okay, done.

---

@doremiyeon can you test it out and make sure everything still works as intended?

jarun:
I am merging this. Can you please list the new keybinds and the intended behaviour so I can try them out?

N-R-K:
New bindings are `M-f`, `M-b`, `M-d` and `M-Backspace/Del`.
Intended behavior can be found here: https://www.gnu.org/software/bash/manual/bash.html#index-forward_002dword-_0028M_002df_0029.

(Remember to compile with `O_NORL=1` if you're going to test it out).

`C-w` was also changed to match readline's behavior. Details in the commit message: https://github.com/jarun/nnn/pull/1971/commits/942afdf4454743be6065e022fc31271c75c0f01c

jarun:
Basic operations work fine for me.

PR #1973, opened 2024-12-19
Check NULL from abspath()

peter15914:
abspath() returns NULL in some cases and its return value is always checked for NULL.


jarun:
Thank you!

PR #1979, opened 2025-01-05
Follow symlink in fzplug plugin.

0fflineuser:
Hi,
my `~/.config/nnn/plugins` folder is symlinked from this repo that I use it as a submodule of my dotfiles.
This make it so that the `find` command in the fzplug plugin follows symlinks.


N-R-K (approved):
Looks fine.

PR #1985, opened 2025-01-27
Allow lowdown to use complete terminal width

psicho2000:
By default, lowdown uses:
* `--term-column`: 72 if running in a pipe
* `--term-width`: max(80, `--term-columns)`

See `man lowdown`.

This PR aims to use the full potential.


N-R-K:
> For the failing GitHub Action I have opened a separate issue: #1986

Don't worry about that. I'll look into it later. Please revert the last two commits.

jarun:
Thank you!

PR #1988, opened 2025-02-02
Allow lowdown to use complete terminal width

psicho2000:
In accordance with 48bc670cac104266395d389f1ddecbd5945be37d, we can maximize lowdown in plugin nuke aswell.


jarun:
Thank you!

PR #1989, opened 2025-02-02
Use jq to preview json if installed

psicho2000:
Currently, we're using bat to preview json (if bat is available). Bat is providing a language for json, but the support is rather poor.
`jq` instead does the job better if it exists.

In comparison, we do something similar in [nuke](https://github.com/jarun/nnn/blob/master/plugins/nuke#L251).


jarun:
Thank you!

PR #1990, opened 2025-02-02
fix: enable flags for NNN_PREVIEWIMGPROG

psicho2000:
Following the [comment](https://github.com/jarun/nnn/blob/master/plugins/preview-tui#L455), we wanted to be able to include command flags.

One use case could be using `timg` inside of a sixel-enabled tmux which needs the flag `-ps`:
`export NNN_PREVIEWIMGPROG="timg -ps"`

With the current solution, we would get:
```
preview-tui: row 456: timg -ps: Command not found.
```

This PR aims to fix that.


N-R-K (approved):

jarun:
Thank you!

PR #1994, opened 2025-02-04
fix/paste from wsl

psicho2000:
1. In Windows, lines are separated by CRLF. We need to remove the CR (carriage return) in order for IFS to work in all cases (i.e., unixoid and bsd like OS on one hand and Windows on the other hand).
2. When calling the script, we never get to know why something breaks. Therefore, we are adding error messages. This is of value if we call the script via fzplug which prints the error (or, wenn quitting from nnn).
3. I am honestly unsure what value this script provides. I see that it successfully writes to the selection file (reproducible), but nnn never seems to read from that file, it only writes into it. @jarun could you provide some insight about that?

### DIFF
--- a/plugins/x2sel
+++ b/plugins/x2sel
@@ -8,7 +8,7 @@
 #   - pbpaste (macOS)
 #   - termux-clipboard-get (Termux)
 #   - powershell (WSL)
-#   - cygwim's /dev/clipboard (Cygwin)
+#   - cygwin's /dev/clipboard (Cygwin)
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

psicho2000:
@jarun: Thanks for merging! 
Regarding my question above: is nnn even supposed to read from the selection file, and if so, how?

jarun:
Please follow `selpath` in the code. It is read in several use cases, e.g. move, copy, list selection etc.

PR #1997, opened 2025-02-08
feat: improve autojump

psicho2000:
In case of z.lua with fzf, the experience can be improved: currently, we need to potentially input twice. First for the initial query, second for filtering. We can improve this in two ways:

1. Delegate the selection process completely to fzf: pipe all dirs into fzf and let it select the entry after the last slash
2. Remove the weightings - they are not needed while selecting.

### DIFF
--- a/plugins/autojump
+++ b/plugins/autojump
@@ -15,7 +15,7 @@
 # to make z.lua work, you need to set $NNN_ZLUA to the path of script z.lua
 #
 # Shell: POSIX compliant
-# Authors: Marty Buchaus, Dave Snider, Tim Adler, Nick Waywood
+# Authors: Marty Buchaus, Dave Snider, Tim Adler, Nick Waywood, Markus Meier
 
 if [ ! -p "$NNN_PIPE" ]; then
     printf 'ERROR: NNN_PIPE is not set!'
@@ -24,13 +24,13 @@ if [ ! -p "$NNN_PIPE" ]; then
 fi
 
 if type jump >/dev/null 2>&1; then
-    printf "jump to : "
+    printf "Jump to: "
     IFS= read -r line
     # shellcheck disable=SC2086
     odir="$(jump cd ${line})"
     printf "%s" "0c$odir" > "$NNN_PIPE"
 elif type autojump >/dev/null 2>&1; then
-    printf "jump to : "
+    printf "Jump to: "
     read -r dir
     odir="$(autojump "$dir")"
     printf "%s" "0c$odir" > "$NNN_PIPE"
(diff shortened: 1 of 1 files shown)


plugins/autojump
```diff
@@ -10,12 +10,16 @@
 #   - OR z (fish) - https://github.com/jethrokuan/z (z requires fzf)
 #   - OR z.lua - https://github.com/skywind3000/z.lua (z.lua can enhanced with fzf)
 #
+# Additionally:
+#   - sed
+#   - awk
```
N-R-K, line 15: sed and awk are standard unix tools. There's no need to document them unless it's using some non standard gnu extension.

N-R-K (changes requested):

plugins/autojump
```diff
@@ -24,33 +28,35 @@ if [ ! -p "$NNN_PIPE" ]; then
 fi
 
 if type jump >/dev/null 2>&1; then
-    printf "jump to : "
+    printf "Jump to: "
```
jarun, line 31: Why capital `J`?

jarun (changes requested):

plugins/autojump
psicho2000: Normally, input forms start with capital letters. But I don't insist on that. Should I revert to small `j`?

jarun: Yes, please. Generally the scripts are written with small.

plugins/autojump
```diff
@@ -24,33 +24,35 @@ if [ ! -p "$NNN_PIPE" ]; then
...
-    	printf "%s" "0c$odir" > "$NNN_PIPE"
+        odir="$(zoxide query -i --)"
+        printf "%s" "0c$odir" > "$NNN_PIPE"
     else
-	printf "jump to : "
-    	read -r dir
-    	odir="$(zoxide query -- "$dir")"
-    	printf "%s" "0c$odir" > "$NNN_PIPE"
+        printf "jump to: "
+        read -r dir
```
jarun, line 43: Can you convert the tabs to 4 spaces in the file?

jarun (changes requested):

plugins/autojump
psicho2000: Oh, that actually already happened in https://github.com/jarun/nnn/pull/1997/commits/c8435a5efd0331366a4cd21d254fc50bf4cbe63a. So the file is uniform now and contains only spaces.

jarun (approved):

jarun:
Thank you!

PR #2004, opened 2025-02-19
Document sort actions

MatthiasKunnen:
The actions in the _Sort toggles_ (`t`) prompt were not very clear to me and lead to a lot of looking up with some confusion:
`'a'u/'d'u/'e'xt/'r'ev/'s'z/'t'm/'v'er/'c'lr/'^T'?`

So I dove into the source to find out what all the options do and added a help option.
Now, when responding to the prompt with `?`, the info of the options is displayed. This follows the _Don't memorize!_ approach taken by nnn.

The same approach as the general help is employed. I have looked into reusing the same code for both help functions but did not find a clean way to do it.

Regarding the order of the usage info, I debated between same order as the prompt, alphabetical, and grouped by type (e.g. all file size/disk usage together) and decided on alphabetical. If you would like prompt order, or any other order, let me know.

The binary size did not increase, to my surprise, it stayed at 126 440 B (compiled on x64 Arch Linux, gcc 14.2.1 20250207).

### DIFF
--- a/misc/auto-completion/fish/nnn.fish
+++ b/misc/auto-completion/fish/nnn.fish
@@ -38,7 +38,7 @@ complete -c nnn -s R    -d 'disable rollover at edges'
 complete -c nnn -s s -r -d 'load session by name' -x -a '@\t"last session" (ls $sessions_dir)'
 complete -c nnn -s S    -d 'persistent session'
 complete -c nnn -s t -r -d 'timeout in seconds to lock'
-complete -c nnn -s T -r -d 'a d e r s t v'
+complete -c nnn -s T -r -d 'sort order' -x -a "a\t'apparent disk usage' d\t'disk usage' e\t'extension' r\t'reverse' s\t'size' t\t'time' v\t'version'"
 complete -c nnn -s u    -d 'use selection (no prompt)'
 complete -c nnn -s U    -d 'show user and group'
 complete -c nnn -s V    -d 'show program version and exit'
--- a/misc/auto-completion/zsh/_nnn
+++ b/misc/auto-completion/zsh/_nnn
@@ -37,7 +37,7 @@ args=(
     '(-s)-s[load session]:session name'
     '(-S)-S[persistent session]'
     '(-t)-t[timeout to lock]:seconds'
-    '(-T)-T[sort order]:key:(( a\:"apparent disk usasge" d\:"disk usage" e\:"extension" r\:"reverse" s\:"size" t\:"time" v\:"version" ))'
+    '(-T)-T[sort order]:key:(( a\:"apparent disk usage" d\:"disk usage" e\:"extension" r\:"reverse" s\:"size" t\:"time" v\:"version" ))'
     '(-u)-u[use selection (no prompt)]'
     '(-U)-U[show user and group]'
     '(-V)-V[show program version and exit]'
--- a/nnn.1
+++ b/nnn.1
@@ -1,4 +1,4 @@
-.Dd Aug 26, 2024
+.Dd February 22, 2025
 .Dt NNN 1
 .Os
 .Sh NAME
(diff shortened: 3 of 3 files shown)


src/nnn.c
```diff
@@ -689,7 +689,7 @@ static const char * const messages[] = {
 	"['l's]/'o'pen/e'x'tract/'m'nt?",
 	"keys:",
 	"invalid regex",
-	"'a'u/'d'u/'e'xt/'r'ev/'s'z/'t'm/'v'er/'c'lr/'^T'?",
+	"'a'u/'d'u/'e'xt/'r'ev/'s'z/'t'm/'v'er/'c'lr/'^T'/?",
```
jarun, line 692: The question mark is literally a question mark. In fact we can remove it.

The short forms are chosen in such a way that it should not take long to get familiar with the them.

Adding the details adds to the binary size and we want to avoid that. Please add a new section `SORT` below the `FILTERS` section in the man page and add the same details there.

misc/auto-completion/zsh/_nnn
```diff
@@ -37,7 +37,7 @@ args=(
     '(-s)-s[load session]:session name'
     '(-S)-S[persistent session]'
     '(-t)-t[timeout to lock]:seconds'
-    '(-T)-T[sort order]:key:(( a\:"apparent disk usasge" d\:"disk usage" e\:"extension" r\:"reverse" s\:"size" t\:"time" v\:"version" ))'
+    '(-T)-T[sort order]:key:(( a\:"apparent disk usage" d\:"disk usage" e\:"extension" r\:"reverse" s\:"size" t\:"time" v\:"version" ))'
```
jarun, line 40: Please add this for the fish shell completion script as well.

src/nnn.c
```diff
@@ -6511,6 +6511,44 @@ static int set_sort_flags(int r)
 	return r;
 }
 
+static void show_sort_help()
+{
+    static const char helpstr[] = {
+        "Set sort order usage:\n"
```
jarun, line 6517: Please move this to the man page.

jarun (changes requested):

src/nnn.c
MatthiasKunnen: I'm aware that question mark was literal and thought to change it to an action. This is similar to other programs such as git where the `?` can be used to explain the options.

> The short forms are chosen in such a way that it should not take long to get familiar with them.

Perhaps this is the case, though in my opinion this prompt is not friendly to new or infrequent users. It promotes memorization of all options rather than having the information at your fingertips as is the case when typing `?` in the main view. Instead of a quick check with `?`, not knowing the option to pick becomes; opening a new terminal, `man nnn`, finding the relevant section, finding the option, closing the terminal, and then using it in nnn. This can be especially unpleasant when SSH'ed into another system.

I am agreeable to your plea to keep the binary size small. This is why I checked how much the binary would increase by. If there were to be a significant increase in size, I would agree that it would be an argument to not include it in the binary. However, my compilation of nnn revealed that the binary size is **not** impacted by this PR. Its size stays the exact same.

If binary size is the only reason for not including this help text in the program, would its non-impact be something that could make you reconsider? 

Eager to hear your thoughts

misc/auto-completion/zsh/_nnn
MatthiasKunnen: I'll have a look

src/nnn.c
MatthiasKunnen: I'll be looking to add this to both wiki and man page

src/nnn.c
jarun: You are not seeing the size increase because the final binary size may  include some padding. But the extra characters and the additional logic would take some bytes.

jarun: We should be fine with the documentation in the man page, wiki and completion scripts. No need to add in the binary as well.

For the wiki, please update here: https://github.com/jarun/nnn/wiki/Usage#configuration

MatthiasKunnen: I'll make the changes as requested.

In addition, would you accept a user patch (https://github.com/jarun/nnn/tree/master/patches) that implements the usage info in the binary? If so, I would add a commit that turns the source changes in this PR into a user patch.

jarun: Not required.

MatthiasKunnen: My question is: If I made one, would you accept/consider it? I'm interested in making one but there is no point if you dont want it

N-R-K: Almost no one who's savvy enough to be compiling their own binary is going to be benefiting from this.

Personally, I think a help string makes sense. But I don't like the situation with creating a tmp file and calling an external pager.

misc/auto-completion/fish/nnn.fish
```diff
@@ -38,7 +38,7 @@ complete -c nnn -s R    -d 'disable rollover at edges'
 complete -c nnn -s s -r -d 'load session by name' -x -a '@\t"last session" (ls $sessions_dir)'
 complete -c nnn -s S    -d 'persistent session'
 complete -c nnn -s t -r -d 'timeout in seconds to lock'
-complete -c nnn -s T -r -d 'a d e r s t v'
+complete -c nnn -s T -r -d 'sort order' -x -a "a\t'apparent disk usage' d\t'disk usage' e\t'extension' r\t'reverse' s\t'size' t\t'time' v\t'version'"
```
N-R-K, line 41: I don't use fish. I'm assuming you tested this to be working, right?

src/nnn.c
jarun: > If I made one, would you accept/consider it?

No.

MatthiasKunnen: > Almost no one who's savvy enough to be compiling their own binary is going to be benefiting from this.

Yeah, that was a concern of mine too. This is not a big feature that would convince someone to go through manual compilation. Thanks for your input.

> > If I made one, would you accept/consider it?
> 
> No.

Thanks for the straightforward, albeit succinct, reply :)

I'll push the man page changes soon, after a last review.

misc/auto-completion/fish/nnn.fish
MatthiasKunnen: I did test it though I don't use fish either. I'll see if I can get you some screenshots

N-R-K: > I'll see if I can get you some screenshots

No need. I'll take your word on it. Just making sure that it's tested by *someone* at least.

MatthiasKunnen: Flag completion
image

Flag option completion
image

src/nnn.c
MatthiasKunnen: Man page changes pushed

src/nnn.c
MatthiasKunnen: The changes to the code have been reverted

jarun:
Can you please squash all the commits into one?

MatthiasKunnen:
> Can you please squash all the commits into one?

Done

jarun:
Thank you!

MatthiasKunnen:
Happy to contribute! :)

src/nnn.c
MatthiasKunnen: @jarun, I thought I was gonna be able to edit the wiki after this PR was merged but that does not seem to be the case. If you'd like to me add it to the wiki, could you grant me access? If you prefer not to do so, all good, no worries

jarun: I've updated the wiki.

PR #2005, opened 2025-02-24
Add opn plugin

MatthiasKunnen:
I would like to add a plugin that integrates with a tool I made called [`opn`](https://github.com/MatthiasKunnen/opn).

example_open_pdf

It is a terminal application picker that works using the [XDG desktop spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/). It can be seen as an alternative to the `nuke` plugin provided by nnn that is out-of-the-box customized to what the user has installed.

One question I had while making the plugin is the name, do users often add the plugin dir to their path? If they do, this will cause an infinite loop where the opn script will call itself. I can change the name to `opn_nnn` or similar should this be desired.

If you have any questions, or even feature ideas, I'll happily answer them

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -52,6 +52,7 @@ Plugins extend the capabilities of `nnn`. They are _executable_ scripts (or bina
 | [nmount](nmount) | Toggle mount status of a device as normal user | sh | pmount (optional), udisks2 |
 | [nuke](nuke) | Sample file opener (CLI-only by default) | sh | _see in-file docs_ |
 | [oldbigfile](oldbigfile) | List large files by access time | sh | find, sort |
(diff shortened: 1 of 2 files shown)


N-R-K:
> a tool I made called [`opn`](https://github.com/MatthiasKunnen/opn).

Looks cool. Though it seems like there isn't many users (yet). I'd recommend hosting the plugin inside opn repo under `extra/nnn/` or some directory like that. If it gains a larger userbase then we can consider adding it here. ATM the plugin is too trivial and for a tool with not much userbase.

MatthiasKunnen:
That's fair, it is very new, I created it because I needed an application picker to use with nnn. I had hoped to share it so others could use it too

N-R-K:
> I had hoped to share it so others could use it too

You can create a "discussion" under [show and tell](https://github.com/jarun/nnn/discussions/categories/show-and-tell) section for this purpose.

PR #2006, opened 2025-02-27
Explicitly declare support for opening a path

MatthiasKunnen:
While the nnn `.desktop` file has `MimeType=inode/directory`, it is missing a field code to specify how the command line should be constructed with the directory's path. This can cause application launchers to disregard nnn as a file manager candidate.

See <https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html>.


N-R-K (approved):

jarun:
Thank you!

Issue #1350, opened 2022-04-28
Open command always in CLI mode, don't prompt for cli/gui


freijon:
I use `nnn` via ssh on a VM without GUI. When I open a file with the "o" key, `nnn` always prompts me to choose if I want to open in _cli_ or _gui_ mode. I don't want to see this option at all, I want `nnn` to always assume _cli_ was chosen.

I read the manual and this is what I tried:

- Run `nnn -c`
- Setting ENV `NNN_OPENER=/home/<user>/.config/nnn/plugins/nuke` (GUI = 0)

Am I missing something here?

N-R-K:
You should open files with `enter` or `l`.
`o` is just a fallback when you don't want to use your default opener.

freijon:
And this is also exactly what I do to edit files.
However, I use `o` to run commands like:

- git log
- git add
- git difftool
- rclone
- custom commands from my .local/bin
- ...

running `o` is one of my most common actions... And I always have to select _cli_ first, which becomes annoying.

N-R-K:
You can easily write a plugin for this. https://github.com/jarun/nnn/tree/master/plugins has detailed instructions on how to do it.

freijon:
Indeed, and I do use plugins. But I need to assign plugins to a letter and I need to remember exactly which letter does what. A lot of letters are already used for other stuff. Also, sometimes I like to pass custom arguments/options like git tags to the open-command that I can't hard-code in a plugin. It seems like a workaround for something that could be (seemingly) easily fixed without much effort and side-effects. I thought about an ENV variable `NNN_OPEN_WITH` which can be set to _CLI_ would that be totally out of the question?

N-R-K:
Plugins don't have to be hardcoded, you can read arbitrary command from the cli into a variable/array and execute it. I can't help you any further at this time as I'm a bit busy with other stuff.

0xACE:
Not near a machine atm, but https://github.com/jarun/nnn/blob/fd998b3045fbc8ef3a97d2f55a80781725f76919/src/nnn.c#L6031-L6042

just change that into

```
mkpath(path, name, newpath);
spawn(tmp, newpath, NULL, NULL, F_CLI);
```

in your local build, you will be able to rebase this on master without much issue whenever you upgrade, otherwise you could add a hook for `getenv(DISPLAY)` or something along those lines to deduce if `GUI` is even a option, but I would read what those flags mean before going so far…

freijon:
You were faster...
[...]

PR #2007, opened 2025-02-27
rework SEL_OPENWITH, add example plugin openwith

Darukutsu:
This solves 3 problems which user might have upon pressing 'o' key:
1) forgot name or typo of gui/cli program, can use fzf, rofi... instead
2) exec arbitrary commands over selection let's say `git add` or anything more complicated
3) no more need for custom patch https://github.com/jarun/nnn/issues/1350

Script `openwith` doesn't clear `.selection` because I was not sure if I should give it `$NNN_SEL`.

What do you think?

### DIFF
--- a/plugins/openwith
+++ b/plugins/openwith
@@ -1,0 +1,80 @@
+#!/usr/bin/env sh
+
+# Description: do operations over file selection
+#
+# After pressing enter \`$EDITOR\` will be opened, so you can edit command. wq to continue.
+# DONT FORGET in place of files you want add \`{}\` instead!!!"
+# Files will be parsed one by one instead of all at once
+#
+# Dependencies: fzf, xargs
+# Shell: sh
+# Author: Arun Prakash Jana, Darukutsu
+
+# setting this to 1 will result in using history as additional source of runners
+# it's recommended to enable editor as well
+history=0
+# setting this to 1 will result in opening editor to further edit command
+editor=0
+selection=${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection}
+
+if [ $editor -eq 0 ]; then
+  header="Choose command to run over your file/selection"
+else
+  header="After pressing enter \`$EDITOR\` will be opened, so you can edit command. Add \`{}\` in place of files!!! If \`{}\` are omitted file/selection will be appended at the end of command-string"
+fi
+
+if type fzf >/dev/null 2>&1; then
+  fuzzy=fzf
+# TODO: more examples with rofi etc
+#if type rofi >/dev/null 2>&1; then
+#  fuzzy=rofi
+else
+  exit 1
+fi
+
+shellname="$(basename "$SHELL")"
+
+tmpfile=$(mktemp)
+echo "$PATH" | xargs -d':' -I{} find {} -type f -perm -1 >>"$tmpfile"
+if [ $history -eq 1 ]; then
+  if [ "$shellname" = "bash" ]; then
+    hist_file=${HISTFILE:-$HOME/.bash_history}
+    cat "$hist_file" >>"$tmpfile"
+  elif [ "$shellname" = "fish" ]; then
+    hist_file="$HOME/.local/share/fish/fish_history"
+    grep "\- cmd: " "$hist_file" | cut -c 8- >>"$tmpfile"
+  fi
+fi
+entry="$("$fuzzy" --header="$header" --tac <"$tmpfile")"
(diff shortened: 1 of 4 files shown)


plugins/openwith
```diff
@@ -0,0 +1,80 @@
...
+    command="$(cat "$tmpfile")"
+    if [ $history -eq 1 ]; then
+      echo "$command" >>"$hist_file"
+    fi
+
+    if ! [ -s "$selection" ]; then
+      printf "%s/%s" "$2" "$1" >"$selection"
+    fi
+
+    xargs -0 -I{} $command <"$selection"
```
Darukutsu, line 72: here we should maybe do to allow using xargs within xargs
```
xargs -0 -I{} sh -c "$command" <"$selection"
```

N-R-K (changes requested):
This takes away a built-in functionality and turns it into a plugin with external dependencies. This is a severe regression. Not everyone has plugins installed or might be using nnn in a remote environment with no configuration.

Why can't this be a regular plugin, which you can bind to `o` and then invoke with alt+key (alt+o in this case)?

Also, nnn already supports running arbitrary shell commands over selection via the builtin shell prompt (bound to `]`). E.g `echo "%j"` will run the command for each file and `"%J"` will run it for all files at once.

Darukutsu:
> This takes away a built-in functionality...

if variable isn't defined it works as previous implementation

> echo "%j" will run the command for each file and "%J" will run it for all files at once.

That's great, didn't know about this. I see now, it's documented [wiki/Concepts#prompts](https://github.com/jarun/nnn/wiki/Concepts#prompts)

Surely this can be plugin but not everyone uses 'o' key and this could extend functionality. I know this is matter of person, for me 'o' was dead key, because of limited auto-completion (almost none). Most of the time in my workflow I do `! ^] Shell` to spawn to original shell (i use [ble.sh](https://github.com/akinomyoga/ble.sh)) which has was more capability.

Plugin itself is half-baked nothing interesting to add, just provides fzf with $PATH or history, then you can edit command in your editor and run it over selection. I just wanted to provide example how to work with selection. I think it solves `1)` and `3)` definitely, no more custom patches.

Maybe you should make repository/directory with not only plugins in one place, but also custom patches, would be nice. I like idea that nnn feels like lego, you start minimal you grow to something bigger. I could go over my free time, collect all interesting patches from issues, discussion... if you're interested

> echo "%j" will run the command for each file and "%J" will run it for all files at once.

You spawned idea in my head. Instead of doing plugin like I did, I could define `%j` and `%J` in blesh, something like this:
```
#.bashrc
  ## we could bind this to enter
  function ble/widget/my_custom/nnn_complete() {
    ble/widget/vi_nmap/copy-current-line
    readline_line="${_ble_edit_kill_ring[0]}"
    ble/widget/kill-line

    if echo "$readline_line" | grep "%j" >/dev/null; then
      ble/widget/insert-string "cat ${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection} | xargs -0 -I%j sh -c \"$readline_line\""
    elif echo "$readline_line" | grep "%J" >/dev/null; then
      #elif echo "$READLINE_LINE" | grep -E "(?<!\\)%J"; then
      readline_line=$(echo "$readline_line" | sed 's/%J//')
      ble/widget/insert-string "cat ${NNN_SEL:-${XDG_CONFIG_HOME:-$HOME/.config}/nnn/.selection} | xargs -0 $readline_line"
    else
      ble/widget/insert-string "$readline_line"
      ## or comment this out and don't bind it to enter
      # echo '%j %J not specified, aborting'
    fi
    unset readline_line
    ble/widget/accept-line
  }

  ble-bind -m vi_nmap -f 'C-/ C-n' my_custom/nnn_complete
  ble-bind -m vi_imap -f 'RET' my_custom/nnn_complete
```

N-R-K:
> if variable isn't defined it works as previous implementation

Ah, you're right, I missed the fall though.

But in any case, I'm still not convinced that this should be built in. A plugin would also solve problem 1 & 3. In fact plugin system exists precisely so that we don't have to support every little preference of every user, they get to write a local plugin that caters to their needs.

N-R-K:
> Instead of doing plugin like I did, I could define `%j` and `%J` in blesh, something like this:

You could also just use a function like this in your shellrc (I'm assuming posix like syntax):

```sh
nnnxargs()
{
    xargs -0 "$@" < "$NNN_SEL" 
}
```

Then use this inside of the nnn spawned shell. Less hacky although a bit more typing.

PR #2035, opened 2025-05-23
Add icon for CommonJS and ES Modules

Juneezee:
- `.cjs` and `.cts` are CommonJS files
- `.mjs` and `.mts` are ECMAScript Modules files

Reference: https://nodejs.org/docs/latest/api/modules.html#enabling
Reference: https://nodejs.org/docs/latest/api/esm.html#enabling
Reference: https://www.typescriptlang.org/docs/handbook/modules/reference.html#module-format-detection


jarun:
Thank you!

PR #2038, opened 2025-06-02
nuke plugin - Faster handle files files by extension

ojio-san:
Hello there,

If we're going to handle the file by extension : don't check mimetype first. The mimetype is not necessary in that use case, and may be costly in a slow connection and/or over wifi (or watever slow media you have)

If we don't handle by extension, mime type will be checked anyway (after handling by extension)

### DIFF
--- a/plugins/nuke
+++ b/plugins/nuke
@@ -545,8 +545,8 @@ handle_bin() {
     esac
 }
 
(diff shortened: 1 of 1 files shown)


N-R-K (approved):

jarun:
Thank you!

ojio-san:
Thanks everyone for this merging this, and also for your amazing work on nnn 👍

PR #2047, opened 2025-07-26
add zellij as a supported multiplexer

TimB87:
Hi all,
this adds [zellij](https://github.com/zellij-org/zellij) as a supported multiplexer, creating a pane for the preview buffer which gets deconstructed when nnn is closed.
image


jarun:
Thank you!

PR #2052, opened 2025-08-02
feat: Improve UX for dragdrop plugin

beast989898:
This pull request introduces some quality-of-life improvements for the dragdrop plugin.

- Similarly to the kdeconnect plugin, there is a new option for system notifications when downloading files from a URL. This is useful when downloading larger files. It is disabled by default.
- With newer fontconfig versions, dragon now gives a warning. I also get some GTK warnings that interfere with the nnn UI. Sending dragon's stderr to /dev/null will prevent that issue.
- Remove unneeded backslashes from regular expression to prevent "stray \ before /" warnings.
- Use quiet flag with grep and silent flag with curl to prevent their outputs and avoid clashing with nnn's UI.
- Some small modifications to the way curl is called inside the script to better handle failed downloads. Errors will still be shown, but adding the file to the selection and the success notification will be skipped on failure.
- Strip query parameters from filenames so they are correctly added to selection.

### Before:
Dragon:
dragon-warnings

Curl:
curl-before

Pasting file with URL parameters:
url-parameters-before

### After:
Dragon:
no-warnings

Curl:
curl-after

Pasting file with URL parameters:
url-parameters-after

### DIFF
--- a/plugins/dragdrop
+++ b/plugins/dragdrop
@@ -66,11 +66,18 @@ elif [ "$resp" = "r" ]; then
     "$dnd" --print-path --target 2>/dev/null | while read -r f
     do
         if printf "%s" "$f" | grep -q '^\(https\?\|ftps\?\|s\?ftp\)://' ; then
-            if curl -fsLJOS "$f"; then
+            output=$(curl -fsLJOS "$f" 2>&1)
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #2061, opened 2025-09-09
Replace bash-specific pattern matching with POSIX-compatible syntax

vladimir-bebeshko:
Convert `+(pattern)` extended glob patterns to standard POSIX conditional expressions using `||` operators for better shell compatibility. Extended glob patterns may not be enabled in all user environments, making this change more reliable.


jarun:
Closing since we need bash for this plugin.

PR #2097, opened 2025-12-02
The →  key acts like the “e” key. useful with preview-tui.

Manix80:
This small patch keeps the normal directory navigation with →
but when the cursor is on a regular file, → behaves like e (open in editor).

This improves UX while keeping backward compatibility and minimal changes.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -7473,6 +7473,11 @@ nochange:
 		if (presel)
 			presel = 0;
 
+		/* CUSTOM: treat right-arrow (SEL_NAV_IN) as edit when hovering a file,
+		 * but keep normal directory navigation when hovering a directory. */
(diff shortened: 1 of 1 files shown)


jarun:
The option `o` is intended to block opening files (including directories) on right or `l`.

Manix80:
I didn't express myself clearly. The → does exactly the same thing as `e`, especially when preview-tui is open. Otherwise, by default, the → leaves preview-tui open, unlike `e`, which closes it.

Please note that if option `o` is not applied, the  → key will continue to navigate through the folders.

jarun:
`e` is a different key. The intention of using `o` is to block any operation.
If `-o` is applied, what is the behaviour with preview-tui and right arrow?

Manix80:
Hmm, yes, the `-o` option no longer works with the patch. I'll take a closer look at it and try to make sure the patch doesn't break the `-o` option.

PR #2102, opened 2025-12-21
fix(xdgdefault): filter application list by current file's MIME type

mysterymarshak:
Currently this plugin lists all installed applications found in `$XDG_DATA_DIRS`, regardless of whether they can actually open the selected file. And I think there's no reason to list all of them.

I updated the `find` command pipeline to filter `.desktop` files based on the detected filetype. So now plugin displays only applications that explicitly support selected file's MIME type.

### DIFF
--- a/plugins/xdgdefault
+++ b/plugins/xdgdefault
@@ -40,6 +40,8 @@ while [ -n "$dirs" ]; do
     dirs=${dirs#*:}
 done
 
+# see PR #2102
+# shellcheck disable=SC2038
 app=$(find "$@" -iname '*.desktop' -exec grep -l "MimeType=.*$ftype" {} + 2>/dev/null \
     | xargs -I {} grep '^Name=' {} /dev/null \
     | sort -u -t ':' -k 1,1 \


jarun:
shellcheck fails on CI with:

```
In plugins/xdgdefault line 43:
app=$(find "$@" -iname '*.desktop' -exec grep -l "MimeType=.*$ftype" {} + 2>/dev/null \
      ^-- SC2038 (warning): Use -print0/-0 or -exec + to allow for non-alphanumeric filenames.

For more information:
  https://www.shellcheck.net/wiki/SC2038 -- Use -print0/-0 or -exec + to allo...

```

mysterymarshak:
yeah i see, i think it triggers because of `find | xargs`
im passing output from `grep -l` (filenames separated by newlines) to `xargs -I {}` to handle filenames with spaces correctly (processing them line-by-line, -I arg)
the only scenario where this breaks is if a .desktop filename contains `\n`, which never happens with those files

N-R-K:
You can add a [comment to suppress](https://github.com/jarun/nnn/blob/9bf833ae0d4aeb366fcd87456680985cc421da39/plugins/.nmv#L20) the warning if it's false positive.

mysterymarshak:
> You can add a [comment to suppress](https://github.com/jarun/nnn/blob/9bf833ae0d4aeb366fcd87456680985cc421da39/plugins/.nmv#L20) the warning if it's false positive.

ty, did it

jarun:
Thank you!

Issue #2076, opened 2025-10-18
Preview-tui Not Able to See Directories


levone1:
`nnn` comes with excellent documentation (including a Troubleshooting section). If you have a _How to do x_ question, it may have already been answered. Wiki: https://github.com/jarun/nnn/wiki

If it looks like a local environment issue, please try to debug yourself. Debugging local setup issues is not our top priority.

If you are looking for a new feature or program option, ensure you have the correct version with the feature installed. Refer to the release notes to confirm.

Before opening an issue, please try to reproduce on latest master. The bug you noticed might have already been fixed.

Useful links:

- compile `nnn` from source - https://github.com/jarun/nnn/wiki/Usage#from-source
- debugging `nnn` - https://github.com/jarun/nnn/wiki/Developer-guides#debugging-nnn

If the issue can be reproduced on master, log it.

Please provide the environment details. **If that's missing, the issue will be closed without any cited reason.**

If we need more information and there is no communication from the bug reporter within 7 days from the date of request, we will close the issue. If you have relevant information, resume discussion any time.

--- PLEASE DELETE THIS LINE AND EVERYTHING ABOVE ---

#### Environment details (Put `x` in the checkbox along with the information)

- [ x] Operating System:  Linux Mint 22
- [ x] Desktop Environment:  Cinnamon
- [ x] Terminal Emulator:  kitty 0.4.31
- [ x] Shell:  bash
- [ ] Custom desktop opener (if applicable):
- [ x] Program options used:  any
- [ x] Configuration options set:  any
- [ x] Plugins are installed
- [ x] Issue exists on `nnn` master

#### Exact steps to reproduce the issue

Open nnn and attempt to use preview-tui plugin.

Files are previewed in preview pane, but directories show only `~`

nnndbg:
`
ln 8911: VERSION=5.1
ln 8689: g_tmpfpath=/tmp
ln 8690: tmpfplen=5
ln 8952: home=/home/disciple
ln 8637: cfgpath=/home/disciple/.config
ln 8643: cfgpath=/home/disciple/.config/nnn
ln 8671: selpath=/home/disciple/.config/nnn/.selection
ln 8960: opener=xdg-open
ln 9074: getenv(envs[ENV_VISUAL])=hx
ln 9075: getenv(envs[ENV_EDITOR])=nano
ln 9076: editor=hx
ln 9080: pager=less
ln 9084: shell=/bin/bash
ln 9086: getenv("PWD")=/home/disciple/src/nnn
[...]

PR #2104, opened 2025-12-26
Fix eza in preview-tui

babbaj:
For some reason eza outputs nothing when run inside preview-tui and only behaved weirder when I tried to debug it. Simply specifying the path as suggested in the eza issue seems to completely fix it.
https://github.com/eza-community/eza/issues/1567
this guy probably also had the same problem https://github.com/jarun/nnn/issues/2076


jarun:
Thank you!

PR #2141, opened 2026-06-07
plugins(dragdrop): add kitty dnd protocol as option

darukutsu:


plugins/dragdrop
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env sh
-
 # Description: Open a Drag and drop window, to drop files onto other programs.
 #              Also provides drag and drop window for files.
 #
-# Dependencies: dragon - https://github.com/mwh/dragon
+# Dependencies: 
+# 	- dragon (https://github.com/mwh/dragon)
+# 	- kitty (https://github.com/kovidgoyal/kitty)
```
N-R-K: It requires both now or either one? Docs should be made clearer.

plugins/dragdrop
```diff
@@ -55,37 +58,59 @@ else
...
+		    else
+			echo "Download Failed: $output"
+		    fi
+		fi
+	    elif [ -e "$f" ]; then
+		add_file "$f"
+	    fi
+	done &
+    elif [ -n "$1" ] && [ -e "$1" ]; then
+	"$dnd" "$1" 2>/dev/null &
```
N-R-K: indentation seems messed up.

N-R-K (changes requested):

plugins/dragdrop
darukutsu: Open raw, github issue. Looks fine on my end.

darukutsu: Oh its probably tabs vs spaces issue. Would love to see clang-format, shfmt config in root repo to prevent this from happening.

darukutsu: nvm figured out where was issue, i was retarded

PR #2142, opened 2026-06-10
build: make icons-hash-gen a proper target

PPN-SD:
EDIT: so the project can be cross-compiled easier

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,8 +224,12 @@ debug: $(BIN)
 norl: $(BIN)
 nolc: $(BIN)
 
-src/$(ICONS_INCLUDE): src/icons-hash.c src/icons.h src/icons-in-terminal.h
-	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -DICONS_GENERATE -o src/icons-hash-gen src/icons-hash.c
+ICONS_SRC = src/icons.h src/icons-in-terminal.h
+
+src/icons-hash-gen: src/icons-hash.c $(ICONS_SRC)
+	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -DICONS_GENERATE -o $@ $<
+
+src/$(ICONS_INCLUDE): src/icons-hash-gen $(ICONS_SRC)
 	./src/icons-hash-gen > $@
 
 install-desktop: $(DESKTOPFILE)


N-R-K:
> so it can be cross-compiled easier

The generator needs to run on the build/host system, not the target one. So I'm not sure why you would want to cross compile it. Can you elaborate a bit more about what problem this solves?

PPN-SD:
It's easier to compile native icons-hash-gen with a proper target.

N-R-K:
> It's easier to compile native icons-hash-gen with a proper target.

If I understood correctly, instead of just running `make`, you're manually first compiling the meta program and then running your cross build?

I've actually thought about using `CC_BUILD`, `CFLAGS_BUILD` etc for the generator so that those can be set to the build machine and cross compilation wouldn't require manual multi-step but I don't do cross builds myself and wasn't sure if there's any conventions around it. Does that work for your use case?

PPN-SD:
> If I understood correctly, instead of just running `make`, you're manually first compiling the meta program and then running your cross build?

That's it

> I've actually thought about using `CC_BUILD`, `CFLAGS_BUILD` etc for the generator so that those can be set to the build machine and cross compilation wouldn't require manual multi-step but I don't do cross builds myself and wasn't sure if there's any conventions around it. Does that work for your use case?

CC_BUILD can't be used as-is and requires additional handling.

jarun:
@N-R-K is this good to go?

Makefile
```diff
@@ -224,8 +224,10 @@ debug: $(BIN)
 norl: $(BIN)
 nolc: $(BIN)
 
-src/$(ICONS_INCLUDE): src/icons-hash.c src/icons.h src/icons-in-terminal.h
-	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -DICONS_GENERATE -o src/icons-hash-gen src/icons-hash.c
+src/icons-hash-gen: src/icons-hash.c
```
N-R-K: This needs to depend on `src/icons.h` and `src/icons-in-terminal.h` as well otherwise generator will be using outdated icons.

N-R-K (changes requested):

N-R-K:
> @N-R-K is this good to go?

Personally think it's better to be explicit about host vs target compiler instead: https://github.com/jarun/nnn/pull/2144.

jarun:
@N-R-K please leave a message when you are good with this.

Issue #2161, opened 2026-09-02
v5.3: leaveok() rendering optimization removes the hardware cursor from the hovered entry (breaks braille/screen reader support)


alex19EP:
#### Environment details (Put `x` in the checkbox along with the information)

- [x] Operating System: Arch Linux
- [x] Desktop Environment: GNOME
- [x] Terminal Emulator: Ghostty (`TERM=xterm-ghostty`), also reproducible under `TERM=xterm-256color`
- [x] Shell: zsh
- [ ] Custom desktop opener (if applicable): n/a
- [x] Program options used: `-d` (any option, detail mode just makes it easier to see)
- [ ] Configuration options set: none relevant
- [ ] Plugins are installed
- [x] Issue exists on `nnn` master (fcd7946)

Additional: ncurses 6.6-2, BRLTTY 6.9.1, packaged `nnn` 5.3-1.

#### Exact steps to reproduce the issue

1. Run `nnn -d` in a directory with a few files.
2. Press <kbd>j</kbd>/<kbd>k</kbd> to move between entries.
3. Observe where the terminal's **hardware cursor** ends up.

Expected: the hardware cursor sits on column 1 of the hovered entry's row.
Actual: the hardware cursor stays on the status bar and never moves.

#### Description

For context: I am blind. I use `nnn` with a screen reader and a braille display
(BRLTTY). The hardware cursor is how both of them know which entry is hovered —
the reverse-video highlight conveys nothing to either. Since v5.3 there is no
longer any way for me to tell which file is selected, so `nnn` went from usable
to unusable for me in that release.

`nnn` deliberately parks the hardware cursor on the hovered entry so that braille
displays and terminal screen readers can follow the selection. That is the
`/* Place HW cursor on current for Braille systems */` comment before the
`tocursor()` call at the end of `statusbar()`, plus the `tocursor()` in
`markhovered()`. It came from #534, the dedicated `-w` option was later removed in
cb5aefa5 specifically to make the behaviour unconditional, and it is advertised in
the README as *"Basic support for screen readers and braille displays"*.

Commit 99a6a2e2 (*"Use leaveok() to improve ncurses performance"*, shipped in v5.3
as *"optimize ncurses rendering with `leaveok()`"*) added to `initcurses()`:

```c
leaveok(stdscr, TRUE); /* Skip precise cursor placement during normal redraws */
```

`leaveok(win, TRUE)` tells ncurses it may leave the physical cursor wherever the
[...]

Issue #534, opened 2020-04-24
The hardware cursor should move to the selected item


tspivey:
**Is your feature request related to a problem? Please describe.**

When nnn starts, the hardware cursor is placed on the status line, and it doesn't move when files are selected.

**Describe the solution you'd like**

I would like the hardware cursor to move to the currently selected file as it changes, along with moving back to the status line when doing an operation such as rename.

**Describe alternatives you've considered**

**Additional context**

While the cursor is hidden, it still has a location. Screen readers and braille displays use it to tell what to focus on.

jarun:
If

1. someone works on this based on a conditional flag
2. There aren't flag checks in every redraw

I can consider this. I will add a line item.

jarun:
@tspivey this feature is added at commit f125802f4f30c5275d4d22b11aba0239f6216d58.

PR #2162, opened 2026-09-02
Revert leaveok() optimization and left-half-block hover marker (restore screen reader / braille support)

alex19EP:
Reverts both changes discussed in #2161, as requested.

**1. `Revert "Use leaveok() to improve ncurses performance"` (99a6a2e)**

`leaveok(stdscr, TRUE)` lets ncurses leave the physical cursor wherever the last
write ended rather than moving it to the logical cursor position on refresh, which
turned the deliberate `tocursor()` calls in `statusbar()` and `markhovered()` into
no-ops for the real terminal cursor. Braille displays and terminal screen readers
follow the hardware cursor to know which row is current — the reverse-video
highlight conveys nothing to them — so `nnn` lost the "Basic support for screen
readers and braille displays" feature (#534) in v5.3.

**2. `Revert "Use left half block to mark hovered file"` (9de5bc8)**

In detail mode that marker is the only *textual* indication of the hovered row.
`>` is spoken by screen readers; U+258C LEFT HALF BLOCK is announced as "left half
block" or dropped entirely, depending on the reader's symbol table. This one is
unreleased, so no released version is affected.

### Testing

Driving `nnn -d` in a pty and dumping the raw byte stream for one `j` keypress.

Before (v5.3 / master) — the redraw ends on the status bar and the cursor stays there:

```
... \r ESC[24d ... 2 ESC[39m ESC(B ESC[m
```

After this PR — a trailing `ESC[4d BS` puts the cursor on column 1 of the hovered row:

```
... \r ESC[24d ... 2 ESC[4d BS ESC[39m ESC(B ESC[m
                     ^^^^^^^^^ lands on hovered entry
```

Builds clean with `make O_NORL=0` and `make O_NERD=1 O_NORL=0` (`-Wall -Wextra
-Wshadow`, no new warnings).

If you would rather keep the optimization and gate the cursor placement behind an
option or an environment variable instead, I am happy to rework this — just say
which you prefer.

## AI Usage Disclosure

This change was developed with assistance from Claude (Anthropic). All code was
reviewed and tested by the author before submission.

### DIFF
--- a/src/nnn.c
+++ b/src/nnn.c
@@ -2439,7 +2439,6 @@ static bool initcurses(void *oldmask)
 	mouseinterval(0);
 #endif
 	curs_set(FALSE); /* Hide cursor */
-	leaveok(stdscr, TRUE); /* Skip precise cursor placement during normal redraws */
 
 	char *colors = getenv(env_cfg[NNN_COLORS]);
 
@@ -3872,7 +3871,6 @@ static int filterentries(char *path, char *lastname)
 
 	cleartimeout();
 	curs_set(TRUE);
-	leaveok(stdscr, FALSE);
 	showfilter(ln);
 
 	while ((r = get_wch(ch)) != ERR) {
@@ -4079,7 +4077,6 @@ end:
 	copycurname(lastname);
 
 	curs_set(FALSE);
-	leaveok(stdscr, TRUE);
 	settimeout();
 
 	/* Return keys for navigation etc. */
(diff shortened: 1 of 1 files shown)


jarun:
On second thoughts, I have added these under a conditional compilation flag _O_NOBRAILLE_ so that people who do not require Braille support can still use them.

Please have a look at commit 1a099bf3082ad677be91e28b0bd54c9223c4ea1f.

Issue #1957, opened 2024-11-14
quitcd not working in fish if path contains backslash


Ordoviz:
If you press `^G` to exit nnn or have [configured quitcd](https://github.com/jarun/nnn/wiki/Basic-use-cases#configure-cd-on-quit), nnn writes a `cd` command in POSIX-shell syntax into `~/.config/nnn/.lastd`. While fish syntax is similar to POSIX, they differ on their handling of [backslashes in single-quoted strings](https://fishshell.com/docs/current/language.html#quotes). In particular, if nnn writes a path ending in a backslash into `.lastd`, and you source `.lastd` in fish you get `Unexpected end of string, quotes are not balanced`.

To fix this, nnn should write the raw (unescaped) path to `~/.config/nnn/.lastd_raw` or stdout so that you don't need to parse the escaped path in fish. This would probably also be helpful for nushell, where you currently need to do [some horrible string parsing](https://github.com/jarun/nnn/blob/master/misc/quitcd/quitcd.nu#L31) to extract the path from `.lastd`.

N-R-K:
ping: @BeyondMagic

BeyondMagic:
> This would probably also be helpful for nushell, where you currently need to do [some horrible string parsing](https://github.com/jarun/nnn/blob/master/misc/quitcd/quitcd.nu?rgh-link-date=2024-11-14T17%3A28%3A15Z#L31) to extract the path from `.lastd`.

Hey, it's not terrible! It's just... tricky!

But yep, it'd be better for Nushell as well the raw path.

N-R-K:
Oh, I just realized the bug is in fish shell not nushell. Anyways...

> or stdout 

Stdout can be used by plugins, so no.

> nnn should write the raw (unescaped) path to ~/.config/nnn/.lastd_raw

The path for last directory is not hardcoded, it depends on `$NNN_TMPFILE`. Randomly adding new temp file or changing the format of the existing file isn't good. Feel free to send a PR fixing the fish shell script instead.

N-R-K:
@PatrickF1 Since you've contributed to `quitcd.fish` before do you want to take a look at this bug? Or maybe you have some ideas on how to fix it?

PatrickF1:
Hi, sorry I don't use nnn anymore.

FugueSoft:
@Ordoviz 
Don't know if this issue is open or closed but I am currently using nnn on Debian 12 and quitcd works without issue (perhaps some screen draw issues from kitty)

[...]

Issue #1546, opened 2022-12-18
Tracker


jarun:
Rolled from #1537.

### Cooking

- [x] create new file or directory (tree) on startup
- [x] improve archive, rename and create new workflows
- [x] optimize link creation
- [x] allow overwriting regular files on new empty file creation
- [x] add correct check for Wayland in clipboard plugins

### Up for grabs

None open at the time.

**Contribution [guideline](https://github.com/jarun/nnn/blob/master/misc/CONTRIBUTING.md).**

PR #2167, opened 2026-09-08
fix: handle backslashes in fish quitcd paths

ryanchou1994:
The fish quitcd wrapper currently sources the POSIX shell command written by nnn directly in fish. Backslashes have different meanings inside single quotes in the two shells: a directory ending in a backslash produces an unbalanced-quote error, while repeated backslashes can change the target path.

Read the existing command with POSIX sh and pass its resulting directory back to fish using a NUL separator. This preserves embedded and trailing newlines, and a guard keeps the current directory unchanged if the target is unavailable. The nnn output format and other shell wrappers are unchanged.

Validation on Linux with fish 3.6.0:
- The regression script covers 12 path cases plus missing-directory and missing-file handling; three backslash cases fail before the fix and all cases pass after it.
- Seven real nnn/PTY exit scenarios pass, including HOME and XDG_CONFIG_HOME configurations; the unchanged wrapper fails four of them.
- All 15 combinations in `make checkpatches` build successfully; default build with `-Werror`, fish syntax validation and Python syntax validation pass.

Run the regression with `python3 misc/test/test-quitcd-fish.py` (fish required). macOS, other fish versions and the complete upstream CI matrix were not exercised.

Refs #1957 and the corresponding item in #1546.

### DIFF
--- a/misc/quitcd/quitcd.fish
+++ b/misc/quitcd/quitcd.fish
@@ -29,8 +29,13 @@ function n --wraps nnn --description 'support nnn quit and change directory'
     # making an infinitely recursive alias
     command nnn $argv
 
-    if test -e $NNN_TMPFILE
-        source $NNN_TMPFILE
-        rm -- $NNN_TMPFILE
+    if test -e "$NNN_TMPFILE"
+        # nnn writes POSIX shell syntax, whose quoting differs from fish.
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

PR #2168, opened 2026-09-09
stop recommending allow_remote_control yes

babbaj:
`allow_remote_control yes` allows kitty to be controlled by any command that prints escape codes, including `cat`ing a file, which is a major security problem and that functionality is not used by preview-tui.

Relevant documentation: https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.allow_remote_control

The [wiki page](https://github.com/jarun/nnn/wiki/Live-previews) doesn't specifically say to use `yes` but it should also be updated to recommend using `socket-only`


jarun:
Thank you!

PR #2171, opened 2026-09-12
fix: sanitize control bytes before measuring filename width

### LINKED ISSUE #2170: Control characters in filenames truncate the rest of the displayed name
`nnn` comes with excellent documentation (including a Troubleshooting section). If you have a _How to do x_ question, it may have already been answered. Wiki: https://github.com/jarun/nnn/wiki

If it looks like a local environment issue, please try to debug yourself. Debugging local setup issues is not our top priority.

If you are looking for a new feature or program option, ensure you have the correct version with the feature installed. Refer to the release notes to confirm.

Before opening an issue, please try to reproduce on latest master. The bug you noticed might have already been fixed.

Useful links:

- compile `nnn` from source - https://github.com/jarun/nnn/wiki/Usage#from-source
- debugging `nnn` - https://github.com/jarun/nnn/wiki/Developer-guides#debugging-nnn

If the issue can be reproduced on master, log it.

Please provide the environment details. **If that's missing, the issue will be closed without any cited reason.**

If we need more information and there is no communication from the bug reporter within 7 days from the date of request, we will close the issue. If you have relevant information, resume discussion any time.

--- PLEASE DELETE THIS LINE AND EVERYTHING ABOVE ---

- [x] Operating System: Linux x86_64
- [x] Desktop Environment: none; PTY test
- [x] Terminal Emulator: PTY replay (`TERM=xterm-256color`)
- [x] Shell: bash
- [x] Custom desktop opener (if applicable): none
- [x] Program options used: `make O_NORL=1`; no runtime options
- [x] Configuration options set: none
- [ ] Plugins are installed

The sanitized build and runtime details are also in
[attachments/environment.txt](attachments/environment.txt).

From this report directory, run:

```sh
bash attachments/repro.sh
```

The attached PTY reproducer builds nnn from audited master and creates a
directory containing a filename whose bytes are
`70 72 65 07 70 6f 73 74`: the printable text `pre`, BEL, and `post`. It
also creates an ordinary sentinel file, starts nnn, captures the file-list row,
and exits. The same run checks DEL, ESC, LF, and TAB controls.

For a manual replay, create the filename with:

```sh
python3 - <<'PY'
import os
fd = os.open(b'pre\x07post', os.O_CREAT | os.O_WRONLY, 0o600)
os.close(fd)
PY
```

Then start nnn in that directory and inspect the file list. Three consecutive
PTY runs produced the same summary:

```text
PASS printable-control ...
FAIL control-byte suffix lost ...
checked=6 failures=5
```

The expected display is the complete filename. nnn's sanitization path replaces
unsafe control bytes with a question mark, so the row should read
`pre?post` rather than losing the printable suffix. The actual row displays
only `pre`; `post` disappears. A filename containing the literal question
mark, `pre?post`, renders correctly in the same replay, which rules out a
general short-name or terminal-width problem.

The display path enters `unescape`, which is documented as replacing escape
characters and limiting the result to the available columns:

[nnn.c](https://github.com/jarun/nnn/blob/6dd7c731eab2785e97f477482ab201ace9040e30/src/nnn.c#L4636-L4675)

In the wide-character build, `unescape` first converts the name and calls
`wcswidth` on the raw wide string at lines 4653–4655. If the calculated length
does not fit, it shortens the wide buffer at lines 4657–4664. Only after that
does the loop replace C0 and DEL characters with question marks at lines
4667–4672.

The normal no-filter name renderer passes every entry through `unescape`
before calling curses `addwstr`:

[nnn.c](https://github.com/jarun/nnn/blob/6dd7c731eab2785e97f477482ab201ace9040e30/src/nnn.c#L4997-L5013)

The ordering matters. A control byte can make the raw `wcswidth` result
invalid before it is sanitized; the suffix is then removed from the buffer
that `printent_name` sends to curses. Sanitizing first, or measuring the
sanitized wide string, would preserve the printable portion.

I searched the existing nnn issues and did not find a report describing this
control-byte suffix truncation. Issues
[#2068](https://github.com/jarun/nnn/issues/2068),
[#279](https://github.com/jarun/nnn/issues/279),
[#273](https://github.com/jarun/nnn/issues/273), and
[#2121](https://github.com/jarun/nnn/issues/2121) cover full-width names,
long-name/filter behavior, or other filename display cases; none describes
this ordering. The C0 and DEL cases fail consistently, while the printable
question-mark control passes.

The audited revision is
`6dd7c731eab2785e97f477482ab201ace9040e30`, the public nnn master revision
checked on 2026-09-12. The captured artifacts are
[attachments/evidence.log](attachments/evidence.log),
[attachments/environment.txt](attachments/environment.txt),
[attachments/repro.sh](attachments/repro.sh), and
[attachments.zip](attachments.zip).

[attachments.zip](https://github.com/user-attachments/files/32144134/attachments.zip)

vulragrag-star:
## Summary

`unescape()` measured display width with `wcswidth()` on the raw wide filename and only afterwards replaced C0/DEL bytes with `?`. A control byte can make that width calculation fail or under-count, so the printable suffix was clipped away.

Example: a name whose bytes are `pre` + BEL + `post` rendered as `pre` instead of `pre?post`.

This change sanitizes control bytes first, then measures width and clips to `maxcols`.

Fixes #2170

## Test plan

- [x] `make` (wide/`ncursesw` build) succeeds
- [x] `make checkpatches` succeeds on the patch framework matrix
- [x] Isolated PTY smoke: directory with `pre\x07post` + `sentinel.txt` lists `pre?post` (not truncated `pre`)
- [ ] CI on the PR


src/nnn.c
```diff
@@ -4645,34 +4649,53 @@ static char *unescape(const char *str, uint_t maxcols)
...
+
+	return wbuf;
+}
 #else
 static wchar_t *unescape(const char *str, uint_t maxcols)
 {
 	wchar_t * const wbuf = (wchar_t *)g_buf;
 	wchar_t *buf = wbuf;
-	size_t len = mbstowcs(wbuf, str, maxcols); /* Convert multi-byte to wide char */
+	size_t len = mbstowcs(wbuf, str, NAME_MAX); /* Convert multi-byte to wide char */
```
N-R-K: Why was the argument changed from `maxcols` to `NAME_MAX`?

vulragrag-star: Good catch — that wasn't needed for the fix.

`maxcols` is the column budget; the real bug was calling `wcswidth()` on the raw controls. Sanitize-first is what matters. Restored `mbstowcs(..., maxcols)` in c324d3b3.

N-R-K: Please don't copy paste LLM generated response here. If you're submitting a pull request you should at least have the basic understanding of the changes made to be able to *humanly* participate in the review. Otherwise you're just wasting our time, we can just use an LLM ourselves without the middleman and that'd be more efficient.

jarun:
@vulragrag-star not sure why you closed the PR, the final version was correct. I have checked in the changes at commit 67961dd14de1173ed90357679abde5b41949fb56.

